PR #7, opened 2020-02-19
Sampling feature

danilohgds:
*Issue #, if available:*
This relates to issue [Enhancement Suggestion - Log Sampling](https://github.com/awslabs/aws-lambda-powertools/issues/6)

*Description of changes:*
This pull request introduces a new variable/ constructor parameter to enable debug log sampling of a given percentage, similar to what DAZN powertools has. POWERTOOLS_LOGGER_SAMPLE_RATE accepts in 0-1 range, to log in debug mode at the given rate.


.gitignore
```diff
@@ -291,5 +291,5 @@ $RECYCLE.BIN/
 # End of https://www.gitignore.io/api/osx,linux,python,windows,pycharm,visualstudiocode
 
 # Misc
-
-test_report
\ No newline at end of file
+test_report
+/.idea/*
```
danilohgds: For some Reason, Idea would add files in a different folder, the current rule from gitignore would not work.

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -52,15 +59,32 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
         >>>
         >>> def handler(event, context):
                 logger.info("Hello")
+                :param service:
```
heitorlessa: 62-64 doesn't seem right - Can you remove it?

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -26,13 +29,17 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
...
     Parameters
     ----------
     service : str, optional
         service name to be appended in logs, by default "service_undefined"
     level : str, optional
         logging.level, by default "INFO"
+    sample_rate: float, optional
+        sample rate for debug calls within execution context defaults to 0
 
     Example
```
heitorlessa: Could you expand the second example to include sampling?  

For example:

    Setups structured logging in JSON for Lambda functions using env vars

        $ export POWERTOOLS_SERVICE_NAME="payment"
        $ export POWERTOOLS_LOGGER_SAMPLE_RATE=0.1 # 10% debug sampling
        >>> from aws_lambda_powertools.logging import logger_setup
        >>> logger = logger_setup()
        >>>
        >>> def handler(event, context):
                logger.info("Hello")

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -26,13 +29,17 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
         service name
     LOG_LEVEL: str
         logging level (e.g. INFO, DEBUG)
+    POWERTOOLS_LOGGER_SAMPLE_RATE: str
```
heitorlessa: I think you meant float in type annotation here - `POWERTOOLS_LOGGER_SAMPLE_RATE: float`

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -26,13 +29,17 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
...
+        samping rate ranging from 0 to 1, float precision
```
heitorlessa: Drop ", float precision" and use "1 being 100% sampling" instead. This along with float type annotation should clear out any ambiguity.

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -52,15 +59,32 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
...
     log_level = os.getenv("LOG_LEVEL") or level
     logger = logging.getLogger(name=service)
+
+    # sampling a small percentage of requests with debug level, using a float value 0.1 = 10%~
+
+    try:
+        if sampling_rate and random.random() <= float(sampling_rate):
+            log_level = logging.DEBUG
+    except ValueError:
+        logger.debug(
```
heitorlessa: Raise the exception instead so they know they need to take corrective actions.

```python
raise ValueError(f"Expected a float value ranging 0 to 1, but received {sampling_rate} instead. Please review POWERTOOLS_LOGGER_SAMPLE_RATE environment variable.")
```

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -110,6 +134,8 @@ def logger_inject_lambda_context(
     -------
     decorate : Callable
         Decorated lambda handler
+        :param log_event:
```
heitorlessa: Drop these two - I believe these are coming from your IDE automatically ;)

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -52,15 +59,32 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
...
+                :param level:
+                :param sampling_rate:
 
     """
     service = os.getenv("POWERTOOLS_SERVICE_NAME") or service
+    sampling_rate = os.getenv("POWERTOOLS_LOGGER_SAMPLE_RATE") or sampling_rate
     log_level = os.getenv("LOG_LEVEL") or level
     logger = logging.getLogger(name=service)
+
+    # sampling a small percentage of requests with debug level, using a float value 0.1 = 10%~
```
heitorlessa: Is this in the right place? ;)

python/tests/functional/test_logger.py
```diff
@@ -80,6 +80,19 @@ def test_setup_service_env_var(monkeypatch, root_logger, stdout):
     assert service_name == log["service"]
 
 
+def test_setup_sampling_rate(monkeypatch, root_logger, stdout):
```
heitorlessa: Could we improve by setting sampling rate to 1 and check if debug logs are always present?

.gitignore
heitorlessa: no problem :)

python/aws_lambda_powertools/logging/logger.py
danilohgds: yup, pycharm being cheeky :)

python/tests/functional/test_logger.py
danilohgds: I've changed the rate to 1, also added an assertion on log level from stdout to confirm DEBUG level being captured. It might also be interesting to do the opposite as well, with 0 sample rate.

python/aws_lambda_powertools/logging/logger.py
danilohgds: removed, good catch.

python/aws_lambda_powertools/logging/logger.py
```diff
@@ -55,12 +63,25 @@ def logger_setup(service: str = "service_undefined", level: str = "INFO", **kwar
...
     log_level = os.getenv("LOG_LEVEL") or level
     logger = logging.getLogger(name=service)
+
+    try:
+        if sampling_rate and random.random() <= float(sampling_rate):
+            log_level = logging.DEBUG
+    except ValueError:
+        raise ValueError(
+            f"Expected a float value ranging 0 to 1, but received {sampling_rate} instead. Please review "
+            f"POWERTOOLS_LOGGER_SAMPLE_RATE environment variable."
```
heitorlessa: You can merge this line with the above instead of having two lines ;)

f"Expected a float value ranging 0 to 1, but received {sampling_rate} instead. Please review POWERTOOLS_LOGGER_SAMPLE_RATE environment variable."

python/tests/functional/test_logger.py
heitorlessa: okay great, can you also check for the message and we can merge this?

```python
`assert "I am being sampled" == log["message"]`
```

heitorlessa: > I've changed the rate to 1, also added an assertion on log level from stdout to confirm DEBUG level being captured. It might also be interesting to do the opposite as well, with 0 sample rate.

I thought about it too, but because you're defaulting to 0 we know this works because other tests that capture the INFO value is still working :)

Issue #9, opened 2020-03-17
Enhancement suggestion: correlation id middleware combined with logger and sampler


marcioemiranda:
Hello Heitor,

Have you considered adding a correlation_id middleware like the one in DAZN power tools?

It could work together with the logger tool by adding correlation_id to the logs automatically, and also be used by the sampling feature (the whole transaction is sampled instead of each function invoke).

Regards,
Marcio.

heitorlessa:
Hey Marcio,

 I did considered though I haven't been able to create a design document to flesh this out yet.

 I'm currently working on another project until mid April, but if that's something you'd like to kick off I'd be super happy to contribute.

Thanks a lot for reaching out

marcioemiranda:
Hello Heitor,

I would like to contribute if you give me some pointers.

Regards,
Marcio

heitorlessa:
Absolutely @marcioemiranda  - Here's a breakdown of how the logger works

* [logger_setup](https://github.com/awslabs/aws-lambda-powertools/blob/develop/python/aws_lambda_powertools/logging/logger.py#L18) returns a Python Logger with minimal structured logging. It also accepts any `key=value` argument that will be present across all log operations
* [logger_inject_lambda_context](https://github.com/awslabs/aws-lambda-powertools/blob/develop/python/aws_lambda_powertools/logging/logger.py#L84) is the decorator that intercepts the Lambda context object, and uses [logger_setup to add key Lambda fn info as part of the struct log](https://github.com/awslabs/aws-lambda-powertools/blob/develop/python/aws_lambda_powertools/logging/logger.py#L149)

I see two ways of implementing it:

* **A)** Extending [logger_inject_lambda_context](https://github.com/awslabs/aws-lambda-powertools/blob/develop/python/aws_lambda_powertools/logging/logger.py#L141) to detect whether there's a correlation ID already (event object)
* **B)** Create a separate module to both detect and hold that correlation ID in memory, so other future modules could make use of it (middleware kinda approach)

I personally prefer B as it's more extensible since we could also use the correlation ID data in Tracer to create an annotation, and in future modules like detecting retries, etc. 

Regardless of the option, it'll have to add a new key `correlation_id` in this line: https://github.com/awslabs/aws-lambda-powertools/blob/develop/python/aws_lambda_powertools/logging/logger.py#L149

Happy to explain further but hopefully that makes sense

heitorlessa:
[...]

PR #17, opened 2020-04-10
Decorator factory Feat: Create your own middleware

heitorlessa:
*Issue #, if available:* #20 

*Description of changes:*

This is required for Serverless Airline project to move from the modified version of powertools, and create own middlewares to improve reusability as much as possible. 

Instead of creating reusable methods, we're at the stage that makes more sense to give flexibility to create own middlewares as #9, so customers can create one middleware with many custom processing before, after and exception handling instead.

This gives flexibility to control **before, after, and exception handling** logic in the way you'd like. It also lay the foundations to allow us to create a base Trace provider so you can use any Tracer that follow our protocol.

* [x] Create decorator factory
* [x] Tests 
* [x] Use existing Tracer class instance to trace custom middlewares
* [x] Update existing middlewares
* [x] Update docs
* [x] Utility to optionally trace middleware
* [x] Optional: Automate API docs

One step closer from GA 🎉 - We'll tackle log level next!

## User Experience

**Create a middleware no params**

```python

@lambda_handler_decorator
def log_response(handler, event, context):
    any_code_to_execute_before_lambda_handler()                
    response = handler(event, context)
    any_code_to_execute_after_lambda_handler()
    print(f"Lambda handler response: {response}")
    return response

@log_response
    return True
```

**Create a middleware with params**

```python

@lambda_handler_decorator
def obfuscate_sensitive_data(handler, event, context, fields=None):
    # Obfuscate email before calling Lambda handler
    if fields:
        for field in fields:
            field = event.get(field, "")
            event[field] = obfuscate_pii(field)

@obfuscate_sensitive_data(fields=["email"])
    return True
```

**Trace execution of custom middleware - Creates sub-segments in X-Ray**

```python
```python
from aws_lambda_powertools.utils import lambda_handler_decorator

@lambda_handler_decorator(trace_execution=True)
def my_custom_middleware(handler, event, context):
    ...
```


heitorlessa:
Local perf tests to rule out major performance impact on multiple callables - Decorator only prints before handler and after.

**No decorator**

50000000 loops, best of 5: 6.88 nsec per loop

**Simpler decorator**

50000000 loops, best of 5: 7.18 nsec per loop

**Decorator factory**

50000000 loops, best of 5: 7.25 nsec per loop

**Simpler Stacked decorator**

50000000 loops, best of 5: 7.05 nsec per loop

**Factory Stacked decorator**

50000000 loops, best of 5: 7.05 nsec per loop
---

```python
def call_handler_simpler(func):
    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        print("SIMPLER - BEGIN")
        response = func(*args, **kwargs)
        print("SIMPLER - END")
        return response
    return wrapper

@lambda_handler_decorator
def call_handler_factory(handler, event, context):
    print("[CALL HANDLER] BEGIN")
    response = handler(event, context)
    print("[CALL HANDLER] END")
    return response

    a = {
        "statusCode": 200,
        "body": "Hello world"
    }
    print("-------------LAMBDA FN CALLED!-------------")
    return a
```

python/README.md
```diff
@@ -85,6 +93,23 @@ def handler(event, context)
   ...
 ```
 
+**Fetching a pre-configured tracer anywhere**
+
+```python
+# handler.py
+from aws_lambda_powertools.tracing import Tracer
+tracer = Tracer(service="payment")
```
nmoutschen, line 101: Does this uses `POWERTOOLS_SERVICE_NAME` if I don't set service or is this mandatory?

python/README.md
```diff
@@ -85,6 +93,23 @@ def handler(event, context)
...
+
+@tracer.capture_lambda_handler
+def handler(event, context)
+  charge_id = event.get('charge_id')
+  payment = collect_payment(charge_id)
+  ...
+
+# another_file.py
+from aws_lambda_powertools.tracing import Tracer
+tracer = Tracer.instance()
```
nmoutschen, line 111: What is the behaviour if `another_file.py` is imported first? Would this fail?

I am thinking about the case where the Tracer is first instanciated in an import file, then re-used using `Trace.instance()` in another one. There shouldn't be a reason to enforce a specific import order.

python/README.md
```diff
@@ -204,6 +229,92 @@ with single_metric(name="ColdStart", unit=MetricUnit.Count, value=1) as metric:
...
+```python
+from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
+
+@lambda_handler_decorator
+def middleware_name(handler, event, context):
+    return handler(event, context)
+
+@lambda_handler_decorator
+def middleware_before_after(handler, event, context):
+    logic_before_handler_execution()
```
nmoutschen, line 254: I'd add comments explaining that this will be run first, etc. to make the order of operations more explicit for people that don't know much about how decorators behave.

python/aws_lambda_powertools/tracing/tracer.py
```diff
@@ -6,106 +6,136 @@
...
 
-    When running locally, it honours POWERTOOLS_TRACE_DISABLED environment variable
+    When running locally, it honours `POWERTOOLS_TRACE_DISABLED` environment variable
     so end user code doesn't have to be modified to run it locally
     instead Tracer returns dummy segments/subsegments.
 
     Tracing is automatically disabled when running locally via via SAM CLI.
 
-    It patches all available libraries supported by X-Ray SDK
+    By default, it patches all available libraries supported by X-Ray SDK. \n
```
nmoutschen, line 25: __Typo__: `\n` at the end of the line

python/aws_lambda_powertools/middleware_factory/factory.py
```diff
@@ -0,0 +1,143 @@
...
+                f"Only keyword arguments is supported for middlewares: {decorator.__qualname__} received {func}"
+            )
+
+        @functools.wraps(func)
+        def wrapper(event, context):
+            try:
+                middleware = functools.partial(decorator, func, event, context, **kwargs)
+                if trace_execution:
+                    with _trace_middleware(middleware=decorator):
+                        response = middleware()
```
nmoutschen, line 121: Is there a reason to use `functools.partial` then call the partial function rather than just do `decorator(func, event, context, **kwargs)`?

python/aws_lambda_powertools/middleware_factory/factory.py
```diff
@@ -0,0 +1,143 @@
...
+                logger.error(f"Caught exception in {decorator.__qualname__}")
+                raise err
+
+        return wrapper
+
+    return final_decorator
+
+
+@contextmanager
+def _trace_middleware(middleware):
```
nmoutschen, line 135: Are there plans to use this beyond the `lambda_handler_decorator` function? Otherwise, shouldn't we just remove this and move the code into the `lambda_handler_decorator` function?

python/README.md
heitorlessa: The former - Explicit first, env second.

python/README.md
heitorlessa: That's exactly why I made `Tracer.instance()` to create an instance if there isn't one so it wouldn't impact ordering :)

There's a **if** though. If you call `.instance()` first, it'll create a Tracer with default values unless there are env vars present. If you then create another `Tracer(explicit_config)` it'll create a new instance with a different config unless they're the same as env vars.

python/README.md
heitorlessa: Makes sense - I'll address that shortl

python/aws_lambda_powertools/middleware_factory/factory.py
heitorlessa: Just personal preference to avoid typing that over and over ;)

python/aws_lambda_powertools/middleware_factory/factory.py
heitorlessa: Good catch! I actually had when i first wrote! I had created another decorator to be used in classes to reduce LOC but since abandoned that.

Will address this shortly.

python/aws_lambda_powertools/tracing/tracer.py
heitorlessa: Not a typo - `pdoc3` uses this to enforce a newline as the link below got crammed

python/README.md
heitorlessa: As per our chat, moved to shared config instead (Borg)

Issue #28, opened 2020-05-04
Feature: Tracing providers


heitorlessa:
Ability to switch Tracing providers while keeping the same UX

heitorlessa:
After research, all tracers have distinct initialization, nomenclature (some OpenTracing, some unique to them like X-Ray), and not all of them handle async, thread-safe operations the same way.

To support this, it'd mean bringing lots of complexity to maintain, and unknown bugs. Customers can choose not to use the Tracer middleware if they already have their own Tracer provider.

Closing as we will not implement and will instead provide X-Ray escape hatches to customers, and simplify Tracer scope

Issue #11, opened 2020-03-25
Feature: Support tracing for async functions


heitorlessa:
## Background

**Issue, if applicable**: https://github.com/awslabs/aws-lambda-powertools/issues/10

As of now, Tracer doesn't support async functions, meaning that any method or library (aioboto3) that makes use of the event loop won't be traced properly.

## Implementation 

Investigation needed - X-Ray SDK supports async function by introducing async context: https://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python-patching.html#xray-sdk-python-patching-async

heitorlessa:
Released on 0.9.1 version

PR #29, opened 2020-05-04
Improv tracer - async support, patch, test coverage and X-Ray escape hatch

heitorlessa:
*Issue #, if available:* #28, #11, #10 

*Description of changes:*

* [x] Accept list of modules to be patched
* [x] Update tests
* [x] Test cold start annotation behaviour
* [x] Create `subsegment` entity/type (put_annotation, etc.)
* [x] Implement thread safe `current_subsegment -> subsegment`
* [x] Update decorators to use own subsegment to prevent race conditions
* [x] Support async
* [x] Support aiohttp
* [x] Expose escape hatch mechanism for X-Ray
* [x] Update docs on async, patch modules, and escape hatch
* [x] Update example
* [x] Update example tests
* [x] Document concurrent and nested async calls
    - Updated reference pointing to https://github.com/aws/aws-xray-sdk-python/issues/164
* [x] Bump version

**UX for async**

Customer needs to control the loop and await all coroutines

```python
@tracer.capture_method      # will prime this coroutine
async def greeting_2(name, message):
    return dummy_response

@tracer.capture_method      # will prime this coroutine
async def greeting(name, message):
    blah = await greeting_2(name, message)
    return dummy_response, blah

def lambda_handler(evt, ctx):
    loop = asyncio.get_event_loop()
    loop.run_until_complete(greeting(name="Foo", message="Bar"))
   # or gather() if multiple need to be awaited
````

### CODE REVIEW

File: python/aws_lambda_powertools/tracing/base.py [hunk 1]
```diff
@@ -0,0 +1,233 @@
...
+    It also allows custom providers to keep lean while Tracer provide:
+
+    * a simplified UX
+    * decorators for Lambda handler and methods
+    * auto-patching, patch all modules by default or a subset
+    * disabling all tracing operations with a single parameter or env var
+
+    Trace providers should implement the following methods:
+
+    * **patch**
```
nmoutschen, line 31: This implementations is inspired by how X-Ray does things. Have you though about how this will look for the most used tracer providers and if people won't have to shoehorn other concepts?

python/aws_lambda_powertools/tracing/base.py
```diff
@@ -0,0 +1,233 @@
...
+
+        Returns
+        -------
+        aws_xray_sdk.core.models.subsegment
+            AWS X-Ray Subsegment
+        """
+        # Will no longer be needed once #155 is resolved
+        # https://github.com/aws/aws-xray-sdk-python/issues/155
+        subsegment = self.client.begin_subsegment(name=name)
+        global is_cold_start
```
nmoutschen, line 163: It makes sense to have a global variable in the context of Lambda execution environments, but wouldn't it make more sense to have it at the class level? (think `self.__class__.__is_cold_start`)

python/aws_lambda_powertools/tracing/base.py
```diff
@@ -0,0 +1,233 @@
...
+    * disabling all tracing operations with a single parameter or env var
+
+    Trace providers should implement the following methods:
+
+    * **patch**
+    * **create_subsegment**
+    * **end_subsegment**
+    * **put_metadata**
+    * **put_annotation**
+    * **disable_tracing_provider**
```
nmoutschen, line 36: Should there be an `enable_tracing_provider`?

python/aws_lambda_powertools/tracing/base.py
heitorlessa: If they instantiate Tracer inside the handler then it'll always be cold start, hence why we have in the module

python/aws_lambda_powertools/tracing/base.py
heitorlessa: totally true and hence why I went back and forth with this - It's great to hear a second opinion as I think it's best to drop the provider support altogether.

heitorlessa:
> Overall, it looks good but I have questions regarding whether the tracer implementation is not too specific to X-Ray.

I dropped additional provider feature. Instead, this PR now tackles enhancements for `Tracer` - Async support, patch specific modules, and an escape hatch mechanism

python/aws_lambda_powertools/tracing/tracer.py
```diff
@@ -78,7 +83,7 @@ def confirm_booking(booking_id: str) -> Dict:
         @tracer.capture_lambda_handler
         def handler(event: dict, context: Any) -> Dict:
             print("Received event from Lambda...")
-            response = greeting(name="Heitor")
+            response = confirm_booking(booking_id=event["booking_id])
```
nmoutschen, line 86: __NIT__: missing quote

python/aws_lambda_powertools/tracing/tracer.py
```diff
@@ -135,7 +152,78 @@ def __init__(
...
+            tracer = Tracer(service="payment")
+            response = collect_payment()
+            tracer.put_metadata("Payment collection", response)
+        """
+        if self.disabled:
+            logger.debug("Tracing has been disabled, aborting put_metadata")
+            return
+
+        namespace = namespace or self.service
+        logger.debug(f"Adding metadata on key '{key}'' with '{value}'' at namespace '{namespace}''")
```
nmoutschen, line 206: __NIT__: are there excess quotes in the log message?

heitorlessa:
Thank you once again @nmoutschen for the review :) It's now published in 0.9.0

PR #65, opened 2020-06-07
chore: rename Makefile target docs-dev to docs-local

cakepietoast:
**Issue #, if available:**

## Description of changes:

> Makefile had 2 targets: docs-dev (which ran the gatsby docs site locally) and dev-docs (which installs required packages for docs build). Renamed the former to better represent what it does.

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

> Ignore if it's not a breaking change

**RFC issue #**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ build-docs-website: dev-docs
 docs-local:
 	cd docs && npm run start
 
-docs-api-dev:
+docs-api-local:
 	poetry run pdoc --http : aws_lambda_powertools
 
 security-baseline:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -43,7 +43,7 @@ build-docs-website: dev-docs
 	cd docs && npm run build
 	cp -R docs/public/* dist/
 
-docs-dev:
+docs-local:
 	cd docs && npm run start
 
 docs-api-dev:
```
heitorlessa, line 49: Should we change the API docs too? It's another local web server run by `pdoc3` 

```suggestion
docs-api-local:
```

PR #71, opened 2020-06-08
feat: improve error handling for log_metrics decorator

cakepietoast:
**Issue #, if available:**

## Description of changes:

> Change default behavior for log_metrics decorator: no longer throws exception if no metrics are emitted during the wrapped function's execution. Added parameter to revert this behavior if exception is desired in this case.

* [X] Update tests
* [X] Update docs

## Breaking change checklist

> Ignore if it's not a breaking change

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/metrics/metrics.py
+++ b/aws_lambda_powertools/metrics/metrics.py
@@ -122,7 +122,11 @@ class Metrics(MetricManager):
         # Return a partial function with args filled
         if lambda_handler is None:
             logger.debug("Decorator called with parameters")
-            return functools.partial(self.log_metrics, capture_cold_start_metric=capture_cold_start_metric)
+            return functools.partial(
+                self.log_metrics,
+                capture_cold_start_metric=capture_cold_start_metric,
+                raise_on_empty_metrics=raise_on_empty_metrics,
+            )
 
         @functools.wraps(lambda_handler)
         def decorate(event, context):
--- a/aws_lambda_powertools/metrics/metrics.py
+++ b/aws_lambda_powertools/metrics/metrics.py
@@ -2,6 +2,7 @@ import functools
 import json
 import logging
 import os
+import warnings
 from typing import Any, Callable
 
 from aws_lambda_powertools.metrics.base import MetricManager
@@ -136,7 +137,7 @@ class Metrics(MetricManager):
                     self.__add_cold_start_metric(context=context)
             finally:
                 if not raise_on_empty_metrics and not self.metric_set:
-                    logger.debug("No metrics to publish, skipping")
+                    warnings.warn("No metrics to publish, skipping")
                 else:
                     metrics = self.serialize_metric_set()
                     self.clear_metrics()
--- a/aws_lambda_powertools/metrics/metrics.py
+++ b/aws_lambda_powertools/metrics/metrics.py
@@ -86,7 +86,7 @@ class Metrics(MetricManager):
         self,
         lambda_handler: Callable[[Any, Any], Any] = None,
         capture_cold_start_metric: bool = False,
-        raise_for_empty_metrics: bool = False,
+        raise_on_empty_metrics: bool = False,
     ):
         """Decorator to serialize and publish metrics at the end of a function execution.
 
@@ -109,7 +109,7 @@ class Metrics(MetricManager):
             Lambda function handler, by default None
         capture_cold_start_metric : bool, optional
             Captures cold start metric, by default False
-        raise_for_empty_metrics : bool, optional
+        raise_on_empty_metrics : bool, optional
             Raise exception if no metrics are emitted, by default False
 
         Raises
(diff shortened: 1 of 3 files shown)


heitorlessa:
I'm trying to think of an actual customer scenario where this would be common to have no validation by default - Could you share one or two use cases where you wouldn't have a metric if you want to use Metrics?

I thought about scenarios where metric value would be 0, but even then it's a data point rather than missing data point.

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -82,7 +82,12 @@ def clear_metrics(self):
         self.metric_set.clear()
         self.dimension_set.clear()
 
-    def log_metrics(self, lambda_handler: Callable[[Any, Any], Any] = None, capture_cold_start_metric: bool = False):
+    def log_metrics(
+        self,
+        lambda_handler: Callable[[Any, Any], Any] = None,
+        capture_cold_start_metric: bool = False,
+        raise_for_empty_metrics: bool = False,
```
heitorlessa, line 89: This should be an opt-in behaviour unless this is valid for 80% of customers and we have feedback in this area already. 

```suggestion
        raise_on_empty_metrics: bool = True,
```

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -122,10 +131,13 @@ def decorate(event, context):
                 if capture_cold_start_metric:
                     self.__add_cold_start_metric(context=context)
             finally:
-                metrics = self.serialize_metric_set()
-                self.clear_metrics()
-                logger.debug("Publishing metrics", {"metrics": metrics})
-                print(json.dumps(metrics))
+                if not raise_for_empty_metrics and not self.metric_set:
```
heitorlessa, line 134: Can we move this to the domain level code (`metrics/base.py`)? That way, we can 

1) quickly implement to `single_metric` if we need to too
2) accommodate those flushing metrics manually like Richard in another issue
3) consider whether this should also be an env var, and possibly in the constructor if point 1 is requested by customers

aws_lambda_powertools/metrics/metrics.py
cakepietoast: In my opinion, metrics are rarely a part of the core functionality of an application - and 

> I'm trying to think of an actual customer scenario where this would be common to have no validation by default - Could you share one or two use cases where you wouldn't have a metric if you want to use Metrics?
> 
> I thought about scenarios where metric value would be 0, but even then it's a data point rather than missing data point.

In the code I wanted to integrate with, I want to emit a metric when a shopping cart gets updated. My lambda function has several scenarios where the cart doesn't get updated (for example, the product doesn't exist, or something else in the request was invalid). Maybe I _should_ be emitting metrics on those events too, but if my code exits without adding a metric, my expectation is that my function would still execute without throwing an exception.

aws_lambda_powertools/metrics/metrics.py
cakepietoast: I added it to the decorator rather than base because I think the behavior should be different for the decorator. If I'm explicitly making a call to serialize metrics when their are none, throwing an exception is fine and I can handle it myself.

heitorlessa:
That’s fair enough - I thought you were also adding metrics on unhappy path
(my expectation).

I’m happy with the change - On my phone now, only change I’d request is to
use “raise_on...” rather than “raise_for...”

cakepietoast:
> That’s fair enough - I thought you were also adding metrics on unhappy path (my expectation). I’m happy with the change - On my phone now, only change I’d request is to use “raise_on...” rather than “raise_for...”
Done!

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -82,7 +82,12 @@ def clear_metrics(self):
         self.metric_set.clear()
         self.dimension_set.clear()
 
-    def log_metrics(self, lambda_handler: Callable[[Any, Any], Any] = None, capture_cold_start_metric: bool = False):
+    def log_metrics(
+        self,
+        lambda_handler: Callable[[Any, Any], Any] = None,
+        capture_cold_start_metric: bool = False,
+        raise_on_empty_metrics: bool = False,
```
nmoutschen, line 89: Shouldn't this be inverted?

Looking at the changes, the current behaviour is to raise an exception on empty metrics, so this would make a change to silently fail on empty metrics. From a developer perspective, I think it'd be better to fail first as it might be due to an error in the code that doesn't trigger a metric under certain conditions. Then a developer can choose to either ignore (set this to False) or fix the issue if there was one.

cakepietoast: Discussed offline, agreed on warning when metric set is empty.

aws_lambda_powertools/metrics/metrics.py
cakepietoast: Discussed offline, agreed on warning when metric set is empty.

PR #96, opened 2020-07-29
feat: add get_parameter utility

nmoutschen:
**Issue #, if available:** https://github.com/awslabs/aws-lambda-powertools-python/issues/94

## Description of changes:

Add a new module `aws_lambda_powertools.utilities` with a first utility function `get_parameter` to retrieve and cache SSM parameter values.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**: https://github.com/awslabs/aws-lambda-powertools-python/issues/94

### DIFF
--- a/aws_lambda_powertools/utilities/parameters.py
+++ b/aws_lambda_powertools/utilities/parameters.py
@@ -1,0 +1,63 @@
+"""
+Parameter retrieval and caching utility
+"""
+
+
+from collections import namedtuple
+from datetime import datetime, timedelta
+
+import boto3
+
+DEFAULT_MAX_AGE = 5
+ExpirableValue = namedtuple("ExpirableValue", ["value", "ttl"])
+PARAMETER_VALUES = {}
+ssm = boto3.client("ssm")
+
+
+def get_parameter(name: str, max_age: int = DEFAULT_MAX_AGE) -> str:
+    """
+    Retrieve a parameter from the AWS Systems Manager (SSM) Parameter Store
+
+    This will keep a local version in cache for `max_age` seconds to prevent
+    overfetching from SSM Parameter Store.
+
+    See the [AWS Systems Manager Parameter Store documentation]
+    (https://docs.aws.amazon.com/systems-manager/latest/userguide/systems-manager-parameter-store.html)
+    for more information.
+
+    Parameters
+    ----------
+    name: str
+        Name of the SSM Parameter
+    max_age: int
+        Duration for which the parameter value can be cached
+
+    Example
+    -------
+
+        from aws_lambda_powertools.utilities import get_parameter
+
+        def lambda_handler(event, context):
+            # This will only make a call to the SSM service every 30 seconds.
+            value = get_parameter("my-parameter", max_age=30)
+
+    Raises
+    ------
+    ssm.exceptions.InternalServerError
+        When there is an internal server error from AWS Systems Manager
+    ssm.exceptions.InvalidKeyId
+        When the key ID is invalid
+    ssm.exceptions.ParameterNotFound
+        When the parameter name is not found in AWS Systems Manager
+    ssm.exceptions.ParameterVersionNotFound
+        When a version of the parameter is not found in AWS Systems Manager
+    """
+
+    if name not in PARAMETER_VALUES or PARAMETER_VALUES[name].ttl < datetime.now():
+        # Retrieve the parameter from AWS Systems Manager
+        parameter = ssm.get_parameter(Name=name)
+        PARAMETER_VALUES[name] = ExpirableValue(
+            parameter["Parameter"]["Value"], datetime.now() + timedelta(seconds=max_age)
+        )
+
+    return PARAMETER_VALUES[name].value
--- a/aws_lambda_powertools/utilities/__init__.py
+++ b/aws_lambda_powertools/utilities/__init__.py
@@ -1,0 +1,8 @@
+# -*- coding: utf-8 -*-
+
+"""General utilities for Powertools"""
+
+
+from .parameters import get_parameter
+
+__all__ = ["get_parameter"]
--- a/poetry.lock
+++ b/poetry.lock
@@ -162,7 +162,7 @@ typed-ast = ">=1.4.0"
 d = ["aiohttp (>=3.3.2)", "aiohttp-cors"]
 
 [[package]]
-category = "dev"
+category = "main"
 description = "The AWS SDK for Python"
 name = "boto3"
 optional = false
@@ -816,7 +816,7 @@ security = ["pyOpenSSL (>=0.14)", "cryptography (>=1.3.4)"]
 socks = ["PySocks (>=1.5.6,<1.5.7 || >1.5.7)", "win-inet-pton"]
 
 [[package]]
-category = "dev"
+category = "main"
 description = "An Amazon S3 Transfer Manager"
 name = "s3transfer"
 optional = false
@@ -951,7 +951,6 @@ multidict = ">=4.0"
 [[package]]
 category = "main"
 description = "Backport of pathlib-compatible object wrapper for zip files"
-marker = "python_version < \"3.8\""
 name = "zipp"
 optional = false
 python-versions = ">=3.6"
@@ -962,7 +961,8 @@ docs = ["sphinx", "jaraco.packaging (>=3.2)", "rst.linker (>=1.9)"]
 testing = ["jaraco.itertools", "func-timeout"]
 
 [metadata]
-content-hash = "a2760fd5f04b7f1841509fcbcb4ccdaf35d92d1395627787e4a11f391a0597d2"
+content-hash = "18607a712e4a4a05de7350ecbcf26327a4fb45bb8609dc7f3d19b7610c2faafc"
+lock-version = "1.0"
 python-versions = "^3.6"
 
 [metadata.files]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,6 +21,7 @@ license = "MIT-0"
 python = "^3.6"
 aws-xray-sdk = "^2.5.0"
 fastjsonschema = "~=2.14.4"
+boto3 = "^1.12"
 
 [tool.poetry.dev-dependencies]
 coverage = {extras = ["toml"], version = "^5.0.3"}
--- a/tests/functional/test_utilities.py
+++ b/tests/functional/test_utilities.py
@@ -1,0 +1,128 @@
+import random
+import string
+from datetime import datetime, timedelta
+
+import pytest
+from botocore import stub
+
+from aws_lambda_powertools import utilities
+from aws_lambda_powertools.utilities import parameters
+
+
+@pytest.fixture(scope="function")
+def mock_name():
+    # Parameter name must match [a-zA-Z0-9_.-/]+
+    return "".join(random.choices(string.ascii_letters + string.digits + "_.-/", k=random.randrange(3, 200)))
+
+
+@pytest.fixture(scope="function")
+def mock_value():
+    # Standard parameters can be up to 4 KB
+    return "".join(random.choices(string.printable, k=random.randrange(100, 4000)))
+
+
+@pytest.fixture(scope="function")
+def mock_version():
+    return random.randrange(1, 1000)
+
+
+def test_get_parameter_new(monkeypatch, mock_name, mock_value, mock_version):
+    """
+    Test get_parameter() with a new parameter name
+    """
+
+    # Patch the parameter value store
+    monkeypatch.setattr(parameters, "PARAMETER_VALUES", {})
+
+    # Stub boto3
+    stubber = stub.Stubber(parameters.ssm)
+    response = {
+        "Parameter": {
+            "Name": mock_name,
+            "Type": "String",
+            "Value": mock_value,
+            "Version": mock_version,
+            "Selector": f"{mock_name}:{mock_version}",
+            "SourceResult": "string",
+            "LastModifiedDate": datetime(2015, 1, 1),
+            "ARN": f"arn:aws:ssm:us-east-2:111122223333:parameter/{mock_name}",
+        }
+    }
+    expected_params = {"Name": mock_name}
+    stubber.add_response("get_parameter", response, expected_params)
+    stubber.activate()
+
+    # Get the parameter value
+    try:
+        value = utilities.get_parameter(mock_name)
+
+        assert value == mock_value
+        stubber.assert_no_pending_responses()
+    finally:
+        stubber.deactivate()
+
+
+def test_get_parameter_cached(monkeypatch, mock_name, mock_value, mock_version):
+    """
+    Test get_parameter() with a cached value for parameter name
+    """
+
+    # Patch the parameter value store
+    monkeypatch.setattr(
+        parameters,
+        "PARAMETER_VALUES",
+        {mock_name: parameters.ExpirableValue(mock_value, datetime.now() + timedelta(seconds=60))},
+    )
+
+    # Stub boto3
+    stubber = stub.Stubber(parameters.ssm)
+    stubber.activate()
+
+    # Get the parameter value
+    try:
+        value = utilities.get_parameter(mock_name)
+
+        assert value == mock_value
+        stubber.assert_no_pending_responses()
+    finally:
+        stubber.deactivate()
+
+
+def test_get_parameter_expired(monkeypatch, mock_name, mock_value, mock_version):
+    """
+    Test get_parameter() with a cached, but expired value for parameter name
+    """
+
+    # Patch the parameter value store
+    monkeypatch.setattr(
+        parameters,
+        "PARAMETER_VALUES",
+        {mock_name: parameters.ExpirableValue(mock_value, datetime.now() - timedelta(seconds=60))},
+    )
+
+    # Stub boto3
+    stubber = stub.Stubber(parameters.ssm)
+    response = {
+        "Parameter": {
+            "Name": mock_name,
+            "Type": "String",
+            "Value": mock_value,
+            "Version": mock_version,
+            "Selector": f"{mock_name}:{mock_version}",
+            "SourceResult": "string",
+            "LastModifiedDate": datetime(2015, 1, 1),
+            "ARN": f"arn:aws:ssm:us-east-2:111122223333:parameter/{mock_name}",
+        }
+    }
+    expected_params = {"Name": mock_name}
+    stubber.add_response("get_parameter", response, expected_params)
+    stubber.activate()
+
+    # Get the parameter value
+    try:
+        value = utilities.get_parameter(mock_name)
+
+        assert value == mock_value
+        stubber.assert_no_pending_responses()
+    finally:
+        stubber.deactivate()


nmoutschen:
OK, this should work better. I've used monkeypatch to inject the environment variables instead. 😅

aws_lambda_powertools/utilities/parameters.py
```diff
@@ -0,0 +1,63 @@
...
+Parameter retrieval and caching utility
+"""
+
+
+from collections import namedtuple
+from datetime import datetime, timedelta
+
+import boto3
+
+DEFAULT_MAX_AGE = 5
```
jplock, line 11: Could this be renamed to DEFAULT_MAX_AGE_SECS or something to reflect the time unit?

aws_lambda_powertools/utilities/parameters.py
```diff
@@ -0,0 +1,63 @@
...
+        When the key ID is invalid
+    ssm.exceptions.ParameterNotFound
+        When the parameter name is not found in AWS Systems Manager
+    ssm.exceptions.ParameterVersionNotFound
+        When a version of the parameter is not found in AWS Systems Manager
+    """
+
+    if name not in PARAMETER_VALUES or PARAMETER_VALUES[name].ttl < datetime.now():
+        # Retrieve the parameter from AWS Systems Manager
+        parameter = ssm.get_parameter(Name=name)
```
jplock, line 58: Should we include WithDecryption=True here?

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+
+        client_kwargs = {}
+        if region:
+            client_kwargs["region_name"] = region
+
+        self.client = boto3.client("ssm", **client_kwargs)
+
+        super().__init__()
+
+    def _get(self, name: str, **kwargs) -> str:
```
heitorlessa, line 48: Shouldn't `decrypt` be an explicit keyword param in the function signature here? 

Maybe I'm misreading the docstring

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+        decrypt: bool
+            If the parameter value should be decrypted
+        """
+
+        # Load kwargs
+        decrypt = kwargs.get("decrypt", False)
+
+        return self.client.get_parameter(Name=name, WithDecryption=decrypt)["Parameter"]["Value"]
+
+    def _get_multiple(self, path: str, **kwargs) -> Dict[str, str]:
```
heitorlessa, line 65: Shouldn't `decrypt`, and `recursive` be explicit keyword params in the function signature here? 

Maybe I'm misreading the docstring. It'd also help autocomplete when using IDEs too

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+
+        client_kwargs = {}
+        if region:
+            client_kwargs["region_name"] = region
+
+        self.client = boto3.client("secretsmanager", **client_kwargs)
+
+        super().__init__()
+
+    def _get(self, name: str, **kwargs) -> str:
```
heitorlessa, line 130: What do you think of standardizing on on additional options via a `sdk_options` keyword param?

If no params are passed we're fine as the call works, otherwise, it still gives them the flexibility to pass in things like [VersionId, VersionStage](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager.html#SecretsManager.Client.get_secret_value)

This would change to

`return self.client.get_secret_value(SecretId=name, **sdk_options)["SecretString"]`

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+    AWS Systems Manager Parameter Store Provider
+    """
+
+    client = None
+
+    def __init__(
+        self, region: Optional[str] = None,
+    ):
+        """
```
heitorlessa, line 36: Could you add an `example section` in the docstring to show the expected usage? i.e. those who might not have access to the docs (IDE shows it)

Logger example as that's also available in the API reference: https://awslabs.github.io/aws-lambda-powertools-python/api/logging/index.html#example

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+
+class SecretsProvider(BaseProvider):
+    """
+    AWS Secrets Manager Parameter Provider
+    """
+
+    client = None
+
+    def __init__(self, region: Optional[str] = None):
+        """
```
heitorlessa, line 118: Could you add an `example section` in the docstring to show the expected usage? i.e. those who might not have access to the docs (IDE shows it)

Logger example as that's also available in the API reference: https://awslabs.github.io/aws-lambda-powertools-python/api/logging/index.html#example

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+    """
+
+    table = None
+    key_attr = None
+    value_attr = None
+
+    def __init__(
+        self, table_name: str, key_attr: str = "id", value_attr: str = "value", region: Optional[str] = None,
+    ):
+        """
```
heitorlessa, line 156: Could you add an `example section` in the docstring to show the expected usage? i.e. those who might not have access to the docs (IDE shows it)

Logger example as that's also available in the API reference: https://awslabs.github.io/aws-lambda-powertools-python/api/logging/index.html#example

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+class DynamoDBProvider(BaseProvider):
+    """
+    Amazon DynamoDB Parameter Provider
+    """
+
+    table = None
+    key_attr = None
+    value_attr = None
+
+    def __init__(
```
heitorlessa, line 153: Same question to the idea of accepting a SDK Options/Config

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+            client_kwargs["region_name"] = region
+        self.table = boto3.resource("dynamodb", **client_kwargs).Table(table_name)
+
+        self.key_attr = key_attr
+        self.value_attr = value_attr
+
+        super().__init__()
+
+    def _get(self, name: str, **kwargs) -> str:
+        """
```
heitorlessa, line 171: Can we expand docstring to explain a bit more this is a `getItem`?

Same question as before, what do you think of a `sdk_options` param

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+        Retrieve a parameter value from Amazon DynamoDB
+        """
+
+        return self.table.get_item(Key={self.key_attr: name})["Item"][self.value_attr]
+
+    def _get_multiple(self, path: str, **kwargs) -> Dict[str, str]:
+        """
```
heitorlessa, line 178: I think this could have an example section as I'm not sure `path` represents your intent, but I could be misreading it

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+
+        return retval
+
+
+# These providers will be dynamically initialized on first use of the helper functions
+_DEFAULT_PROVIDERS = {}
+
+
+def get_parameter(name: str, transform: Optional[str] = None) -> Union[str, list, dict, bytes]:
+    """
```
heitorlessa, line 213: Could you expand the docstring on Parameters and Example? Specially `transform` - It's clear in the RFC but not when consuming it on what's available

tests/functional/test_utilities_parameters.py
```diff
@@ -0,0 +1,941 @@
...
+    Test DynamoDBProvider.get() with a non-cached value
+    """
+
+    table_name = "TEST_TABLE"
+
+    # Create a new provider
+    provider = parameters.DynamoDBProvider(table_name, region="us-east-1")
+
+    # Stub the boto3 client
+    stubber = stub.Stubber(provider.table.meta.client)
```
heitorlessa, line 44: nitpick: I think you might be able to turn this Stubber into a ctx manager

aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -0,0 +1,247 @@
...
+            If the parameter values should be decrypted
+        recursive: bool
+            If this should retrieve the parameter values recursively or not
+        """
+
+        # Load kwargs
+        decrypt = kwargs.get("decrypt", False)
+        recursive = kwargs.get("recursive", False)
+
+        response = self.client.get_parameters_by_path(Path=path, WithDecryption=decrypt, Recursive=recursive)
```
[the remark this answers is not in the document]
nmoutschen: You mean uppercase? As in, should we pass all kwargs directly?

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -0,0 +1,124 @@
...
+        if key not in self.store or self.store[key].ttl < datetime.now():
+            try:
+                value = self._get(name, **kwargs)
+            # Encapsulate all errors into a generic GetParameterError
+            except Exception as exc:
+                raise GetParameterError(str(exc))
+
+            if transform == "json":
+                value = json.loads(value)
+            elif transform == "binary":
```
heitorlessa, line 80: **suggestion** : Raise an exception if transform is not JSON or BINARY?

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -0,0 +1,124 @@
...
+        if key not in self.store or self.store[key].ttl < datetime.now():
+            try:
+                values = self._get_multiple(path, **kwargs)
+            # Encapsulate all errors into a generic GetParameterError
+            except Exception as exc:
+                raise GetParameterError(str(exc))
+
+            if transform == "json":
+                values = {k: json.loads(v) for k, v in values.items()}
+            elif transform == "binary":
```
heitorlessa, line 112: Same as before - Perhaps raise an exception if string isn't `json` or `binary`? 

I was gonna suggest an enum to make this easier but it'll further complicate the code for only two options (not worth it)

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -0,0 +1,164 @@
...
+    def _get(self, name: str, **sdk_options) -> str:
+        """
+        Retrieve a parameter value from Amazon DynamoDB
+
+        Parameters
+        ----------
+        name: str
+            Name of the parameter
+        sdk_options: dict, optional
+            Dictionary of options that will be passed to the get_item call
```
heitorlessa, line 126:
```suggestion
            Dictionary of options that will be passed to DynamoDB get_item API call
```

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -0,0 +1,164 @@
...
+        """
+
+        # Explicit arguments will take precedence over keyword arguments
+        sdk_options["Key"] = {self.key_attr: name}
```
heitorlessa, line 130: resolve me - Just wanted to drop saying I loved this new iteration ;)  It gives customers flexibility without taking the burden on maintaining that long term

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -0,0 +1,164 @@
...
+    table_name: str
+        Name of the DynamoDB table that stores parameters
+    key_attr: str, optional
+        Hash key for the DynamoDB table
+    value_attr: str, optional
+        Attribute that contains the values in the DynamoDB table
+    config: botocore.config.Config, optional
+        Botocore configuration to pass during client initialization
+
+    Example
```
heitorlessa, line 30: **nitpick:** Could you add one more line in the first example showing how to access the returned parameter?

That way it's clearer and you only do that twice - One for `get` and one for `get_multiple`

```
print(f"Parameter value -> { }")
```

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -0,0 +1,164 @@
...
+    the hood. This expects that the sort key is named `sk`.
+
+        >>> from aws_lambda_powertools.utilities.parameters import DynamoDBProvider
+        >>> ddb_provider = DynamoDBProvider("ParametersTable")
+        >>>
+        >>> ddb_provider.get_multiple("my-parameters")
+
+    **Retrieves multiple values from a DynamoDB table with a custom sort key**
+
+    In this case, the provider will use a sort key to retrieve multiple values using a query under
```
heitorlessa, line 82: I think this might require an example Item object to make `sort_attr` clearer

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+    Retrieve a parameter value from AWS Secrets Manager
+
+    Parameters
+    ----------
+    name: str
+        Name of the parameter
+    transform: str, optional
+        Transforms the content from a JSON object ('json') or base64 binary string ('binary')
+    sdk_options: dict, optional
+        Dictionary of options that will be passed to the get_secret_value call
```
heitorlessa, line 98:
```suggestion
        Dictionary of options that will be passed to the Secrets Manager get_secret_value API call
```

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -0,0 +1,164 @@
...
+        Retrieve multiple parameter values from Amazon DynamoDB
+
+        Parameters
+        ----------
+        path: str
+            Path to retrieve the parameters
+        sort_attr: str, optional
+            Name of the DynamoDB table sort key (defaults to 'sk')
+        sdk_options: dict, optional
+            Dictionary of options that will be passed to the query call
```
heitorlessa, line 145:
```suggestion
            Dictionary of options that will be passed to the DynamoDB query API call
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -0,0 +1,125 @@
...
+    def _get(self, name: str, **sdk_options) -> str:
+        """
+        Retrieve a parameter value from AWS Systems Manager Parameter Store
+
+        Parameters
+        ----------
+        name: str
+            Name of the parameter
+        sdk_options: dict
+            Dictionary of options that will be passed to the get_secret_value call
```
heitorlessa, line 72:
```suggestion
            Dictionary of options that will be passed to the Secrets Manager get_secret_value API call
```

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -0,0 +1,212 @@
...
+        Retrieve a parameter value from AWS Systems Manager Parameter Store
+
+        Parameters
+        ----------
+        name: str
+            Parameter name
+        decrypt: bool, optional
+            If the parameter value should be decrypted
+        sdk_options: dict, optional
+            Dictionary of options that will be passed to the get_parameter call
```
heitorlessa, line 82:
```suggestion
            Dictionary of options that will be passed to the Parameter Store get_parameter API call
```

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -0,0 +1,212 @@
...
+        Parameters
+        ----------
+        path: str
+            Path to retrieve the parameters
+        decrypt: bool, optional
+            If the parameter values should be decrypted
+        recursive: bool, optional
+            If this should retrieve the parameter values recursively or not
+        sdk_options: dict, optional
+            Dictionary of options that will be passed to the get_parameters_by_path call
```
heitorlessa, line 104:
```suggestion
            Dictionary of options that will be passed to the Parameter Store get_parameters_by_path API call
```

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -0,0 +1,212 @@
...
+
+        >>> from botocore.config import Config
+        >>> from aws_lambda_powertools.utilities.parameters import SSMProvider
+        >>>
+        >>> config = Config(region_name="us-west-1")
+        >>> ssm_provider = SSMProvider(config=config)
+        >>>
+        >>> ssm_provider.get("/my/parameter")
+
+    **Retrieves multiple parameter values from Systems Manager Parameter Store using a path prefix**
```
heitorlessa, line 42: Could you add an example of how to access the parameters retrieved? same as the suggestion on Dynamo

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -0,0 +1,212 @@
...
+        # Explicit arguments will take precedence over keyword arguments
+        sdk_options["Path"] = path
+        sdk_options["WithDecryption"] = decrypt
+        sdk_options["Recursive"] = recursive
+
+        response = self.client.get_parameters_by_path(**sdk_options)
+        parameters = response.get("Parameters", [])
+
+        # Keep retrieving parameters
+        while "NextToken" in response:
```
alexcasalboni, line 116: I always recommend using actual paginators:

```python
response = self.client.get_paginator('get_parameters_by_path').paginate(**sdk_options)
```

so you don't have to reinvent the "while" :)

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -0,0 +1,212 @@
...
+
+    # Only create the provider if this function is called at least once
+    if "ssm" not in DEFAULT_PROVIDERS:
+        DEFAULT_PROVIDERS["ssm"] = SSMProvider()
+
+    return DEFAULT_PROVIDERS["ssm"].get(name, transform=transform)
+
+
+def get_parameters(
+    path: str, transform: Optional[str] = None, recursive: bool = False, decrypt: bool = False, **sdk_options
```
alexcasalboni, line 175: I've had many customers complaining about the default value of `recursive` in the CLI/SDK.

Imho, for 99% of use cases customers want `recursive=True`, even though the default is false.

For example, if you have these parameters:
- /app/dev/db/collection
- /app/dev/db/host
- /app/dev/bucket

A developer would expect `get_parameters("/app/dev/")` to return three parameters, but it won't return anything. This is very confusing and frustrating.

You'd need to `get_parameters("/app/dev/", recursive=True)` to fetch everything.

That's why I made recursive=True the default value in https://github.com/alexcasalboni/ssm-cache-python

heitorlessa:
Reviewing the docs now - Thanks for implementing the other changes so quickly @nmoutschen ;)

Can't wait for 3.6 to reach EOL, so we can use the new Protocols more effectively instead of all-in on ABC: https://www.python.org/dev/peps/pep-0544/

Issue #97, opened 2020-07-31
Get the logger to respect module hierarchy as a normal logger


alexanderluiscampino:
Feature request to have this logger have the same multiple module inheritance present on the normal logging module.

When we create a normal log on the main.py, then subsequence invocations of `logging.getLogger(__name__) `on submodule will inherit the handlers and levels set on the main.py, due to the singleton nature of the logging.Logger class , hence using the logging.getLogger(name) to get the correct instance of the logger.

This allows for any log message in sub-modules to have the same characteristics a the parent logger. In this specific case, since this formatter transforms everything into a nice JSON with the possibility of adding structure logs, would be nice for those structure logs to persist and be inherited by the child logging instances. 

If I need to create another `logger = Logger()` this will indeed create a whole new instance which has no relationship with the logger on `main.py`, hence all the structure logs setup on main are now lost and need to be recreated, besides being repetitive, it might not be possible.

For example, these two logs, one from the lambda handler, another one from a submodule. The runID was lost, since it was added via `logger.structure_logs(append=True, runID=run_id)`
```json
{"timestamp": "2020-07-31 06:29:32,699", "level": "INFO", "location": "lambda_handler:46", "service": "myService", "sampling_rate": 0.0, "cold_start": true, "function_name": "name, "function_memory_size": "512", "runID": "1b250a6732", "message": "my message on main.py"}
```
```json
{"timestamp": "2020-07-31 06:29:33,161", "level": "INFO", "location": "start_execution:49", "service": "core.handlers.handler", "sampling_rate": 0.0, "message": "my message on a sub module"}
```

It would be ideal that some basic keys would be inherited, such that querying the logs afterwards can be easy, such as getting all logs where runID = 'abc'.

Re setup the structure_logs on every module is not possible, since modules might be agnostic to the concept of runID

heitorlessa:
Hey @alexanderluiscampino - Thanks for opening a feature request!

### Background

As you pointed out, `Logger` doesn't follow a compliant hierarchy because we don't use `getLogger` under the hood - It's not a singleton per se as we return different instances. 

[...]

PR #99, opened 2020-08-09
feat: logger inheritance

heitorlessa:
**Issue #, if available:** #97 

## Description of changes:

This addresses a common issue when trying to use `Logger` across multiple modules while wanting to:

* **Create a new child Logger** to inherit parent Logger's configuration (handler, formatter, custom keys, sampling)
* **Reuse the same Logger**, where Logger is enriched iteratively across the code base

For this, we've introduced a new parameter, `child`, when initializing a new Logger, and some internal changes while keeping the same UX and backwards compatibility.

* **[Internal]** Direct access to methods, handlers, etc work as before, except that we proxy properties access to `._logger`

## UX

**Key aspects**:

* We support logging inheritance via the `child` parameter, and the `Logger` will be named after `<service>.<caller_file_name>`
* Calling `Logger` with the same `service` name (explicit or via environment variable) returns the same `Logger`
* Changes to either Child `Logger` or Parent `Logger` will be propagated to all instances

### Creating a child Logger

Child logger can be created in any other as long as a parent is created too - This supports customers using Layers or different file names to create child loggers.

```python
# lambda_handler.py
import collect

logger = Logger(service="payment") # parent Logger just like before, named 'payment'
...

# collect.py

logger = Logger(service="payment", child=True) # child Logger named 'payment.collect'

# a_new_key will now be available both in this child logger as well as the parent
logger.structure_logs(append=True, a_new_key="value") 
...
```

### Reusing the same Logger

```python
# lambda_handler.py
import collect

logger = Logger(service="payment") # parent Logger just like before, named 'payment'

# collect.py

logger = Logger(service="payment") # returns the existing 'payment' Logger
...
```

**Checklist**

 

- [x] Update tests
- [ ] Update docs
- [x] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/376ec0a2ac0d2a40e0af5717bef42ff84ca0d1b9/.github/semantic.yml#L2)

**Changes**

- [x] Register logger via `getLogger` to allow inheritance
- [x] Encapsulate logger initialization/config in a new method
- [x] Remove Logger sub-classing
  - [x] Proxy calls to underneath Logger instead
- [x] ~~Create `name` parameter for hierarchy~~
  - [x] ~~Use `service` if `name` is empty~~
- [x] Test on Lambda locally
  - [x] Single Logger, no code changes
  - [x] Two Loggers in different files
  - [x] Additional Logger in a package that is imported on main entry-point
  - [x] Lambda Layer
- [x] Only add a handler if there isn't one
- [x] Only add a handler if parent is root
- [x] Update Formatter to append new keys
- [x] Make `handler` private
- [x] Remove `log_keys`
- [x] Remove duplicate Parameters in docstring
- [x] Double check Debug w/ getattr perf impact
- [x] Docs: Document inheriting Loggers
- [x] Docs: Changelog (doc random logger for tests)
- [x] Test: Logger hierarchy via logger.parent
- [x] Remove `name` parameter
- [x] Allow logger child propagation
- [x] Introduce `child` boolean param
- [x] Don't add a handler nor evaluate sampling if it's a `child` Logger
- [x] Name `child` Loggers based on `<service>.<caller_file_name>` to mimick `getLogger(__name__)`
- [x] Update tests to remove Handler on every test
- [x] Update tests to remove `getRandomLogger` utility function

**Optional**

- [x] Ensure log level input is case insensitive

## Benchmark

> **TL;DR**: For the average use case we have a slight performance decrease, but all negligible in all cases (0.10 nano seconds). 

Quick checks as we've removed `logging.Logger` sub-classing, respect Logging inheritance, and are proxying calls to inner Logger.

### Basic logger

* **Script**: `python -m timeit -s 'from aws_lambda_powertools import Logger; logger = Logger(); logger.info("hello world")'`
* **Before**: 50000000 loops, best of 5: **6.12 nsec** per loop
* **After with Logger proxy object, no sub-classing**: 50000000 loops, best of 5: **6.22 nsec** per loop

### Inheritance

* **Script**: `python -m timeit -s 'from aws_lambda_powertools import Logger; logger = Logger(); l = Logger(); l.info("hello world")'`
* **Same logger**: 50000000 loops, best of 5: **6.28 nsec** per loop
* **Random child logger name + fixup**: 50000000 loops, best of 5: **6.14 nsec** per loop
    - **Script**: `python -m timeit -s 'from aws_lambda_powertools import Logger; logger = Logger(child=True); l = Logger(); l.info("hello world")'`

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -1,4 +1,5 @@
 import functools
+import inspect
 import logging
 import os
 import random
@@ -54,8 +55,8 @@ class Logger:
         service name to be appended in logs, by default "service_undefined"
     level : str, optional
         logging.level, by default "INFO"
-    name: str, optional
-        Logger name, "{service}" by default
+    child: bool, optional
+        create a child Logger named <service>.<caller_file_name>, False by default
     sample_rate: float, optional
         sample rate for debug calls within execution context defaults to 0.0
     stream: sys.stdout, optional
@@ -90,15 +91,16 @@ class Logger:
                 logger.structure_logs(append=True, payment_id=event["payment_id"])
                 logger.info("Hello")
 
-    **Create child Logger using logging inheritance via name param**
+    **Create child Logger using logging inheritance via child param**
 
         >>> # app.py
+        >>> import another_file
         >>> from aws_lambda_powertools import Logger
-        >>> logger = Logger(name="payment")
+        >>> logger = Logger(service="payment")
         >>>
         >>> # another_file.py
         >>> from aws_lambda_powertools import Logger
-        >>> logger = Logger(name="payment.child)
+        >>> logger = Logger(service="payment", child=True)
 
     Raises
     ------
@@ -110,18 +112,18 @@ class Logger:
         self,
         service: str = None,
         level: Union[str, int] = None,
-        name: str = None,
+        child: bool = False,
         sampling_rate: float = None,
         stream: sys.stdout = None,
         **kwargs,
     ):
         self.service = service or os.getenv("POWERTOOLS_SERVICE_NAME") or "service_undefined"
-        self.name = name or self.service
         self.sampling_rate = sampling_rate or os.getenv("POWERTOOLS_LOGGER_SAMPLE_RATE") or 0.0
         self.log_level = level or os.getenv("LOG_LEVEL".upper()) or logging.INFO
+        self.child = child
         self._handler = logging.StreamHandler(stream) if stream is not None else logging.StreamHandler(sys.stdout)
         self._default_log_keys = {"service": self.service, "sampling_rate": self.sampling_rate}
-        self._logger = logging.getLogger(self.name)
+        self._logger = self._get_logger()
 
         self._init_logger(**kwargs)
 
@@ -130,15 +132,32 @@ class Logger:
         # https://github.com/awslabs/aws-lambda-powertools-python/issues/97
         return getattr(self._logger, name)
 
+    def _get_logger(self):
+        """ Returns a Logger named {self.service}, or {service.filename} for child loggers"""
+        logger_name = self.service
+        if self.child:
+            logger_name = f"{self.service}.{self._get_caller_filename()}"
+
+        return logging.getLogger(logger_name)
+
+    def _get_caller_filename(self):
+        """ Return caller filename by finding the caller frame """
+        # Current frame         => _get_logger()
+        # Previous frame        => logger.py
+        # Before previous frame => Caller
+        frame = inspect.currentframe()
+        caller_frame = frame.f_back.f_back.f_back
+        filename = caller_frame.f_globals["__name__"]
+
+        return filename
+
     def _init_logger(self, **kwargs):
         """Configures new logger"""
-        # Ensure logger children remains independent
-        self._logger.propagate = False
 
-        # Skip configuration if it's a child logger, or a handler is already present
-        # to prevent multiple Loggers with the same name or its children having different sampling mechanisms
+        # Skip configuration if it's a child logger to prevent
+        # multiple handlers being attached as well as different sampling mechanisms
         # and multiple messages from being logged as handlers can be duplicated
-        if self._logger.parent.name == "root" and not self._logger.handlers:
+        if not self.child:
             self._configure_sampling()
             self._logger.setLevel(self.log_level)
             self._logger.addHandler(self._handler)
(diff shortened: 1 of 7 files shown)


aws_lambda_powertools/logging/logger.py
```diff
@@ -112,19 +110,48 @@ def __init__(
         self,
         service: str = None,
         level: Union[str, int] = None,
+        name: str = None,
         sampling_rate: float = None,
         stream: sys.stdout = None,
         **kwargs,
     ):
         self.service = service or os.getenv("POWERTOOLS_SERVICE_NAME") or "service_undefined"
+        self.name = name or self.service
```
nmoutschen, line 119: I'd have a slight preference to do something like `${name}.${self.service}`, seems easier from an experience point of view to me.

aws_lambda_powertools/logging/logger.py
```diff
@@ -89,18 +90,15 @@ class Logger(logging.Logger):
                 logger.structure_logs(append=True, payment_id=event["payment_id"])
                 logger.info("Hello")
 
-    Parameters
-    ----------
-    logging : logging.Logger
-        Inherits Logger
```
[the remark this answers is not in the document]
heitorlessa: No, because I'm proxying all not found attributes directly to the inner logger. It's the same effect of sub-classing for advanced customers who were making changes directly to the logger afterwards , except done right now

tests/functional/utility_functions.py
```diff
@@ -0,0 +1,8 @@
+import secrets
+
+from aws_lambda_powertools import Logger
+
+
+def get_random_logger(**kwargs):
+    """Return vanilla logger w/ random name"""
+    return Logger(name=secrets.token_urlsafe(), **kwargs)
```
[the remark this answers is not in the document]
heitorlessa: Good call - I accidentally moved all tests to fetch a random Logger to prevent errors/side effects. I'll create one

aws_lambda_powertools/logging/logger.py
heitorlessa: The `dot` represents inheritance here, and there could be a limit on how many chars you could have in a Logger at some point, hence why either name or service. I actually had it like that but saw this quickly growing when tried something like `name=__name__` in a module

aws_lambda_powertools/logging/logger.py
```diff
@@ -55,6 +54,8 @@ class Logger(logging.Logger):
         service name to be appended in logs, by default "service_undefined"
     level : str, optional
         logging.level, by default "INFO"
+    name: str
+        Logger name, "{service}" by default
```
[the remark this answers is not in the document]
heitorlessa: Nope, not required. Just changed the docstring to reflect that - good catch.

aws_lambda_powertools/logging/logger.py
```diff
@@ -112,19 +112,65 @@ def __init__(
...
+
+        self._init_logger(**kwargs)
+
+    def __getattr__(self, name):
+        # Proxy attributes not found to actual logger to support backward compatibility
+        # https://github.com/awslabs/aws-lambda-powertools-python/issues/97
+        return getattr(self._logger, name)
+
+    def _get_logger(self):
+        """ Returns a Logger named {self.service}, or {service.filename} for child loggers"""
```
alexanderluiscampino, line 136: for consistency: `or {self.service.filename}`

aws_lambda_powertools/logging/logger.py
```diff
@@ -118,7 +118,7 @@ def __init__(
     ):
         self.service = service or os.getenv("POWERTOOLS_SERVICE_NAME") or "service_undefined"
         self.sampling_rate = sampling_rate or os.getenv("POWERTOOLS_LOGGER_SAMPLE_RATE") or 0.0
-        self.log_level = level or os.getenv("LOG_LEVEL") or logging.INFO
+        self.log_level = level or os.getenv("LOG_LEVEL".upper()) or logging.INFO
```
alexanderluiscampino, line 122: Not sure what is the purpose of  `.upper()` in here since `"LOG_LEVEL"` is explicitly set as all uppercase

heitorlessa: `LOG_LEVEL` is the env variable, and one particular customer ended up using "info" as the value for the env var - This solves that. 

In fact, I'm gonna turn this into a private method to cover the explicit `level` too

alexanderluiscampino: Gotcha, your explanation makes sense, then you need the .upper outside of the getenv(LOG_LEVEL).UPPER()

heitorlessa:
Just addressed @alexanderluiscampino commentson consistency, and took the opportunity to turn`log level` input to case insensitive to prevent common mistakes for newcomers.

I'll start working on the docs this afternoon or tomorrow, and should release a minor version by EOW

heitorlessa:
Yep it’s fixed in the latest commit ;)

Issue #92, opened 2020-07-20
RFC: SQS partial batch failure middleware


gmcrocetti:
First things first: Congratulations for the amazing repo.

## Key information

* RFC PR: (leave this empty)
* Related issue(s), if known:
* Area: Utilities
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-python/#tenets): Yes

## Summary
[summary]: #summary
A lambda processing a batch of messages from SQS is a very common approach and it works smooth for most use cases. Now, for the sake of example, suppose we're processing a batch and one of the messages failed, lambda is going to redrive this batch to the queue again, including the successful ones ! Re-running successful messages is not acceptable for all use cases.

## Motivation
[motivation]: #motivation

A very common execution pattern is running a lambda connected in sqs, in most cases with a batch size not equal to one. In such cases, an error to one of the processed messages will cause the whole batch to return to the queue. For some use cases, it's impossible to rely on such behavior - non idempotent actions. A solution for this problem would improve the experience of using a lambda with SQS.

## Proposal
[proposal]: #proposal
I'm going to propose a very simple code that's not complete.

```python
from aws_lambda_powertools.middleware_factory import lambda_handler_decorator

@lambda_handler_decorator
def sqs_partial_batch_failure(handler, event, context):
	sent_records = event['Records']
	sqs_client = boto3.client('sqs')

	response = handler(event, context)

	successful_messages = get_successful(response)
	sqs_client.delete_message_batch(successful_messages)  # deletes 3rd and 7th messages


# batchsize of 10, fails for 3rd and 7th message
@sqs_partial_batch_failure
def handler(event, context):
	for record in event['Records']:
		do_sth(record)
```

## Drawbacks
- Add boto3 as dependency;
- It may be unrelated with the proposal of this package;
- Adds a little performance overhead to track failed records and delete the successful ones.

## Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

* **What other designs have been considered? Why not them?**
The lambda may be invoked with a single message (batch size one) but it just feels like ill-use of the full power of this integration. Running as much as possible message per lambda call is the best scenario.
Inspired by (middy)[https://github.com/middyjs/middy/tree/master/packages]

* **What is the impact of not doing this?**
It may attract more users to use such a powerful option: batch processing with sqs.

## Unresolved questions
[unresolved-questions]: #unresolved-questions

> Optional, stash area for topics that need further development e.g. TBD

[...]

PR #100, opened 2020-08-13
feat: SQS Partial failure

gmcrocetti:
#92 : SQS Partial batch

## Description of changes:

Hi everyone ! Here's a **very initial** design proposal to work with batches. This draft intends to work as our communication channel to hear your toughts. Before putting more effort in everything that's missing (docs, tests, typing, etc)  I look forward hearing opinions from issue's participants: @heitorlessa , @nmoutschen , @Nr18 .

**Checklist**

 

* [-] Update tests
* [-] Update docs

## Breaking change checklist

### DIFF
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -27,7 +27,7 @@ class BaseProcessor(ABC):
         raise NotImplementedError()
 
     @abstractmethod
-    def _process_record(self):
+    def _process_record(self, record):
         raise NotImplementedError()
 
     def process(self) -> List[Tuple]:
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -72,5 +72,5 @@ class BasePartialProcessor(BaseProcessor):
         Failure callback
         """
         entry = ("fail", exception.args, record)
-        self.fail_messages.append(entry)
+        self.fail_messages.append(record)
         return entry
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -9,9 +9,6 @@ from typing import Any, Callable, Iterable, List, MutableSequence, Tuple
 
 
 class BaseProcessor(ABC):
-
-    # init with lambda's context ?
-
     @abstractmethod
     def _prepare(self):
         """
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -5,13 +5,17 @@ Batch processing utilities
 """
 
 from abc import ABC, abstractmethod
-from typing import Any, Callable, Iterable, List, MutableSequence, Tuple
+from typing import Any, Callable, Iterable, List, Tuple
 
 
-class BaseProcessor(ABC):
+class BasePartialProcessor(ABC):
     """
     Abstract class for batch processors.
     """
+
+    def __init__(self):
+        self.success_messages: List = []
+        self.fail_messages: List = []
 
     @abstractmethod
     def _prepare(self):
@@ -62,12 +66,6 @@ class BaseProcessor(ABC):
         self.handler = handler
         return self
 
-
-class BasePartialProcessor(BaseProcessor):
-
-    success_messages: MutableSequence = None
-    fail_messages: MutableSequence = None
-
     def success_handler(self, record: Any, result: Any):
         """
         Success callback
@@ -75,7 +73,7 @@ class BasePartialProcessor(BaseProcessor):
         Returns
         -------
         tuple
-            "success", record processing result and original record
+            "success", result, original record
         """
         entry = ("success", result, record)
         self.success_messages.append(record)
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -63,14 +63,14 @@ class BasePartialProcessor(BaseProcessor):
         """
         Success callback
         """
-        entry = (result, record)
+        entry = ("success", result, record)
         self.success_messages.append(record)
         return entry
 
-    def failure_handler(self, record, error):
+    def failure_handler(self, record, exception):
         """
         Failure callback
         """
-        entry = (error, record)
+        entry = ("fail", exception.args, record)
         self.fail_messages.append(entry)
         return entry
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -1,0 +1,79 @@
+# -*- coding: utf-8 -*-
+
+"""
+Batch processing utilities
+"""
+
+from abc import ABC, abstractmethod
+from typing import Any, Callable, Iterable, List, MutableSequence, Tuple
+
+
+class BaseProcessor(ABC):
+
+    # init with lambda's context ?
+
+    @abstractmethod
+    def _prepare(self):
+        """
+        Prepare context manager.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _clean(self):
+        """
+        Clear context manager.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def _process_record(self):
+        raise NotImplementedError()
+
+    def process(self) -> List[Tuple]:
+        return [self._process_record(record) for record in self.records]
+
+    def __enter__(self):
+        self._prepare()
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        self._clean()
+
+    def __call__(self, records: Iterable[Any], handler: Callable):
+        """
+        Set instance attributes before execution
+
+        Parameters
+        ----------
+
+        records: Iterable[Any]
+            Iterable with objects to be processed.
+        handler: Callable
+            Callable to process "records" entries.
+        """
+        self.records = records
+        self.handler = handler
+        return self
+
+
+class BasePartialProcessor(BaseProcessor):
+
+    success_messages: MutableSequence = None
+    fail_messages: MutableSequence = None
+
+    def success_handler(self, record, result):
+        """
+        Success callback
+        """
+        entry = (result, record)
+        self.success_messages.append(record)
+        return entry
+
+    def failure_handler(self, record, error):
+        """
+        Failure callback
+        """
+        entry = (error, record)
+        self.fail_messages.append(entry)
+        return entry
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -9,6 +9,10 @@ from typing import Any, Callable, Iterable, List, MutableSequence, Tuple
 
 
 class BaseProcessor(ABC):
+    """
+    Abstract class for batch processors.
+    """
+
     @abstractmethod
     def _prepare(self):
         """
@@ -24,10 +28,16 @@ class BaseProcessor(ABC):
         raise NotImplementedError()
 
     @abstractmethod
-    def _process_record(self, record):
+    def _process_record(self, record: Any):
+        """
+        Process record with handler.
+        """
         raise NotImplementedError()
 
     def process(self) -> List[Tuple]:
+        """
+        Call instance's handler for each record.
+        """
         return [self._process_record(record) for record in self.records]
 
     def __enter__(self):
@@ -43,7 +53,6 @@ class BaseProcessor(ABC):
 
         Parameters
         ----------
-
         records: Iterable[Any]
             Iterable with objects to be processed.
         handler: Callable
@@ -59,7 +68,7 @@ class BasePartialProcessor(BaseProcessor):
     success_messages: MutableSequence = None
     fail_messages: MutableSequence = None
 
-    def success_handler(self, record, result):
+    def success_handler(self, record: Any, result: Any):
         """
         Success callback
         """
@@ -67,7 +76,7 @@ class BasePartialProcessor(BaseProcessor):
         self.success_messages.append(record)
         return entry
 
-    def failure_handler(self, record, exception):
+    def failure_handler(self, record: Any, exception: Exception):
         """
         Failure callback
         """
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -71,6 +71,11 @@ class BasePartialProcessor(BaseProcessor):
     def success_handler(self, record: Any, result: Any):
         """
         Success callback
+
+        Returns
+        -------
+        tuple
+            "success", record processing result and original record
         """
         entry = ("success", result, record)
         self.success_messages.append(record)
@@ -79,6 +84,11 @@ class BasePartialProcessor(BaseProcessor):
     def failure_handler(self, record: Any, exception: Exception):
         """
         Failure callback
+
+        Returns
+        -------
+        tuple
+            "fail", exceptions args, original record
         """
         entry = ("fail", exception.args, record)
         self.fail_messages.append(record)
--- a/aws_lambda_powertools/utilities/batch/middlewares.py
+++ b/aws_lambda_powertools/utilities/batch/middlewares.py
@@ -1,0 +1,29 @@
+# -*- coding: utf-8 -*-
+
+"""
+Middlewares for batch utilities
+"""
+from typing import Callable, Dict
+
+from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
+
+from .sqs import PartialSQSProcessor
+
+
+@lambda_handler_decorator
+def partial_sqs_processor(
+    handler: Callable, event: Dict, context: Dict, record_handler: Callable, processor: PartialSQSProcessor = None
+):
+    """
+
+    Examples
+    --------
+
+    """
+    records = event["Records"]
+    processor = processor or PartialSQSProcessor()
+
+    with processor(records, record_handler) as ctx:
+        ctx.process()
+
+    return handler(event, context)
--- a/aws_lambda_powertools/utilities/batch/middlewares.py
+++ b/aws_lambda_powertools/utilities/batch/middlewares.py
@@ -38,7 +38,7 @@ def batch_processor(
         >>> def record_handler(record):
         >>>     return record["body"]
         >>>
-        >>> @batch_processor(record_handler=record_handler, processor=PartialSQSProcessor)
+        >>> @batch_processor(record_handler=record_handler, processor=PartialSQSProcessor())
         >>> def handler(event, context):
         >>>     return {"StatusCode": 200}
 
--- a/aws_lambda_powertools/utilities/batch/middlewares.py
+++ b/aws_lambda_powertools/utilities/batch/middlewares.py
@@ -3,6 +3,7 @@
 """
 Middlewares for batch utilities
 """
+
 from typing import Callable, Dict
 
 from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
@@ -49,7 +50,7 @@ def batch_processor(
     """
     records = event["Records"]
 
-    with processor(records, record_handler) as ctx:
-        ctx.process()
+    with processor(records, record_handler):
+        processor.process()
 
     return handler(event, context)
--- a/aws_lambda_powertools/utilities/batch/middlewares.py
+++ b/aws_lambda_powertools/utilities/batch/middlewares.py
@@ -8,12 +8,12 @@ from typing import Callable, Dict
 
 from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
 
-from .base import BaseProcessor
+from .base import BasePartialProcessor
 
 
 @lambda_handler_decorator
 def batch_processor(
-    handler: Callable, event: Dict, context: Dict, record_handler: Callable, processor: BaseProcessor = None
+    handler: Callable, event: Dict, context: Dict, record_handler: Callable, processor: BasePartialProcessor = None
 ):
     """
     Middleware to handle batch event processing
--- a/aws_lambda_powertools/utilities/batch/middlewares.py
+++ b/aws_lambda_powertools/utilities/batch/middlewares.py
@@ -7,21 +7,47 @@ from typing import Callable, Dict
 
 from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
 
-from .sqs import PartialSQSProcessor
+from .base import BaseProcessor
 
 
 @lambda_handler_decorator
-def partial_sqs_processor(
-    handler: Callable, event: Dict, context: Dict, record_handler: Callable, processor: PartialSQSProcessor = None
+def batch_processor(
+    handler: Callable, event: Dict, context: Dict, record_handler: Callable, processor: BaseProcessor = None
 ):
     """
+    Middleware to handle batch event processing
+
+    Parameters
+    ----------
+    handler: Callable
+        Lambda's handler
+    event: Dict
+        Lambda's Event
+    context: Dict
+        Lambda's Context
+    record_handler: Callable
+        Callable to process each record from the batch
+    processor: PartialSQSProcessor
+        Batch Processor to handle partial failure cases
 
     Examples
     --------
+    **Processes Lambda's event with PartialSQSProcessor**
+        >>> from aws_lambda_powertools.utilities.batch import batch_processor
+        >>>
+        >>> def record_handler(record):
+        >>>     return record["body"]
+        >>>
+        >>> @batch_processor(record_handler=record_handler, processor=PartialSQSProcessor)
+        >>> def handler(event, context):
+        >>>     return {"StatusCode": 200}
+
+    Limitations
+    -----------
+    * Async batch processors
 
     """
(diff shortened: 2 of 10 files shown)


heitorlessa:
@cakepietoast could you help review when you get a chance :)?

I'll finish mine tomorrow, and have bandwidth next week if you can't

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -0,0 +1,66 @@
...
+    def _prepare(self):
+        """
+        """
+        self.success_messages.clear()
+        self.fail_messages.clear()
+
+    def _clean(self):
+        """
+        """
+        if not self.fail_messages:
```
cakepietoast, line 47: We can expand on this check: I don't think we need to delete messages from SQS if we have ONLY success (no fail_messages) or ONLY failures (no success_messages)

```suggestion
        if not self.fail_messages or not self.success_messages:
```

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -0,0 +1,66 @@
...
+            return
+
+        queue_url = self.get_queue_url()
+        entries_to_remove = self.get_entries_to_clean()
+
+        return self._client.delete_message_batch(QueueUrl=queue_url, Entries=entries_to_remove)
+
+
+class DefaultPartialSQSProcessor(BasePartialSQSProcessor):
```
cakepietoast, line 56: I think we can simplify the implementation by just having a single class here. Perhaps just rename `BasePartialSQSProcessor` to `DefaultPartialSQSProcessor` instead of having the extra class just to change the name.

gmcrocetti: Yeah, see your point but I think the prefix `Base` is more appealing from a dev point of view. "It just" makes sense to extend a class with such name, IMO, more than something named `DefaultPartial`... I'll stick with `PartialSQSProcessor`, alright ?

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -0,0 +1,66 @@
...
+    pass
+
+
+@lambda_handler_decorator
+def partial_sqs_processor(handler, event, context, record_handler, processor=None):
+    records = event["Records"]
+    processor = processor or DefaultPartialSQSProcessor()
+
+    with processor(records, record_handler) as ctx:
+        ctx.process()
```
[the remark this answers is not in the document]
gmcrocetti: That's a very good point, let me try to explain. I'm not calling `handler` directly because this decorator must receive a `record_handler`, a Callable which knows how to handle a single record, such that your implementation would look something like that:
```python
def pass_through_handler(record):
   return record

@lambda_handler_decorator(pass_through_handler)
def handler(event, context):
  pass
```
The computation will be done by this `record_handler`, not the original handler. 

BTW you may have noticed a comment in `BaseProcessor`. I've implemented `__call__` to allow multiple runs of the same processor but with different data or handler. One missing param is`context`, do you see any use case where user may change it during execution ?

gmcrocetti:
> Thanks for the awesome contribution so far @gmcrocetti! Looks like its on the right track to me. I added a few inline comments, happy to take another look when we have tests/docs for the functionality.

Thx for your fast review. I've left some replies :).

Nr18:
Currently on vacation I will have a more in-depth look next week, but in general I am missing docs and tests. 

But I assume you concentrated on the implementation first?

aws_lambda_powertools/utilities/batch/sqs.py
cakepietoast: > there's no need to clean because there was no failure

There's also no need to clean if there was no success - we only need to handle the "special case" which is partial failure. Does that make sense?

aws_lambda_powertools/utilities/batch/sqs.py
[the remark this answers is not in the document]
cakepietoast: The `@lambda_handler_decorator` middleware allows you to create a decorator which in turn can be used to decorate a lambda handler function. As such, you're expected to call the lambda handler - I think I understand your intent, but with the implementation as it stands right now, the lambda handler won't actually be called at all. To use your example:

```
def pass_through_handler(record):
   return record  # this will be executed for each SQS record

@lambda_handler_decorator(pass_through_handler)
def handler(event, context):
  other_logic()  # Not executed
  return {"statusCode": 200, "result": "whatever"}  # Not executed
```

With regards to the lambda context, I can't think of a scenario where the record processor would need knowledge of it so I think its fine to leave it out.

aws_lambda_powertools/utilities/batch/sqs.py
cakepietoast: I think this one might be a matter of taste, I just felt like the class hierarchy was a bit more complex than necessary. `PartialSQSProcessor` sounds good to me!

aws_lambda_powertools/utilities/batch/sqs.py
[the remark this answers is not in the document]
gmcrocetti: Nice catch !! I'll call the handler

aws_lambda_powertools/utilities/batch/sqs.py
gmcrocetti: Got it ! Nice point, I'll add this check.

gmcrocetti:
> hey @gmcrocetti don't worry about the docs, any basic text for us to understand the UX would be helpful - If you allow us to make commits directly to you fork, we can help you craft the docs ;). Implementation and tests are the most important
> 
> This is the first utility we launched yesterday ICYMI: https://awslabs.github.io/aws-lambda-powertools-python/utilities/parameters/

Hi @heitorlessa , sorry to take so long, it was quite a busy week. I'm hoping to finish tests and docs this week (tomorrow in fact), is it possible to follow this schedule :) ?

I'll push my progress ASAP so you can follow.

heitorlessa:
Absolutely - There’s no rush ;)

I’m not sure if it was discussed in the review yet, but I don’t think
DynamoDB and Kinesis can have their records deleted in this manner, so this
would only work for SQS.

Super looking forward to reviewing it, this will be awesome

gmcrocetti:
Yeap, you're totally right ! It wasn't touched by no means but I think we must at least discuss this point.

How can we proceed if we'll not be able to delete these failed records for both cases ? Do you have any idea ?

Even we don't tackle this right now, it's a good point because we may avoid design flaws.

heitorlessa:
My suggestion would be to focus on SQS only, thus simplifying the
implementation - I'm happy to write a recommendation for customers using
Kinesis and DynamoDB streams under a Tips section in your docs. The new
async controls for stream processing in Lambda allow them to prevent poison
pills, and also be more flexible on not allowing a retry, or simply
bisecting a failed record.

I'll also write that later in the year when we refresh Serverless best
practices given the new controls that are now in place. SQS is the
exception as you can't control that, so you need something like your
implementation if you don't want to implement idempotency downstream and
onwards (harder).

Does that make sense?

gmcrocetti:
Yes, makes a lot of sense. In fact, that's why I was working solely with the sqs processor. These async controllers are pretty awesome, and even smth "simple" like "destination on failure" helps a lot. A "tips" section would be great !
Thanks

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -0,0 +1,77 @@
...
+class PartialSQSProcessor(BasePartialProcessor):
+    def __init__(self, config: Optional[Config] = None):
+        config = config or Config()
+        self.client = boto3.client("sqs", config=config)
+        self.success_messages: List = []
+        self.fail_messages: List = []
+
+        super().__init__()
+
+    def get_queue_url(self):
```
Nr18, line 26:
```suggestion
    def _get_queue_url(self) -> str:
```

Other than a unit test there is no need to expose this method right?

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -0,0 +1,77 @@
...
+        """
+        Format QueueUrl from first records entry
+        """
+        if not getattr(self, "records", None):
+            return
+
+        *_, account_id, queue_name = self.records[0]["eventSourceARN"].split(":")
+        return f"{self.client._endpoint.host}/{account_id}/{queue_name}"
+
+    def get_entries_to_clean(self):
```
Nr18, line 36:
```suggestion
    def _get_entries_to_clean(self) -> List:
```

Other than a unit test there is no need to expose this method right?

aws_lambda_powertools/utilities/batch/sqs.py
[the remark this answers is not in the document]
Nr18: So looking at this we now have a decorator that will be invoked and process all received records with the given `record_handler` method the return value is captured by the `PartialSQSProcessor` class but not passed to the original handler.

In other words, it does not make sense to store the results in Lists if we are not providing the results to a consumer. So either we inject the results of the processed messages in the event before calling the handler or we can simplify the logic for failures to a boolean flag.

I think from a flexibility point of view you want to enrich the event? Thoughts anyone?

Nr18 (commented):
@gmcrocetti nice implementation I left some questions/remarks hopefully you will find them helpful.

aws_lambda_powertools/utilities/batch/sqs.py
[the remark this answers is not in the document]
gmcrocetti: Yeap ! It works like a charm if we run as a context. I was expecting someone to raise this discussion, thx ! 

My initial tought was to enrich the event, but I'm not a huge fan of this approach since we're going to change the interface of the integration. I'm more inclined to follow a "kwargs" approach.

Opnions would be great !

aws_lambda_powertools/utilities/batch/sqs.py
gmcrocetti: Absolutely. They were private methods in my original approach, changed for some reason but forgot to go back, thx.

About the annotation, I'm about to review everything to write docs.

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -0,0 +1,76 @@
...
+        entry = ("success", result, record)
+        self.success_messages.append(record)
+        return entry
+
+    def failure_handler(self, record, exception):
+        """
+        Failure callback
+        """
+        entry = ("fail", exception.args, record)
+        self.fail_messages.append(entry)
```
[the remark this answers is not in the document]
gmcrocetti: In fact it's wrong, it should append the record for internal usage. `entry` should be used by the caller

aws_lambda_powertools/utilities/batch/sqs.py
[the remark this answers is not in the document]
Nr18: So how would the `kwargs` approach look like, something like this then?

```pyhton
def pass_through_handler(record):
   return {"Foo": "Bar"}

@lambda_handler_decorator(pass_through_handler)
def handler(event, context, **kwargs):
  assert kwargs["record_results"][0] == {"Foo": "Bar"}
```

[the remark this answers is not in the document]
gmcrocetti: @heitorlessa , @cakepietoast. Can you please take a look ?

gmcrocetti:
@heitorlessa , finally did it ! Phewwww !! Waiting reviews. Thx

aws_lambda_powertools/utilities/batch/middlewares.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    Middleware to handle batch event processing
+
+    Parameters
+    ----------
+    handler: Callable
+        Lambda's handler
+    event: Dict
+        Lambda's Event
+    context: Dict
```
Nr18, line 26: Theoretically it's not a dict it's has methods, I always use:

```python
from aws_lambda_context import LambdaContext

def handler(event: Dict, context: LambdaContext):
    pass
```

@heitorlessa I think it would be a good addition to provide a typing for the context parameter. 

Surely not in scope for this PR but I can write a RFC if you think it's useful?

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+import json
+
+from aws_lambda_powertools.utilities.batch import BaseProcessor, batch_processor
+import boto3
+
+class DynamoDBProcessor(BaseProcessor):
+
+    def __init__(self, queue_url: str):
+        self.queue_url = queue_url
+        self.client = boto3.client("sqs")
```
Nr18, line 63: I think you copied the wrong class the sample is about dynamodb this is sqs 😉

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+Here is an example implementation of a DynamoDBStream custom processor:
+
+```python:title=custom_processor.py
+import json
+
+from aws_lambda_powertools.utilities.batch import BaseProcessor, batch_processor
+import boto3
+
+class DynamoDBProcessor(BaseProcessor):
```
Nr18, line 59:
```suggestion
class DynamoDBStreamProcessor(BaseProcessor):
```

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+# As context
+
+processor = DynamoDBProcessor("dummy")
+records = {"Records": []}
+
+with processor(records=records, handler=record_handler) as ctx:
+    result = ctx.process()
+
+# As middleware
+@batch.batch_processor(record_handler=record_handler, processor=DynamoDBProcessor("dummy"))
```
Nr18, line 92:
```suggestion
@batch.batch_processor(record_handler=record_handler, processor=DynamoDBStreamProcessor("dummy"))
```

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+        body = json.dumps(result)
+        self.client.send_message(QueueUrl=self.queue_url, MessageBody=body)
+        return result
+
+def record_handler(record):
+    return record["Keys"]
+
+# As context
+
+processor = DynamoDBProcessor("dummy")
```
Nr18, line 85:
```suggestion
processor = DynamoDBStreamProcessor("dummy")
```

Nr18, line 85:
```suggestion
processor = DynamoDBStreamProcessor("dummy")
```

Nr18 (commented):
@gmcrocetti nice work! Coming along nicely did spot a naming improvement and I believe a copy paste error

aws_lambda_powertools/utilities/batch/middlewares.py
heitorlessa: definitely @Nr18 - Here's a good start as it's 90% done: https://gist.github.com/alexcasalboni/a545b68ee164b165a74a20a5fee9d133

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+# highlight-start
+@batch_processor(record_handler=record_handler, processor=PartialSQSProcessor())
+# highlight-end
+def lambda_handler(event, context):
+    return {"statusCode": 200}
+```
+
+## Create your own processor
+
+You can create your own batch processor by inheriting the `BaseProcessor` class, and implementing `_prepare()`, `_clean` and `_process_record()`.
```
heitorlessa, line 48: could you please expand what these methods should be responsible for w/ an example?

Feel free to create a sub-heading `###` for each so you have more room to explain these

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+import Note from "../../src/components/Note"
+
+The batch utility provides an abstraction to process a batch event. Useful for lambda integrations with [AWS SQS](https://aws.amazon.com/sqs/), [AWS Kinesis](https://aws.amazon.com/kinesis/) and [AWS DynamoDB Streams](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html).
+It also provides base classes (`BaseProcessor`, `BasePartialProcessor`) allowing you to create your **own** batch processor.
+
+**Key Features**
+
+* Run batch processing logic with a clean interface;
+* Middleware and context to handle a batch event;
+* Removal of successful messages for [AWS SQS](https://aws.amazon.com/sqs/) batch - in case of partial failure.
```
heitorlessa, line 15: This is the selling point, it should come first.

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+---
+
+import Note from "../../src/components/Note"
+
+The batch utility provides an abstraction to process a batch event. Useful for lambda integrations with [AWS SQS](https://aws.amazon.com/sqs/), [AWS Kinesis](https://aws.amazon.com/kinesis/) and [AWS DynamoDB Streams](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html).
+It also provides base classes (`BaseProcessor`, `BasePartialProcessor`) allowing you to create your **own** batch processor.
+
+**Key Features**
+
+* Run batch processing logic with a clean interface;
```
heitorlessa, line 13: Expand this - I think what you meant is:

"Bring your own batch processor with <some included features/ctx manager/failure/success etc>"

aws_lambda_powertools/utilities/batch/middlewares.py
```diff
@@ -0,0 +1,55 @@
...
+        >>>     return {"StatusCode": 200}
+
+    Limitations
+    -----------
+    * Async batch processors
+
+    """
+    records = event["Records"]
+
+    with processor(records, record_handler) as ctx:
```
heitorlessa, line 52: it could be, but unsure what you meant by `ctx` here - u mean `as _processor.. _processor.process()`?

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+**IAM Permissions**
+
+This utility requires additional permissions to work as expected. See the following table:
+
+Processor | Function/Method | IAM Permission
+|---------|-----------------|---------------|
+PartialSQSProcessor | `_clean` | `sqs:DeleteMessageBatch`
+
+### PartialSQSProcessor
+
```
heitorlessa, line 26: add the problem statement - Why would I want to use this over native SQS capabilities?

I can make a suggestion if you like

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
...
+import Note from "../../src/components/Note"
+
+The batch utility provides an abstraction to process a batch event. Useful for lambda integrations with [AWS SQS](https://aws.amazon.com/sqs/), [AWS Kinesis](https://aws.amazon.com/kinesis/) and [AWS DynamoDB Streams](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html).
+It also provides base classes (`BaseProcessor`, `BasePartialProcessor`) allowing you to create your **own** batch processor.
+
+**Key Features**
+
+* Run batch processing logic with a clean interface;
+* Middleware and context to handle a batch event;
```
heitorlessa, line 14: What do you mean by "context" here?

heitorlessa (changes requested):
This is a huge improvement from the initial version - **Thank you for all effort you've put in already**

Made some comments to help clarify a few areas - I'm happy to contribute some docs later if you don't have the bandwidth, but the implementation itself looks solid. 

I might make a comment on `middlewares.py` to reorganize that into a `processor.py`/something else I can't think of yet, but we can contribute to that later too, don't worry.

aws_lambda_powertools/utilities/batch/middlewares.py
Nr18: Ah thanks, I created #144 and will have a closer look at the gist you provided later on.

docs/content/utilities/batch.mdx
```diff
@@ -0,0 +1,95 @@
+---
+title: Batch
+description: Utility
+---
+
+import Note from "../../src/components/Note"
+
+The batch utility provides an abstraction to process a batch event. Useful for lambda integrations with [AWS SQS](https://aws.amazon.com/sqs/), [AWS Kinesis](https://aws.amazon.com/kinesis/) and [AWS DynamoDB Streams](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Streams.html).
```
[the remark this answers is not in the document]
gmcrocetti: Got your point, let me try to explain mine. As you said in one of the comments, the selling point is definitely "partial batch processing" but *what if* we also provided some a default batch processor (without the "partial behavior") ? We would be able to offer an interface for anyone doing batch processing. Personally I'm always doing code similar to this:
```python
def handle_record(record):
    return record

def handler(event, context):
   records = event["Records"]
   for record in records:
      do_something()
   return True
```
*If* we follow the suggest approach, we'll be able to provide a standard not only from sqs, but all current/future providers.
```python
from aws_lambda_powertools.utilities.batch import BatchProcessor
# BatchProcessor doesn't exist, we may be able to create now or in a next PR

def handle_record(record)
   return record

@batch_processor(handle_record, BatchProcessor())
def handler(event, context):
    return True
```
Do you think is a good idea ? Not for now, of course :).

docs/content/utilities/batch.mdx
gmcrocetti: Erased.

docs/content/utilities/batch.mdx
gmcrocetti: Improved, can you take a look ?

aws_lambda_powertools/utilities/batch/middlewares.py
gmcrocetti: Cleaned

docs/content/utilities/batch.mdx
gmcrocetti: Nope ! Improved the example with some comments. Can you take a look again ? Case it's unclear I'll switch the example.

docs/content/utilities/batch.mdx
gmcrocetti: Didn't create sub-heads, improved the example instead.

docs/content/utilities/batch.mdx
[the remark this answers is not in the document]
Nr18: I like the idea, a sample that would benefit from this would be an AWS Lambda function that handles SNS for example. But as long as that functionality is not in the codebase I have to agree with @heitorlessa, currently we only do partial SQS processing so the documentation should reflect that.

docs/content/utilities/batch.mdx
Nr18: Right, now I get the intention would `DDBStreamToSQSProcessor` be a better name then? Or something among those lines `DDBStreamProcessor` is very generic and the sample has some specific SQS behaviour in there.

heitorlessa: Thanks for working on the docs @gmcrocetti, much appreciated ;) It's clearer now, and it helped me confirm my initial concerns with it:

Shouldn't this utility be primarily focused on SQS Partial batch failures? 

I understand the appeal for having an interface to handle batch processing, though it makes more sense in non-Lambda function's environment. If customers have to implement their own batch processors, then this means we (as a Provider) should step back and work on better constructs to allow customers to have less code since code is liability - For example, DynamoDB, Kinesis, S3 Batch Operations all benefit from async controls, max retries, etc. requiring zero code to be maintained.

Right now, my vote would be for:

* [ ] Have only `BasePartialProcessor` so that customers could opt-in for your great SQS partial implementation, or create their own
* [ ] Remove any non-partial, non-SQS code unless we have a demand for it (80% tenet)
* [ ] Reword the custom processor doc section to focus on a custom SQS partial failure processor, where a customer could add additional logging or tracing annotations, etc.

cc @cakepietoast for a second opinion as I'd really like to include this in our next release this week, but am conflicted on the value of creating a custom processor for non-partial, non-SQS processing vs maintaining it.

Once again @gmcrocetti, I **really appreciate all your efforts, and sorry this is taking longer than you might have expected** - We want to make sure we strike a balance on customers knowing how easy it is to see the value prop of your utility and use it vs maintenance and customer demand.

cakepietoast: @heitorlessa I agree. This is an awesome PR with "just" the functionality to gracefully handle partial failures in SQS. I think its best that we focus on the pain point that was raised in the RFC, which is very specific to SQS. I'm open to implementing other stream processing functionality in the future - perhaps as a separate utility - but we need to give that some more thought beforehand.

gmcrocetti: Hehehhe, thanks for worrying so much, @heitorlessa , no problem at all. Sent a new commit ! :)

Nr18 (approved):
LGTM

heitorlessa:
Thanks a lot @gmcrocetti for the changes - Much cleaner, and we appreciate very much all effort you've put on to this!

We're merging as-is. @cakepietoast will create another PR to clean up the docs for consistency, and will cut a **1.5.0** release on Friday that will contain this and [other enhancements we've made](https://github.com/awslabs/aws-lambda-powertools-python/issues?q=is%3Aissue+is%3Aopen+label%3Astatus%2Fstaged-next-release) ;)

heitorlessa (approved):

cakepietoast (approved):

gmcrocetti:
Happy to see it merged ! Thx for all support !

PR #105, opened 2020-08-14
fix: regression 104

heitorlessa:
**Issue #, if available:** #105 104

## Description of changes:

Corrects regression where `Logger` constructor wasn't taking `int` as a level.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -1,5 +1,5 @@
-import copy
 import functools
+import inspect
 import logging
 import os
 import random
@@ -34,7 +34,7 @@ def _is_cold_start() -> bool:
     return cold_start
 
 
-class Logger(logging.Logger):
+class Logger:
     """Creates and setups a logger to format statements in JSON.
 
     Includes service name and any additional key=value into logs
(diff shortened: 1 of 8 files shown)


aws_lambda_powertools/logging/logger.py
```diff
@@ -132,8 +132,11 @@ def __getattr__(self, name):
         # https://github.com/awslabs/aws-lambda-powertools-python/issues/97
         return getattr(self._logger, name)
 
-    def _get_log_level(self, level: str):
+    def _get_log_level(self, level: Union[str, int]) -> Union[str, int]:
         """ Returns preferred log level set by the customer in upper case """
+        if isinstance(level, int):
+            return level
+
         log_level: str = level or os.getenv("LOG_LEVEL")
```
zroger, line 140: The docstring on line 47 above declares the LOG_LEVEL environment variable as capable of accepting an integer. That probably needs to be handled here as well.

heitorlessa: handling that in another PR before publishing it, as the test didn't go through

Issue #112, opened 2020-08-16
@tracer.capture_lambda_handler does not support functions that return generator expressions


michaelbrewer:
I have an async lambda listening to S3 bucket notifications and i wanted to track the performance of downloading a large json file. Original this was a regular function that worked but later when refactoring it to yield the file pointer it failed when deployed. Nothing was found during unit testing.

**What were you trying to accomplish?**
Add a `@tracer.capture_method`  to a function that `yields` a result.

## Expected Behavior

Either fail the same way during testing when `POWERTOOLS_TRACE_DISABLED='true'` or support generator expression

## Current Behavior

Either results in a `RuntimeError("generator didn't yield")` or a `Task timed out `

## Possible Solution

Detect and support generator expression, or have a better emulation during testing which returns an error when a return
returns a generator expression.

## Steps to Reproduce (for bugs)

```python3
import contextlib
import json
import tempfile
from typing import IO

import boto3
from aws_lambda_powertools import Tracer

bucket_name = "some_bucket"
key = "file.json"
tracer = Tracer()
s3 = boto3.client("s3")


@contextlib.contextmanager
def yield_without_capture() -> str:
    yield "[]"


@tracer.capture_method
@contextlib.contextmanager
def yield_with_capture() -> str:
    yield "[]"


@tracer.capture_method
@contextlib.contextmanager
def boto_yield_with_capture() -> IO:
    with tempfile.TemporaryFile() as fp:
        s3.download_fileobj(bucket_name, key, fp)
        print("download complete")  # Never prints
        fp.seek(0)
        yield fp


@contextlib.contextmanager
def boto_yield_without_capture() -> IO:
    with tempfile.TemporaryFile() as fp:
        s3.download_fileobj(bucket_name, key, fp)
        fp.seek(0)
        print("download complete")
        yield fp


@tracer.capture_lambda_handler
def lambda_handler(event: dict, _):
    if event["case"] == "yield_without_capture":
        with yield_without_capture() as yielded_value:
            result = yielded_value
    if event["case"] == "yield_with_capture":
        with yield_with_capture() as yielded_value:
            result = yielded_value
    if event["case"] == "boto_yield_without_capture":
        with boto_yield_without_capture() as fp:
            result = fp.read().decode("UTF-8")
    if event["case"] == "boto_yield_with_capture":
        with boto_yield_with_capture() as fp:
            result = fp.read().decode("UTF-8")  # "Task timed out after 10.01 seconds"

[...]

PR #113, opened 2020-08-17
feat: add support for tracing of generators using capture_method decorator

cakepietoast:
**Issue #, if available:** #112 

## Description of changes:

Added checks to capture_method to test if decorated function is a generator (and/or a context manager) and handle those cases appropriately.

**Checklist**

 

* [X] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/tracing/tracer.py
+++ b/aws_lambda_powertools/tracing/tracer.py
@@ -281,7 +281,7 @@ class Tracer:
 
         return decorate
 
-    def capture_method(self, method: Callable = None):  # noqa: C901
+    def capture_method(self, method: Callable = None):
         """Decorator to create subsegment for arbitrary functions
 
         It also captures both response and exceptions as metadata
@@ -320,6 +320,39 @@ class Tracer:
             def lambda_handler(event: dict, context: Any) -> Dict:
                 booking_id = event.get("booking_id")
                 asyncio.run(confirm_booking(booking_id=booking_id))
+
+        **Custom generator function using capture_method decorator**
+
+            from aws_lambda_powertools import Tracer
+            tracer = Tracer(service="booking")
+
+            @tracer.capture_method
+            def bookings_generator(booking_id):
+                resp = call_to_booking_service()
+                yield resp[0]
+                yield resp[1]
+
+            def lambda_handler(event: dict, context: Any) -> Dict:
+                gen = bookings_generator(booking_id=booking_id)
+                result = list(gen)
+
+        **Custom generator context manager using capture_method decorator**
+
+            from aws_lambda_powertools import Tracer
+            tracer = Tracer(service="booking")
+
+            @tracer.capture_method
+            @contextlib.contextmanager
+            def booking_actions(booking_id):
+                resp = call_to_booking_service()
+                yield "example result"
+                cleanup_stuff()
+
+            def lambda_handler(event: dict, context: Any) -> Dict:
+                booking_id = event.get("booking_id")
+
+                with booking_actions(booking_id=booking_id) as booking:
+                    result = booking
 
         **Tracing nested async calls**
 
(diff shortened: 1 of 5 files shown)


cakepietoast:
This solves for the example reported in the issue, but I'd like to get some more eyes on it to make sure we aren't introducing unwanted side effects.

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -280,7 +281,7 @@ def decorate(event, context):
 
         return decorate
 
-    def capture_method(self, method: Callable = None):
+    def capture_method(self, method: Callable = None):  # noqa: C901
```
cakepietoast, line 284: I'd like to avoid suppressing the "too complex" warning from flake8, but I'm not sure how to refactor this - open to suggestions!

example/tests/test_handler.py
```diff
@@ -4,6 +4,14 @@
 
 import pytest
 
+from aws_lambda_powertools import Tracer
+
+
+@pytest.fixture(scope="function", autouse=True)
```
cakepietoast, line 10: In case its not obvious - I had to add this to avoid side effects from tests in other modules.

aws_lambda_powertools/tracing/tracer.py
heitorlessa: Now that we've grown to support generators, **what do you think of encapsulating the inner decorating logic into their own methods?**

That way we can have a cleaner `if/elif`, ease maintenance, and pass the linter -- There must be an even cleaner way if we think hard, but this is good enough (tm) for now IMHO.

```python
        if inspect.iscoroutinefunction(method):
            decorate = self._decorate_async_function(method=method)
        elif inspect.isgeneratorfunction(method):
            decorate = self._decorate_generator_function(method=method)
        elif hasattr(method, "__wrapped__") and inspect.isgeneratorfunction(method.__wrapped__):
            decorate = self._decorate_generator_function_with_context_manager(method=method)
        else:
            decorate = self._decorate_sync_function(method=method)
``` 

The private methods are basically a copy/paste of that logic we had:

```python
    def _decorate_async_function(self, method: Callable = None):
        method_name = f"{method.__name__}"

        @functools.wraps(method)
        async def decorate(*args, **kwargs):
            async with self.provider.in_subsegment_async(name=f"## {method_name}") as subsegment:
                try:
                    logger.debug(f"Calling method: {method_name}")
                    response = await method(*args, **kwargs)
                    self._add_response_as_metadata(function_name=method_name, data=response, subsegment=subsegment)
                except Exception as err:
                    logger.exception(f"Exception received from '{method_name}' method")
                    self._add_full_exception_as_metadata(function_name=method_name, error=err, subsegment=subsegment)
                    raise

                return response

        return decorate

    def _decorate_generator_function(self, method: Callable = None):
        method_name = f"{method.__name__}"

        @functools.wraps(method)
        def decorate(*args, **kwargs):
            with self.provider.in_subsegment(name=f"## {method_name}") as subsegment:
                try:
                    logger.debug(f"Calling method: {method_name}")
                    result = yield from method(*args, **kwargs)
                    self._add_response_as_metadata(function_name=method_name, data=result, subsegment=subsegment)
                except Exception as err:
                    logger.exception(f"Exception received from '{method_name}' method")
                    self._add_full_exception_as_metadata(function_name=method_name, error=err, subsegment=subsegment)
                    raise

                return result

        return decorate

    def _decorate_generator_function_with_context_manager(self, method: Callable = None):
        method_name = f"{method.__name__}"

        @functools.wraps(method)
        @contextlib.contextmanager
        def decorate(*args, **kwargs):
            with self.provider.in_subsegment(name=f"## {method_name}") as subsegment:
                try:
                    logger.debug(f"Calling method: {method_name}")
                    with method(*args, **kwargs) as return_val:
                        result = return_val
                    self._add_response_as_metadata(function_name=method_name, data=result, subsegment=subsegment)
                except Exception as err:
                    logger.exception(f"Exception received from '{method_name}' method")
                    self._add_full_exception_as_metadata(function_name=method_name, error=err, subsegment=subsegment)
                    raise

                yield result

        return decorate

    def _decorate_sync_function(self, method: Callable = None):
        method_name = f"{method.__name__}"

        @functools.wraps(method)
        def decorate(*args, **kwargs):
            with self.provider.in_subsegment(name=f"## {method_name}") as subsegment:
                try:
                    logger.debug(f"Calling method: {method_name}")
                    response = method(*args, **kwargs)
                    self._add_response_as_metadata(function_name=method_name, data=response, subsegment=subsegment)
                except Exception as err:
                    logger.exception(f"Exception received from '{method_name}' method")
                    self._add_full_exception_as_metadata(function_name=method_name, error=err, subsegment=subsegment)
                    raise

                return response

        return decorate

```

example/tests/test_handler.py
heitorlessa: yep :( I wish there was a cleaner way to keep config across instances and not requiring a fixture

aws_lambda_powertools/tracing/tracer.py
cakepietoast: Great idea, implemented as suggested

heitorlessa (approved):
Changes look great, and docs are clear to me :)

Thanks a lot Tom for pulling this together.

Issue #95, opened 2020-07-28
RFC: Validate incoming and outgoing events utility


heitorlessa:
## Key information

* RFC PR: (leave this empty)
* Related issue(s), if known: #30
* Area: Utilities
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-python/#tenets): Yes

## Summary
[summary]: #summary

This utility helps you validate incoming events from Lambda event sources as well as the response from your Lambda handler - All based on JSON Schemas. You can either validate an entire incoming event including the chosen event source wrapper, or only the event payload/body if you will.

## Motivation
[motivation]: #motivation

Well-Architected Serverless Lens recommends validating events under the Security pillar. As of now, customers have to implement their own validation mechanisms, bring additional dependencies, and end up crafting multiple JSON Schemas for popular event sources.

We could ease that by including in Powertools and lowering the bar for entry. 

## Proposal
[proposal]: #proposal

This utility would use the already present [Fast JSON Schema](https://github.com/horejsek/python-fastjsonschema/) lib to validate both incoming and outgoing events using a preferred JSON Schema.

Another plus is that we can use our [Middleware Factory](https://awslabs.github.io/aws-lambda-powertools-python/utilities/middleware_factory/) to make this easier to implement, and automatically trace it with X-Ray to profile performance impact when necessary.

**Validating both inbound/outbound**

```python
from aws_lambda_powertools.utilities import validator

@validator(inbound=inbound_schema_dict, outbound=outbound_schema_dict)
def lambda_handler(evt, ctx):
    ...
``` 

For customers wanting to validate only the payload of popular event sources, say API Gateway, this utility will work in tandem with an **extractor** utility - That will provide the following benefits:

* Validate only the actual event payload of a popular event source
  * e.g. validate API GW POST payload event (event['body']) itself not the whole API GW event 
* Craft your own JSON Schema to help you validate payload as well as things like headers/message attributes, etc.

> By default, envelopes will pluck only the payload of a message within the event. Allowing multiple paths can easily add complexity, so we will defer to customers creating their own envelopes if they want to.

**Validating inbound with built-in popular event source schemas**

```python
from aws_lambda_powertools.utilities import validator
from aws_lambda_powertools.utilities.extractor import envelopes

@validator(inbound=inbound_schema, envelope=envelopes.api_gateway_rest)
def lambda_handler(evt, ctx):
    ...
```

## Drawbacks
[drawbacks]: #drawbacks

[...]

PR #118, opened 2020-08-19
feat: Advanced parser utility (pydantic)

risenberg-cyberark:
**Issue, if available:** #147, #95

## Description of changes:

Added a new validation module. It has the validator decorator code with 3 envelopes (eventbridge, dynamoDB and custom user) and tests. Also has the eventbridge & dynamoDB schemas. 

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #95  :


risenberg-cyberark:
open issues:
1. update documentation.
2. add SQS/SNS support
3. decided whether the UX is good enough, i.e the lambda handler gets a new dict with two keys: custom and orig, where orig is the original event unparsed (should it be parsed?) and custom is the pydantic parsed object that user defined.
Note that for the dynamoDB you get a list of dicts where each dict has a new and old objects (new & old stream for each record because dynamodb sends a list of records).
4. how will the user let the decorator know in sns/sqs that the message fields is string or a string that represents a json with a custom schema.
5. add more tests?

aws_lambda_powertools/validation/schemas.py
```diff
@@ -0,0 +1,56 @@
+from datetime import date, datetime
+from typing import Any, Dict, List, Optional
+
+from pydantic import BaseModel, root_validator
+from typing_extensions import Literal
+
+
+class DynamoScheme(BaseModel):
```
alexcasalboni, line 8: Maybe let's use `DynamoDBSchema`?

aws_lambda_powertools/validation/validator.py
```diff
@@ -0,0 +1,126 @@
...
+    def parse(self, event: Dict[str, Any], inbound_schema_model: BaseModel) -> Any:
+        try:
+            parsed_envelope = EventBridgeSchema(**event)
+        except (ValidationError, TypeError):
+            logger.exception("Valdation exception received from input eventbridge event")
+            raise
+        return self._parse_user_dict_schema(parsed_envelope.detail, inbound_schema_model)
+
+
+class DynamoEnvelope(Envelope):
```
alexcasalboni, line 54: Maybe let's use `DynamoDBEnvelope`?

risenberg-cyberark: sure, np

aws_lambda_powertools/validation/schemas.py
risenberg-cyberark: sure, np

risenberg-cyberark: fixed.

aws_lambda_powertools/validation/validator.py
risenberg-cyberark: fixed.

heitorlessa:
**Notes to self to review on Friday**

* [ ] Customers shouldn't need to import `pydantic` to define a model/schema
* [ ] Revisit Schema, Envelope and Model naming convention - There could be confusion here
* [ ] Customers shouldn't need to import `ValidationError` from `pydantic`
* [ ] Revisit DynamoDB Schema names - Schema, Scheme, Record, and DynamoDBStream might be a better name
* [ ] Revisit DynamoDB streams complexity baseline (failing)
* [ ] Review high level function that parses and validates (not a decorator)
* [ ] Revisit `validator` to return parsed event by default instead of `custom/orig` event keys
* [ ] Revisit `validator` option to not return parsed event but original one (prevent breaking those relying on event being a dictionary)
* [ ] Read docs and derive ideas to simplify usage

jplock:
Is there a possibility of implementing event schema validation using the existing fastjsonschema dependency instead of introducing pydantic?

risenberg-cyberark:
@jplock  yes you can, but pydantic is much more than just a schema validator. It gives the users advanced validation capabilities, more type checking (types that JSON doesnt provide), custom logic checks on values (other than empty or null), validate custom relationships between parameters and more. In this PR , we dont only validate but give back the parsed object (a pydantic dataclass basically).

heitorlessa:
> cc @cakepietoast @alexcasalboni your thoughts are mostly welcome here too

@jplock suppose we were to use Pydantic (parser+validation), how would you see yourself using this feature? If not, what would be helpful?

I'm still awe with Pydantic flexibility, the ability to provide runtime safety for customers, and providing auto-complete for customers on common event sources they typically hop onto multiple doc pages to find the right format. 

**However**,  the UX is not easy to use yet. I'm looking at the PR more carefully tomorrow, and will block Monday afternoon to think about ways to simplify this, as Ran did a great job in pulling this together already.

---
**Pros**

* Parses popular event sources and provides auto-complete to event data attributes instead of a dict
* Allows customers to define validation more easily, including runtime type checker with helpful errors
* Plays nicely with popular IDEs as well as deriving JSON Schemas from models

**Cons**

* Envelop, Schema, and Model are intertwined and a bit confusing to use (UX)
* Requires customers to understand Pydantic to use it effectively (learning curve, which is what we aim to lower with this lib)
* Additional dependency (8.2MB), and need to profile/benchmark cold start impact
    - Perhaps we can make an optional dependency using Extras?
---

I fear we might add a powerful feature that not many customers will be able to leverage it to its full extent. If we can't think of a way to find a middle ground by early next week, this will be a good opportunity to break this into two simpler utilities (parser, validator), and add simplicity to this project Tenets.

jplock:
@heitorlessa i like the idea of event schema validation but pydantic feels too heavy for me. All I’d want to do is validate a schema and get a dict. I wouldn’t personally use the full models. That’s why I thought we could build this feature using fastjsonschema as a first release.

michaelbrewer:
@heitorlessa  - i don't like any having any kind of extra dependencies. Currently with no extra dependencies we are already at 8.3M 

I see a few options:

1. Docs: A directory listing of companion libraries like a `Lambda Powertools Awesome Collection`, with a very wrapper implementation. Like getting Chalice + Powertools to work or  Lambda Proxy + Powertools etc... This could be a great resource of cookbook examples.
2. Extras: An extension utility that works when you opt in for that library dependency, but does not require common use cases to include any extra dependencies. 
3. CLI tool: Powertools CLI tool that generates models from a schema file. Then the CLI tool can be quite bloated and generated rich but lightweight model objects from a schema file with a collection of common usage cases like a API Gateway Request / Response. In fact these generate model objects should not need any extra dependencies, and should be treated as read only resource built as part of the makefile.

I prefer a combination of 1 and 3.

----

Current dependencies:

```text
name                                         summary
-------------------------------------------  --------------------------------------------------------------------------------------------------------------------------------------------------------
aws_lambda_powertools                        Python utilities for AWS Lambda functions including but not limited to tracing, logging and custom metric
├── aws-xray-sdk<3.0.0,>=2.5.0               The AWS X-Ray SDK for Python (the SDK) enables Python developers to record and emit information from within their applications to the AWS X-Ray service.
│   ├── botocore>=1.11.3                     Low-level, data-driven core of boto 3.
│   │   ├── docutils<0.16,>=0.10             Docutils -- Python Documentation Utilities
│   │   ├── jmespath<1.0.0,>=0.7.1           JSON Matching Expressions
│   │   ├── python-dateutil<3.0.0,>=2.1      Extensions to the standard Python datetime module
│   │   │   └── six>=1.5                     Python 2 and 3 compatibility utilities
│   │   └── urllib3<1.26,>=1.20              HTTP library with thread-safe connection pooling, file post, and more.
│   ├── future                               Clean single-source support for Python 3 and 2
│   ├── jsonpickle                           Python library for serializing any arbitrary object graph into JSON
│   │   └── importlib-metadata               Read metadata from Python packages
│   │       └── zipp>=0.5                    Backport of pathlib-compatible object wrapper for zip files
│   └── wrapt                                Module for decorators, wrappers and monkey patching.
├── boto3<2.0,>=1.12                         The AWS SDK for Python
│   ├── botocore<1.18.0,>=1.17.50            Low-level, data-driven core of boto 3.
│   │   ├── docutils<0.16,>=0.10             Docutils -- Python Documentation Utilities
│   │   ├── jmespath<1.0.0,>=0.7.1           JSON Matching Expressions
│   │   ├── python-dateutil<3.0.0,>=2.1      Extensions to the standard Python datetime module
│   │   │   └── six>=1.5                     Python 2 and 3 compatibility utilities
│   │   └── urllib3<1.26,>=1.20              HTTP library with thread-safe connection pooling, file post, and more.
│   ├── jmespath<1.0.0,>=0.7.1               JSON Matching Expressions
│   └── s3transfer<0.4.0,>=0.3.0             An Amazon S3 Transfer Manager
│       └── botocore<2.0a.0,>=1.12.36        Low-level, data-driven core of boto 3.
│           ├── docutils<0.16,>=0.10         Docutils -- Python Documentation Utilities
│           ├── jmespath<1.0.0,>=0.7.1       JSON Matching Expressions
│           ├── python-dateutil<3.0.0,>=2.1  Extensions to the standard Python datetime module
│           │   └── six>=1.5                 Python 2 and 3 compatibility utilities
│           └── urllib3<1.26,>=1.20          HTTP library with thread-safe connection pooling, file post, and more.
└── fastjsonschema<2.15.0,>=2.14.4           Fastest Python implementation of JSON schema
```

heitorlessa:
Thanks @jplock and @michaelbrewer for your inputs, much appreciated. 

We'll review this today with the core team and circle back with a solution - As of now, this doesn't meet our Keep It Lean tenet however useful this is.

risenberg-cyberark:
I'll add my two cents here if you dont mind :)
 I've introduced pydantic to my organisation half a year ago. It's been well received and we have an excellent experience with it. 
It's very easy to define a schema. You dont have to use the bells & whistles. You just inherit from BaseModel class and define parameters with the Python.Typing module. It produces a much readable code than a dict schema. You can also extend it and import and have a repo just for schemas. What I like here is that as an end user I have an option. I can define a basic schema or a more sophisticated schema.

Regarding the usage -I added per @heitorlessa 's request the validate function. It's not the decorator. it's a simpler version that take a schema and returns a parsed object, just like you wanted.

Regarding the extra size, as an end user, if i'm getting good performance and more usability options, i really dont mind.

I think this repo can set a high standard for validation. It's a great opportunity and I believe that pydantic can help that.

cakepietoast:
I'll weigh in on this one too. Firstly, pydantic looks like an awesome library, and I can definitely see it being useful in the context of Lambda. With that said, I feel like this PR in its current state is breaking the "keep it lean" tenet - if I've understood @michaelbrewer correctly above, the pydantic dependency roughly doubles the size of the powertools package. I think it would also mean dedicating a significant part of the powertools documentation to explaining pydantic specific concepts.

We don't spell it out explicitly in the tenets (perhaps we should?), but for me "easing the adoption of best practices" means thinking about how to do that for 80%+ of the users. My feeling is that a json validation based on the already included fastjsonschema would be the simplest implementation here, and have the smallest learning curve for the majority of users. Users who need a more involved validation solution could make use of the pydantic library in their own custom middlewares. To help those users, I like the idea of having companion libraries and some examples documented in a cookbook.

jplock:
@michaelbrewer is it necessary to bundle boto3 in favor of people bundling themselves in their applications?

heitorlessa:
Hi Ran - As we've discussed, we've decided to put this PR on hold while we hear from more customers on their use of Pydantic for parsing and business rules validation. 

I've created a [dedicated feature request](https://github.com/awslabs/aws-lambda-powertools-python/issues/147) for this, and added labels to remind us to revisit this in the next 3 months. 

As of now, we'll work on a simple validation tool that uses JSON Schemas to meet the 80% of customers who are used to this mechanism. We hope to hear from more customers interested in using Pydantic, and recommend customers using the future simple validation tool to take a look at Pydantic in the hope to build awareness and educate on what's possible.

As the usage of Pydantic grows for Lambda functions and we hear from more customers the need for this feature, we'd like to resume collaborating on this PR with you.

We appreciate all efforts you've done so far on pulling this together, and we look forward to chatting soon about this topic in the future.

Thank you

heitorlessa:
Sharing it here the UX we discussed today - We will include the Parser utility in 1.7.0 as a featured new utility ;)

This should:

1. Make the implementation and maintenance easier (tests, etc.)
2. Allow advanced customers to keep their workflows while having access to popular event sources schemas/models
3. Make UX simpler to remember and use

----

### Parsing with built-in envelopes

We'll parse the wrapped event for popular event sources using the data schema provided:

```python
class MyBusiness(BaseModel):
    message: str
    message_id: int

@parser(schema=MyBusiness, envelope=envelopes.EVENTBRIDGE)
def handler(event: MyBusiness, context: LambdaContext):
    # MyBusiness data with no EventBridge data
    # We’ll parse event[“detail”] with the schema provided
    event.message
    event.message_id
```

### Extending schemas

For advanced customers who know Pydantic well, or want to get access to the whole payload plus their data parsed, they will be able to extend official schemas:

```python
class MyBusiness(BaseModel):
    message: str
    message_id: int

class MyBusinessEventBridge(EventBridgeSchema):
    detail: MyBusiness # override detail key with MyBusiness data as opposed to a simple Dict

@parser(schema=MyBusinessEventBridge)
def handler(event: MyBusinessEventBride, context: LambdaContext):
     # Event Bridge fields
     event.account_id
     event.region
     # MyBusiness data inside detail
     event.detail.message
```

risenberg-cyberark:
Does anybody know why I'm getting this build error:
All local linter/format section work just fine on my dev env.

Run make lint
poetry run isort -rc aws_lambda_powertools tests example
Creating virtualenv aws-lambda-powertools-s3ycVuVL-py3.6 in /home/runner/.cache/pypoetry/virtualenvs

[FileNotFoundError]
[Errno 2] No such file or directory
Makefile:13: recipe for target 'format' failed
make: *** [format] Error 1
Error: Process completed with exit code 2.

risenberg-cyberark:
@heitorlessa @cakepietoast @jplock  hey guys, do you have any idea why the makefile doesnt pass the format stage? i'm stuck :(

heitorlessa:
> @heitorlessa @cakepietoast @jplock hey guys, do you have any idea why the makefile doesnt pass the format stage? i'm stuck :(

looking into it..

heitorlessa:
The format is now solved upstream in develop - rebasing should work.

For the Pytest on examples, it's best to completely remove the example folder from the repo, since we now have a cookiecutter for that purpose.

I'll push that change to develop later and ping here for rebasing

risenberg-cyberark:
if anyone knows to reduce the complexity of:
ERROR:xenon:module ‘aws_lambda_powertools/utilities/advanced_parser/envelopes/dynamodb.py’ has a rank of B

I'm open to suggestions.. it's just a pydantic data class.

risenberg-cyberark:
@heitorlessa  all ready for ya ;)

aws_lambda_powertools/utilities/advanced_parser/schemas/sqs.py
```diff
@@ -0,0 +1,61 @@
...
+
+class SqsMsgAttributeSchema(BaseModel):
+    stringValue: Optional[str]
+    binaryValue: Optional[str]
+    stringListValues: List[str] = []
+    binaryListValues: List[str] = []
+    dataType: str
+
+    @validator("dataType")
+    def valid_type(cls, v):  # noqa: VNE001
```
heitorlessa, line 28: docs: could you add a comment as a context for this validation?

aws_lambda_powertools/utilities/advanced_parser/schemas/sqs.py
```diff
@@ -0,0 +1,61 @@
...
+        pattern = re.compile("Number.*|String.*|Binary.*")
+        if not pattern.match(v):
+            raise TypeError("data type is invalid")
+        return v
+
+    @root_validator
+    def check_str_and_binary_values(cls, values):
```
heitorlessa, line 35: docs: could you add a comment as a context for this validation?

aws_lambda_powertools/utilities/advanced_parser/envelopes/base.py
```diff
@@ -0,0 +1,37 @@
...
+            return schema(**user_event)
+        except (ValidationError, TypeError):
+            logger.exception("Validation exception while extracting user custom schema")
+            raise
+
+    def _parse_user_json_string_schema(self, user_event: str, schema: BaseModel) -> Any:
+        if user_event is None:
+            return None
+        logger.debug("parsing user dictionary schema")
+        if schema == str:
```
heitorlessa, line 25: docs: could you add a comment as a context for why we need this? one could also pass `schema=bool()` and would pass and fail down there, so I guess it's not a type check only

aws_lambda_powertools/utilities/advanced_parser/envelopes/base.py
```diff
@@ -0,0 +1,37 @@
...
+            logger.debug("input is string, returning")
+            return user_event
+        logger.debug("trying to parse as json encoded string")
+        try:
+            return schema.parse_raw(user_event)
```
heitorlessa, line 30: docs: could you add a comment as a context for parse_raw for non-pydantic maintainers?

tests/functional/parser/test_sqs.py
```diff
@@ -0,0 +1,68 @@
...
+
+from aws_lambda_powertools.utilities.advanced_parser.envelopes.envelopes import Envelope
+from aws_lambda_powertools.utilities.advanced_parser.parser import parser
+from aws_lambda_powertools.utilities.typing import LambdaContext
+from tests.functional.parser.schemas import MyAdvancedSqsBusiness, MySqsBusiness
+from tests.functional.parser.utils import load_event
+from tests.functional.validator.conftest import sqs_event  # noqa: F401
+
+
+@parser(schema=str, envelope=Envelope.SQS)
```
heitorlessa, line 11: docs: could you add a comment as a context as to why someone would pass a schema as a str? should we expect something else?

pyproject.toml
```diff
@@ -23,6 +23,8 @@ aws-xray-sdk = "^2.5.0"
 fastjsonschema = "^2.14.5"
 boto3 = "^1.12"
 jmespath = "^0.10.0"
+pydantic = {version = "^1.6.0", optional = true }
+typing_extensions = {version = "^3.7.4.2", optional = true }
```
heitorlessa, line 27: docs: could you add a comment as a context for this additional optional lib?

aws_lambda_powertools/utilities/advanced_parser/parser.py
```diff
@@ -0,0 +1,56 @@
...
+        -------
+        **Lambda function using validation decorator**
+
+            @parser(schema=MyBusiness, envelope=envelopes.EVENTBRIDGE)
+            def handler(event: inbound_schema_model , context: LambdaContext):
+                ...
+
+        Parameters
+        ----------
+        todo add
```
heitorlessa, line 36: TODO

aws_lambda_powertools/utilities/advanced_parser/parser.py
```diff
@@ -0,0 +1,56 @@
...
+            logger.debug("Parsing and validating event schema, no envelope is used")
+            parsed_event = schema(**event)
+        except (ValidationError, TypeError):
+            logger.exception("Validation exception received from input event")
+            raise
+    else:
+        parsed_event = parse_envelope(event, envelope, schema)
+
+    logger.debug(f"Calling handler {lambda_handler_name}")
+    handler(parsed_event, context)
```
heitorlessa, line 56: `return handler(parsed_event, context)`

aws_lambda_powertools/utilities/advanced_parser/envelopes/dynamodb.py
```diff
@@ -0,0 +1,27 @@
...
+
+from pydantic import BaseModel, ValidationError
+
+from aws_lambda_powertools.utilities.advanced_parser.envelopes.base import BaseEnvelope
+from aws_lambda_powertools.utilities.advanced_parser.schemas import DynamoDBSchema
+
+logger = logging.getLogger(__name__)
+
+
+class DynamoDBEnvelope(BaseEnvelope):
```
heitorlessa, line 12: docs: comment or docstring on the return if possible

aws_lambda_powertools/utilities/advanced_parser/envelopes/sqs.py
```diff
@@ -0,0 +1,22 @@
...
+
+from pydantic import BaseModel, ValidationError
+
+from aws_lambda_powertools.utilities.advanced_parser.envelopes.base import BaseEnvelope
+from aws_lambda_powertools.utilities.advanced_parser.schemas import SqsSchema
+
+logger = logging.getLogger(__name__)
+
+
+class SqsEnvelope(BaseEnvelope):
```
heitorlessa, line 12: docs: comment or docstring on the return if possible

aws_lambda_powertools/utilities/advanced_parser/envelopes/event_bridge.py
```diff
@@ -0,0 +1,19 @@
...
+
+from pydantic import BaseModel, ValidationError
+
+from aws_lambda_powertools.utilities.advanced_parser.envelopes.base import BaseEnvelope
+from aws_lambda_powertools.utilities.advanced_parser.schemas import EventBridgeSchema
+
+logger = logging.getLogger(__name__)
+
+
+class EventBridgeEnvelope(BaseEnvelope):
```
heitorlessa, line 12: docs: comment or docstring on the return if possible

heitorlessa (changes requested):
New user experience looks GREAT! Added some comments to help us maintain as not all maintainers are well versed with Pydantic, so any additional context helps.

aws_lambda_powertools/utilities/advanced_parser/envelopes/base.py
risenberg-cyberark: added

aws_lambda_powertools/utilities/advanced_parser/schemas/sqs.py
risenberg-cyberark: added

aws_lambda_powertools/utilities/advanced_parser/schemas/sqs.py
risenberg-cyberark: added

aws_lambda_powertools/utilities/advanced_parser/schemas/dynamodb.py
```diff
@@ -0,0 +1,43 @@
...
+class DynamoScheme(BaseModel):
+    ApproximateCreationDateTime: Optional[date]
+    Keys: Dict[str, Dict[str, Any]]
+    NewImage: Optional[Dict[str, Any]]
+    OldImage: Optional[Dict[str, Any]]
+    SequenceNumber: str
+    SizeBytes: int
+    StreamViewType: Literal["NEW_AND_OLD_IMAGES", "KEYS_ONLY", "NEW_IMAGE", "OLD_IMAGE"]
+
+    @root_validator
```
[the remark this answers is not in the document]
risenberg-cyberark: added

Makefile
```diff
@@ -5,6 +5,7 @@ target:
 dev:
 	pip install --upgrade pip poetry pre-commit
 	poetry install
+	poetry install --extras "pydantic"
```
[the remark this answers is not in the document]
risenberg-cyberark: fixed

aws_lambda_powertools/utilities/advanced_parser/parser.py
risenberg-cyberark: fixed

aws_lambda_powertools/utilities/advanced_parser/parser.py
risenberg-cyberark: fixed

aws_lambda_powertools/utilities/advanced_parser/envelopes/dynamodb.py
risenberg-cyberark: fixed.

aws_lambda_powertools/utilities/advanced_parser/envelopes/sqs.py
risenberg-cyberark: the type hint is good here but i'll add a comment.

aws_lambda_powertools/utilities/advanced_parser/envelopes/event_bridge.py
risenberg-cyberark: fixed.

heitorlessa:
**THANK YOU** @risenberg-cyberark for all those changes, UX improvements, and additional comments you've made. 

I'm gonna merge this as-is now. As agreed, I'll make a PR to write the docs for this feature and get your review before merging it. 

As you also spent a ton of effort already, I'll create another PR the following minor changes:

* [ ] **Remove validators**. We need an universal schema owned by Lambda for each event source, or else we'll likely break some customers if a few services change, as we won't be notified.
* [ ] **Snake_case**. IIRC, EventBridge `detailtype` was the only odd one out but I'll double check if there are others
* [ ] **Refactor test event data**. Data_classes and validator utility are partially using the same events, so we'll refactor tests to provide events to all tests as fixtures.
* [ ] **Kwargs over args**. Some internal code is using args over kwargs - Moving to kwargs make it easier to refactor, and order of args wouldn't be impacted.
* [ ] **Docstrings**. Enhance Example section in docstrings.
* [ ] **Remove flake8 polyfill**. Double check whether a fresh Python env hits the issue you hit with that, since we don't use it
* [ ] **Raise own Validation Exception**. For customers wanting to catch that exception (e.g. custom middlewares) they'll have to find Pydantic exception to catch it properly
* [ ] **Enhance imports**. Envelopes and a few others can be turned into high level imports to clean up customers' code, plus DX.
* [ ] **Test coverage**. Validation errors aren't being tested as of now.
* [ ] **[Optional] Enhancing envelope selection**. Time allowing, I'll experiment with providing a similar experience we have in the `validator` envelope utility `envelope.<event_source>`.
* [ ] **[Optional] Create a standalone parser**. Time allowing, I'll experiment creating a standalone `parser` function with the same functionality the decorator
* [ ] **[Optional] Support for API Gateway, Kinesis, S3, and SNS**. Likely to be after 1.7.0, but time allowing we can do that sooner.

heitorlessa (approved):

risenberg-cyberark:
@heitorlessa can you maybe put them in comments? i'd hate for that logic to be forgotten. These are connections that json schemas cant really define (only documentation).

heitorlessa:
> @heitorlessa can you maybe put them in comments? i'd hate for that logic to be forgotten. These are connections that json schemas cant really define (only documentation).

Sure thing - Commented them out, and added a link to this discussion for history purposes.

mwarkentin:
@heitorlessa are the extras required for this functionality included in the SAR app / lambda layer? Or do those need to be handled separately? Might be worth adding a note in the docs, not sure if that would be under the parser docs or the lambda layer installation docs?

heitorlessa:
> @heitorlessa are the extras required for this functionality included in the SAR app / lambda layer? Or do those need to be handled separately? Might be worth adding a note in the docs, not sure if that would be under the parser docs or the lambda layer installation docs?

Hey @mwarkentin - the latter, handled separately as described in the parser docs.

Unsure whether it worth creating a separate Layer for that tbh given the additional operational overhead. In V2, we'd like to create each utility to be pip installable if a theory works, then we could consider multiple layers I suppose.

Happy to discuss more in a separate issue tho ;)

PR #133, opened 2020-08-26
feat(parameters): transform = "auto"

michaelbrewer:
**Issue #, if available:** #136

When fetching multiple values from the parameter store or dynamodb often each record might have different encoding and not all have uniform encoding. eg: 

| key   | value                               |
|------|------------------------|
| A      | json string                       |
| B      | base64 decoded image |
| C      | might be a plain string   |

## Description of changes:

Add a new tranform type called "auto", which looks at the
key name to autodetect the encoding of the values in dynamodb or
the parameter store.

* Keys ending with ".json" are assumed to be "json" string
* Keys ending with ".binary" are assumed to be "binary" base64 decode
strings
* Other Keys are assumed to be plain text values.

| key        | value                                    |
|---------|---------------------------|
| A.json    | ["one", "two"]                      |
| B.binary| b'base64 decoded image'  |
| C            | plain text string                   |

**Checklist**

 

* [X] Update tests
* [ ] Update docs


michaelbrewer:
@nmoutschen - if it is a matter of flexibility on the naming pattern of the keys, we could create patterns for them. 

Another optional it giving the new user an option to set a customer converter that accepts "key" and "value". So you could provide an auto-formatter, a yaml one, xml etc.

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -159,6 +163,45 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
         raise NotImplementedError()
 
 
+def get_transform_method(key: str, transform: str) -> Optional[str]:
```
nmoutschen, line 166: Should `transform` be an optional parameter here or a nullable value?
E.g. similar to how `transform` get passed to the get() and get_multiple() methods.

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -159,6 +163,45 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+    """
+    if transform != "auto":
+        return transform
+
+    if key.endswith(".json"):
+        return "json"
+    elif key.endswith(".binary"):
+        return "binary"
+    else:
+        return None
```
nmoutschen, line 202: Not critical now, but it would be good to parameterise this across the utility. If we add new transformers in the future, it'd be good to have only one place where we need to make a chance to reduce mistakes.

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -145,7 +145,11 @@ def get_multiple(
 
         if transform is not None:
             for (key, value) in values.items():
-                values[key] = transform_value(value, transform, raise_on_transform_error)
+                _transform = get_transform_method(key, transform)
+                if _transform is None:
+                    continue
+
+                values[key] = transform_value(value, _transform, raise_on_transform_error)
```
nmoutschen, line 152: __Open question:__ Should we change the key name based on the transform?

E.g., let's say we add a yaml/json5/etc. transform in the future, what if someone wants to change a parameter from JSON to YAML but minimise code change?

nmoutschen (changes requested):
Hey @michaelbrewer !

Thanks a lot for your contribution! 😄 I have one open question regarding the user experience for this feature: should we rename the key by dropping the extension?

It would also be good to parameterise the transforms just to have a single place where to add/change transforms to prevent errors in the future if we want to add new transforms. However, this aspect is good as-is for my point of view.

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen - is currently only called from `get_multiple` and when the `transform is not None`

```python3
        if transform is not None:
            for (key, value) in values.items():
                _transform = get_transform_method(key, transform)
```

I am not sure if it makes sense to use `auto` for single items.

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen  - maybe we introduce a constant of all of the supported `transform_value` and they can be used as the extension name for the `auto` mode?

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen  - in `auto` method, their key name is explicitly setting the transform method, then `.json` and `.json5` would have different extensions. And for new methods like `.xml`, `.properties` and  `.yaml` we would use different transform_methods.

michaelbrewer:
@nmoutschen would this patch make this more flexibile?

```patch
diff --git a/aws_lambda_powertools/utilities/parameters/base.py b/aws_lambda_powertools/utilities/parameters/base.py
index e1ba9ac..1269432 100644
--- a/aws_lambda_powertools/utilities/parameters/base.py
+++ b/aws_lambda_powertools/utilities/parameters/base.py
@@ -15,6 +15,9 @@ DEFAULT_MAX_AGE_SECS = 5
 ExpirableValue = namedtuple("ExpirableValue", ["value", "ttl"])
 # These providers will be dynamically initialized on first use of the helper functions
 DEFAULT_PROVIDERS = {}
+TRANSFORM_METHOD_JSON = "json"
+TRANSFORM_METHOD_BINARY = "binary"
+SUPPORTED_TRANSFORM_METHODS = [TRANSFORM_METHOD_JSON, TRANSFORM_METHOD_BINARY]
 
 
 class BaseProvider(ABC):
@@ -194,12 +197,10 @@ def get_transform_method(key: str, transform: str) -> Optional[str]:
     if transform != "auto":
         return transform
 
-    if key.endswith(".json"):
-        return "json"
-    elif key.endswith(".binary"):
-        return "binary"
-    else:
-        return None
+    for transform_method in SUPPORTED_TRANSFORM_METHODS:
+        if key.endswith("." + transform_method):
+            return transform_method
+    return None
 
 
 def transform_value(value: str, transform: str, raise_on_transform_error: bool = True) -> Union[dict, bytes, None]:
@@ -223,9 +224,9 @@ def transform_value(value: str, transform: str, raise_on_transform_error: bool =
     """
 
     try:
-        if transform == "json":
+        if transform == TRANSFORM_METHOD_JSON:
             return json.loads(value)
-        elif transform == "binary":
+        elif transform == TRANSFORM_METHOD_BINARY:
             return base64.b64decode(value)
         else:
             raise ValueError(f"Invalid transform type '{transform}'")

```

aws_lambda_powertools/utilities/parameters/base.py
nmoutschen: Sorry, I think I was not clear here so let me put an example.

Let's say I have a DynamoDB like this:

| id | sk | value |
|----|----|--------|
| my-param | param.json | {"message": "Hello"} |

I'm wondering, should the return value be A or B between:

```
a = {
    "param.json": {
        "message": "Hello"
    }
}
```

```
b = {
    "param": {
        "message": "Hello"
    }
}
```

aws_lambda_powertools/utilities/parameters/base.py
nmoutschen: My idea here was that when I pass `transform=None`, I'd get `None` as a return value.

It might not be needed today, but I'm thinking from two point of view: consistency with other function signatures (excepted behaviours for contributors), and if we want to move this transform if it's useful for other utilities in the future.

Looking at the code today, it would only change the signature since we already return the transform if it's not "auto":

```python
    if transform != "auto":
        return transform
```

aws_lambda_powertools/utilities/parameters/base.py
nmoutschen: Not sure if I follow you here, what would it look like?

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen  it should be the same as if all of them where `json` to be consistent with the `json` transform method.

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen  - maybe look at the later version of the code.

nmoutschen: Got you! :) Looks good to me there!

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen I made this change. You can check again.

aws_lambda_powertools/utilities/parameters/base.py
michaelbrewer: @nmoutschen i think it should be `a`, because if you turn off auto. The keys should be the same as before. 

But I see your point. If might be nicer for it to be `b` for the end user.

michaelbrewer: @nmoutschen Actually I take that back it should be `a`. Because `b` might conflict when an existing value with the same name.

michaelbrewer: @nmoutschen  as an example:

id | sk | value
-- | -- | --
my-param | param.json | {"message": "Hello"}
my-param | param| Plain text

nmoutschen: Good point! 👍 I agree with you on that. Since that's the current behaviour, no change needed here.

nmoutschen (approved):
LGTM!

cakepietoast (approved):

PR #134, opened 2020-08-26
docs: add info how to get layer versions from the SAR for the layers 

am29d:
**Issue #, if available:**

## Description of changes:

The AWS console does not show the available versions of the SAR app for the layers we have recently published. I added the CLI command this the ARN to fetch the available versions.

**Checklist**

 

* [x] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [ ] Update tests
* [x] Update docs
* [x] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/376ec0a2ac0d2a40e0af5717bef42ff84ca0d1b9/.github/semantic.yml#L2)

## Breaking change checklist

**RFC issue #**:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #145, opened 2020-08-28
docs: add description where to find the layer arn

am29d:
**Issue #, if available:**

## Description of changes:

The AWS console does not show the available versions of the SAR app for the layers we have recently published. I added the CLI command this the ARN to fetch the available versions.

I also add instruction on how to find the Layern ARN after deploying the SAR App.

Unfortunately, you have first to deploy the SAR app to get the LayerVersionArn as an exported value that can be referenced in a Lambda function resource definition. You can not do this within one SAM template and deploy the app and the lambda in one go :(.

I added two options in the documentation, either way you need to deploy two times.

Reopened PR because I screw up my fork, this was previous PR #134 

**Checklist**

 

* [ ] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


docs/content/index.mdx
```diff
@@ -44,6 +44,16 @@ If using SAM, you can include this SAR App as part of your shared Layers stack,
         SemanticVersion: 1.3.1 # change to latest semantic version available in SAR
 ```
 
+This will add a nested app stack with an output parameter `LayerVersionArn`. You can not reference layer arn within the template on the first try, this is currently not supported.
```
heitorlessa, line 47: Why is it not supported?

How did you try to reference it?

heitorlessa (changes requested):

heitorlessa:
Here's an example where I'm referencing the outputs of a SAR App within the same stack


https://github.com/aws-samples/aws-serverless-airline-booking/blob/develop/src/backend/payment/template.yaml

Trying to understand whether this is a new limitation OR how we built the App

docs/content/index.mdx
am29d: I tried to reference it via !Ref and !GetAtt via the logical resource name within the template:

```
   Layers:
     - !GetAtt AwsLambdaPowertoolsPythonLayer.Arn
```

going via outputs as you have showed in your example is working:

```
   Layers:
     - !GetAtt AwsLambdaPowertoolsPythonLayer.Outputs.LayerVersionArn
```

I have updated the docs and added this to the `example` project we have under `template.yml`

heitorlessa (approved):
Thanks for making the changes, Alex :)

There's a minor piece in the example, since that's now using both the Layer and requirements. 

It needs a more elaborate fix there, so I'll merge as-is, and we can deal with that later

PR #151, opened 2020-08-30
docs: add troubleshooting section

Nr18:
**Issue #, if available:**

## Description of changes:

Add a troubleshooting section in the CONTRIBUTING.md

**Checklist**

 

* [ ] ~Update tests~
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -71,7 +71,14 @@ If you discover a potential security issue in this project we ask that you notif
 
 ## Troubleshooting
 
-* The local **API reference** throws a `Module aws_lambda_powertools not found` error when an exception is raised in the code-base, when you run `poetry run pdoc --pdf aws_lambda_powertools` the exception is shown.
+### API reference documentation
+
+When you are working on the codebase and you use the local API reference documentation to preview you changes you might se the following message: `Module aws_lambda_powertools not found`.
+
+This happens when:
+
+* You did not install the local dev environment yet. The solution for that would be running: `make dev`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -69,6 +69,9 @@ opensource-codeofconduct@amazon.com with any additional questions or comments.
 ## Security issue notifications
 If you discover a potential security issue in this project we ask that you notify AWS/Amazon Security via our [vulnerability reporting page](http://aws.amazon.com/security/vulnerability-reporting/). Please do **not** create a public github issue.
 
+## Troubleshooting
+
+* The local **API reference** throws a `Module aws_lambda_powertools not found` error when an exception is raised in the code-base, when you run `poetry run pdoc --pdf aws_lambda_powertools` the exception is shown.
```
heitorlessa, line 74: is there a fix for it? Because it's under Troubleshooting, I thought there was :)

heitorlessa (changes requested):
Let me know if I misunderstood the guidance - I suspect it's missing a resolution besides naming the problem. I suspect you were also trying to run `make docs-api-local`

CONTRIBUTING.md
Nr18: Ah I will rephrase it

What I ment to write is that if the local api webserver doesn't work and throwing 👆 that error. You can then use the pdf generation option to display the exception that is raised.

heitorlessa: Ahhh, gotcha! I only saw that error when the package wasn't installed locally (`make dev`).

Nr18: Ah nice I added the `make dev` as a solution as well, the `pdoc3` path can't be wrong anymore since we run everything in a poetry shell right?

heitorlessa:
Yes, unless you somehow (on purpose, pure evil) change directory and then
use poetry shell to run pdoc3 haha

heitorlessa (approved):
Thanks a lot for adding the troubleshooting section - It'll help us provide more prescriptive guidance when hitting these pesky issues ;)

Issue #147, opened 2020-08-28
Feature: Advanced parser utility


heitorlessa:
**Is your feature request related to a problem? Please describe.**

We've heard from a small number of customers that parsing Lambda Event Source payloads require a considerable effort since these don't have official schemas for Python.

With parsing and classes modelled after these schemas, they can have the benefits of runtime type safety, custom validations on possible values pertinent to their use case, autocomplete, and only parse fields they're interested in.

**Describe the solution you'd like**

Solution is two-fold:

* A new `parser` utility that uses Pydantic to parse and validate incoming/outgoing events, and allow customers to use their own data models
* Pre-defined schemas and event envelopes for popular event sources, so one can apply and validate their models against where they payload is

This would reduce the amount of time developers invest searching for official data structure for each event source, improve their security posture on incoming and outgoing events, and increased developer productivity.

**Describe alternatives you've considered**

* Implement simple validation using JSON Schemas as well as an extractor utility to retrieve the payload only
* Bring Pydantic as an optional package to prevent bloating the library for those not using it

Challenge with JSON Schemas is they typically don't validate business rules for incoming/outgoing events, but merely a schema.

**Additional context**

Initial implementation that lacked customer data points as of now, but could be revisited depending on interest for thisfeature: https://github.com/awslabs/aws-lambda-powertools-python/pull/118

jplock:
I like the idea of using JSON schemas strictly for validating AWS managed event types that don’t change. Validating the payload would be the responsibility of the consumer (using pydantic or something similar).

Nr18:
My preference would be to use pydantic having the ability to add business rule validation would be beneficial to me.

I actually ran into the issue of having to set my date field to a string in the JSON schema because the date is optional... With pydantic, you can have an `Optional[date]` without having to write an additional statement just to check that the string is actually a date.

koxudaxi:
I always use pydantic for Lambda for events and payload.
Pydantic is a very great solution for validation and parsing.

However, The library is a little heavy for lambda. 
Pydantic is compiled by Cython which is created some *.so files.
I calculate these file sizes. It's about 76MB https://pypi.org/project/pydantic/#files

I suggest Pydantic is supplied as an option.
And I think powertools should provide two type validator(decorator) that are JSON Schema and Pydantic.
Users can select one.

Also, I develop a code-generator that generates pydantic models from JSON Schema.
https://github.com/koxudaxi/datamodel-code-generator
[...]

PR #153, opened 2020-08-31
feat: simple JSON Schema validator utility

heitorlessa:
**Issue #, if available:** #95, and related to #147

## Description of changes:

**Checklist**

 

* [x] Update tests
* [x] Update docs
* [x] `validator` decorator for both inbound/outbound schema
* [x] `validate` standalone function
* [x] support JMESPath for flexible envelope expressions (a.k.a data selectors)
* [x] make `jmespath` an optional dependency (pip install aws-lambda-powertools[jmespath])
* [x] Docstrings
* [x] Double check whether we want to throw data into exception as it might contain sensitive data
  - Trade-off on having helpful exception
* [x] Experiment w/ closure/cache envelope expr
  - maybe premature optimization given it's quite fast (need benchmark)
* [x] Bring envelopes for popular event sources to retrieve data
  * [x] API Gateway REST API
  * [x] API Gateway HTTP API
  * [x] SQS
  * [x] SNS
  * [x] EventBridge
  * [x] Kinesis
  * [x] CloudWatch Events Scheduled Event
  * [x] CloudWatch Logs
* [x] Custom JMESPath function for JSON deserialization
* [x] Custom JMESPath function for Base64 decoding
* [x] Custom JMESPath function for Base64 and gzip decompression (cloudwatch logs)

## Open questions

None

### Decorator

```python
from aws_lambda_powertools.utilities.validation import validate, validator

hello_schema = {
    "$schema": "http://json-schema.org/draft-07/schema",
    "$id": "http://example.com/example.json",
    "type": "object",
    "title": "The root schema",
    "description": "The root schema comprises the entire JSON document.",
    "examples": [
        {
            "message": "hello world",
            "username": "lessa"
        }
    ],
    "required": [
        "message",
        "username"
    ],
    "properties": {
        "message": {
            "$id": "#/properties/message",
            "type": "string",
            "title": "The message schema",
            "description": "Message field",
        },
        "username": {
            "$id": "#/properties/username",
            "type": "string",
            "title": "The username schema",
            "description": "Username field",
        }
    },
    "additionalProperties": True
}

sample_wrapped_event = {
    "data": {
        "payload": {
            "message": "hello hello",
            "username": "blah blah"
        }
    }
}

# Either Inbound or Outbound schema, or both
# Envelope uses JMESPath expression to extract 
## only what data should be validated against schema
@validator(inbound_schema=hello_schema, envelope="data.payload")
def handler(evt, ctx):
    ...
```

### Deserializing JSON Strings before validation with powertools_json

We'll provide a function to deserialize JSON Strings that are commonly present in API Gateway events as well as any arbitrary payload. This JMESPath function will deserialize any value you pass as an argument.

```python
from aws_lambda_powertools.utilities.validation import validate, validator

sample_wrapped_event_json_string = {
    "data": json.dumps({"payload": {"message": "hello hello", "username": "blah blah"}})
}

@validator(inbound_schema=hello_schema, envelope="powertools_json(data).payload")
def handler(evt, ctx):
    ...
```

### Deserializing Base64 JSON Strings before validation with powertools_base64 and powertools_json

We'll provide a function to decode base64 data that can be used in tandem with our function to deserialize JSON Strings - This is typically the case for events coming from Kinesis Records, CloudWatch Logs, etc.

An example of a raw JMESPath expression to decode and deserialize JSON Strings sent to Kinesis before applying validation - The validation has to take into account an array of the data, in this case

```python
from aws_lambda_powertools.utilities.validation import validate, validator

sample_wrapped_event_base64_json_string = {
    "data": "eyJtZXNzYWdlIjogImhlbGxvIGhlbGxvIiwgInVzZXJuYW1lIjogImJsYWggYmxhaCJ9=")
}

@validator(inbound_schema=hello_schema, envelope=envelopes.KINESIS_DATA_STREAM) # when the event is a Kinesis record (array)
@validator(inbound_schema=hello_schema, envelope="powertools_base64(data) | powertools_json(@)") # sample event above
def handler(evt, ctx):
    ...
```

### Decompressing and deserializing GZIP Base64 JSON Strings before validation with powertools_base64_gzip

AWS services like CloudWatch Logs compress log events as ZIP archives and base64 its binary content - Similar to base64 and JSON deserialization above, we'll provide a `powertools_base64_gzip` function.

An example of a [CloudWatch Logs event already compressed](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/SubscriptionFilters.html#LambdaFunctionExample) to demonstrate this function:

```python
from aws_lambda_powertools.utilities.validation import validate, validator

sample_wrapped_event_base64_json_string = {
    "awslogs": {
        "data": "H4sIACZAXl8C/52PzUrEMBhFX2UILpX8tPbHXWHqIOiq3Q1F0ubrWEiakqTWofTdTYYB0YWL2d5zvnuTFellBIOedoiyKH5M0iwnlKH7HZL6dDB6ngLDfLFYctUKjie9gHFaS/sAX1xNEq525QxwFXRGGMEkx4Th491rUZdV3YiIZ6Ljfd+lfSyAtZloacQgAkqSJCGhxM6t7cwwuUGPz4N0YKyvO6I9WDeMPMSo8Z4Ca/kJ6vMEYW5f1MX7W1lVxaG8vqX8hNFdjlc0iCBBSF4ERT/3Pl7RbMGMXF2KZMh/C+gDpNS7RRsp0OaRGzx0/t8e0jgmcczyLCWEePhni/23JWalzjdu0a3ZvgEaNLXeugEAAA=="
    }
}

@validator(inbound_schema=hello_schema, envelope=envelopes.CLOUDWATCH_LOGS) # when the event is a CloudWatch Logs record (array)
@validator(inbound_schema=hello_schema, envelope="awslogs.powertools_base64_gzip(data) | powertools_json(@).logEvents[*]") # sample event above
def handler(evt, ctx):
    ...
```

### Standalone function

```python
from aws_lambda_powertools.utilities.validation import validate, validator

sample_wrapped_event_json_string = {
    "data": json.dumps({"payload": {"message": "hello hello", "username": "blah blah"}})
}

def handler(evt, ctx):
    validate(event=evt, schema=hello_schema, envelope="powertools_json(data).payload")
```

#### Built-in envelopes

```python
from aws_lambda_powertools.utilities.validation import validator, envelopes

sample_wrapped_event_json_string = {
    "data": json.dumps({"message": "hello hello", "username": "blah blah"})
}

@validator(inbound_schema=hello_schema, envelope=envelopes.API_GATEWAY_REST)
def handler(evt, ctx):
    return True
```

## Breaking change checklist

**RFC issue #**:


jplock:
I like this a lot 🎊

nmoutschen:
Idea on how to handle API Gateway events and other events that might contain serialized data:

We could add a function to JMESPath named `powertools_json` that will take a string as input and deserialize it. This would allow to use these expressions as envelopes:

```
"powertools_json(body)" -> Deserialize the body and return it
"body" -> Keep the body as-is (e.g. string)
"powertools_json(body).payload" -> Deserialize the body and return the value for the payload key

# We could have nested JSON bodies
"powertools_json(body).payload | powertools_json(@)"
```

__Why prefix with powertools?__ The documentation for JMESPath states that "If you have a custom function that you’ve found useful, consider submitting it to jmespath.site and propose that it be added to the JMESPath language", therefore new functions could be added over time. If JMESPath adds a "json()" function and this function behaves in a slightly different way, there should be no conflict with the Powertools implementation.

heitorlessa:
@nmoutschen @cakepietoast could you have a pass at the UX and help me answer the questions I put at the description, please?

:)

nmoutschen:
Overall the UX looks good. Quick questions:

* Will this also support outbound validation through the same decorator?
* (Nitpick) Should the envelope for API_GATEWAY_REST should be v1? Since HTTP API supports both v1 and v2.

heitorlessa:
Yes to the first question, and for API GW payload v1 and v2 kept the same
“body” key (luckily for us)

heitorlessa:
Update based on discussions we had on Slack

## JMESPath as a project dependency

Initially started as an optional dependency, but we'll need a standard across Powertools as customers request an utility to extract as well as validate their events against JSONSchema. 

JMESPath is literally 24KB as it uses AST and a Lexer, so there is no impact on an additional dep, nor at runtime. It also provides the flexibility customers want to use their own JMESPath expressions to work with their events in the same way they're used with AWS CLI and other related projects -- Making it a standard.

## Custom function to deserialize base64 data

Kinesis is a popular event sources and customers will not be able to validate JSON payloads without decoding it first - This problem also occurs when working with the event, where customers need to create a `for loop` and decode the event before working with it.

We'll create a custom JMESPath function named `powertools_base64` to help address this while making it easier to maintain and remove complexity from customers. They'll be able to either use `envelope=envelopes.KINESIS_DATA_STREAM` or use the function for any base64 payload via `envelope=powertools_base64(key)`

@nmoutschen @cakepietoast need your public sign-off to this for historical reasons :)

nmoutschen:
If you meant `envelope="powertools_base64(key)"` then I'm OK with it :)

cakepietoast:
> @nmoutschen @cakepietoast need your public sign-off to this for historical reasons :)

You have it!

brysontyrrell:
This is shockingly similar to code we wrote for our APIs to handle schema validations.

Question on use as a decorator: it looks like there's nothing to handle validator exceptions so a 4xx response could be returned instead of a 5xx. Is the intended use case here to use middleware to wrap the handler so those exceptions can be handled?

nmoutschen:
@brysontyrrell This utility should support both sync and async workflows. The error handling would be different between the two. And for sync, it could be AppSync, ALB, API Gateway, etc. where the return values might vary.

I think the use-case where someone would like to use the decorator over the handler is for cases where the result should be the Lambda execution failing. In async cases, this would trigger the OnFailure/DLQ/etc. flow.

For API Gateway cases, I think this could be done like this:

```python
def handler(evt, ctx):
    try:
        validate(event=evt, schema=hello_schema, envelope="powertools_json(data).payload")
    except ValidationException:
        return {
            "statusCode": 400,
            "Invalid payload"
        }
```

The ValidationException would come from validation errors, while internal errors would result in an invocation failure.

aws_lambda_powertools/utilities/validation/validator.py
```diff
@@ -0,0 +1,40 @@
...
+
+    if inbound_schema:
+        logger.debug("Validating inbound event")
+        validate_data_against_schema(data=event, schema=inbound_schema)
+
+    response = handler(event, context)
+
+    if outbound_schema:
+        logger.debug("Validating outbound event")
+        validate_data_against_schema(data=event, schema=outbound_schema)
```
jplock, line 31: Should event be response here?

heitorlessa: Yep - Just finished writing tests and caught this ;)

docs/content/utilities/validation.mdx
```diff
@@ -0,0 +1,236 @@
...
+
+handler(event=sample_event, context={})
+```
+
+## Bring your own JMESPath function
+
+<Note type="warning">
+    This should only be used for advanced use cases where you have special formats not covered by the built-in functions.
+    <br/><br/>
+    This will <strong>replace all built-in functions provided with what you provide</strong>.
```
cakepietoast, line 214:
```suggestion
    This will <strong>replace all provided built-in functions such as `powertools_json`, so you will no longer be able to use them</strong>.
```

cakepietoast (approved):

PR #155, opened 2020-09-02
fix: batch processing util

cakepietoast:
**Issue #, if available:**

## Description of changes:

Simplified the documentation, primarily by focusing more on the SQS specific implementation. Added a sqs_batch_processor interface to simplify the common use case. Fixed an issue where no exception was being raised for failed messages, which would have caused failed messages to be deleted from the queue.

**Checklist**

 

* [X] Update tests
* [X] Update docs

## Breaking change checklist

**RFC issue #**:


aws_lambda_powertools/utilities/batch/middlewares.py
```diff
@@ -54,3 +56,58 @@ def batch_processor(
         processor.process()
 
     return handler(event, context)
+
+
+@lambda_handler_decorator
+def sqs_batch_processor(
```
heitorlessa, line 62: Suggestion: Bring this middleware under `sqs.py` as it's closer to its implementation.

aws_lambda_powertools/utilities/batch/middlewares.py
```diff
@@ -54,3 +56,58 @@ def batch_processor(
         processor.process()
 
     return handler(event, context)
+
```
heitorlessa, line 59: I can't seem to comment on the other `batch_processor` middleware... **suggestion** Bring the `batch_processor` under `base.py`

aws_lambda_powertools/utilities/batch/middlewares.py
```diff
@@ -54,3 +56,58 @@ def batch_processor(
...
+    * Async batch processors
+
+    """
+    config = config or Config()
+    processor = PartialSQSProcessor(config=config, suppress_exception=suppress_exception)
+
+    records = event["Records"]
+
+    with processor(records, record_handler):
+        processor.process()
```
heitorlessa, line 111: Can you add a `logger.debug` here to provide a message that you're processing another record. We don't need to log the record for security reasons, just a message suffices.

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -89,5 +90,6 @@ def failure_handler(self, record: Any, exception: Exception):
             "fail", exceptions args, original record
         """
         entry = ("fail", exception.args, record)
+        self.exceptions.append(exception)
```
heitorlessa, line 93: Can you add a `logger.debug` here? No need to log the record/data for security reasons.

docs/content/utilities/batch.mdx
```diff
@@ -1,73 +1,163 @@
...
 
-* Removal of successful messages for [AWS SQS](https://aws.amazon.com/sqs/) batch - in case of partial failure;
-* Build your own batch processor using the base classes.
+* Prevent succesfully processed messages being returned to SQS
+* Simple interface for individually processing messages from a batch
+* Build your own batch processor using the base classes
+
+**Background**
+
+When using SQS as a Lambda event source mapping, functions can be triggered with a batch of messages from SQS. If the Lambda function fails when processing the batch,
```
heitorlessa, line 18: Let's add some line breaks to make it easier to read. The docs theme kinda forces you to do that or else it cramps the lines

docs/content/utilities/batch.mdx
```diff
@@ -1,73 +1,163 @@
...
-Processor | Function/Method | IAM Permission
-|---------|-----------------|---------------|
-PartialSQSProcessor | `_clean` | `sqs:DeleteMessageBatch`
+There are 2 ways to use this utility for processing SQS messages:
 
-### PartialSQSProcessor
+**With a decorator:**
 
-SQS integration with Lambda is one of the most well established ones and pretty useful when building asynchronous applications. One common approach to maximize the performance of this integration is to enable the batch processing feature, resulting in higher throughput with less invocations.
+Using the `sqs_batch_processor` decorator with your lambda handler function, you provide a `record_handler` which is responsible for processing individual messages. It should raise an exception if
```
heitorlessa, line 37: Let's add line breaks here to make it easier to read in our docs theme

docs/content/utilities/batch.mdx
```diff
@@ -1,73 +1,163 @@
...
+    return_value = do_something_with(record["body"])
+    return return_value
+
+@sqs_batch_processor(record_handler=record_handler)
+def lambda_handler(event, context):
+    return {"statusCode": 200}
+```
 
-A *naive* approach to solving this problem is to delete successful records from the queue before redriving's phase. The `PartialSQSProcessor` class offers this solution both as context manager and middleware, removing all successful messages from the queue case one or more failures occurred during lambda's execution. Two examples are provided below, displaying the behavior of this class.
+**With a context manager:**
```
heitorlessa, line 56: Use a sub-heading `####` as opposed to bold text - It allows customers and us to easily reference it in conversations

heitorlessa (changes requested):
Suggestions to break `middlewares.py` functionalities into `base.py` and `sqs.py` to be closer to their respective implementations, debug logging, **and** docs to make it easier to read. 

I'll work on the docs as it's a quick one for me

aws_lambda_powertools/utilities/batch/middlewares.py
cakepietoast: There's no access to individual records here, but I've added logging elsewhere which should suffice.

cakepietoast: This is outside of the loop for processing single records, but I've added debug logging elsewhere which should suffice.

Issue #185, opened 2020-09-30
Case insensitive lookup of header values in http proxy events


michaelbrewer:
**Is your feature request related to a problem? Please describe.**

Http headers are meant to be case insensitive, but lambda events use case sensitive dicts for the header keys.

For example getting the `Authorization` header you will need to do a case insensitive look up to support if the header is `authorization`. So you end up doing something like this.

```python3
authorization = None
for key, value in self.headers.items():
    if name.lower() == 'authorization'
        authorization = value
```

**Describe the solution you'd like**

> Linked pull request : #185

Update `BaseProxyEvent.get_header_value` method to support case insensitive lookups by default

```python3
authorization = event.get_header_value("authorization")
```

```python3
    def get_header_value(
        self, name: str, default_value: Optional[str] = None, case_sensitive: Optional[bool] = False
    ) -> Optional[str]:
        """Get header value by name

        Parameters
        ----------
        name: str
            Header name
        default_value: str, optional
            Default value if no value was found by name
        case_sensitive: bool
            Whether to use a case sensitive look up
        Returns
        -------
        str, optional
            Header value
        """
        if case_sensitive:
            return self.headers.get(name, default_value)
        
        return next((value for key, value in self.headers.items() if name.lower() == key.lower()), default_value)
```

**Questions**

* Should we default to case insensitive look ups? This is what most people would want?
* Just return the first match? Or support returning all matches?

**Describe alternatives you've considered**

Use the requests CaseInsensitiveDict or port it :
* https://github.com/psf/requests/blob/333fa8d3f201f68b82d9e8f26be4987e869fd3bf/requests/structures.py#L15

**Additional context**

From the Http 1.1 spec, https headers are not supposed to be case sensitive:  https://www.w3.org/Protocols/rfc2616/rfc2616.html

PR #186, opened 2020-10-01
feat(data_classes): case insensitive header lookup

michaelbrewer:
**Issue #, if available:**
#185

## Description of changes:

Update BaseProxyEvent `get_header_value` method to be case insensitive
by default and add a flag for case sensitive lookups

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -48,7 +48,9 @@ class BaseProxyEvent(DictWrapper):
         params = self.query_string_parameters
         return default_value if params is None else params.get(name, default_value)
 
-    def get_header_value(self, name: str, default_value: Optional[str] = None) -> Optional[str]:
+    def get_header_value(
+        self, name: str, default_value: Optional[str] = None, case_sensitive: Optional[bool] = False
+    ) -> Optional[str]:
         """Get header value by name
 
         Parameters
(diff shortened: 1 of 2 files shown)


aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -48,7 +48,9 @@ def get_query_string_value(self, name: str, default_value: Optional[str] = None)
         params = self.query_string_parameters
         return default_value if params is None else params.get(name, default_value)
 
-    def get_header_value(self, name: str, default_value: Optional[str] = None) -> Optional[str]:
+    def get_header_value(
```
heitorlessa, line 51: Shouldn't this be available only for where there are `self.headers` instead of being a common method  
e.g. API GW?

michaelbrewer: > Shouldn't this be available only for where there are `self.headers` instead of being a common method
> e.g. API GW?

`BaseProxyEvent` is the base class for `APIGatewayProxyEvent`, `APIGatewayProxyEventV2`, and `ALBEvent` which all support `self.headers`. See test events `apiGatewayProxyEvent.json`, `apiGatewayProxyV2Event.json` and `albEvent.json`.

michaelbrewer: @heitorlessa ☝️  - does this make sense?

heitorlessa: It does now, thanks for clarifying! Could you then add that in the docstring so one doesn't inadvertently use it for AppSync, for example?

PR #188, opened 2020-10-03
test(parser): Add missing test coverage and lint changes

michaelbrewer:
**Issue #, if available:**

## Description of changes:

* Make some methods static
* Run code formatter
* Some lint based changes
* Add missing test cases

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/advanced_parser/envelopes/base.py
+++ b/aws_lambda_powertools/utilities/advanced_parser/envelopes/base.py
@@ -8,7 +8,8 @@ logger = logging.getLogger(__name__)
 
 
 class BaseEnvelope(ABC):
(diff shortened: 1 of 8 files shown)


michaelbrewer:
@heitorlessa - one thing I did pick up from the "schema" package was user specific examples. Should we move some of that into an examples or docs ?

heitorlessa:
> @heitorlessa - one thing I did pick up from the "schema" package was user specific examples. Should we move some of that into an examples or docs ?

They'll be in the docs. They're similar to the data classes, except they can be inherited and overridden so we can parse the entire payload + user defined models for autocompletion, etc. 

Here's the list of other changes to be made before we launch 1.7.0 with it - Coverage and linting were some of them, so big thank you ;): https://github.com/awslabs/aws-lambda-powertools-python/pull/118#issuecomment-702701896

PR #189, opened 2020-10-04
chore: ease maintenance of upcoming parser #118

heitorlessa:
**Issue #, if available:** #118 

## Description of changes:

Minor internal changes to ensure consistency across utilities, and ease maintenance before we write docs.

**Checklist**

 

* [ ] Update docs
* [x] Update tests
* [x] **Remove validators**. We need an universal schema owned by Lambda for each event source, or else we'll likely break some customers if a few services change, as we won't be notified.
* [x] **Snake_case**. IIRC, EventBridge `detailtype` was the only odd one out but I'll double check if there are others
* [x] **Kwargs over args**. Some internal code is using args over kwargs - Moving to kwargs make it easier to refactor, and order of args wouldn't be impacted.
* [x] **Docstrings**. Enhance Example section in docstrings, and areas where we need more clarity for future maintenance
* [x] **Raise own Validation Exception**. For customers wanting to catch that exception (e.g. custom middlewares) they'll have to find Pydantic exception to catch it properly
* [x] **Enhance imports**. Envelopes and a few others can be turned into high level imports to clean up customers' code, plus DX.
* [x] **Simplify internals of envelope**: Parsing dict event, obj event, and string event share similar logic; wrap them under a single `_parse`
* [x] **Fix PyCharm inspect issues**
* [x] **Test coverage**. Validation errors aren't being tested as of now.
* [x] **Refactor seemingly duplicates to ease maintenance**
* [x] **Chain exceptions with own exception to ease debugging**
* [x] **[Optional] Enhancing envelope selection**. Time allowing, I'll experiment with providing a similar experience we have in the `validator` envelope utility `envelope.<event_source>`.
* [x] **Revisit chain of try/catch**: Allow exceptions to propagate, and use exception chaining to provide more details instead.
* [x] **Expose Pydantic import system**: Common imports from Pydantic such as validator, root_validator, and BaseModel should be part of the high level import system. Uncommon imports such as json, dataclasses, should be available nevertheless but ring fenced.
* [x] **[Optional] Create a standalone parser**. Time allowing, I'll experiment creating a standalone `parser` function with the same functionality the decorator
* [x] **Remove flake8 polyfill**. Double check whether a fresh Python env hits the issue you hit with that, since we don't use it
* [ ] **[Optional] Support for API Gateway, Kinesis, S3, and SNS**. Likely to be after 1.7.0, but time allowing we can do that sooner.
* [x] **Revisit Schemas naming**. DynamoDBScheme vs DynamoDBSchema - might be best looked at when writing docs

> Note: Decided to refactor test event data as a separate PR to prevent cluttering non-parser changes here. Data_classes and validator utility are partially using the same events, so we'll refactor tests to provide events to all tests as fixtures.

## Breaking change checklist

**RFC issue #**:


heitorlessa:
@cakepietoast ready when you are -- this is mostly semantics to ease maintenance after publishing the parser utility -- Could you take a look and see if the docstrings, import system, etc. are to what you'd expect too?

The additional `conftest` for parser, test data from data_classes, etc. will be dealt with in a separate PR - We need to consolidate all those JSON into fixtures available to all functional tests to prevent unnecessary I/O + easier to use.

heitorlessa:
Next steps before 1.7.0: 

* Review by @cakepietoast 
* Parser docs

aws_lambda_powertools/utilities/parser/pydantic.py
```diff
@@ -0,0 +1,8 @@
+# Pydantic has many utilities that some advanced customers typically use.
+# Importing what's currently in the docs would likely miss something.
+# As Pydantic export new types, new utilities, we will have to keep up
+# with a project that's not used in our core functionalities.
+# For this reason, we're relying on Pydantic's __all__ attr to allow customers
+# to use `from from aws_lambda_powertools.utilities.parser.pydantic import <anything>`
```
Nr18, line 6:
```suggestion
# to use `from aws_lambda_powertools.utilities.parser.pydantic import <anything>`
```

aws_lambda_powertools/utilities/parser/parser.py
```diff
@@ -0,0 +1,156 @@
...
+        When schema given does not implement BaseModel
+    InvalidEnvelopeError
+        When envelope given does not implement BaseEnvelope
+    """
+    parsed_event = parse(event=event, schema=schema, envelope=envelope)
+    logger.debug(f"Calling handler {handler.__name__}")
+    return handler(parsed_event, context)
+
+
+def parse(event: Dict[str, Any], schema: BaseModel, envelope: Optional[BaseEnvelope] = None) -> Any:
```
heitorlessa, line 87: @cakepietoast tempted to change "schema" parameter name to "model", as it best reflect Pydantic models.

You can derive a Schema from a model (JSON Schema, OpenAPI).

aws_lambda_powertools/utilities/parser/pydantic.py
heitorlessa: Thanks for catching that ;)

aws_lambda_powertools/utilities/parser/parser.py
cakepietoast: I vote for changing it to model. Makes sense to stick to the Pydantic terminology unless we have good reason not to.

aws_lambda_powertools/utilities/parser/parser.py
```diff
@@ -0,0 +1,156 @@
...
+from ...middleware_factory import lambda_handler_decorator
+from ..typing import LambdaContext
+from .envelopes.base import BaseEnvelope
+from .exceptions import InvalidEnvelopeError, InvalidSchemaTypeError, SchemaValidationError
+
+logger = logging.getLogger(__name__)
+
+
+@lambda_handler_decorator
+def parser(
```
cakepietoast, line 15: How would you feel about renaming this to `event_parser`? That makes it less likely for people to accidentally do `from aws_lambda_powertools.utilities import parser` when they mean `from aws_lambda_powertools.utilities.parser import parser`

cakepietoast, line 15: How would you feel about renaming this to `event_parser`? Naming the module and the contained function the same makes it easier for people to make a mistake and accidentally do `from aws_lambda_powertools.utilities import parser` when they mean `from aws_lambda_powertools.utilities.parser import parser`

heitorlessa: Good point! Perhaps "...utilities.**parsing** import parser" ?

event_parser wouldn't do justice to its features IMHO, as I can parse any str/obj into classes, enforce type hints at runtime, etc

cakepietoast: I think event_parser makes sense given that this decorator will always have a lambda event as its input. Happy to disagree and commit though if you prefer renaming the module to parsing! Both are better than having `parser.parser` as the path.

heitorlessa: Ah yes, didn't consider it from that angle - I'll rename it to event_parser

PR #192, opened 2020-10-14
docs: new parser utility

heitorlessa:
**Issue #, if available:** #118 

## Description of changes:

Documentation for upcoming advanced parser.

Rendered doc: https://github.com/heitorlessa/aws-lambda-powertools-python/blob/docs/parser/docs/content/utilities/parser.mdx

**Checklist**

 

* [x] Update tests
* [x] Update docs
* [x] Document key features
* [x] Document how to install additional dependencies (`pydantic`, `typing_extension`)
* [x] Document `event_parser` decorator
* [x] Document `parse` standalone function
* [x] Document models vs envelopes terminology
* [x] Document built-in envelopes
* [x] Document inheriting models to access all their properties e.g. EventBridgeModel
* [x] Document how to use `root_validator` and `validator`
* [x] Document exception handling
* [x] Document cold start impact
* [x] Review by Ran (utility author)
* [x] If possible, get a review by Koudai
* [x] revert: Use Pydantic's `ValidationError` as opposed to ours given its utility functions `json()`, `errors()`, etc. Original implementation also provided a better experience for those migrating existing Pydantic models and validations, which I can only see why not.

## Breaking change checklist

**RFC issue #**:


docs/content/utilities/parser.mdx
```diff
@@ -0,0 +1,349 @@
...
+# Parsed model only contains our actual model, not the entire EventBridge + Payload parsed
+assert ret.password1 == ret.password2
+```
+
+**What's going on here, you might ask**:
+
+1. We imported built-in `models` from the parser utility
+2. Used `envelopes.EventBridgeModel` as the envelope for our `UserModel` model
+3. Parser parsed the original event against the EventBridge model
+4  Parser then parsed the `detail` key using `UserModel`
```
Nr18, line 276:
```suggestion
4. Parser then parsed the `detail` key using `UserModel`
```

heitorlessa: Thanks as always @Nr18 ;)

Nr18: So did not comment on it because I was not sure but my assumption is that the parser that is used to build the documentation can handle the `3.1. This confirms that data received conforms with how EventBridge wraps events` notation used on line 348?

Because GitHub is rending it incorrect

heitorlessa:
hey @risenberg-cyberark @koxudaxi - Would appreciate if you could review the new parser's docs before we release it. 

Parser is a thin wrapper on top of Pydantic

docs/content/utilities/parser.mdx
heitorlessa: Aha! good catch, I can see that the malformed rendering now - Merged the two in a single sentence now :) Thank you!

aws_lambda_powertools/utilities/parser/envelopes/sqs.py
```diff
@@ -19,7 +19,7 @@ class SqsEnvelope(BaseEnvelope):
     all items in the list will be parsed as str and npt as JSON (and vice versa)
     """
 
-    def parse(self, data: Dict[str, Any], model: Union[BaseModel, str]) -> List[Union[BaseModel, str]]:
+    def parse(self, data: Dict[str, Any], model: Union[BaseModel, str]) -> List[BaseModel]:
```
koxudaxi, line 22: The abstract method doesn't accept `str` for the model arguments. 
I guess the model is not `BaseModel` `Type[BaseModel]`
Also, I must recommend using TypeVar for parse methods and _parse method.

example
```python
Model = TypeVar("Model", bound=BaseModel)
...
def parse(self, data: Dict[str, Any], model: Type[Model]) -> List[Model]:
```
I found a good example of this case in this repo.
https://github.com/heitorlessa/aws-lambda-powertools-python/blob/9e4286399527b32a0cc2fee44362ea4864e46383/aws_lambda_powertools/utilities/parser/parser.py#L89

aws_lambda_powertools/utilities/parser/envelopes/sqs.py
```diff
@@ -34,8 +34,10 @@ def parse(self, data: Dict[str, Any], model: Union[BaseModel, str]) -> List[Unio
         List
             List of records parsed with model provided
         """
+        logger.debug(f"Parsing incoming data with SQS model {SqsModel}")
         parsed_envelope = SqsModel(**data)
         output = []
+        logger.debug(f"Parsing SQS records in `body` with {model}")
         for record in parsed_envelope.Records:
             output.append(self._parse(record.body, model))
```
koxudaxi, line 42: self._parse can not get `str` type as a model argument.

aws_lambda_powertools/utilities/parser/envelopes/dynamodb.py
```diff
@@ -32,8 +32,10 @@ def parse(self, data: Dict[str, Any], model: BaseModel) -> List[Dict[Literal["Ne
         List
             List of records parsed with model provided
         """
+        logger.debug(f"Parsing incoming data with DynamoDB Stream model {DynamoDBStreamModel}")
         parsed_envelope = DynamoDBStreamModel(**data)
```
koxudaxi, line 36: It's better to use `DynamoDBStreamModel.parse_obj(data)`.

aws_lambda_powertools/utilities/parser/parser.py
```diff
@@ -84,7 +86,7 @@ def handler(event: Order, context: LambdaContext):
     return handler(parsed_event, context)
 
 
-def parse(event: Dict[str, Any], model: BaseModel, envelope: Optional[BaseEnvelope] = None) -> Any:
+def parse(event: Dict[str, Any], model: Type[Model], envelope: Optional[Type[Envelope]] = None) -> Any:
```
koxudaxi, line 89: Will this method return `Any` object?
If the method only returns the parsed model then the type should be `Model`.

koxudaxi (commented):
@heitorlessa 
Thank you for requesting a review.
I wrote some comments.
I think `TypeVar` can support that users decide types of models.
I guess `_parse` and `parse` methods should have TypeVar arguments for Models.

heitorlessa:
> @heitorlessa 
> 
> Thank you for requesting a review.
> 
> I wrote some comments.
> 
> I think `TypeVar` can support that users decide types of models.
> 
> I guess `_parse` and `parse` methods should have TypeVar arguments for Models.

All great comments; I'll reply to them but first a question @koxudaxi:

What's the value in using parser with "str" as a model?

If nothing will be parsed, why not use the data class utility we provide?

We had support for str in the original implementation by Ran but I struggled to see the value if you know the value will be just a str - I might me missing something hence the question 

Thanks a lot everyone

aws_lambda_powertools/utilities/parser/envelopes/sqs.py
heitorlessa: Ah yes! That's leftover from the refactor - Addressing this now

aws_lambda_powertools/utilities/parser/envelopes/sqs.py
heitorlessa: Reverted to kwargs to avoid confusion - `output.append(self._parse(data=record.body, model=model))`.

heitorlessa:
All addressed @koxudaxi and I've just learned we don't need the additional `Type[Model]`, `Model` does it just fine according to PyCharm Code inspection feature.

aws_lambda_powertools/utilities/parser/envelopes/base.py
```diff
@@ -11,14 +11,14 @@ class BaseEnvelope(ABC):
     """ABC implementation for creating a supported Envelope"""
 
     @staticmethod
-    def _parse(data: Union[Dict[str, Any], str], model: BaseModel) -> Any:
+    def _parse(data: Union[Dict[str, Any], str], model: Model) -> Union[Model, None]:
```
koxudaxi, line 14: I guess `Model` argument is `Type[Model]`. these parse methods need a defined Model class. It does not need to be instanced.
I found a near example in the official repo.
https://github.com/samuelcolvin/pydantic/blob/29e3877a448e9cb37e058c0a1e10e47a61676b18/pydantic/tools.py#L33

koxudaxi:
> All addressed @koxudaxi and I've just learned we don't need the additional Type[Model], Model does it just fine according to PyCharm Code inspection feature.

@heitorlessa 
It's great.
Also, I'm developing a PyCharm plugin of Pydantic. https://github.com/koxudaxi/pydantic-pycharm-plugin
About the plugin is written in an official pydantic document too:smile_cat:

aws_lambda_powertools/utilities/parser/envelopes/base.py
koxudaxi: Also, I recommend to run mypy checking types.

docs/content/utilities/parser.mdx
```diff
@@ -0,0 +1,508 @@
...
+    id: int
+    description: str
+    items: List[OrderItem] # nesting models are supported
+    optional_field: Optional[str] # this field may or may not be available when parsing
+```
+
+These are simply Python classes that inherit from BaseModel. **Parser** enforces type hints declared in your model at runtime.
+
+<!-- TODO: We need a dedicated section to expand on these great features from Pydantic -->
+<!-- The advantage here is that they can be [recursive, dumped as JSON, JSON Schema, Dicts, have validation and more](https://pydantic-docs.helpmanual.io/usage/models/). -->
```
risenberg-cyberark, line 52: yeah, root_validator & validator are very useful.

docs/content/utilities/parser.mdx
```diff
@@ -0,0 +1,508 @@
...
+    "items": [
+        {
+            "id": 1015938732,
+            "quantity": 1,
+            "description": "item xpto"
+        }
+    ]
+}
+
+@event_parser(model=Order) # highlight-line
```
risenberg-cyberark, line 172: what about an example of model + envelope together? that's the most powerful & easy to use combination.

risenberg-cyberark (changes requested):

docs/content/utilities/parser.mdx
risenberg-cyberark: other than that, amazing job to document this feature and pydantic

aws_lambda_powertools/utilities/parser/envelopes/base.py
heitorlessa: Address that and other findings with mypy - Thanks for the suggestion :) 

I'll create a maintenance issue to discuss with the core maintainers on the feasibility of adding it as part of the code checks - Reason we didn't add earlier was to lower the bar for contributions.

docs/content/utilities/parser.mdx
heitorlessa: Made 3 improvements there to address your feedback ;)

- Expanded that section to include `event_parser` using envelope param
- Explicit call out that both `event_parser` and `parse` support envelope param
- Note to call out `event_parser` will replace the event with the parsed model, if successful, in case someone is trying to use it with other nested decorators

heitorlessa:
Thanks a lot @risenberg-cyberark and @koxudaxi for the review - I'm merging this now with all feedbacks being addressed, and will work on the 1.7.0 release tomorrow by EOD to include this publicly ;)

Issue #206, opened 2020-11-15
Feature: Add sns notification support to Parser utility


risenberg-cyberark:
Validation utility support sns notification schema, while the parser does not.
This issue will add the requires schema, envelope and tests for sns notification events.

heitorlessa:
Resolving as it's now merged in Develop 🎉

PR #207, opened 2020-11-15
feat: Add sns notification support to Parser utility #206

risenberg-cyberark:
**Issue #, if available:**

## Description of changes:

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/parser/envelopes/sns.py
+++ b/aws_lambda_powertools/utilities/parser/envelopes/sns.py
@@ -1,0 +1,42 @@
+import logging
+from typing import Any, Dict, List, Optional, Union
+
+from ..models import SnsModel
+from ..types import Model
+from .base import BaseEnvelope
+
+logger = logging.getLogger(__name__)
+
+
+class SnsEnvelope(BaseEnvelope):
+    """SNS Envelope to extract array of Records
+
+    The record's body parameter is a string, though it can also be a JSON encoded string.
+    Regardless of its type it'll be parsed into a BaseModel object.
+
+    Note: Records will be parsed the same way so if model is str,
+    all items in the list will be parsed as str and npt as JSON (and vice versa)
+    """
+
+    def parse(self, data: Optional[Union[Dict[str, Any], Any]], model: Model) -> List[Optional[Model]]:
+        """Parses records found with model provided
+
+        Parameters
+        ----------
+        data : Dict
+            Lambda event to be parsed
+        model : Model
+            Data model provided to parse after extracting data using envelope
+
+        Returns
+        -------
+        List
+            List of records parsed with model provided
+        """
+        logger.debug(f"Parsing incoming data with SNS model {SnsModel}")
+        parsed_envelope = SnsModel.parse_obj(data)
+        output = []
+        logger.debug(f"Parsing SNS records in `body` with {model}")
+        for record in parsed_envelope.Records:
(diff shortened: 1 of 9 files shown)


aws_lambda_powertools/utilities/parser/envelopes/sns.py
```diff
@@ -0,0 +1,42 @@
...
+        -------
+        List
+            List of records parsed with model provided
+        """
+        logger.debug(f"Parsing incoming data with SNS model {SnsModel}")
+        parsed_envelope = SnsModel.parse_obj(data)
+        output = []
+        logger.debug(f"Parsing SNS records in `body` with {model}")
+        for record in parsed_envelope.Records:
+            output.append(self._parse(data=record.Sns.Message, model=model))
```
heitorlessa, line 41: Getting confirmation from SNS team whether there could be more than one record when using SNS as an event source for Lambda. 

SNS FAQ says otherwise despite the Records schema being an actual array - I'll revert back with an approval once I have it.

```
Q: Will a notification contain more than one message?
No, all notification messages will contain a single published message.
```

aws_lambda_powertools/utilities/parser/models/sns.py
```diff
@@ -0,0 +1,36 @@
+from datetime import datetime
+from typing import Dict, List, Optional
+
+from pydantic import BaseModel
+from pydantic.networks import HttpUrl
+from typing_extensions import Literal
+
+
+class SqsMsgAttributeModel(BaseModel):
```
heitorlessa, line 9: Meant SNS instead of SQS?

```suggestion
class SnsMsgAttributeModel(BaseModel):
```

aws_lambda_powertools/utilities/parser/envelopes/sns.py
heitorlessa: Resolving it - SNS confirmed that they might change this in the future, let's keep the implementation as-is on the record array.

aws_lambda_powertools/utilities/parser/models/sns.py
```diff
@@ -0,0 +1,36 @@
...
+    Type: str
+    Value: str
+
+
+class SnsNotificationModel(BaseModel):
```
heitorlessa, line 14: We need `groupId` and `dedupId` optional fields that might be added when FIFO is enabled - Haven't tried yet myself.

Publisher reference: https://docs.aws.amazon.com/sns/latest/dg/fifo-topic-code-examples.html#fifo-topic-java-publish

heitorlessa (changes requested):
It looks good to me except FIFO if we want to be future proof

aws_lambda_powertools/utilities/parser/models/sns.py
risenberg-cyberark: fixed!

aws_lambda_powertools/utilities/parser/models/sns.py
heitorlessa: Resolving it as @risenberg-cyberark confirmed Lambda cannot directly subscribe to SNS FIFO topics, it'd have to go through SQS FIFO, in which case our model already covers it.

heitorlessa (approved):
Thanks for adding support to SNS @risenberg-cyberark :) much appreciated. Merging it now

Issue #224, opened 2020-11-29
feat: Add S3 lambda event support to Parser utility 


risenberg-cyberark:
This schema is currently supported in the validator utility.
This issue will add the schema to the parser utility.
Please note that there's no need for an envelope in this case, just a schema class.

heitorlessa:
Launched in 1.9.0 now

PR #225, opened 2020-11-29
feat: Add S3 lambda event support to Parser utility #224

risenberg-cyberark:
**Issue #, if available:**
feat: Add S3 lambda event support to Parser utility #224

Added schema files and tests to parser utility

### DIFF
--- a/aws_lambda_powertools/utilities/parser/models/__init__.py
+++ b/aws_lambda_powertools/utilities/parser/models/__init__.py
@@ -1,5 +1,6 @@
 from .dynamodb import DynamoDBStreamChangedRecordModel, DynamoDBStreamModel, DynamoDBStreamRecordModel
 from .event_bridge import EventBridgeModel
+from .s3 import S3Model, S3RecordModel
 from .ses import SesModel, SesRecordModel
 from .sns import SnsModel, SnsNotificationModel, SnsRecordModel
 from .sqs import SqsModel, SqsRecordModel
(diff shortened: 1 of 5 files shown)


heitorlessa (changes requested):
Awesome - two questions as with the other PRs:

1. Could you please update the docs on this new model?
2. Do we want to cover S3 Batch Operations too?

Docs portion: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/docs/content/utilities/parser.mdx#extending-built-in-models

On 2, here's the example payload which seems pretty quick to create a model from: https://docs.aws.amazon.com/lambda/latest/dg/services-s3-batch.html

risenberg-cyberark:
> Awesome - two questions as with the other PRs:
> 
> 1. Could you please update the docs on this new model?
> 2. Do we want to cover S3 Batch Operations too?
> 
> Docs portion: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/docs/content/utilities/parser.mdx#extending-built-in-models
> 
> On 2, here's the example payload which seems pretty quick to create a model from: https://docs.aws.amazon.com/lambda/latest/dg/services-s3-batch.html

Sure, i'll try my best

heitorlessa:
thanks a lot as always @risenberg-cyberark - As discussed, I'll commit these suggestions, and will try pushing the docs to your branch then merge. Failing that, I'll merge as-is, then send another PR with docs and suggestions implemented ;)

ETA: This Friday EOD

heitorlessa (approved):
Docs added - Merging

PR #229, opened 2020-11-30
feat: Add alb lambda event support to Parser utility

risenberg-cyberark:
feat: Add alb lambda event support to Parser utility #228

### DIFF
--- a/aws_lambda_powertools/utilities/parser/models/__init__.py
+++ b/aws_lambda_powertools/utilities/parser/models/__init__.py
@@ -1,3 +1,4 @@
+from .alb import AlbModel, AlbRequestContext, AlbRequestContextData
 from .dynamodb import DynamoDBStreamChangedRecordModel, DynamoDBStreamModel, DynamoDBStreamRecordModel
 from .event_bridge import EventBridgeModel
 from .ses import SesModel, SesRecordModel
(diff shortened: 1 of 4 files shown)


heitorlessa (changes requested):
Looks great!! Two questions:

1. Could you update the docs too?
2. Do we need an envelope for the `body` field?

Here's the line where we tell customers what models Parser offers: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/docs/content/utilities/parser.mdx#extending-built-in-models

risenberg-cyberark:
> Looks great!! Two questions:
> 
> 1. Could you update the docs too?
> 2. Do we need an envelope for the `body` field?
> 
> Here's the line where we tell customers what models Parser offers: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/docs/content/utilities/parser.mdx#extending-built-in-models

I didnt add the body parsing because i suppose in most cases you might want to use also the other params in the model so using an envelope would not be helpful.

heitorlessa:
thanks a lot as always @risenberg-cyberark - As discussed, I'll commit these suggestions, and will try pushing the docs to your branch then merge. Failing that, I'll merge as-is, then send another PR with docs and suggestions implemented ;)

ETA: This Friday EOD

heitorlessa (approved):
Doc added - Merging

PR #233, opened 2020-12-04
improv: Added __eq__ function to DictWrapper for better equality checks

GroovyDan:
**Issue #, if available:**

## Description of changes:

This change is to allow easier testing of data classes by adding an __eq__ function to the base DictWrapper class. This new function will check the equality of the data contained in the object opposed to the memory location.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -12,7 +12,7 @@ class DictWrapper:
 
     def __eq__(self, other: Any) -> bool:
         if not isinstance(other, DictWrapper):
-            return NotImplemented
+            return False
 
         return self._data == other._data
 
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -10,6 +10,12 @@ class DictWrapper:
     def __getitem__(self, key: str) -> Any:
         return self._data[key]
 
+    def __eq__(self, other):
+        if not isinstance(other, DictWrapper):
+            return NotImplemented
+
+        return self._data == other._data
+
     def get(self, key: str) -> Optional[Any]:
         return self._data.get(key)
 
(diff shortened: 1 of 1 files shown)


heitorlessa:
hey @GroovyDan thanks a lot for taking the time in making your first contribution, we appreciate it very much :)

LGTM at first glance and can see how it can make tests easier (it'd be great to have one test updated to cover it too). However, I need to triple check the side effects of not having `__hash__`, as this will make the class not hashable too. 

what do you think @michaelbrewer?

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -10,6 +10,12 @@ def __init__(self, data: Dict[str, Any]):
     def __getitem__(self, key: str) -> Any:
         return self._data[key]
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, DictWrapper):
+            return NotImplemented
```
Nr18, line 15:
```suggestion
            return False
```

Shouldn't the `__eq__` method just return `False` instead of a `NotImplemented`? If the Type is not a `DictWrapper` it's never equal hence it should return False?

Nr18 (changes requested):
@GroovyDan nice addition indeed, it did raise a question while looking at it though.

aws_lambda_powertools/utilities/data_classes/common.py
heitorlessa: It should indeed return a bool not an Exception.

heitorlessa (changes requested):
One small change required as per @Nr18 pointed out

I've also double checked about equality and hash theory. `__hash__` won't be necessary because Dicts are not hashable, so `DictWrapper` not being hashable with `__eq__` isn't an issue in this case.

heitorlessa (approved):
Thanks a lot for making the change @GroovyDan - Merging this now.

PR #235, opened 2020-12-07
chore: implement phony targets correctly

Nr18:
**Issue #, if available:**

## Description of changes:

From the manual of make

> A phony target is one that is not really the name of a file; rather it is just a name for a recipe to be executed when you make an explicit request. There are two reasons to use a phony target: to avoid a conflict with a file of the same name, and to improve performance.

See: https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+.PHONY: target dev dev-docs format lint test coverage-html pr  build build-docs build-docs-api build-docs-website
+.PHONY: docs-local docs-api-local security-baseline complexity-baseline release-prod release-test release
 
 target:
 	@$(MAKE) pr


heitorlessa:
I need to have a read on the Phony targets more carefully, then merge this as it seems like a no brainier - But first I need to make sure I understand how this improves performance OR if this is to avoid filename conflicts (do we have any?)

Nr18:
The main reason why I always use it because I had conflicts where I accidentally created a file that had the same name as a target and then you get a "Nothing to do" message from Make as the file is already there. `.PHONY` will prevent that by always executing the target.

So after hitting that one and a good amount of troubleshooting I am always adding those phony definitions to prevent it from happening again.

cakepietoast (approved):

heitorlessa:
Understood, thanks a lot @Nr18 !!

Issue #187, opened 2020-10-01
Add AppConfig parameter provider


thomasklinger1234:
**Is your feature request related to a problem? Please describe.**
Currently, the parameter provider supports SSM, SecretsManager and DynamoDB. However, there is another configuration service which currently is not implemented: [AppConfig](https://docs.aws.amazon.com/appconfig/latest/userguide/what-is-appconfig.html). 

**Describe the solution you'd like**
Create a native implementation for `AppConfigProvider`. It will fetch the current configuration version based on application, environment and (optionally) client configuration version. If the client is not set, it should default to `POWERTOOLS_SERVICE_NAME`.  

**NOTE** For the `_get_multiple` function I do not have a clever semantic mapping for AppConfig. This is something that should be discussed.

For DynamoDB this is multiple items, for SSM multiple parameters. Some considerations for multiple parameters:

- Multiple environments
- Multiple keys extracted from the underlying JSON object in AppConfig using something like JSONpath (or JMESPath) for extracting the values

**Describe alternatives you've considered**
I have implemented it on my own following https://awslabs.github.io/aws-lambda-powertools-python/utilities/parameters/#create-your-own-provider but it would be nice to have this in powertools integrated.

**Additional context**

- AppConfig boto3 API: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/appconfig.html

nmoutschen:
Hi @thomasklinger1234 !

Sounds interesting. Thinking about how the developer experience would be like, AppConfig has a few moving parts compared to the other solutions.

To retrieve a value, AppConfig needs the following parameters: Application, Environment, Configuration and ClientId. Semantically speaking, Configuration is the configuration name to get, equivalent to the name of the parameter when we do `get_parameter`. However, the combination between Application/Environment/Configuration is a bit trickier.

I think Application should correspond to `POWERTOOLS_SERVICE_NAME`, and Environment should be the deployment environment (dev, test, prod, feat-123, etc.). This would correspond to the way these resources are created within AppConfig: you need to have an ApplicationId to create an Environment, and you need an ApplicationId too to create a configuration profile.

For ClientId, the [Lambda extension for AppConfig](https://aws.amazon.com/blogs/mt/introducing-aws-appconfig-lambda-extension-deploying-application-configuration-serverless/) dynamically generate a new ClientId for every Lambda execution environment. So if a function has 7 execution environment at a given time, that would be 7 different client ids.

## Developer experience

[...]

PR #236, opened 2020-12-08
feat: Add AppConfig parameter provider

risenberg-cyberark:
#187 
this is a poc for the app conf provider.
there's a lot of work in the docs department and in the tests.
There's just one test at the moment but it works. 
There's also the issue of handling yaml confs. This might require a new transform with pyyaml package maybe.

Is the UX ok?

### DIFF
--- a/aws_lambda_powertools/utilities/parameters/__init__.py
+++ b/aws_lambda_powertools/utilities/parameters/__init__.py
@@ -4,6 +4,7 @@
 Parameter retrieval and caching utility
 """
 
+from .appconfig import AppConfigProvider, get_app_config
 from .base import BaseProvider
 from .dynamodb import DynamoDBProvider
 from .exceptions import GetParameterError, TransformParameterError
@@ -11,12 +12,14 @@ from .secrets import SecretsProvider, get_secret
 from .ssm import SSMProvider, get_parameter, get_parameters
 
 __all__ = [
+    "AppConfigProvider",
     "BaseProvider",
     "GetParameterError",
     "DynamoDBProvider",
     "SecretsProvider",
     "SSMProvider",
     "TransformParameterError",
+    "get_app_config",
     "get_parameter",
     "get_parameters",
     "get_secret",
--- a/aws_lambda_powertools/utilities/parameters/appconfig.py
+++ b/aws_lambda_powertools/utilities/parameters/appconfig.py
@@ -1,0 +1,158 @@
+"""
+AWS App Config configuration retrieval and caching utility
+"""
+
+
+import os
+from typing import Dict, Optional, Union
+from uuid import uuid4
+
+import boto3
+from botocore.config import Config
+
+from .base import DEFAULT_PROVIDERS, BaseProvider
+
+CLIENT_ID = str(uuid4())
+
+
+class AppConfigProvider(BaseProvider):
+    """
+    AWS App Config Provider
+
+    Parameters
+    ----------
+    environment: str
+        Environment of the configuration to pass during client initialization
+    application: str, optional
+        Application of the configuration to pass during client initialization
+    config: botocore.config.Config, optional
+        Botocore configuration to pass during client initialization
+
+    Example
+    -------
+    **Retrieves the latest configuration value from App Config**
+
+        >>> from aws_lambda_powertools.utilities.parameters import AppConfigProvider
+        >>> appconf_provider = parameters.AppConfigProvider(environment="my_env", application="my_app")
+        >>>
+        >>> value : bytes = appconf_provider.get("my_conf")
+        >>>
+        >>> print(value)
+        My configuration value
+
+    **Retrieves a configuration value from App Config in another AWS region**
+
+        >>> from botocore.config import Config
+        >>> from aws_lambda_powertools.utilities.parameters import AppConfigProvider
+        >>>
+        >>> config = Config(region_name="us-west-1")
+        >>> appconf_provider = parameters.AppConfigProvider(environment="my_env", application="my_app", config=config)
+        >>>
+        >>> value : bytes = appconf_provider.get("my_conf")
+        >>>
+        >>> print(value)
+        My configuration value
+
+    """
+
+    client = None
+
+    def __init__(
+        self, environment: str, application: Optional[str] = None, config: Optional[Config] = None,
+    ):
+        """
+        Initialize the App Config client
+        """
+
+        config = config or Config()
+        self.client = boto3.client("appconfig", config=config)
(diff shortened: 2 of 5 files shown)


aws_lambda_powertools/utilities/parameters/__init__.py
```diff
@@ -4,19 +4,22 @@
...
 
 __all__ = [
+    "AppConfigProvider",
     "BaseProvider",
     "GetParameterError",
     "DynamoDBProvider",
     "SecretsProvider",
     "SSMProvider",
     "TransformParameterError",
+    "get_configuration",
```
nmoutschen, line 22: I'm wondering if this could cause confusion. Should this be `get_app_config`?

aws_lambda_powertools/utilities/parameters/appconf.py
```diff
@@ -0,0 +1,185 @@
...
+        >>> from aws_lambda_powertools.utilities.parameters import SSMProvider
+        >>> ssm_provider = SSMProvider()
+        >>>
+        >>> values = ssm_provider.get_multiple("/my/path/prefix", MaxResults=10)
+        >>>
+        >>> for key, value in values.items():
+        ...     print(key, value)
+        /my/path/prefix/a   Parameter value a
+        /my/path/prefix/b   Parameter value b
+        /my/path/prefix/c   Parameter value c
```
nmoutschen, line 76: These examples use the SSMProvider instead of AppConfigProvider.

aws_lambda_powertools/utilities/parameters/appconf.py
```diff
@@ -0,0 +1,185 @@
...
+import os
+from typing import Dict, Optional, Union
+from uuid import uuid4
+
+import boto3
+from botocore.config import Config
+
+from .base import DEFAULT_PROVIDERS, BaseProvider
+
+CLIENT_ID = str(uuid4())
```
nmoutschen, line 15: __Question__: should we keep this configurable when initializing the class?

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -230,6 +231,8 @@ def transform_value(value: str, transform: str, raise_on_transform_error: bool =
             return json.loads(value)
         elif transform == TRANSFORM_METHOD_BINARY:
             return base64.b64decode(value)
+        elif transform == TRANSFORM_METHOD_YAML:
+            return value  # todo
```
nmoutschen, line 235: __For tracking__: TODO

nmoutschen (changes requested):
Hey Ran!

Thanks a lot for your PR! Overall, from an UX perspective, I just have two main questions:

* Should we rename the helper function to `get_app_config()` instead?
* Should we allow developers to control the client ID?

I've also flagged 2-3 things just for tracking purposes (e.g examples, a todo).

aws_lambda_powertools/utilities/parameters/__init__.py
risenberg-cyberark: sure, good idea.

aws_lambda_powertools/utilities/parameters/appconf.py
risenberg-cyberark: yep, i know, i wrote that this is a draft pr. need to sort out the comments, docs and tests. Just wanted to know if the general direction/implementation is fine.

aws_lambda_powertools/utilities/parameters/appconf.py
risenberg-cyberark: is that auditable in some way ? if not, i suppose it's best to keep things simple.

aws_lambda_powertools/utilities/parameters/base.py
risenberg-cyberark: @nmoutschen 
this might mean increasing the package size by adding pyyaml or soemthing of the sort. is that ok with you guys?

nmoutschen: Fair point - I'm wondering how frequently is YAML used and if there's any need for it. Since this is modular, I think we should keep it for another RFC unless there's a strong need to do it for AppConfig.

Size wise, PyYAML is 444KB, versus 62MB for Powertools with all dependencies (with 52MB for Botocore).

aws_lambda_powertools/utilities/parameters/appconf.py
nmoutschen: I think this is a good default, but I'm wondering if there could be cases where someone might want to share a client ID across multiple execution environment.

risenberg-cyberark:
@heitorlessa @nmoutschen added docs and more tests. Please review :)

aws_lambda_powertools/utilities/parameters/appconfig.py
```diff
@@ -0,0 +1,158 @@
...
+        >>>
+        >>> print(value)
+        My configuration's JSON value
+    """
+
+    # Only create the provider if this function is called at least once
+    if "appconfig" not in DEFAULT_PROVIDERS:
+        DEFAULT_PROVIDERS["appconfig"] = AppConfigProvider(environment=environment, application=application)
+
+        sdk_options["ClientId"] = CLIENT_ID
```
nmoutschen, line 156: This will only run during the first call. Is this over-indented?

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -230,6 +231,8 @@ def transform_value(value: str, transform: str, raise_on_transform_error: bool =
             return json.loads(value)
         elif transform == TRANSFORM_METHOD_BINARY:
             return base64.b64decode(value)
+        elif transform == TRANSFORM_METHOD_YAML:
+            raise NotImplementedError
```
nmoutschen, line 235: We're adding the `TRANSFORM_METHOD_YAML` on line 20, but the implementation is not defined. Should we remove both and put it in a different PR at a later stage?

nmoutschen (changes requested):
Hey Ran!

Overall looks good. I just have two remarks. On one, I think there's an indentation issue. On the second one, I think we should remove the mention of the yaml transform since it's not supported.

heitorlessa:
I know Ran is currently ill (wishing him a speedy recovery), so I'll wait until Friday morning before we send those fixes ourselves to ensure this gets shipped in 1.10.0 this week -- ;)

heitorlessa:
@nmoutschen can we merge to release 1.10 with this feat?

nmoutschen:
Yep! LGTM here. I'll let you do it since I pushed changes.

Issue #218, opened 2020-11-25
RFC: Function Idempotency Helper


igorlg:
## Key information

* RFC PR: #245 
* Related issue(s), if known: _<none>_
* Area: _Utilities_
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-python/#tenets): _Yes_

## Summary
[summary]: #summary

Helper to facilitate writing [**Idempotent**](https://en.wikipedia.org/wiki/Idempotence) Lambda functions.
The developer would specify (via JMESPath) which value from the **event** will be used as a **unique** execution identifier, then this helper would search a persistence layer (e.g. DynamoDB) for that ID; if present, get the return value and **skip** the function execution, otherwise run the function normally and **persist** the return value + execution ID.

## Motivation
[motivation]: #motivation

Idempotency is a very useful design characteristic of any system. It enables the seamless separation of successful and failed executions, and is particularly useful in Lambdas used by [AWS Step Functions](https://aws.amazon.com/step-functions/). It is also a design principle on the [AWS Well Architected Framework - Serverless Lens](https://docs.aws.amazon.com/wellarchitected/latest/serverless-applications-lens/wellarchitected-serverless-applications-lens.pdf)

Broader description of this idea can be found [here](https://aws.amazon.com/premiumsupport/knowledge-center/lambda-function-idempotent/)

## Proposal
[proposal]: #proposal

Define a Python Decorator `@idempotent` which would receive as arguments a) the JMESPath of the _event_ key to use as execution ID, b) {optional} storage backend configuration, e.g. DynamoDB table name, or ElasticSearch URL + Index).

This decorator would wrap the function execution in the following way (_pseudo-python_):

```python
from aws_lambda_powertools.itempotent import PersistenceLayer

def idempotent(func, event_key, persistence_config):
  def wrapper(*args, **kwargs):
    persistence = PersistenceLayer(persistence_config)
    key = jmespath.find(event_key, **kwargs['event'])
    persistence.find(key)

    if persistence.executed_successfully():
      return persistence.result()

    try:
      result = func(*args, **kwargs)
      persistence.save_success(key, result)
      return result
    except Exception => e:
      persistence.save_error(key, e)

  return wrapper
```

Usage then would be similar to:

```python
from aws_lambda_powertools.itempotent import itempotent

@idempotent(event_key='Event.UniqueId', persistence_config='dynamodb://lambda-idemp-table')
def handler(event, context):
  # Normal function code here
[...]

PR #245, opened 2020-12-16
feat: Idempotency helper utility

cakepietoast:
**Issue #, if available:** #218

## Description of changes:

Implementing RFC #218. I'll be working together with @igorlg on this.

**Todos**

* [x] Implement caching
* [x] Update caching implementation with LRU / max size
* [x] Review/replace error handling implementation (especially the pickle bit, which currently causes security baseline to fail the build)
* [x] Add unit tests and complete integration tests
* [x] Review/remove table creation logic/implementation
* [x] Add logging
* [x] Add docs
* [x] Add payload validation
* [x] Improve exception handling for unrecoverable errors from persistence layer

**Checklist**

 

* [x] Update tests
* [x] Update docs


aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> @idempotent(persistence=persistence_store)
+        >>> def handler(event, context):
+        >>>     return {"StatusCode": 200}
+    """
+
+    persistence_instance = persistence
+    try:
+        event_record = persistence_instance.get_record(event)
+    except ItemNotFoundError:
+        persistence_instance.save_inprogress(event=event)
```
michaelbrewer, line 57: @cakepietoast - A ItemAlreadyExistsError may be raised here, and it is not really a bad thing. 

For example 2 calls are made close to each other. The first call has not yet put in progress record and the second call gets ItemNotFoundError and tries to also put in a record.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,96 @@
...
+        >>> persistence_store = DynamoDBPersistenceLayer(event_key="body", table_name="idempotency_store")
+        >>>
+        >>> @idempotent(persistence=persistence_store)
+        >>> def handler(event, context):
+        >>>     return {"StatusCode": 200}
+    """
+
+    persistence_instance = persistence
+    try:
+        event_record = persistence_instance.get_record(event)
```
michaelbrewer, line 55: Alternatively you could call `save_inprogress` first and assume if the save was successful then process with the call and then save the result, otherwise load the latest version of what was saved. 

In most cases this will result in 2 calls (1 in progress save and a success save) to the db and not 3 calls (1 get, 1 in progress save and 1 success save).

```python3
try:
    persistence_instance.save_inprogress(event=event)
    return _call_lambda(handler=handler, persistence_instance=persistence_instance, event=event, context=context)

except ItemAlreadyExistsError:
    event_record = persistence_instance.get_record(event)

    if event_record.status == STATUS_CONSTANTS["EXPIRED"]:
        return _call_lambda(handler=handler, persistence_instance=persistence_instance, event=event, context=context)
    
    if event_record.status == STATUS_CONSTANTS["INPROGRESS"]:
        raise AlreadyInProgressError(
            f"Execution already in progress with idempotency key: "
            f"{persistence_instance.event_key}={event_record.idempotency_key}"
        )
    
    if event_record.status == STATUS_CONSTANTS["COMPLETED"]:
        return event_record.response_json_as_dict()
```

aws_lambda_powertools/utilities/idempotency/persistence.py
```diff
@@ -0,0 +1,560 @@
...
+from .cache_dict import LRUDict
+from .exceptions import IdempotencyValidationerror, InvalidStatusError, ItemAlreadyExistsError, ItemNotFoundError
+
+logger = logging.getLogger(__name__)
+
+STATUS_CONSTANTS = {
+    "NOTEXISTING": "DOESNOTEXIST",
+    "INPROGRESS": "INPROGRESS",
+    "COMPLETED": "COMPLETED",
+    "EXPIRED": "EXPIRED",
```
michaelbrewer, line 25: If we use TTL the record might be deleted by Dynamdob (or Redis). So i guess we could have a TTL record which is longer than the EXPIRED record?

michaelbrewer (commented):
Just some comments so far.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Agreed, this is a good optimization for the happy path. Addressed this in latest commit.

aws_lambda_powertools/utilities/idempotency/persistence.py
cakepietoast: Yep, the TTL set in DDB or Redis needs to be >= the setting in the persistence instance. I'll make a note to add that to the documentation.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,101 @@
...
+
+    try:
+        # We call save_inprogress first as an optimization for the most common case where no idempotent record already
+        # exists. If it succeeds, there's no need to call get_record.
+        persistence_store.save_inprogress(event=event)
+    except ItemAlreadyExistsError:
+        try:
+            event_record = persistence_store.get_record(event)
+        except ItemNotFoundError:
+            return _call_lambda(handler=handler, persistence_store=persistence_store, event=event, context=context)
```
michaelbrewer, line 61: @cakepietoast we might want to log a warning message here as this is an edge case where the record was deleted between the `save_inprogress` and `get_record` calls.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,101 @@
...
+
+        if event_record.status == STATUS_CONSTANTS["EXPIRED"]:
+            return _call_lambda(handler=handler, persistence_store=persistence_store, event=event, context=context)
```
michaelbrewer, line 64: @cakepietoast possibly log a debug/info message here, as this might be useful to known when debugging expired records

michaelbrewer: @cakepietoast also, should this not be an error?

aws_lambda_powertools/utilities/idempotency/idempotency.py
michaelbrewer: @cakepietoast This might also raise an error?

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,432 @@
...
+        exception
+            The exception raised by the lambda handler
+        """
+        data_record = DataRecord(idempotency_key=self._get_hashed_idempotency_key(event))
+
+        logger.debug(
+            f"Lambda raised an exception ({type(exception).__name__}). Clearing in progress record in persistence "
+            f"store for idempotency key: {data_record.idempotency_key}"
+        )
+        self._delete_record(data_record)
```
michaelbrewer, line 332: @cakepietoast  - i thought you wanted to save the record as completed, but with an error? 

This would mean any retries with the same idempotent key would not execute the lambda again.

michaelbrewer (commented):
@cakepietoast  I added some updated comments.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: I'll add a log statement. This code path is unlikely to be hit often, but I can think of a couple of scenarios where it will occur (data layer expiring records or another concurrent execution raising an exception and deleting the record). In both of those cases, I think it makes sense proceed with executing the lambda handler.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: I'll add a log statement. We don't want to raise an error if the idempotency record has expired - we should treat it the same as a missing record and proceed with executing the lambda handler.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
cakepietoast: That was how the initial implementation looked, but there were too many gotchas to work around. We'd have to break the contract of "an invocation with the same payload (idempotency key) gets the same response" which means complicating customers' client implementations or orchestration logic.

The solution as its implemented now is simple. If the lambda handler raises an unhandled exception, we will allow it to be invoked again without "protecting" it with idempotency logic. This happens because we delete the idempotency record for that key. If you don't want the handler to be called again on a retry, you can catch the exception in your lambda handler and return a response instead of raising an exception.

aws_lambda_powertools/utilities/idempotency/idempotency.py
pcolazurdo: Shouldn't a call similar to save_inprogress (with extra error handling) happen here before the call to the handler? In this case another concurrent call can happen and execute the handler before this one is recorded duplicating the call.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,109 @@
...
+        Lambda handler
+    persistence_store: BasePersistenceLayer
+        Instance of persistence layer
+    event
+        Lambda event
+    context
+        Lambda context
+    """
+    try:
+        handler_response = handler(event, context)
```
pcolazurdo, line 103: 1. Let's say that lambda timeouts before the handler returns, you may end up with with partial information in the persistence store but with an executed (or partially executed) operation. Probably this is an use case that's need to be clearly documented.
1. If the persistence store has a problem (i.e. throttling) you may end up with a situation where the operation was finished, the status is never recorded, so once the expiration time passes the operation may be called again, as there was no record this has already been properly executed.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,109 @@
...
+        >>> from aws_lambda_powertools.utilities.idempotency import idempotent, DynamoDBPersistenceLayer
+        >>>
+        >>> persistence_store = DynamoDBPersistenceLayer(event_key="body", table_name="idempotency_store")
+        >>>
+        >>> @idempotent(persistence_store=persistence_store)
+        >>> def handler(event, context):
+        >>>     return {"StatusCode": 200}
+    """
+
+    try:
```
pcolazurdo, line 53: It looks like only ItemAlreadyExistsError is being expected as possible error. As the persistence error can have other multiple exceptions, should these be managed too?

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,432 @@
...
+            expiry_timestamp=self._get_expiry_timestamp(),
+            payload_hash=self._get_hashed_payload(event),
+        )
+
+        logger.debug(f"Saving in progress record for idempotency key: {data_record.idempotency_key}")
+
+        if self.use_local_cache:
+            record = self._retrieve_from_cache(idempotency_key=data_record.idempotency_key)
+            if record:
+                raise ItemAlreadyExistsError
```
pcolazurdo, line 306: Looks like this exception is only raised if use_local_cache is true. I didn't find this exception being raised in _put_record

pcolazurdo:
@cakepietoast I've left some comments above. Sorry if some have been already discussed as I may have missed the conversation as I joined this thread very late (but basically because I'm very interested in this util) - let me know if I can help in any other way.

aws_lambda_powertools/utilities/idempotency/idempotency.py
pcolazurdo: If it's treated as a missing record, shouldn't the call to save_inprogress happen as in that case?

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: For exceptions that aren't handled as ItemAlreadyExistsError, I think allowing the exception to bubble up is the correct approach in order to maintain the idempotency guarantees. We could wrap those exceptions into a custom exception and re-raise, but I don't think this adds much utility.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Yep, you're right - we need to retry the save_inprogress call here. This code path would be triggered rarely (when the record isn't expired at initial save_inprogress but is expired at get_record), which would cause a bug. Will fix this.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Yep, you're right - we need to retry the save_inprogress call here same as your other comment.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: The handler cannot start executing until a lock is obtained ("in progress" record in the persistence layer). If we can't obtain that lock, lambda_handler will not be executed. If the lock was obtained but the operation did not complete / Lambda times out, the "in progress" record will persist, preventing any further executions of the function with the same idempotency key. When the record expires - regardless of what status it was in the event can be retried: this will be the documented behavior.

I'm not sure there's much we can do (aside from clearly documenting the behavior) to handle partially executed operations where the lambda times out during the handler. If you have any improvement ideas on that they're very welcome!

aws_lambda_powertools/utilities/idempotency/persistence/base.py
cakepietoast: `_put_record` does raise the same error. I'll clarify in the ABC abstractmethod docstring though.

https://github.com/awslabs/aws-lambda-powertools-python/blob/c4856fdd0386460b329711907cf4b5b8466ba31e/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py#L124

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,432 @@
...
+    """
+    Data Class for idempotency records.
+    """
+
+    def __init__(
+        self,
+        idempotency_key,
+        status: str = "",
+        expiry_timestamp: int = None,
+        response_data: str = "",
```
michaelbrewer, line 42: @cakepietoast  i think we should give the option pretty early on the also save the request (or parts of it)

For example you might need some of the original request to cancel an idempotent request (https://developer.squareup.com/reference/square/payments-api/cancel-payment-by-idempotency-key)

michaelbrewer (commented):
Add another note about also saving part (or all of the request/event)

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Latest commit addresses this

michaelbrewer:
@cakepietoast @heitorlessa Once you think the api for this is stable enough, i don't mind running some beta testing against a build of this.

michaelbrewer:
@cakepietoast  maybe we can later reference this utility (or to a blog post) from the warning in the [SQS Batch Processing docs](https://awslabs.github.io/aws-lambda-powertools-python/utilities/batch/)

Screen Shot 2021-02-14 at 9 28 32 AM

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,441 @@
...
+from aws_lambda_powertools.utilities.idempotency.exceptions import (
+    IdempotencyInvalidStatusError,
+    IdempotencyItemAlreadyExistsError,
+    IdempotencyValidationError,
+)
+
+logger = logging.getLogger(__name__)
+
+STATUS_CONSTANTS = {
+    "NOTEXISTING": "DOESNOTEXIST",
```
nmoutschen, line 24: __Nitpick__: Is there a reason why the names here are different?

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,221 @@
...
+
+        if event_record.status == STATUS_CONSTANTS["INPROGRESS"]:
+            raise IdempotencyAlreadyInProgressError(
+                f"Execution already in progress with idempotency key: "
+                f"{self.persistence_store.event_key_jmespath}={event_record.idempotency_key}"
+            )
+
+        return event_record.response_json_as_dict()
+
+    def _call_lambda(self) -> Any:
```
nmoutschen, line 190: __Nitpick__: Since we're already inside a Lambda function, should this be called `_call_handler`?

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,221 @@
...
+        Lambda's Context
+    persistence_store: BasePersistenceLayer
+        Instance of BasePersistenceLayer to store data
+
+    Examples
+    --------
+    **Processes Lambda's event in an idempotent manner**
+        >>> from aws_lambda_powertools.utilities.idempotency import idempotent, DynamoDBPersistenceLayer
+        >>>
+        >>> persistence_store = DynamoDBPersistenceLayer(event_key="body", table_name="idempotency_store")
```
nmoutschen, line 52: This mentions `event_key` but the base class has `event_key_jmespath` as argument.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,441 @@
...
+        data: Any
+            The data to hash
+
+        Returns
+        -------
+        str
+            Hashed representation of the provided data
+
+        """
+        hashed_data = self.hash_function(json.dumps(data).encode())
```
nmoutschen, line 210: __Question__: This assumes that the data is always serializable. Are there any situation where an event payload or function response could contain non-serializable data?

E.g. `decimal.Decimal` is not serializable by default, but you can return a `decimal.Decimal` from a Lambda function.

Here's a Lambda function to extract how the Python runtime does it:

```python
def handler(event, context):
    with open("/var/runtime/lambda_runtime_marshaller.py") as f:
        print(f.read())
```

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -0,0 +1,163 @@
...
+from aws_lambda_powertools.utilities.idempotency.exceptions import (
+    IdempotencyItemAlreadyExistsError,
+    IdempotencyItemNotFoundError,
+)
+from aws_lambda_powertools.utilities.idempotency.persistence.base import DataRecord
+
+logger = logging.getLogger(__name__)
+
+
+class DynamoDBPersistenceLayer(BasePersistenceLayer):
```
nmoutschen, line 18: __Question__: DynamoDB only supports items up to 400 KB, while Lambda supports payloads up to 6 MB. This will not be a problem for event payloads since we hash the record, but this could be on the response side. Should we document this somewhere?

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,441 @@
...
+        self.use_local_cache = use_local_cache
+        if self.use_local_cache:
+            self._cache = LRUDict(max_items=local_cache_max_items)
+        self.payload_validation_enabled = False
+        if payload_validation_jmespath:
+            self.validation_key_jmespath = jmespath.compile(payload_validation_jmespath)
+            self.payload_validation_enabled = True
+        self.hash_function = getattr(hashlib, hash_function)
+
+    def _get_hashed_idempotency_key(self, lambda_event: Dict[str, Any]) -> str:
```
nmoutschen, line 155: __Question__: Should we perform this operation only once at init or is this over-optimisation?

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,221 @@
...
+            A lambda execution is already in progress
+        IdempotencyInconsistentStateError
+            The persistence store reports inconsistent states across different requests. Retryable.
+        """
+        # This code path will only be triggered if the record becomes expired between the save_inprogress call and here
+        if event_record.status == STATUS_CONSTANTS["EXPIRED"]:
+            raise IdempotencyInconsistentStateError("save_inprogress and get_record return inconsistent results.")
+
+        if event_record.status == STATUS_CONSTANTS["INPROGRESS"]:
+            raise IdempotencyAlreadyInProgressError(
```
nmoutschen, line 183: __Question__: This is fine as long as the caller is not expecting an answer from the Lambda function. Should we implement a waiting logic that will poll DynamoDB regularly to get the result?

michaelbrewer: The callee should retry with an incremental backoff. Polling in the lambda would be alot more costly.

aws_lambda_powertools/shared/cache_dict.py
```diff
@@ -0,0 +1,26 @@
+from collections import OrderedDict
+
+
+class LRUDict(OrderedDict):
```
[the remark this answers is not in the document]
michaelbrewer: I assume this is only used internally? I guess if documented, it might be assumed to always be part of the implementation?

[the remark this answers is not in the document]
heitorlessa: Internal only (shared) - Adding a docstring explaining what's its intended usage and notable places it's used can help us with further maintenance ;)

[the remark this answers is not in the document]
cakepietoast: Docstring added

aws_lambda_powertools/utilities/idempotency/persistence/base.py
cakepietoast: Yes, but not a good one! I removed a couple of unused constants now.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Good idea - fixed!

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Fixed!

aws_lambda_powertools/utilities/idempotency/persistence/base.py
cakepietoast: Fixed - json serialization supports decimals now

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Agree with @michaelbrewer here. Adding an option to poll instead could be a future feature addition, but I think relying on the client to handle this case is best in most cases.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
cakepietoast: The persistence layer is initialized outside of the Lambda handler, so it has no awareness of the event when we create it. This method will be called with lots of different events during the instance's lifetime, so any caching would need to maintain a map of event -> key. There would be a tradeoff in terms of complexity (as dict isn't hashable so can't be used as a key by default) and memory usage. In my opinion its an unnecessary optimisation.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -0,0 +1,221 @@
...
+    Examples
+    --------
+    **Processes Lambda's event in an idempotent manner**
+        >>> from aws_lambda_powertools.utilities.idempotency import idempotent, DynamoDBPersistenceLayer
+        >>>
+        >>> persistence_store = DynamoDBPersistenceLayer(event_key="body", table_name="idempotency_store")
+        >>>
+        >>> @idempotent(persistence_store=persistence_store)
+        >>> def handler(event, context):
+        >>>     return {"StatusCode": 200}
```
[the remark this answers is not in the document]
cakepietoast: I think if we address this it should be in the documentation. Happy to make changes to the docs based on feedback!

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
cakepietoast: Yep, I'll add this to the docs.

aws_lambda_powertools/utilities/idempotency/idempotency.py
nmoutschen: Agreed there, especially as processing could take between seconds and minutes depending on what the function does. I think it'd be good to have it documented as a recommended practice when we go into the pattern.

docs/utilities/idempotency.md
```diff
@@ -0,0 +1,351 @@
...
+
+```python hl_lines="4"
+DynamoDBPersistenceLayer(
+    event_key_jmespath="body",
+    table_name="IdempotencyTable",
+    expires_after_seconds=5*60  # 5 minutes
+    )
+
+```
+This will mark any records older than 5 minutes expired, and the lambda handler will be executed as normal if it is
```
nmoutschen, line 175: __Typo__: missing *as* in "minutes expired"

docs/utilities/idempotency.md
```diff
@@ -0,0 +1,351 @@
...
+* Prevent Lambda handler code executing more than once on the same event payload during a time window
+* Ensure Lambda handler returns the same result when called with the same payload
+* Select a subset of the event as the idempotency key using JMESpath expressions
+* Set a time window in which records with the same payload should be considered duplicates
+
+## Getting started
+
+### Required resources
+
+Before getting started, you need to create a DynamoDB table to store state used by the idempotency utility. Your lambda
```
nmoutschen, line 27: Should we put this as the general case? I can quickly see an ask for ElastiCache in the future for example, so wouldn't it be better to put it under a DynamoDB section?

nmoutschen (approved):
LGTM for the doc. Just 2 nitpicks, but these can be changed in the future.

docs/utilities/idempotency.md
cakepietoast: You're right, but I'll save a bigger restructuring of the docs until we have more support for other persistence layers.

aws_lambda_powertools/utilities/idempotency/idempotency.py
cakepietoast: Added a section to the docs on handling concurrent executions to address this.

aws_lambda_powertools/shared/cache_dict.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+    Cache implementation based on ordered dict with a maximum number of items. Last accessed item will be evicted
+    first. Currently used by idempotency utility.
+    """
+
+    def __init__(self, max_items=1024, *args, **kwds):
+        self.max_items = max_items
+        super().__init__(*args, **kwds)
```
michaelbrewer, line 12:
```suggestion
        super().__init__(*args, **kwargs)
```

aws_lambda_powertools/shared/cache_dict.py
```diff
@@ -0,0 +1,31 @@
+from collections import OrderedDict
+
+
+class LRUDict(OrderedDict):
+    """
+    Cache implementation based on ordered dict with a maximum number of items. Last accessed item will be evicted
+    first. Currently used by idempotency utility.
+    """
+
+    def __init__(self, max_items=1024, *args, **kwds):
```
michaelbrewer, line 10:
```suggestion
    def __init__(self, max_items=1024, *args, **kwargs):
```

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,436 @@
...
+
+        Returns
+        -------
+        bool
+            Whether the record is currently expired or not
+        """
+        if self.expiry_timestamp:
+            if int(datetime.datetime.now().timestamp()) > self.expiry_timestamp:
+                return True
+        return False
```
michaelbrewer, line 74:
```suggestion
        return bool(self.expiry_timestamp and int(datetime.datetime.now().timestamp()) > self.expiry_timestamp)
```

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -0,0 +1,436 @@
...
+        self._cache[data_record.idempotency_key] = data_record
+
+    def _retrieve_from_cache(self, idempotency_key: str):
+        cached_record = self._cache.get(idempotency_key)
+        if cached_record:
+            if cached_record.is_expired:
+                logger.debug(f"Removing expired local cache record for idempotency key: {idempotency_key}")
+                self._delete_from_cache(idempotency_key)
+            else:
+                return cached_record
```
michaelbrewer, line 253:
```suggestion
            if not cached_record.is_expired:
                return cached_record
            logger.debug(f"Removing expired local cache record for idempotency key: {idempotency_key}")
            self._delete_from_cache(idempotency_key)
```

tests/functional/idempotency/test_idempotency.py
```diff
@@ -0,0 +1,586 @@
...
+    IdempotencyInconsistentStateError,
+    IdempotencyPersistenceLayerError,
+    IdempotencyValidationError,
+)
+from aws_lambda_powertools.utilities.idempotency.idempotency import idempotent
+from aws_lambda_powertools.utilities.validation import envelopes, validator
+
+TABLE_NAME = "TEST_TABLE"
+
+
```
michaelbrewer, line 18: Maybe add a missing test for the invalid status returned from the data store

```suggestion

def test_data_record_invalid_status_value():
    data_record = DataRecord("key", status="UNSUPPORTED_STATUS")

    with pytest.raises(IdempotencyInvalidStatusError) as e:
        data_record.status

    assert e.value.args[0] == "UNSUPPORTED_STATUS"
```

michaelbrewer (changes requested):
Small review changes

PR #254, opened 2020-12-24
docs: add missing parser models

risenberg-cyberark:
Added missing documentation - sns & ses parser models


heitorlessa:
Thanks a lot @risenberg-cyberark !! It definitely got lost in between merges last time - Thanks for sending the fix!

PR #259, opened 2021-01-12
improv: override Tracer auto-capture response/exception via env vars

heitorlessa:
**Issue #, if available:** Feature parity with Java powertools PR https://github.com/awslabs/aws-lambda-powertools-java/pull/249

## Description of changes:

Adds env var support to override default behaviour in Tracer of auto-capturing both response and error:

Environment variable | Description
------------------------------------------------- | --------------------------------------------------------------------------------- 
**POWERTOOLS_TRACER_CAPTURE_RESPONSE** | disable auto-capture response as metadata (e.g. `"true", "True", "TRUE"`)
**POWERTOOLS_TRACER_CAPTURE_ERROR** | disable auto-capture error as metadata (e.g. `"true", "True", "TRUE"`)

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


docs/content/index.mdx
```diff
@@ -118,15 +118,16 @@ Utility | Description
...
-**POWERTOOLS_LOGGER_LOG_EVENT** | Logs incoming event | [Logging](./core/logger)
-**POWERTOOLS_LOGGER_SAMPLE_RATE** | Debug log sampling | [Logging](./core/logger)
-**LOG_LEVEL** | Sets logging level | [Logging](./core/logger)
+Environment variable | Description | Utility | Default
+------------------------------------------------- | --------------------------------------------------------------------------------- | --------------------------------------------------------------------------------- | ------------------------------------------------- | -------------------------------------------------
+**POWERTOOLS_SERVICE_NAME** | Sets service name used for tracing namespace, metrics dimension and structured logging | All | `"service_undefined"`
+**POWERTOOLS_METRICS_NAMESPACE** | Sets namespace used for metrics | [Metrics](./core/metrics) | `None`
+**POWERTOOLS_TRACE_DISABLED** | Disables tracing | [Tracing](./core/tracer) | `false`
+**POWERTOOLS_TRACER_CAPTURE_RESPONSE** | Overrides tracer from auto-capturing response as metadata | [Tracing](./core/tracer) | `true`
```
nmoutschen, line 126: Is this missing a `POWERTOOLS_TRACER_CAPTURE_ERROR` row?

__NIT__: "Overrides" is not 100% clear to me. Does this means the environment overrides explicit behaviour? Is it in all cases (both true/false) or only in one? Does setting to true means that we override? Since `resolve_env_var_choice` will take the code first then the environment variables, I think we should have something like "Adds the Lambda function response in trace's metadata. Passing it explicitly to the Tracer overrides the environment variable."

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -449,51 +472,65 @@ async def async_tasks():
...
-                method=method, capture_response=capture_response, method_name=method_name
+                method=method, capture_response=capture_response, capture_error=capture_error, method_name=method_name
             )
 
-    def _decorate_async_function(self, method: Callable = None, capture_response: bool = True, method_name: str = None):
+    def _decorate_async_function(
+        self,
+        method: Callable = None,
+        capture_response: Optional[bool] = None,
+        capture_error: Optional[bool] = None,
```
nmoutschen, line 505: __NOT BLOCKING__: Is there a better way to do this? As time goes and we add new flags/features, this signature will become rather long.

heitorlessa: I couldn't stop to think of that yet, but definitely an area of concern to optimize it for future maintenance. Lemme know if you have ideas - Sounds like a builder or factory pattern could work here.

docs/content/index.mdx
heitorlessa: Good call - It's from a previous behaviour I've changed at last minute (explicit wins over env var). Will update it shortly.

tests/functional/test_shared_functions.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+from aws_lambda_powertools.shared.functions import resolve_env_var_choice
+
+
+def test_explicit_wins_over_env_var():
+    choice_env = os.getenv("CHOICE", True)
```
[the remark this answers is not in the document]
heitorlessa: last minute draft that shouldn't be in the commit - will remove it. Any str value for `env` would be the right place, and it's also missing a test for when choice isn't passed.

heitorlessa:
Addressed all your comments and some other oversights I've noticed too - Here's the doc too. Over to you @nmoutschen 

image

nmoutschen (approved):
LGTM

PR #261, opened 2021-01-13
test: add benchmark on AWS Lambda

nmoutschen:
**Issue #, if available:** None

## Description of changes:

Add a benchmark that deploys Lambda functions with/without Lambda Powertools to measure the performance impact of using it.

**Checklist**

* [*] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [ ] Update tests
* [ ] Update docs
* [*] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/376ec0a2ac0d2a40e0af5717bef42ff84ca0d1b9/.github/semantic.yml#L2)

## Breaking change checklist

None

**RFC issue #**: None


benchmark/benchmark.sh
```diff
@@ -0,0 +1,77 @@
...
+process_id=$!
+for i in {0..20}; do
+    run_function $REFERENCE_FUNCTION
+done &
+wait $process_id
+wait $!
+echo
+
+# Gather statistics
+sleep 150
```
heitorlessa, line 70: add a note as to why 150 e.g. CWL Insights ingestion time

benchmark/benchmark.sh
```diff
@@ -0,0 +1,77 @@
...
+export BENCHMARK_STACK_NAME=${BENCHMARK_STACK_NAME:-"powertools-benchmark"}
+
+function cleanup {
+    echo "Cleaning up stack..."
+    aws cloudformation delete-stack --stack-name $BENCHMARK_STACK_NAME
+}
+
+function run_function {
+    # Update function to force a cold start
+    aws lambda update-function-configuration --function-name $1 --memory-size 256 >/dev/null
```
heitorlessa, line 15: can we add 512 too? It's quite common to see 256, 512, and 1024. can have skewed results

benchmark/benchmark.sh
```diff
@@ -0,0 +1,77 @@
+#!/bin/bash
```
heitorlessa, line 1: could you add a `README` under `benchmark` so maintainers have more info on how and when to run it?

heitorlessa (changes requested):
Looks great!! Only two remarks (Memory, Readme), and a minor (comment on sleep 150)

benchmark/benchmark.sh
nmoutschen: Will do

benchmark/benchmark.sh
nmoutschen: This doesn't really test at 256, it's just a trick to force a cold start. 😄

benchmark/benchmark.sh
nmoutschen: Will do

heitorlessa:
Haha I glanced too quickly at them then

michaelbrewer:
@heitorlessa  You might want to set a baseline and publish results between each release to track regressions. Either as a simple json file or something funky like this: https://simonwillison.net/2020/Oct/9/git-scraping/

nmoutschen:
Before we merge this, open question:

Right now, this depends on `sam deploy --guided` which require user inputs. Should I create an S3 bucket for deployment artefacts (similar to [what I did here](https://github.com/aws-samples/aws-serverless-ecommerce-platform/blob/master/tools/package#L50-L54)) so that this can be fully automated?

heitorlessa:
Create a bucket under Powertools account - this will get us closer to
automate and check for regressions

nmoutschen:
> Create a bucket under Powertools account - this will get us closer to automate and check for regressions

For the generic case, that means just checking for an environment variable and letting other users manage it?

heitorlessa:
Oh, I see. You're thinking of customers also running this benchmark by
themselves, and not us (maintainers) only - Did I get this right?

If that's the case, yes, it'd have to be a default for us (our S3 bucket in
our Powertools account), and if customers want to run that they can
override it. I think for customers in general, it might be best to provide
guidance on how to use Python Tuna to generate flame graphs to measure
their overall import time, since they might use more than just the Core
utilities + other libraries like Requests which increase a lot in startup

benchmark/README.md
```diff
@@ -0,0 +1,24 @@
+# Cold Start Benchmark
+
+The [benchmark.sh script](./benchmark.sh) is a bash script to compare the cold-start time of using the AWS Lambda Powertools in a semi-automated way. It does so by deploying two Lambda functions which both have the aws-lambda-powertools module installed. One Lambda function will import and initialize the three main utilities (`Metrics`, `Logger`, `Tracer`), while the other one will not.
```
heitorlessa, line 3:
```suggestion
The [benchmark.sh script](./benchmark.sh) is a bash script to compare the cold-start time of using the AWS Lambda Powertools in a semi-automated way. It does so by deploying two Lambda functions which both have the aws-lambda-powertools module installed. One Lambda function will import and initialize the three core utilities (`Metrics`, `Logger`, `Tracer`), while the other one will not.
```

benchmark/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+Please note that this requires the [SAM CLI](https://github.com/aws/aws-sam-cli) version 1.2.0 or later.
+
+## Usage
+
+To use the script, you should move into the benchmark folder and run the benchmark script:
```
heitorlessa, line 9:
```suggestion
> **NOTE**: This script is expected to run in Unix-based systems only, and can incur charges on your AWS account.

To use the script, you should move into the benchmark folder and run the benchmark script:
```

benchmark/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+cd benchmark
+./benchmark.sh
+```
+
+This will:
+
+* Deploy a CloudFormation stack using guided SAM deployment (*you will need to answer a few questions*).
+* Run loops to update the memory setting of the functions to force a cold start, then invoke them. This process is repeated a number of time to get more consistent results.
+* Wait for the data to propagate from CloudWatch Logs to CloudWatch Logs Insights.
```
heitorlessa, line 22:
```suggestion
* Wait 2.5 minutes to ensure data propagates from CloudWatch Logs to CloudWatch Logs Insights.
```

heitorlessa (commented):
Made some tiny adjustments to ensure people are aware this runs on Unix based systems only, it can incur charges, and how long the script waits for data to propagate from CWL to CWLI

heitorlessa (approved):
Thanks a lot again @nmoutschen - Merging this now.

Issue #262, opened 2021-01-13
version 1.7.0 and above broke pytest verbose output


silvestriluca:
## Problem description
During local tests is not infrequent to use Pytest with **live log call** output to watch logs as they are emitted and to help in debugging activities. The command to have this "verbose" output is the following  
`pytest -o log_cli=1`  

Until aws-lambda-powertools 1.6.1 the command worked fine.
Since version 1.7.0 all pytest live log call outputs are missing.

## Expected behaviour
All function/test **live log call** outputs should be visible when issuing `pytest -o log_cli=1`

## Current behaviour
No **live log call** outputs when issuing `pytest -o log_cli=1` and using powertools 1.7.0 or above

## How to reproduce
I've provided a simple gist to reproduce the issue.

1. Clone the following gist: https://gist.github.com/silvestriluca/95352eb38587cb6bc333f8ec790fb22a
2. Navigate to the repo folder
### Working scenario
3. Set a virtual environment: `python3 -m venv v-env-pt161`
4. Activate the `v-env-pt161` virtual environment
5. `pip install -r requirements_powertools161.txt` to install pytest and powertools v1.6.1
6. Run `pytest -o log_cli=1` 
7. You should see the following output:  
<img width="1128" alt="Good" src="https://user-images.githubusercontent.com/43203075/104462602-141dc600-55b1-11eb-82cb-a9a8479bcff8.png">

   As you can see, all log lines appear as expected, with proper errorlevel => this powertool version works
### Broken scenario
8. Now set a new virtual environment: `python3 -m venv v-env-ptlatest`
9. Activate the `v-env-ptlatest` virtual environment
10. `pip install -r requirements.txt` to install pytest and latest powertools (v1.9.1)
11. Run `pytest -o log_cli=1` 
12. You should see the following output. <img width="1128" alt="Bad" src="https://user-images.githubusercontent.com/43203075/104462441-dd47b000-55b0-11eb-8cf1-68e3cf6c9183.png">As you can see, the logs are missing => this powertool version doesn't work
13. Going backwards with powertools versions, you can see that the issue starts with version 1.7.0

## Possible cause
According to Powertools changelog, [version 1.7.0](https://github.com/awslabs/aws-lambda-powertools-python/releases/tag/v1.7.0) introduced the following bugfix:
>improv: keeps Lambda root logger handler intact, and add log filter instead to prevent child log records duplication (#198)

[...]

PR #268, opened 2021-01-17
feat: toggle to disable log deduplication locally for pytest live log #262

heitorlessa:
**Issue #, if available:** #262 

## Description of changes:

This allows customers to disable log deduplication filter protection to use Pytest Live Log feature (and similar). When root logger is configured, Powertools Logger will intercept any propagated message to prevent duplicates. This feature overrides this protection.

**Pytest Live Log feature with this feature**

image

**Pytest Live Log feature without the explicit override**

image

**Docs**

image

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:


aws_lambda_powertools/logging/logger.py
```diff
@@ -129,12 +129,14 @@ def __init__(
         self.sampling_rate = resolve_env_var_choice(
             choice=sampling_rate, env=os.getenv(constants.LOGGER_LOG_SAMPLING_RATE, 0.0)
         )
+        self._is_deduplication_disabled = resolve_truthy_env_var_choice(
```
nmoutschen, line 132: __NIT__: We store whether the deduplication is disabled rather than enabled, then (line 175) we check against this flag not being true. It makes this flag a bit hard to read IMHO. Should we invert this?

aws_lambda_powertools/logging/logger.py
```diff
@@ -167,12 +169,16 @@ def _init_logger(self, **kwargs):
...
-            handler.addFilter(SuppressFilter(self.service))
+        # Pytest Live Log feature duplicates log records for colored output
+        # but we explicitly add a filter for log deduplication.
+        # This flag disables this protection when you explicit want logs to be duplicated (#262)
+        if not self._is_deduplication_disabled:
+            logger.debug("Adding filter in root logger to suppress child logger records to bubble up")
+            for handler in logging.root.handlers:
+                # It'll add a filter to suppress any child logger from self.service
+                # Where service is Order, it'll reject parent logger Order,
+                # and child loggers such as Order.checkout, Order.shared
```
nmoutschen, line 180: __NITPICK__: This seems to be an example but it's not marked explicitly as such.

nmoutschen (approved):
LGTM - Just some small nitpicks.

Issue #275, opened 2021-01-21
Exception handling in batch processing doesn't work as expected


pcolazurdo:
In this code:

```python
processor = PartialSQSProcessor(suppress_exception=False)
with processor(records, record_handler) as proc:
    result = proc.process()
```

If there is an exception raised inside the record_handler function, I expect the exception to be raised in the function and get some details about what caused the exception. The code in https://github.com/awslabs/aws-lambda-powertools-python/blob/e104bfb5dcedf7b52d4a7925ae446ff719b8c110/aws_lambda_powertools/utilities/batch/sqs.py#L126 tries to be smart about it and raise a list of all the exceptions in the batch, but this doesn't work as expected, as the Exception code is not smart enough to show a list of exceptions. The actual result looks like this:

```
Traceback (most recent call last):
  File "/var/task/aws_lambda_powertools/logging/logger.py", line 270, in decorate
    return lambda_handler(event, context)
  File "/var/task/aws_lambda_powertools/tracing/tracer.py", line 303, in decorate
    response = lambda_handler(event, context, **kwargs)
  File "/var/task/aws_lambda_powertools/metrics/metrics.py", line 144, in decorate
    response = lambda_handler(event, context)
  File "/var/task/app.py", line 179, in lambda_handler
    result = proc.process()  # Returns a list of all results from record_handler
  File "/var/task/aws_lambda_powertools/utilities/batch/base.py", line 58, in __exit__
    self._clean()
  File "/var/task/aws_lambda_powertools/utilities/batch/sqs.py", line 126, in _clean
    raise SQSBatchProcessingError(list(self.exceptions))
```

**What were you trying to accomplish?**

## Expected Behavior
I would except the Exception to be raised in a manner that all the exceptions are shown together.
Also, reviewing the code I see there is a fail_message property that doesn't seem to be exposed anywhere. Is there a plan for this? This may be very useful in some cases.

## Current Behavior
Please review the Stack trace above

## Possible Solution
I think re raising an exception may need some extra work like what is commented here: https://nedbatchelder.com/blog/200711/rethrowing_exceptions_in_python.html

## Steps to Reproduce (for bugs)
```
Raise an exception inside a function called from the record handler. Without debug mode, you can't see the details of this exception.
```

## Environment

* **Powertools version used**:  aws-lambda-powertools-1.10.1
* **Packaging format (Layers, PyPi)**: PyPi
* **AWS Lambda function runtime**: Python 3.8
[...]

PR #276, opened 2021-01-26
fix: batch processing exceptions

cakepietoast:
**Issue #, if available:** #275

## Description of changes:

Update batch exception handling to include more detail of the actual exceptions raised in users processing code.

_Why not just log the exception one-by-one as we catch them?_
By default, the aws_lambda_powertools package logger is suppressed. Logging from package modules would require users to take [additional steps](https://awslabs.github.io/aws-lambda-powertools-python/#debug-mode) to see those logs.

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/batch/exceptions.py
+++ b/aws_lambda_powertools/utilities/batch/exceptions.py
@@ -1,7 +1,28 @@
 """
 Batch processing exceptions
 """
+import traceback
 
 
 class SQSBatchProcessingError(Exception):
     """When at least one message within a batch could not be processed"""
+
+    def __init__(self, msg="", child_exceptions=()):
+        super().__init__(msg)
+        self.msg = msg
+        self.child_exceptions = child_exceptions
+
+    def __str__(self):
+        parent_exception_str = super(SQSBatchProcessingError, self).__str__()
+        exception_list = [f"{parent_exception_str}\n"]
+
+        for exception in self.child_exceptions:
+            extype, ex, tb = exception
+            formatted = "".join(traceback.format_exception(extype, ex, tb))
+            exception_list.append(formatted)
+
+        return "\n".join(exception_list)
+
+    def log(self):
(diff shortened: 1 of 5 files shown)


heitorlessa:
Will look into it tomorrow :)

heitorlessa:
I'm late on reviewing this... I'll try tomorrow evening - Been a roller-coaster these past days

aws_lambda_powertools/utilities/batch/exceptions.py
```diff
@@ -1,7 +1,23 @@
...
 class SQSBatchProcessingError(Exception):
     """When at least one message within a batch could not be processed"""
+
+    def __init__(self, msg="", child_exceptions=()):
+        super().__init__(msg)
+        self.msg = msg
+        self.child_exceptions = child_exceptions
+
+    def __str__(self):
+        parent_exception_str = super(SQSBatchProcessingError, self).__str__()
```
heitorlessa, line 16: Could you add a comment as to why we're doing this? Maybe a ref to the issue, so it's easier to remember why

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -92,8 +93,8 @@ def _process_record(self, record) -> Tuple:
         try:
             result = self.handler(record)
             return self.success_handler(record, result)
-        except Exception as exc:
-            return self.failure_handler(record, exc)
+        except Exception:
+            return self.failure_handler(record, sys.exc_info())
```
heitorlessa, line 97: nitpick: could you use kwargs in case we refactor in the future?

heitorlessa (changes requested):
LGTM - minor changes so we remember why we're making some of these changes, and not a generic exception processing class

cakepietoast:
Implemented requested changes from review

Issue #280, opened 2021-02-03
feat: only install typing-extensions dependency when necessary


nadobando:
## Key information

* RFC PR: (leave this empty)
* Related issue(s), if known:
* Area: Parser
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-python/#tenets): Yes

## Summary
[summary]: #summary

typing-extensions module is a backport for typing hint for old versions of python <= 3.5
in this project many times typing-extensions is imported to use the Literal type hint which is already natively included in python >= 3.8

## Motivation
[motivation]: #motivation
There is no need for this package in Python >= 3.8 and we are currently enforced to install it

## Proposal
[proposal]: #proposal
add if try/except blocks based on the python version
```python
# aws_lambda_powertools/utilities/parser/types.py
import sys

if sys.version_info[0] >= 3 and sys.version_info[1] >= 8:
    from typing import Literal  # noqa: F401
else:
    try:
        from typing_extensions import Literal  # noqa: F401
    except ImportError:
        raise Exception("please install typing-extensions or upgrade to Python >= 3.8")
```
```python
# aws_lambda_powertools/utilities/parser/models/dynamodb.py
from datetime import date
from typing import Any, Dict, List, Optional

from pydantic import BaseModel

from ..types import Literal


class DynamoDBStreamChangedRecordModel(BaseModel):
    ApproximateCreationDateTime: Optional[date]
    Keys: Dict[str, Dict[str, Any]]
    NewImage: Optional[Dict[str, Any]]
    OldImage: Optional[Dict[str, Any]]
    SequenceNumber: str
    SizeBytes: int
    StreamViewType: Literal["NEW_AND_OLD_IMAGES", "KEYS_ONLY", "NEW_IMAGE", "OLD_IMAGE"]
```
this is not really a design, it's more a check to wether import typing-extensions or not based on the system python version 

in order to use it right instead of importing typing_extensions we can include the dependency in the file types.py  and then import that dependency from the types.py as showed in the code blocks

## Drawbacks
[drawbacks]: #drawbacks
since this is not a design it's more an improvement of a dependency import  I don't see a drawback

## Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

* **What is the impact of not doing this?**
The impact of not doing this will be the used of a backport library which is not needed by the interpreter as it has the native implementation
## Unresolved questions
[unresolved-questions]: #unresolved-questions
no questions

[...]

PR #281, opened 2021-02-03
fix: remove unnecessary typing-extensions for py3.8

nadobando:
**Issue #280**

## Description of changes:
1. Added if block to know if typing_extensions should be imported based on the python version.
if python >= 3.8 typing extensions will not be imported and the native implementation will be used
else will try to import typing_extensions and catch the ImportError

2. Updated modules which imported this dependency.

3. Updated pyproject.toml to only install typing-extensions when python version is <3.8
**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/parser/types.py
+++ b/aws_lambda_powertools/utilities/parser/types.py
@@ -1,6 +1,13 @@
 """Generics and other shared types used across parser"""
+
 from typing import TypeVar
 
 from pydantic import BaseModel
+
+# We only need typing_extensions for python versions <3.8
+try:
+    from typing import Literal  # noqa: F401
+except ImportError:
+    from typing_extensions import Literal  # noqa: F401
 
 Model = TypeVar("Model", bound=BaseModel)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * **Utilities**: Correctly handle and list multiple exceptions in SQS batch processing utility.
 * **Docs*:: Fix typos on AppConfig docstring import, and `SnsModel` typo in parser.
+* **Utilities**: `typing_extensions` package is now only installed in Python < 3.8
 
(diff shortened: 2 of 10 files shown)


aws_lambda_powertools/utilities/parser/types.py
```diff
@@ -1,6 +1,16 @@
 """Generics and other shared types used across parser"""
+import sys
```
cakepietoast, line 2:
```suggestion
```

aws_lambda_powertools/utilities/parser/types.py
```diff
@@ -1,6 +1,16 @@
...
 from pydantic import BaseModel
 
+# Workaround for not importing typing_extensions on python ^3.8
+if sys.version_info[0] > 3 or (sys.version_info[0] == 3 and sys.version_info[1] >= 8):
+    from typing import Literal  # noqa: F401
+else:
+    try:
+        from typing_extensions import Literal  # noqa: F401
+    except ImportError:
+        raise Exception("please install typing-extensions or upgrade to Python >= 3.8")
```
cakepietoast, line 14: We can simplify this and just fall back to typing_extensions if the import of `typing.Literal` fails. No need to check the python version or re-raise the ImportError.


```suggestion
# We only need typing_extensions for python versions <3.8
try:
    from typing import Literal  # noqa: F401
except ImportError:
    from typing_extensions import Literal  # noqa: F401
```

cakepietoast (changes requested):
Nice change, great to keep the package size down where we can! Added a small suggestion to change - otherwise LGTM!

aws_lambda_powertools/utilities/parser/types.py
nadobando: Sure, I will do it 👍

cakepietoast (approved):

heitorlessa:
Thanks a lot for this contribution @nadobando !!! And thanks @cakepietoast for the review. Merging this and publishing a new release

PR #288, opened 2021-02-13
docs(data-classes): Fix anchor tags to be lower case

michaelbrewer:
## Description of changes:

Anchors casing need to match what is generated in the html, which is lower case.

eg: `S3` should be `s3`

> **NOTE:** This does not fix the anchor links on the logger.md as the toc level is 3, and no anchor tags are generated for the following

```markdown
[Service parameter](#the-service-parameter), [Inheriting Loggers](#inheriting-loggers), [Overriding Log records](#overriding-log-records), and [Logging exceptions](#logging-exceptions)
```

mkdocs.yml
```yml
  - toc:
      permalink: true
      toc_depth: 3
```

**Checklist**

 

* [X] Update tests
* [X] Update docs


heitorlessa:
Another issue I miss having a proper strict mode - not sure if it's worth investing a few days on markdownlint custom rules + setup.

Thanks for picking up those!

PR #295, opened 2021-02-19
fix: correct behaviour to avoid caching "INPROGRESS" records

cakepietoast:
**Issue #, if available:**

## Description of changes:

This fixes an issue with caching in the unreleased idempotency util. We were mistakenly caching in progress records, meaning execution environments could get "stuck" with in progress records in the cache. Records in this state shouldn't be included in the cache, as we've no way to reflect changes from outside the execution environment in the local caches.

**Checklist**

 

* [X] Update tests
* [X] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -7,6 +7,7 @@ import hashlib
 import json
 import logging
 from abc import ABC, abstractmethod
+from types import MappingProxyType
 from typing import Any, Dict
 
 import jmespath
@@ -21,7 +22,7 @@ from aws_lambda_powertools.utilities.idempotency.exceptions import (
 
 logger = logging.getLogger(__name__)
 
-STATUS_CONSTANTS = {"INPROGRESS": "INPROGRESS", "COMPLETED": "COMPLETED", "EXPIRED": "EXPIRED"}
+STATUS_CONSTANTS = MappingProxyType({"INPROGRESS": "INPROGRESS", "COMPLETED": "COMPLETED", "EXPIRED": "EXPIRED"})
 
 
 class DataRecord:
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -82,8 +82,7 @@ class DataRecord:
         """
         if self.is_expired:
             return STATUS_CONSTANTS["EXPIRED"]
-
-        if self._status in STATUS_CONSTANTS.values():
+        elif self._status in STATUS_CONSTANTS.values():
             return self._status
         else:
             raise IdempotencyInvalidStatusError(self._status)
@@ -222,7 +221,7 @@ class BasePersistenceLayer(ABC):
         """
         if self.payload_validation_enabled:
             lambda_payload_hash = self._get_hashed_payload(lambda_event)
-            if not data_record.payload_hash == lambda_payload_hash:
+            if data_record.payload_hash != lambda_payload_hash:
                 raise IdempotencyValidationError("Payload does not match stored record for this event key")
 
     def _get_expiry_timestamp(self) -> int:
@@ -263,7 +262,7 @@ class BasePersistenceLayer(ABC):
     def _retrieve_from_cache(self, idempotency_key: str):
         cached_record = self._cache.get(idempotency_key)
         if cached_record:
-            if all((not cached_record.is_expired, not cached_record.status == "INPROGRESS")):
+            if not cached_record.is_expired:
                 return cached_record
             logger.debug(f"Removing expired local cache record for idempotency key: {idempotency_key}")
             self._delete_from_cache(idempotency_key)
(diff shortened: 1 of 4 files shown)


michaelbrewer:
@cakepietoast haha. that was quick :)

michaelbrewer (approved):
LGTM

michaelbrewer:
@cakepietoast @heitorlessa maybe to help with code review we can try to use more GIVEN, WHEN & THENs.

At least just to cover the basic behavors we put into the RFCs. Coz then we are less likely to miss things

nmoutschen (approved):
LGTM!

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -243,7 +243,7 @@ def _save_to_cache(self, data_record: DataRecord):
     def _retrieve_from_cache(self, idempotency_key: str):
         cached_record = self._cache.get(idempotency_key)
         if cached_record:
-            if not cached_record.is_expired:
+            if all((not cached_record.is_expired, not cached_record.status == "INPROGRESS")):
```
michaelbrewer, line 246: It might be more correct to use the `STATUS_CONSTANTS`

```suggestion
            if all((not cached_record.is_expired, not cached_record.status == STATUS_CONSTANTS["INPROGRESS"])):
```

OR remove the check all together, as we are already preventing this from happening in `get_record` 

```suggestion
            if not cached_record.is_expired:
```

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -365,7 +360,9 @@ def get_record(self, event: Dict[str, Any]) -> DataRecord:
 
         record = self._get_record(idempotency_key)
 
-        if self.use_local_cache:
+        # We can't cache "INPROGRESS" records as we have no way to reflect updates that can happen outside of the
+        # execution environment
+        if self.use_local_cache and not record.status == STATUS_CONSTANTS["INPROGRESS"]:
```
michaelbrewer, line 365: @cakepietoast  alternatively you could just modify `_save_to_cache` to implement this check. 
Easier to test and guarantee we never save "INPROGRESS" records

michaelbrewer: Updates to _retrieve_from_cache:

```python3
    def _retrieve_from_cache(self, idempotency_key: str):
        cached_record = self._cache.get(idempotency_key)
        if cached_record:
            if not cached_record.is_expired:
                return cached_record
            logger.debug(f"Removing expired local cache record for idempotency key: {idempotency_key}")
            self._delete_from_cache(idempotency_key)
```

Update `_save_to_cache` would be 

```python3
    def _save_to_cache(self, data_record: DataRecord):
        # We can't cache "INPROGRESS" records as we have no way to reflect updates that can happen outside of the
        # execution environment
        if data_record.status == STATUS_CONSTANTS["INPROGRESS"]:
            return

        self._cache[data_record.idempotency_key] = data_record
```

Modify existing test `test_idempotent_lambda_in_progress_with_cache` to do:
```python3
save_to_cache_spy.assert_called()
```

Add new test for `_save_to_cache`

```python3
@pytest.mark.parametrize("persistence_store", [{"use_local_cache": True}], indirect=True)
def test_in_progress_never_saved_to_cache(persistence_store):
    # GIVEN a data record with status "INPROGRESS"
    # and persistence_store has use_local_cache = True
    data_record = DataRecord("key", status="INPROGRESS")

    # WHEN saving to local cache
    persistence_store._save_to_cache(data_record)

    # THEN don't save to local cache
    assert persistence_store._cache.get("key") is None
```

michaelbrewer (changes requested):
I have attached my proposal changes 

[changes.diff.zip](https://github.com/awslabs/aws-lambda-powertools-python/files/6015911/changes.diff.zip)

michaelbrewer:
@cakepietoast @heitorlessa 

I spend a little time on this and attached a diff of changes

### Changes:

base.py
* Make `STATUS_CONSTANTS` a frozen dict (by using `MappingProxyType`)
* `status` - Make `status` call internal mutually exclusive
* `_validate_payload` - Fix doc string typo 
* `_validate_payload` - Simplifies conditional logic using De Morgan's Identity
* `_save_to_cache`, `_retrieve_from_cache`, `_delete_from_cache` to check `use_local_cache` and exit early
* `_save_to_cache` - to not save for status == "INPROGRESS"
* `_save_to_cache` - add doc string to explain has "INPROGRESS" is not saved
* `save_success` - Fix doc string typo
* `save_success` - Just call `_save_to_cache` directly
* `save_inprogress` - Just call `_retrieve_from_cache` directly
* `delete_record` - Just call `_delete_from_cache` directly
* `get_record` - Simplify logic by calling `_retrieve_from_cache` and `_save_to_cache` directly

test_idempotency.py
* `test_idempotent_lambda_in_progress_with_cache` update check to `save_to_cache_spy.assert_called()`
* `test_in_progress_never_saved_to_cache` - add new test to check "INPROGRESS" is not saved to local cache
* `test_user_local_disabled` - add new test to check if all local cache options work when use_local_cache  is False

test_json_encoder.py
* `test_jsonencode_calls_default` - Add missing test case for unhandled types


### Attached diff
* [proposed_changes.diff.zip](https://github.com/awslabs/aws-lambda-powertools-python/files/6016089/proposed_changes.diff.zip)

cakepietoast:
> @cakepietoast @heitorlessa maybe to help with code review we can try to use more GIVEN, WHEN & THENs.
> 
> At least just to cover the basic behavors we put into the RFCs. Coz then we are less likely to miss things

Yep, you're completely right on this. Its been on my todo list with this PR for ages but I should've prioritised it! I'll add this as soon as I get chance.

PR #296, opened 2021-02-21
test(general): Add some missing code coverage

michaelbrewer:
## Description of changes:

* test(DictWrapper): Add check for not isinstance
* test(tracing): Add mock to test aiohttp_trace_config
* test(appconfig): Add test for get_app_config() without a default provider

**Checklist**

 

* [X] Update tests
* [X] Update docs


michaelbrewer:
@cakepietoast some more clean up for the next release.

tests/functional/test_tracing.py
```diff
@@ -212,3 +214,15 @@ def handler(event, context):
...
+def test_aiohttp_trace_config():
+    aiohttp_mock = Mock()
+    sys.modules["aiohttp"] = aiohttp_mock
+    from aws_lambda_powertools.tracing import aiohttp_trace_config
+
+    aiohttp_trace_config()
+
+    from aws_xray_sdk.ext.aiohttp.client import aws_xray_trace_config
+
+    assert aws_xray_trace_config.__doc__ == "aiohttp extension for X-Ray (aws_xray_trace_config)"
```
heitorlessa, line 228: What's going on here? not sure I follow

tests/functional/test_utilities_parameters.py
```diff
@@ -1534,6 +1534,36 @@ def _get_multiple(self, path: str, **kwargs) -> Dict[str, str]:
     assert str_value == json.dumps(mock_body_json)
 
 
+def test_appconf_get_app_config_new(monkeypatch, mock_name, mock_value):
+    """
+    Test get_app_config() without a default provider
+    """
```
heitorlessa, line 1540:
```suggestion

```

No need for docstring as we ensure the name is as explicit as it can be (plus GIVEN/WHEN/THEN which you already have)

heitorlessa (changes requested):
One clarification and one tiny change - LGTM after that

tests/functional/test_tracing.py
michaelbrewer: @heitorlessa  we are just not hitting the part where we modify the `__doc__` (i am not sure why this is happening either, i was hoping you would know)

heitorlessa, line 228:
```suggestion
    assert aiohttp_trace_config.__doc__ == "aiohttp extension for X-Ray (aws_xray_trace_config)"
```

I believe you got this backward -- You are asserting what we export has the right docstring, not what X-Ray exports per se.

heitorlessa: We're basically only injecting the doctstring for IDE Intellisense when people are about to import that, nothing else. Testing it can be called or simply ignored from test coverage altogether suffices tbh

heitorlessa (approved):

heitorlessa:
for the record https://github.com/awslabs/aws-lambda-powertools-python/pull/296/commits/faf75ca7ef17d32e86d6ebdef18b0ae9c47845a4 😈

Issue #294, opened 2021-02-19
Exception on import when AWS_PROFILE is invalid/null


lorengordon:
## Expected Behavior

`import aws_lambda_powertools` should still succeed even if the ENV contains `AWS_PROFILE` with a bad value.

## Current Behavior

It appears that during import, under at least some conditions, aws_lambda_powertools will attempt to setup an aws client session? I'm not sure I understand why. So if the ENV contains an invalid AWS config, in particular `AWS_PROFILE=` or `AWS_PROFILE=fakeprofile`, then the import fails with an exception. Which, yes, it's invalid, and I now that I know `aws_lambda_powertools` has this behavior I can work around it easily by unsetting the envs. But it still seems like it shouldn't be setting up a session on import?

## Possible Solution

Not sure on a solution, but a workaround is to make sure your development environment unsets bad AWS_ envs before attempting to run tests on your python code. :)

## Steps to Reproduce (for bugs)

1. `AWS_PROFILE= python -c 'import aws_lambda_powertools'`

## Environment

* **Powertools version used**: 1.10.5
* **Packaging format (Layers, PyPi)**: PyPi
* **AWS Lambda function runtime:**: n/a
* **Debugging logs**

```python
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/home/loren/.local/lib/python3.6/site-packages/aws_lambda_powertools/__init__.py", line 9, in <module>
    from .tracing import Tracer  # noqa: F401
  File "/home/loren/.local/lib/python3.6/site-packages/aws_lambda_powertools/tracing/__init__.py", line 6, in <module>
    from .tracer import Tracer
  File "/home/loren/.local/lib/python3.6/site-packages/aws_lambda_powertools/tracing/tracer.py", line 10, in <module>
    import aws_xray_sdk.core
  File "/home/loren/.local/lib/python3.6/site-packages/aws_xray_sdk/core/__init__.py", line 10, in <module>
    xray_recorder = AsyncAWSXRayRecorder()
  File "/home/loren/.local/lib/python3.6/site-packages/aws_xray_sdk/core/recorder.py", line 66, in __init__
    self._sampler = DefaultSampler()
  File "/home/loren/.local/lib/python3.6/site-packages/aws_xray_sdk/core/sampling/sampler.py", line 25, in __init__
    self._connector = ServiceConnector()
[...]

PR #301, opened 2021-03-01
fix: import time latency by lazily loading high level modules

heitorlessa:
**Issue #, if available:** #294

## Description of changes:

**Checklist**

 

* [x] Update tests
* [x] Update docs
* [x] Create generic types for Tracer provider
* [x] Create generic types for Tracer subsegments
* [x] Convert EMF JSON Schema into native object to reduce I/O

**Changes table**

Scenario | Result | Change
------------------------------------------------- | --------------------------------------------------------------------------------- | -------------------------------------------------
Logger import only | **0.33s** user 0.15s system 82% cpu 0.582 total | **No change**
Logger import only | **0.07s** user 0.02s system 97% cpu 0.102 total | **Tracer lazily loaded**
Tracer imported but not initialized | **0.08s** user 0.02s system 84% cpu 0.115 total | **Tracer lazily loaded**
Tracer imported and initialized | **0.32s** user 0.11s system 92% cpu 0.469 total | **Tracer lazily loaded**
Logger import only | negligible difference | **Metrics EMF JSON Schema as dict**

## Notes

* LazyLoader adds roughly 10ms due to class instantiation and globals manipulation
* Hard lesson: Type annotations load lazily loaded library ahead of time diminishing its effectiveness
    - Turns out there's an [explicit mechanism to do that safely](https://www.python.org/dev/peps/pep-0563/#runtime-annotation-resolution-and-type-checking), though we now have a Tracer Provider ABC which can be extended to bring any other Tracer Provider later

## Breaking change checklist

**RFC issue #**:


michaelbrewer:
@heitorlessa  nice, definitely handy for lambdas that just want structured logging or want to toggle tracing only for dev and test and disable on production.

heitorlessa:
New flamegraph with these patches - Tracer no longer appears and removes ~400ms for all imports and customers once merged

image

heitorlessa:
Done - @nmoutschen, do you have the bandwidth to review?

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -512,8 +516,8 @@ async def async_tasks():
     def _decorate_async_function(
         self,
         method: Callable = None,
-        capture_response: Optional[bool] = None,
-        capture_error: Optional[bool] = None,
+        capture_response: Optional[Union[bool, str]] = None,
+        capture_error: Optional[Union[bool, str]] = None,
```
nmoutschen, line 520: __Question__: Not sure what this change does. Why do we need to add str as a possible value here?

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -625,8 +629,8 @@ def _add_response_as_metadata(
         self,
         method_name: str = None,
         data: Any = None,
-        subsegment: aws_xray_sdk.core.models.subsegment = None,
-        capture_response: Optional[bool] = None,
+        subsegment: BaseSegment = None,
```
nmoutschen, line 632: __Question__: could this break anything since aws_xray_sdk.core.models.subsegment does not inherit BaseSegment, but fulfill its methods properly?

tests/performance/test_high_level_imports.py
```diff
@@ -0,0 +1,90 @@
...
+    # GIVEN Core utilities are imported
+    # WHEN none are used
+    # THEN import and any global initialization perf should be below 30ms
+    # though we adjust to 35ms to take into account different CI machines, etc.
+    # instead of re-running tests which can lead to false positives
+    with timing() as t:
+        core_utilities()
+
+    elapsed = t()
+    if elapsed > 0.035:
```
nmoutschen, line 45: __Question__: Putting a static value could break on older/low-power devices, is there a better way to decide if the elapsed time is acceptable based on the capabilities of the current machine?

tests/performance/test_high_level_imports.py
```diff
@@ -0,0 +1,90 @@
...
+        pytest.fail(f"High level imports should be below 40ms: {elapsed}")
```
nmoutschen, line 46: __Nitpick__: the value here does not match the constant float the line above.

_(there's the same issue a few more times in this file)_

nmoutschen (commented):
I have a few questions/remarks before I can approve. We also need to include the proper licensing for the class taken from Terraform (Apache-2.0 license).

aws_lambda_powertools/tracing/tracer.py
heitorlessa: It's because of the`resolve_env_choice..` function that can return either bool or str (1, 0)

tests/performance/test_high_level_imports.py
heitorlessa: Sadly not, I had various attempts on finding a better mechanism - I've come to the conclusion that was best to try on a 128M function, add some room for lower powered machines and keep it static

aws_lambda_powertools/tracing/tracer.py
heitorlessa: not in this case no, it's primarily for type hinting since adding `aws_xray_sdk.core.models.subsegment` triggers a full load throwing the lazy import effectiveness out of the window. This custom type/ABC is an entry level change to something bigger - Bring Your Own Tracer Provider in 2.0

nmoutschen (approved):
LGTM

PR #302, opened 2021-03-03
feat(idempotent): Add support for jmespath_options

michaelbrewer:
**Issue #, if available:**

## Description of changes:

Like for the validator, idempotent utility should allow for extracting an idempotent key using custom functions

Example usage:

We have an api gateway proxy event:

Where event is:

```json5
{
   "body": "{\"id\": \"c299229\"}"
}
```

And `c299229` is the idempotent key we want to use:

```python3
@idempotent(DynamoDBPersistenceLayer(table_name="foo", event_key_jmespath="powertools_json(body).id"))
def handler(event, context):
    pass
```

**Checklist**

 

* [x] Update tests
* [ ] Update docs


michaelbrewer:
@cakepietoast @heitorlessa  i can update the docs, if this feature makes sense?

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -19,6 +19,7 @@
     IdempotencyItemAlreadyExistsError,
     IdempotencyValidationError,
 )
+from aws_lambda_powertools.utilities.validation.jmespath_functions import PowertoolsFunctions
```
heitorlessa, line 22: Could you leave a comment as `Maintenance: Make PowertoolsFunctions a shared class`? That way we can improve it later

heitorlessa (changes requested):
LGTM after docs mention this -- In all fairness, simply copy and replace what's available in Validations doc, and we can refactor that to make it reusable later.

https://awslabs.github.io/aws-lambda-powertools-python/utilities/validation/#built-in-jmespath-functions

aws_lambda_powertools/utilities/idempotency/persistence/base.py
michaelbrewer: @heitorlessa i moved it into the shared dir.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -135,6 +137,8 @@ def __init__(
             Function to use for calculating hashes, by default md5.
         raise_on_no_idempotency_key: bool, optional
             Raise exception if no idempotency key was found in the request, by default False
+        jmespath_options : Dict
+            Alternative JMESPath options to be included when filtering expr
```
heitorlessa, line 141: no-op: leaving a note to remember to move all of these outside persistence, so we keep persistence layer with one task only.

aws_lambda_powertools/utilities/validation/base.py
```diff
@@ -5,8 +5,9 @@
 import jmespath
 from jmespath.exceptions import LexerError
 
+from aws_lambda_powertools.shared.jmespath_functions import PowertoolsFunctions
```
heitorlessa, line 8: 🎉 thank you!!!

tests/functional/idempotency/test_idempotency.py
```diff
@@ -690,3 +691,33 @@ def test_raise_on_no_idempotency_key(persistence_store):
 
     # THEN raise IdempotencyKeyError error
     assert "No data found to create a hashed idempotency_key" in str(excinfo.value)
+
+
+@pytest.mark.parametrize("persistence_store", [{"use_local_cache": True}], indirect=True)
+def test_jmespath_with_powertools_json(persistence_store):
+    # GIVEN an event_key_jmespath with powertools_json custom function
+    persistence_store.event_key_jmespath = "[requestContext.authorizer.claims.sub, powertools_json(body).id]"
+    persistence_store.event_key_compiled_jmespath = jmespath.compile(persistence_store.event_key_jmespath)
```
heitorlessa, line 700: do you actually need this? shouldn't this be covered here:

```python
        if self.event_key_jmespath:
            self.event_key_compiled_jmespath = jmespath.compile(event_key_jmespath)
```

This key `event_key_compiled_jmespath` sounds like it's primarily cache -- JMESPath search function already caches it after the first result, so calling it `compile` doesn't gain much tbh (feel free to do a perf test if you'd like)

https://github.com/jmespath/jmespath.py/blob/develop/jmespath/parser.py#L85

heitorlessa (changes requested):
Last change and I'll merge it :) 

Here's the inner cache of jmespath parser - used both by compile and search: https://github.com/jmespath/jmespath.py/blob/599b0f7ea650a60f7512233301e70ff8151de0f0/jmespath/parser.py#L84

Here's the subtle difference between compile and search: https://github.com/jmespath/jmespath.py/blob/599b0f7ea650a60f7512233301e70ff8151de0f0/jmespath/__init__.py#L23

Contrary to fastjsonschema that doesn't cache, JMESPath does on both occasions, and keep up to 128 expressions.

heitorlessa:
Thanks Michael, again! I'm merging as-is, and create a reminder to come back to simplify the internals once I have a better understanding of Tom's implementation.

PR #315, opened 2021-03-07
fix(idempotency): TypeError when calling is_missing_idempotency_key with an int

michaelbrewer:
## Description of changes:

Fix for `is_missing_idempotency_key` which  can raise a `TypeError: 'int' object is not iterable` 
for the follow code

```python3
BasePersistenceLayer.is_missing_idempotency_key(1000)
```

**Checklist**

 

* [x] Update tests
* [x] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -183,7 +183,14 @@ class BasePersistenceLayer(ABC):
 
     @staticmethod
     def is_missing_idempotency_key(data) -> bool:
-        return data is None or not data or (type(data).__name__ in ("tuple", "list") and all(x is None for x in data))
+        if data is None:
+            return True
+        elif not data:
+            return True
+        elif type(data).__name__ in ("tuple", "list", "dict"):
+            return all(x is None for x in data)
+        else:
+            return False
 
     def _get_hashed_payload(self, lambda_event: Dict[str, Any]) -> str:
         """
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -183,11 +183,9 @@ class BasePersistenceLayer(ABC):
 
     @staticmethod
     def is_missing_idempotency_key(data) -> bool:
-        if not data:
-            return True
         if type(data).__name__ in ("tuple", "list", "dict"):
             return all(x is None for x in data)
-        return False
+        return not data
 
     def _get_hashed_payload(self, lambda_event: Dict[str, Any]) -> str:
         """
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -183,7 +183,7 @@ class BasePersistenceLayer(ABC):
 
     @staticmethod
     def is_missing_idempotency_key(data) -> bool:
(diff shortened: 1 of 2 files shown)


aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -183,7 +183,7 @@ def _get_hashed_idempotency_key(self, lambda_event: Dict[str, Any]) -> str:
 
     @staticmethod
     def is_missing_idempotency_key(data) -> bool:
-        return data is None or not data or all(x is None for x in data)
+        return data is None or not data or (type(data).__name__ in ("tuple", "list") and all(x is None for x in data))
```
heitorlessa, line 186: Can we break this into a multi-line thing?

It's quite complex now and we might find another bug here in the future I'm sure

heitorlessa (changes requested):
Great catch! One minor thing to break that into a multi-line statement as that would've been easier to spot this issue too.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -183,7 +183,14 @@ def _get_hashed_idempotency_key(self, lambda_event: Dict[str, Any]) -> str:
 
     @staticmethod
     def is_missing_idempotency_key(data) -> bool:
-        return data is None or not data or all(x is None for x in data)
+        if data is None:
+            return True
+        elif not data:
+            return True
```
heitorlessa, line 189: This can be a single block:

if not data:
    return True

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -183,7 +183,14 @@ def _get_hashed_idempotency_key(self, lambda_event: Dict[str, Any]) -> str:
...
+        elif type(data).__name__ in ("tuple", "list", "dict"):
+            return all(x is None for x in data)
+        else:
+            return False
```
heitorlessa, line 193: Perhaps no else and simply return False if flow above doesn't yield anything?

return False

I think we can do with two Ifs only

heitorlessa (changes requested):
Control flow simplification with two Ifs

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -183,7 +183,11 @@ def _get_hashed_idempotency_key(self, lambda_event: Dict[str, Any]) -> str:
 
     @staticmethod
     def is_missing_idempotency_key(data) -> bool:
-        return data is None or not data or all(x is None for x in data)
+        if data is None or not data:
```
heitorlessa, line 186: if not data

Sorry Michael, I should've been clearer. Since we're not reassigning a default value, we don't need to check if data is None. If not data should account for None and falsy values

heitorlessa (changes requested):

aws_lambda_powertools/utilities/idempotency/persistence/base.py
michaelbrewer: No problem (personally i do hate Truthy stuff (but i understand some languages have it))

heitorlessa (approved):

michaelbrewer:
@heitorlessa  funny way of doing it:

```python3
assert type(None).__name__ == "NoneType"
```

heitorlessa:
First name I actually see type(obj).__name__ tbh as opposed to instance ;)

PR #321, opened 2021-03-09
feat(logging): inject correlation_id

michaelbrewer:
**Issue #, if available:**

#9 

## Description of changes:

Prototype idea for correlation_id support in logger.

> NOTE: I am just using `inject_lambda_context` for demo purposes, i am not sure if this should be a new decorator?

Example API Gateway Proxy Event
```json5
{
    "requestContext":{
        "requestId":"13B398ED-FABA-4D50-A4A1-6CD39BDA85E5"
    }
}
```

Usage:

```python3
@logger.inject_lambda_context(correlation_id_path="requestContext.requestId")
def handler(_1, _2):
    logger.info("My message")

```

Cloudwatch log
```json
{
    "level":"INFO",
    "location":"handler:450",
    "message":"My message",
    "timestamp":"2021-03-09 15:12:32,652",
    "service":"81shAbHB2IPQTex",
    "sampling_rate":0.0,
    "cold_start":true,
    "function_name":"test",
    "function_memory_size":128,
    "function_arn":"arn:aws:lambda:eu-west-1:809313241:function:test",
    "function_request_id":"52fdfc07-2182-154f-163f-5f0f9a621d72",
    "correlation_id":"13B398ED-FABA-4D50-A4A1-6CD39BDA85E5"
}
```

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -4,7 +4,9 @@ import logging
 import os
 import random
 import sys
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Dict, Optional, Union
+
+import jmespath
 
 from ..shared import constants
 from ..shared.functions import resolve_env_var_choice, resolve_truthy_env_var_choice
@@ -34,6 +36,27 @@ def _is_cold_start() -> bool:
         is_cold_start = False
 
     return cold_start
+
+
+def _get_correlation_id_path(path: Union[str, int]) -> Optional[str]:
+    """
+
+    Parameters
+    ----------
+    path :
+        JMESPath
+    Returns
+    -------
+    str, optional
+        Returns the JMESPath
+    """
+    if path is None:
+        return None
+
+    # NOTE: We could have a enum of standard event types
+    # like "requestContext.requestId" or API Gateway Proxy Events
+
+    return path
 
 
 # PyCharm does not support autocomplete via getattr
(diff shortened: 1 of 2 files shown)


heitorlessa:
@michaelbrewer could you push an empty commit to test the label one more time, plz?

heitorlessa:
That's great, thank you! While we think through the other RFC to allow a more cohesive correlation ID across utilities, may I ask two enhancements where one is optional?

1. A pre-defined set of correlation ID keys based on Event Sources we know? e.g. API Gateway, ALB, and EventBridge to start with [validation envelopes example](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/validation/envelopes.py)
2. [Optional] Allow a JMESPath value to cover use cases where correlation_id might be deeper in the stack as well as combine/filter/slice values

heitorlessa:
Scratch that.. too early in the day for me. It's indeed better to handle that in the Uber Supreme Leader Handler, where we could have a Config available to all utilities, where Logger could use, Metrics could use to set correlation ID as metadata, Tracer as annotation, etc. 

It kinda sucks though having to ask customers to add an additional line to set correlation ID when they likely already have `inject_lambda_context` - What do you think of also adding an extra parameter in that [decorator](https://github.com/awslabs/aws-lambda-powertools-python/blob/7ec0648419cf491dd41fc64a8cb6dd7bf1f7ffae/aws_lambda_powertools/logging/logger.py#L264)?

docs/core/logger.md
```diff
@@ -188,6 +188,47 @@ You can append your own keys to your existing Logger via `structure_logs(append=
 
 	This example will add `order_id` if its value is not empty, and in subsequent invocations where `order_id` might not be present it'll remove it from the logger.
 
+#### set_correlation_id method
```
heitorlessa, line 191:
```suggestion
#### Setting correlation ID
```

heitorlessa (changes requested):
One tiny change to improve SEO on docs, and we're good to merge 🎉 !

I'll write another PR to include the new correlation_path in the docs too

heitorlessa:
@pankajagrawal16 FYI - We're adding the first iteration of setting correlation ID. This is before we work on an uber duper decorator to rule them all with a single config to prevent the confusion of many decorators while also allowing customers to create their own. 

The latter doesn't necessarily translates to Java, but the correlation ID is a good one we want to have in the core utilities

heitorlessa (approved):

PR #323, opened 2021-03-10
feat(data-classes): AppSync Resolver Event

michaelbrewer:
## Description of changes:

Add support for AppSync Resolver Events either Direct or Amplify

```python3
from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent


def handler(event: dict, context: LambdaContext):
    event = AppSyncResolverEvent(event)
    assert event.type_name == "Merchant"
    assert event.field_name == "locations"
    assert event.info.parent_type_name == event.type_name
    assert event.info.field_name == event.field_name
```

Can also be used as a decorator

```python3
from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent
from aws_lambda_powertools.utilities.data_classes.appsync_resolver_utils import (
    AppSyncResolver,
    make_id,
    aws_date,
    aws_datetime,
    aws_time,
    aws_timestamp,
)

logger = Logger()
app = AppSyncResolver()


@app.resolver(field_name="listLocations", include_event=True)
@app.resolver(type_name="Merchant", field_name="locations", include_event=True)
def get_locations(event: AppSyncResolverEvent, name: str = None):
    logger.info(event)
    return f"returning locations for {name}"


@app.resolver(type_name="Merchant", field_name="anotherField")
def another_field_resolver(count: int):
    return {
        "id": make_id(),
        "date": aws_date(),
        "date_time": aws_datetime(),
        "time": aws_time(),
        "ts": aws_timestamp(),
        "message": f"another_field_resolver with parameter count={count}",
    }


@app.resolver(type_name="Query", field_name="noParams")
def no_params():
    return "no_params has no params"


def handler(event: dict, context):
```

This is based on what Amplify produces by default:
* https://docs.amplify.aws/cli/graphql-transformer/function#structure-of-the-function-event
* https://docs.amplify.aws/guides/api-graphql/lambda-resolvers/q/platform/js

Key | Description
-- | --
typeName | The name of the parent object type of the field being resolver.
fieldName | The name of the field being resolved.
arguments | A map containing the arguments passed to the field being resolved.
identity | A map containing identity information for the request. Contains a nested key ‘claims’ that will contains the JWT claims if they exist.
source | When resolving a nested field in a query, the source contains parent value at runtime. For example when resolving Post.comments, the source will be the Post object.
request | The AppSync request object. Contains header information.
prev | When using pipeline resolvers, this contains the object returned by the previous function. You can return the previous value for auditing use cases.

Which is a simplification of is possible in the app sync $context : 
* https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html

Key | Description
-- | --
arguments | A map containing the arguments passed to the field being resolved.
identity | A map containing identity information for the request. 
source | When resolving a nested field in a query, the source contains parent value at runtime. For example when resolving Post.comments, the source will be the Post object.
request | The AppSync request object. Contains header information.
prev | A container for the results of this resolver. This field is only available to response mapping templates. (like source)
info | An object that contains information about the GraphQL request.
stash | The stash is a map that is made available inside each resolver and function mapping template.

**Checklist**

* [x] Implement the identity class (API_KEY, AWS_IAM vs AMAZON_COGNITO_USER_POOLS)
* [x] Decide whether to create a data class specifically for Amplify ?
* [x] Add support for correlation id
* [X] Update tests
* [x] Update docs


heitorlessa:
hey @awsed - could you take a quick peek at this PR in case we're missing anything obvious from the Lambda Resolver event object?

aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py
```diff
@@ -0,0 +1,157 @@
...
+    @property
+    def source(self) -> Dict[str, any]:
+        """A map that contains the resolution of the parent field."""
+        return self["source"]
+
+    @property
+    def request_headers(self) -> Dict[str, str]:
+        """Request headers"""
+        return self["request"]["headers"]
+
```
heitorlessa, line 133: What do you think of adding a few helper methods for date timestamp like this project did?  I'm sure customers will appreciate this touch, as not everyone can remember how to return the correct timestamp since it's not available in the SDK.

https://github.com/twkiiim/sleemo/blob/46c416577033567b5b15fdd2018775f8692c4c70/sleemo/utils.py#L8

aws_lambda_powertools/logging/correlation_paths.py
```diff
@@ -2,5 +2,6 @@
 
 API_GATEWAY_REST = "requestContext.requestId"
 API_GATEWAY_HTTP = API_GATEWAY_REST
+APP_SYNC_RESOLVER = "request.headers.x-amzn-trace-id"
```
heitorlessa, line 5: fix `APPSYNC_RESOLVER`

heitorlessa (changes requested):
It looks mostly great - only one-liner that could be more readable to ease maintenance by others.

Docs are lacking on confirming the two variants of object, as you pointed out - Amplify GraphQL Transformer vs AppSync Direct Lambda Resolvers. 

Hopefully @awsed or @dabit3 have a few minutes to confirm whether there is a good complete event available for each in the docs somewhere

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -20,6 +20,14 @@ def get(self, key: str) -> Optional[Any]:
         return self._data.get(key)
 
 
+def get_header_value(headers: dict, name: str, default_value: str, case_sensitive: bool) -> Optional[str]:
+    """Get header value by name"""
+    if case_sensitive:
+        return headers.get(name, default_value)
+
+    return next((value for key, value in headers.items() if name.lower() == key.lower()), default_value)
```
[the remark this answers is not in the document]
michaelbrewer: @heitorlessa  i just moved it from the old location into a reusable function, but i can clarify it :)

heitorlessa:
**Note to self** - Validate how AppSync Direct Lambda resolvers event look like compared to what Amplify `@function` does. AppSync is unlikely to break contract while Amplify `@function` as at first glance it isn't formalized.

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -20,6 +20,21 @@ def get(self, key: str) -> Optional[Any]:
...
+        return headers.get(name, default_value)
+
+    name_lower = name.lower()
+
+    return next(
+        # Iterate over the dict and do a case insensitive key comparison
+        (value for key, value in headers.items() if key.lower() == name_lower),
+        # Default value is returned if no matches was found
+        default_value,
+    )
```
michaelbrewer, line 35: @heitorlessa i am not sure if this is better or worse

awsed:
This is a better event with the info object: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/generated_sample_events/events/appsync/AppSyncDirectResolver.json

michaelbrewer:
> This is a better event with the info object: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/generated_sample_events/events/appsync/AppSyncDirectResolver.json

Thanks @awsed i will add that to the test suite, interesting how it differs from what Amplify generates.

awsed:
@michaelbrewer the ```info``` object is missing, check line 60 here https://github.com/aws/aws-sam-cli/blob/ab3dcfae17f971cfbfea3f459ed872ba14cead83/samcli/lib/generated_sample_events/events/appsync/AppSyncDirectResolver.json#L60

michaelbrewer:
> @michaelbrewer the `info` object is missing, check line 60 here https://github.com/aws/aws-sam-cli/blob/ab3dcfae17f971cfbfea3f459ed872ba14cead83/samcli/lib/generated_sample_events/events/appsync/AppSyncDirectResolver.json#L60

yes, that does not come through for Amplify AppSync events.

aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py
michaelbrewer: @heitorlessa maybe we can have a separate PR for these kind of utility methods for response processing.

michaelbrewer:
@heitorlessa can you run through this again 

@awsed - i have added all of the optional fields that might be in direct appsync calls.

michaelbrewer (commented):
@heitorlessa hopefully this is looking for merging

heitorlessa:
Looking :) Is the PR description ready? I'll use that for the docs later

michaelbrewer:
> Looking :) Is the PR description ready? I'll use that for the docs later

The description progressed over time as i found more things. I can put in a checklist of changes. For the data-classes  docs in general we could start to expand on the examples, but that could warrant a who cookbook section.

docs/utilities/data_classes.md
```diff
@@ -101,6 +103,100 @@ Typically used for API Gateway REST API or HTTP API using v1 proxy event.
             do_something_with(event.body, query_string_parameters)
     ```
 
+### AppSync Resolver
+
+Used when building a Lambda GraphQL Resolvers with [Amplify GraphQL Transform Library](https://docs.amplify.aws/cli/graphql-transformer/function){target="_blank"}
+and can also be used for [AppSync Direct Lambda Resolvers](https://aws.amazon.com/blogs/mobile/appsync-direct-lambda/){target="_blank"}.
+
+=== "lambda_app.py"
```
heitorlessa, line 111:
```suggestion
=== "app.py"
```

tests/functional/appsync/test_appsync_resolver_utils.py
```diff
@@ -0,0 +1,219 @@
...
+        return json.load(fp)
+
+
+def test_direct_resolver():
+    # Check whether we can handle an example appsync direct resolver
+    mock_event = load_event("appSyncDirectResolver.json")
+
+    app = AppSyncResolver()
+
+    @app.resolver(field_name="createSomething", include_context=True)
```
heitorlessa, line 33: I wonder if this would be considered a Dataclass. I see two options:

1. Discuss some design options in a RFC for lightweight API handling
2. Expand Event Source Data Classes description to include utilities in it - Current description:`The event source data classes utility provides classes describing the schema of common Lambda events triggers.`

The first is my preference as we might find other use cases, UX, as we put this into written form.

aws_lambda_powertools/utilities/data_classes/appsync_resolver_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    return now.strftime("%Y-%m-%dT%H:%M:%SZ")
+
+
+def aws_timestamp():
+    """AWSTimestamp - An integer value representing the number of seconds before or after 1970-01-01-T00:00Z."""
+    return int(time.time())
+
+
+class AppSyncResolver:
+    """AppSync resolver decorator utility"""
```
heitorlessa, line 40: Could you include an Example docstring section like we have in others?

heitorlessa (changes requested):
Some suggestions on wording, a suggestion to follow what you did with Cognito data classes so including utils is cleaner, and a question on whether it's best to leave AppSyncResolver decorator out until we have a more complete written RFC for discussion?

Whatever we decide on AppSyncResolver will open precedent to discuss API Gateway Lightweight decorators too, for things like CORS, HTTP Methods, etc.

aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py
```diff
@@ -0,0 +1,232 @@
...
+    @property
+    def variables(self) -> Dict[str, str]:
+        """A map which holds all variables that are passed into the GraphQL request."""
+        return self.get("variables")
+
+    @property
+    def selection_set_list(self) -> List[str]:
+        """A list representation of the fields in the GraphQL selection set. Fields that are aliased will
+        only be referenced by the alias name, not the field name. The following example shows this in detail."""
```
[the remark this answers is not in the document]
michaelbrewer: Sorry that was lifted from the aws docs

aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py
```diff
@@ -0,0 +1,232 @@
...
+
+    @property
+    def field_name(self) -> str:
+        """The name of the field that is currently being resolved."""
+        return self.info.field_name
+
+    @property
+    def arguments(self) -> Dict[str, any]:
+        """A map that contains all GraphQL arguments for this field."""
+        return self["arguments"]
```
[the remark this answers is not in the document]
michaelbrewer: As long as we don't get to the size of boto3 ;)

michaelbrewer:
> Some suggestions on wording, a suggestion to follow what you did with Cognito data classes so including utils is cleaner, and a question on whether it's best to leave AppSyncResolver decorator out until we have a more complete written RFC for discussion?
> 
> Whatever we decide on AppSyncResolver will open precedent to discuss API Gateway Lightweight decorators too, for things like CORS, HTTP Methods, etc.

I was thinking of leaving it as an undocumented or beta feature (but i will feature restructure the locations)

tests/functional/appsync/test_appsync_resolver_utils.py
michaelbrewer: @heitorlessa  kind of started that here:
https://github.com/awslabs/aws-lambda-powertools-python/issues/324

But yes, in general having lightweight decorators help. Maybe like the SQS batch processing?

heitorlessa (changes requested):
After some thought, let's leave it as-is `AppSyncResolver()` and undocumented. 

We can document that later once that RFC is written. That way, we add AppSync support within the data class, and have a good initial draft implementation of an AppSync Resolver ready to go once we find a better home for it.

To account for the `utils()` bit in AppSync data class, please make those changes in the Event Source Data Classes top description to include `utilities` too, so it doesn't get ruled out of scope for anything beyond a schema like this now

tests/functional/appsync/test_appsync_resolver_utils.py
heitorlessa: It's missing the RFC body like summary, drawbacks, alternatives, etc ;) one step a time

heitorlessa (approved):
Great, thank you!! I'll merge it as soon as the top text in the Data classes change to increase its scope to support utilities for event sources

PR #326, opened 2021-03-11
feat(idempotent): Include function name in the idempotent key

michaelbrewer:
**Issue #, if available:**
https://github.com/awslabs/aws-lambda-powertools-python/issues/318

## Description of changes:

Prefix the `hashed_idempotency_key` with the context function name:

```python
f"{context.function_name}#{self._generate_hash(data)}"
```

> **NOTE:** Rather than hash this prefix, I kept it in the clear to make debugging a little easier and I did not make 
this configurable, as this makes sense as the default

HOWEVER: You _MIGHT_  get edge cases where the function name has been changed while keeping an existing datastore

**Checklist**

 

* [X] Update tests
* [x] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/config.py
+++ b/aws_lambda_powertools/utilities/idempotency/config.py
@@ -33,8 +33,6 @@ class IdempotencyConfig:
             Max number of items to store in local cache, by default 1024
         hash_function: str, optional
             Function to use for calculating hashes, by default md5.
-        include_function_name: bool, optional
-            Whether to include the function name in the idempotent key
         """
         self.event_key_jmespath = event_key_jmespath
         self.payload_validation_jmespath = payload_validation_jmespath
@@ -44,4 +42,3 @@ class IdempotencyConfig:
         self.use_local_cache = use_local_cache
         self.local_cache_max_items = local_cache_max_items
         self.hash_function = hash_function
-        self.include_function_name = include_function_name
--- a/aws_lambda_powertools/utilities/idempotency/config.py
+++ b/aws_lambda_powertools/utilities/idempotency/config.py
@@ -12,6 +12,7 @@ class IdempotencyConfig:
         use_local_cache: bool = False,
         local_cache_max_items: int = 256,
         hash_function: str = "md5",
+        include_function_name: bool = True,
     ):
         """
         Initialize the base persistence layer
@@ -32,6 +33,8 @@ class IdempotencyConfig:
             Max number of items to store in local cache, by default 1024
         hash_function: str, optional
             Function to use for calculating hashes, by default md5.
+        include_function_name: bool, optional
+            Whether to include the function name in the idempotent key
         """
         self.event_key_jmespath = event_key_jmespath
         self.payload_validation_jmespath = payload_validation_jmespath
@@ -41,3 +44,4 @@ class IdempotencyConfig:
         self.use_local_cache = use_local_cache
         self.local_cache_max_items = local_cache_max_items
         self.hash_function = hash_function
+        self.include_function_name = include_function_name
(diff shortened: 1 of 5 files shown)


michaelbrewer:
@heitorlessa  what do you think of this ?

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -151,7 +151,7 @@ def _get_idempotency_record(self) -> DataRecord:
 
         """
         try:
-            event_record = self.persistence_store.get_record(self.event)
+            event_record = self.persistence_store.get_record(self.event, self.context)
```
Nr18, line 154:
```suggestion
            event_record = self.persistence_store.get_record(event=self.event, context=self.context)
```

Makes sense to remain consistent

tests/functional/idempotency/conftest.py
```diff
@@ -34,6 +35,18 @@ def lambda_apigw_event():
     return event
 
 
+@pytest.fixture
+def lambda_context():
+    lambda_context = {
+        "function_name": "test-func",
+        "memory_limit_in_mb": 128,
+        "invoked_function_arn": "arn:aws:lambda:eu-west-1:809313241:function:test",
```
Nr18, line 43:
```suggestion
        "invoked_function_arn": "arn:aws:lambda:eu-west-1:809313241234:function:test-func",
```

Nit-picky, but the account id is 12 chars and the `function_name` is also in the ARN.

That makes me think I have the following in my project:

```python
import uuid


class LambdaTestContext(LambdaContext):
    def __init__(
        self,
        name: str,
        version: int = 1,
        region: str = "us-east-1",
        account_id: str = "111122223333",
    ):
        self._function_name = name
        self._function_version = str(version)
        self._memory_limit_in_mb = 128
        self._invoked_function_arn = f"arn:aws:lambda:{region}:{account_id}:function:{name}:{version}"
        self._aws_request_id = str(uuid.uuid4())
        self._log_group_name = f"/aws/lambda/{name}"
        self._log_stream_name = str(uuid.uuid4())
```
 
I use this for my tests to supply it with a valid context during testing, not sure how often we need a context but it could be useful I guess.

michaelbrewer: @Nr18  something like this needs to be in the Powertools shared lib. It comes up a lot :)

Nr18: Yeah thats what I figured :P

heitorlessa (changes requested):
Great, thanks!!! Only change here is that we shouldn't add this extra parameter to include function_name, but make it the default. 

Because we're in Beta this won't cause any distress, and due to the fact changing a Function name is not possible from an API perspective - IaC triggers a replacement instead.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -151,35 +153,40 @@ def configure(self, config: IdempotencyConfig) -> None:
...
-            )
+            data = self.event_key_compiled_jmespath.search(event, options=jmespath.Options(**self.jmespath_options))
 
         if self.is_missing_idempotency_key(data):
             if self.raise_on_no_idempotency_key:
                 raise IdempotencyKeyError("No data found to create a hashed idempotency_key")
             warnings.warn(f"No value found for idempotency_key. jmespath: {self.event_key_jmespath}")
 
-        return self._generate_hash(data)
+        generated_hard = self._generate_hash(data)
```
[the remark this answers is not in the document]
michaelbrewer: :) already changed

aws_lambda_powertools/utilities/idempotency/config.py
```diff
@@ -12,6 +12,7 @@ def __init__(
         use_local_cache: bool = False,
         local_cache_max_items: int = 256,
         hash_function: str = "md5",
+        include_function_name: bool = True,
```
heitorlessa, line 15: 💀

heitorlessa (approved):
Thanks a lot for the changes @michaelbrewer, as always ;)

Added a note to update docs to emphasize you can now use a single DDB table to store idempotence state if you wish

PR #341, opened 2021-03-14
feat(parameters): Add force_fetch option

michaelbrewer:
**Issue #, if available:**

https://github.com/awslabs/aws-lambda-powertools-python/issues/131

## Description of changes:

Changes:
- Add new `force_fetch` to always fetch the latest value, even if there is a cached value

Example based on https://github.com/alexcasalboni/ssm-cache-python#complex-invalidation-based-on-signals
```python3
import os


class InvalidCredentials(Exception):
    """Raised when client get an invalid credentials error"""


class Client:
    password = None

    def configure(self, password):
        self.password = password

    def read_record(self):
        """Might raise an InvalidCredentials exception """


DB_PASS_PARAM = os.environ["DB_PASS_PARAM"]
db_client = Client()
db_client.configure(parameters.get_parameter(DB_PASS_PARAM))


def refresh_db_password():
    db_client.configure(parameters.get_parameter(DB_PASS_PARAM, force_fetch=True))


def read_record(is_retry=False):
    try:
        return db_client.read_record()
    except InvalidCredentials:
        if not is_retry:  # avoid infinite recursion
            refresh_db_password()  # force parameter refresh
            return read_record(is_retry=True)


def handler(event, context):
    return {"record": read_record()}
```

**Checklist**

 

* [x] Update tests
* [ ] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/parameters/appconfig.py
+++ b/aws_lambda_powertools/utilities/parameters/appconfig.py
@@ -107,7 +107,12 @@ class AppConfigProvider(BaseProvider):
 
 
 def get_app_config(
-    name: str, environment: str, application: Optional[str] = None, transform: Optional[str] = None, **sdk_options
+    name: str,
+    environment: str,
+    application: Optional[str] = None,
+    transform: Optional[str] = None,
+    force_update: bool = False,
+    **sdk_options
 ) -> Union[str, list, dict, bytes]:
     """
     Retrieve a configuration value from AWS App Config.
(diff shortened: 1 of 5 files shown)


michaelbrewer:
@heitorlessa I added a `force_update` option but i did notice that the `sdk_options` is confusing things for PyCharm

Screen Shot 2021-03-14 at 3 13 10 AM

michaelbrewer:
@heitorlessa I made it an explicit option throughout, the question is what should it be called?

`force_update`, `refresh`, `force_fetch` ?

heitorlessa:
For completion, as @nmoutschen suggested, let's go with `force_fetch` as that doesn't make assumption whether we have something cache to be ignored.

michaelbrewer:
> For completion, as @nmoutschen suggested, let's go with `force_fetch` as that doesn't make assumption whether we have something cache to be ignored.

Ok @nmoutschen  @heitorlessa  i have made the update to `force_fetch`

heitorlessa (approved):

PR #345, opened 2021-03-17
docs(tracer): Fix Tracer typing hinting for Pycharm

michaelbrewer:
**Issue #, if available:**

Typing for Tracer seems to be broken on PyCharm, see screenshot below:

Screen Shot 2021-03-17 at 6 39 54 AM

The location of the `Returns` doc string is causing the PyCharm linter to thing that `Tracer` is  a tuple (possibly also due to the `_default_config` and `_config` before the `__init__` ? 

## Description of changes:

Changes:
* Move up `Returns` to just below the `Parameters`
* Add type hinting for `provider` 
* Add docstring for `provider`

**Checklist**

 

* [x] Update tests
* [x] Update docs

### DIFF
--- a/aws_lambda_powertools/tracing/tracer.py
+++ b/aws_lambda_powertools/tracing/tracer.py
@@ -155,7 +155,7 @@ class Tracer:
         self.__build_config(
             service=service, disabled=disabled, auto_patch=auto_patch, patch_modules=patch_modules, provider=provider
         )
-        self.provider = self._config["provider"]
+        self.provider: BaseProvider = self._config["provider"]
         self.disabled = self._config["disabled"]
         self.service = self._config["service"]
         self.auto_patch = self._config["auto_patch"]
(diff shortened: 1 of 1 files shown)


michaelbrewer:
@heitorlessa  this seems to fix the issue with PyCharm and Tracer

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -153,7 +155,7 @@ def __init__(
         self.__build_config(
             service=service, disabled=disabled, auto_patch=auto_patch, patch_modules=patch_modules, provider=provider
         )
-        self.provider = self._config["provider"]
+        self.provider: BaseProvider = self._config["provider"]
```
heitorlessa, line 158: could you remove this as it has an unintended side effect for escape hatches in X-Ray? e.g. auto-completing methods that we're not providing in the BaseProvider.

heitorlessa (changes requested):
tiny change and it should be good to release!

aws_lambda_powertools/tracing/tracer.py
heitorlessa: Backing up as it's a Pycharm thing only, not VSCode. Best to keep it, and then later I'll add non-ABC methods for the sake of completion

PR #350, opened 2021-03-18
docs: Fix layers example template config

BrockReece:
## Description of changes:

Remove `Location` key from the `MyLambdaFunction` resource in the example template.yml.
According to the [ref docs](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html), `Layers` shouldn't be nested under `Location` 

This tripped me up so I thought I would offer to fix it.

**Checklist**

 

* [ ] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -49,7 +49,7 @@ If using SAM, you can include this SAR App as part of your shared Layers stack, 
 
 === "template.yml"
 
-```yaml hl_lines="5-6 12-14"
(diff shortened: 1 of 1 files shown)


heitorlessa (approved):

heitorlessa:
Thank you so much @BrockReece for this. We definitely slipped that one.

PR #353, opened 2021-03-20
feat(data-classes): Add S3 Object Lambda Event

michaelbrewer:
**Issue #, if available:**

## Description of changes:

Add data class for Amazon S3 Object Lambda

Example usage based on the [AWS Blog Post on S3 Object Lambda](https://aws.amazon.com/blogs/aws/introducing-amazon-s3-object-lambda-use-your-code-to-process-data-as-it-is-being-retrieved-from-s3/)

```python3
import requests
from aws_lambda_powertools.logging.correlation_paths import S3_OBJECT_LAMBDA
from aws_lambda_powertools.utilities.data_classes.s3_object_event import S3ObjectLambdaEvent

logger = Logger()

@logger.inject_lambda_context(correlation_id_path=S3_OBJECT_LAMBDA, log_event=True)
    event = S3ObjectLambdaEvent(event)

    object_context = event.object_context
    request_route = object_context.output_route
    request_token = object_context.output_token
    s3_url = object_context.input_s3_url

    # Get object from S3
    response = requests.get(s3_url)
    original_object = response.content.decode("utf-8")

    # Transform object
    transformed_object = original_object.upper()

    # Write object back to S3 Object Lambda
    s3 = boto3.client("s3")
    s3.write_get_object_response(Body=transformed_object, RequestRoute=request_route, RequestToken=request_token)

    return {"status_code": 200}
```

**Checklist**

 

* [x] Update tests
* [x] Update docs


michaelbrewer:
@heitorlessa  still need to add `S3ObjectSessionContext` class potentially, I just wanted to check if this was something reusable as it would also help with CloudTrail event logs processing

aws_lambda_powertools/utilities/data_classes/s3_event.py
```diff
@@ -187,3 +187,208 @@ def bucket_name(self) -> str:
...
+        return self["arn"]
+
+    @property
+    def session_context(self) -> Optional[Dict[str, Any]]:
+        """ If the request was made with temporary security credentials,
+        this element provides information about the session that was created for those credentials."""
+        return self.get("sessionContext")
+
+
+class S3ObjectEvent(DictWrapper):
```
heitorlessa, line 351: Is this the new S3 Object Lambda?

It looks like it from the structure - can we rename it to S3ObjectLambdaEvent?

Reason being is that searching for S3 Object docs yield S3 Notifications Event. Perhaps a SEO thing that needs propagating but it's easy to get confused given the naming convention it was employed.

aws_lambda_powertools/utilities/data_classes/s3_event.py
```diff
@@ -187,3 +187,208 @@ def bucket_name(self) -> str:
...
+    """S3 object event notification
+
```
heitorlessa, line 353: Could you add an Example section like we have in others?

When looking at the API docs for data classes we have a link to the docs only - we could have an usable or two showing a common use case, which we can point from docs too

heitorlessa (changes requested):
This is awesome and so timely - added one major ask and a nitpick

heitorlessa:
On the readability thing, I need to get my laptop to take a proper look - We can do tomorrow (your morning) as part of an experimental office hours

I'm all for reusing as long as UX doesn't confuse S3 Object with S3 Object Lambda (naming challenging that will confuse for sure)

aws_lambda_powertools/utilities/data_classes/s3_event.py
michaelbrewer: done

aws_lambda_powertools/utilities/data_classes/s3_event.py
michaelbrewer: @heitorlessa  I do have it in the `data_classes.md` but I can see how to do it in general for data-classes package?

heitorlessa:
Let’s discuss some ideas on Monday ;)

aws_lambda_powertools/logging/correlation_paths.py
```diff
@@ -5,3 +5,4 @@
 APPSYNC_RESOLVER = 'request.headers."x-amzn-trace-id"'
 APPLICATION_LOAD_BALANCER = 'headers."x-amzn-trace-id"'
 EVENT_BRIDGE = "id"
+S3_OBJECT = "xAmzRequestId"
```
heitorlessa, line 8: S3 Object Lambda? Unless it's also the same for S3 Event Notifications

heitorlessa (changes requested):
Is correlation path S3_OBJECT or supposed to be S3_OBJECT_LAMBDA?

aws_lambda_powertools/utilities/data_classes/s3_object_event.py
```diff
@@ -0,0 +1,271 @@
...
+        return S3ObjectSessionContext(session_context)
+
+
+class S3ObjectLambdaEvent(DictWrapper):
+    """S3 object lambda event
+
+    Documentation:
+    -------------
+    - https://docs.aws.amazon.com/AmazonS3/latest/userguide/olap-writing-lambda.html
+    """
```
heitorlessa, line 234: Can we add PR example here to shorten the cognitive load when doing the most expected use case? e.g. fetch and transform?

Note that I added three new properties to save some keystrokes as customers will have to fetch these 3 properties anyway

```suggestion


    Example
    -------
    **Fetch and transform original object from Amazon S3**

        import boto3
        import requests
        from aws_lambda_powertools.utilities.data_classes.s3_object_event import S3ObjectLambdaEvent

        session = boto3.Session()

        def lambda_handler(event, context):
            event = S3ObjectLambdaEvent(event)

            # Get object from S3
            response = requests.get(event.input_s3_url)
            original_object = response.content.decode("utf-8")

            # Make changes to the object about to be returned
            transformed_object = original_object.upper()

            # Write object back to S3 Object Lambda
            s3 = session.client("s3")
            s3.write_get_object_response(
                Body=transformed_object, RequestRoute=event.request_route, RequestToken=event.request_token
            )
    """
```

aws_lambda_powertools/utilities/data_classes/s3_object_event.py
```diff
@@ -0,0 +1,271 @@
...
+    @property
+    def user_request(self) -> S3ObjectUserRequest:
+        """Information about the original call to S3 Object Lambda."""
+        return S3ObjectUserRequest(self["userRequest"])
+
+    @property
+    def user_identity(self) -> S3ObjectUserIdentity:
+        """Details about the identity that made the call to S3 Object Lambda."""
+        return S3ObjectUserIdentity(self["userIdentity"])
+
```
heitorlessa, line 260: Suggestion - Add these three required properties that are necessary to fetch and transform objects when using this event to shorten the cognitive load (and keystrokes).

```suggestion
    @property
    def request_route(self) -> str:
        """A routing token that is added to the S3 Object Lambda URL when the Lambda function
        calls `WriteGetObjectResponse`."""
        return S3ObjectContext(self["getObjectContext"]).output_route

    @property
    def request_token(self) -> str:
        """An opaque token used by S3 Object Lambda to match the WriteGetObjectResponse call
        with the original caller."""
        return S3ObjectContext(self["getObjectContext"]).output_token

    @property
    def input_s3_url(self) -> str:
        """A pre-signed URL that can be used to fetch the original object from Amazon S3.

        The URL is signed using the original caller’s identity, and their permissions
        will apply when the URL is used. If there are signed headers in the URL, the
        Lambda function must include these in the call to Amazon S3, except for the Host.

        Example
        -------
        **Fetch original object from Amazon S3**

            import requests
            from aws_lambda_powertools.utilities.data_classes.s3_object_event import S3ObjectLambdaEvent

            def lambda_handler(event, context):
                event = S3ObjectLambdaEvent(event)

                response = requests.get(event.input_s3_url)
                original_object = response.content.decode("utf-8")
                ...
        """
        return S3ObjectContext(self["getObjectContext"]).input_s3_url

```

heitorlessa (changes requested):
two suggestions to make this feature and docs even more awesome ;)

heitorlessa:
Example of UX with these additional changes ;)

image

michaelbrewer:
@heitorlessa quick question why `request_route` over `output_route` for the name?

heitorlessa:
The name was to match SDK params

heitorlessa:
> @heitorlessa quick question why `request_route` over `output_route` for the name?

That's to match writer get object response params - RequestRoute & RequestToken.

When you're writing now you'll instinctively look up the event to find these, and they'll find them now effortlessly to comply with the SDK ;-)

heitorlessa (approved):
Thanks a lot for the super timely addition ;-)

Will release it this week

PR #357, opened 2021-03-24
fix(metrics): AttributeError raised by MediaManager and Typing and docs

michaelbrewer:
**Issue #, if available:**

## Description of changes:

Changes:
- Correct typing to make mypy happier
- Correct highlighted lines
- fix(metrics): AttributeError raised when setting metadata_set
- fix(parser): Use sys.version_info to determine where we need typing_extensions

Before 
```
$ mypy aws_lambda_powertools
....
Found 126 errors in 29 files (checked 99 source files)
```

After 
```
$ mypy aws_lambda_powertools
....
Found 52 errors in 18 files (checked 99 source files)
```

and ignore missing imports:
```
Found 32 errors in 10 files (checked 99 source files)
```

**Checklist**

 

* [x] Update tests
* [x] Update docs

### DIFF
--- a/aws_lambda_powertools/logging/lambda_context.py
+++ b/aws_lambda_powertools/logging/lambda_context.py
@@ -1,3 +1,6 @@
+from typing import Any
+
+
 class LambdaContextModel:
     """A handful of Lambda Runtime Context fields
 
@@ -31,7 +34,7 @@ class LambdaContextModel:
         self.function_request_id = function_request_id
 
 
-def build_lambda_context_model(context: object) -> LambdaContextModel:
+def build_lambda_context_model(context: Any) -> LambdaContextModel:
     """Captures Lambda function runtime info to be used across all log statements
 
     Parameters
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -4,7 +4,7 @@ import logging
 import os
 import random
 import sys
-from typing import Any, Callable, Dict, Union
+from typing import Any, Callable, Dict, Optional, Union
 
 import jmespath
 
@@ -318,12 +318,12 @@ class Logger(logging.Logger):  # lgtm [py/missing-call-to-init]
         self.structure_logs(append=True, correlation_id=value)
 
     @staticmethod
-    def _get_log_level(level: Union[str, int]) -> Union[str, int]:
+    def _get_log_level(level: Union[str, int, None]) -> Union[str, int]:
         """ Returns preferred log level set by the customer in upper case """
         if isinstance(level, int):
             return level
 
-        log_level: str = level or os.getenv("LOG_LEVEL")
+        log_level: Optional[str] = level or os.getenv("LOG_LEVEL")
         if log_level is None:
             return logging.INFO
 
--- a/aws_lambda_powertools/metrics/base.py
+++ b/aws_lambda_powertools/metrics/base.py
@@ -88,7 +88,7 @@ class MetricManager:
         self.service = resolve_env_var_choice(choice=service, env=os.getenv(constants.SERVICE_NAME_ENV))
         self._metric_units = [unit.value for unit in MetricUnit]
         self._metric_unit_options = list(MetricUnit.__members__)
-        self.metadata_set = self.metadata_set if metadata_set is not None else {}
+        self.metadata_set = metadata_set if metadata_set is not None else {}
 
     def add_metric(self, name: str, unit: Union[MetricUnit, str], value: float):
         """Adds given metric
--- a/aws_lambda_powertools/metrics/metric.py
+++ b/aws_lambda_powertools/metrics/metric.py
@@ -1,7 +1,7 @@
 import json
 import logging
 from contextlib import contextmanager
-from typing import Dict
+from typing import Dict, Optional, Union
 
 from .base import MetricManager, MetricUnit
 
@@ -42,7 +42,7 @@ class SingleMetric(MetricManager):
         Inherits from `aws_lambda_powertools.metrics.base.MetricManager`
     """
 
-    def add_metric(self, name: str, unit: MetricUnit, value: float):
+    def add_metric(self, name: str, unit: Union[MetricUnit, str], value: float):
         """Method to prevent more than one metric being created
 
         Parameters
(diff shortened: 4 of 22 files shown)


michaelbrewer:
@heitorlessa  MyPy found a bug for us :)

aws_lambda_powertools/logging/lambda_context.py
```diff
@@ -31,7 +34,7 @@ def __init__(
         self.function_request_id = function_request_id
 
 
-def build_lambda_context_model(context: object) -> LambdaContextModel:
+def build_lambda_context_model(context: Any) -> LambdaContextModel:
```
heitorlessa, line 37: `context: Any`

I wonder if we could use this implementation as a type to have a more realistic representation of what we expect as input: https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/main/awslambdaric/lambda_context.py

aws_lambda_powertools/logging/logger.py
```diff
@@ -318,12 +318,12 @@ def set_correlation_id(self, value: str):
...
 
     @staticmethod
-    def _get_log_level(level: Union[str, int]) -> Union[str, int]:
+    def _get_log_level(level: Union[str, int, None]) -> Union[str, int]:
         """ Returns preferred log level set by the customer in upper case """
         if isinstance(level, int):
             return level
 
-        log_level: str = level or os.getenv("LOG_LEVEL")
+        log_level: Optional[str] = level or os.getenv("LOG_LEVEL")
```
heitorlessa, line 326: Shouldn't this be 

```suggestion
        log_level: Union[str, int, None] = level or os.getenv("LOG_LEVEL")
```

aws_lambda_powertools/metrics/base.py
```diff
@@ -88,7 +88,7 @@ def __init__(
         self.service = resolve_env_var_choice(choice=service, env=os.getenv(constants.SERVICE_NAME_ENV))
         self._metric_units = [unit.value for unit in MetricUnit]
         self._metric_unit_options = list(MetricUnit.__members__)
-        self.metadata_set = self.metadata_set if metadata_set is not None else {}
+        self.metadata_set = metadata_set if metadata_set is not None else {}
```
heitorlessa, line 91: Nice!!!!

aws_lambda_powertools/metrics/metric.py
```diff
@@ -109,11 +109,11 @@ def single_metric(name: str, unit: MetricUnit, value: float, namespace: str = No
     SchemaValidationError
         When metric object fails EMF schema validation
     """
-    metric_set = None
+    metric_set: Optional[Dict] = None
```
heitorlessa, line 112: We can get rid of this - It was added in error. `metric_set` being created within `try` block will also be available in `finally`.

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -71,15 +71,15 @@ def do_something():
         When metric object fails EMF schema validation
     """
 
-    _metrics = {}
-    _dimensions = {}
-    _metadata = {}
+    _metrics: Dict[str, Any] = {}
+    _dimensions: Dict[str, Any] = {}
```
heitorlessa, line 75: Did MyPy caused issues with `Dict[str, str]`?

Dimensions can only be strings.

```
>>> my_metrics._dimensions
{'test_dimension': 'test', 'test_dimension_2': 'test'}
```

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -71,15 +71,15 @@ def do_something():
         When metric object fails EMF schema validation
     """
 
-    _metrics = {}
-    _dimensions = {}
-    _metadata = {}
+    _metrics: Dict[str, Any] = {}
```
heitorlessa, line 74: This is possible the most complex, so happy to leave as Any for now until there is a cleaner way.

For example, it might deserve a generic like MetricsContainer, since at the end this holds:

```
>>> my_metric._metrics
{'metric_one': defaultdict(<class 'list'>, {'Unit': 'Count', 'Value': [1.0]}), 'metric_two': defaultdict(<class 'list'>, {'Unit': 'Count', 'Value': [1.0]})}
```

It's a nested dict, where the dictionary is a default dict with `Unit` holding a `str`, and `Value` holding a `List[Number]`

aws_lambda_powertools/utilities/parser/types.py
```diff
@@ -1,13 +1,14 @@
...
 from typing import TypeVar
 
 from pydantic import BaseModel
 
 # We only need typing_extensions for python versions <3.8
-try:
+if sys.version_info >= (3, 8):
     from typing import Literal  # noqa: F401
-except ImportError:
+else:
```
heitorlessa, line 11: That's an interesting one, perhaps `noqa` might not be needed after all?

heitorlessa (changes requested):
Looks great!!!!! Asking two minor changes only

aws_lambda_powertools/logging/logger.py
michaelbrewer: @heitorlessa  i can't be an int at this point as we already exist early before this.

michaelbrewer: ie: 
```python3
if isinstance(level, int):
```

aws_lambda_powertools/metrics/metric.py
michaelbrewer: If we do this then PyCharm complains and wants it to be a global

Screen Shot 2021-03-25 at 1 09 38 AM

aws_lambda_powertools/utilities/parser/types.py
michaelbrewer: @heitorlessa  that is for the unused imports in `types.py`

michaelbrewer: @heitorlessa adding `if sys.version_info >= (3, 8):` is a recommendation from MyPy to aid the detection some how

aws_lambda_powertools/metrics/metrics.py
michaelbrewer: Yep, this is kind of where typing falls apart, but also it does highlight the inherit complexity too.

aws_lambda_powertools/logging/lambda_context.py
michaelbrewer: Future idea? Maybe part of the testing libs as well?

aws_lambda_powertools/logging/logger.py
heitorlessa: missed it despite bring right in front of me.

heitorlessa (approved):

PR #368, opened 2021-03-27
build(pre-commit): Add pre-commit to `make pr`

michaelbrewer:
**Issue #, if available:**

## Description of changes:

`pre-commit` runs additional checks that `make pr` does not perform like `trailing-whitespace`

Changes:
- Add `pre-commit` as a target in the Makefile

**Checklist**

 

* [x] Update tests
* [x] Update docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,10 @@ test:
 coverage-html:
 	poetry run pytest -m "not perf" --cov=aws_lambda_powertools --cov-report=html
 
-pr: lint test security-baseline complexity-baseline
+pre-commit:
+	pre-commit run --show-diff-on-failure
+
+pr: lint pre-commit test security-baseline complexity-baseline
 
 build: pr
 	poetry build


heitorlessa:
Hey Michael, out of curiosity, did you run into those before?

`make dev` installs pre-commit hooks so they're run upon a commit.

That said, these are not run in CI and this is an improvement - is that what you intended?

Thanks ;)

michaelbrewer:
@michaelbrewer yes I do run "make dev" and the pre-commit does run, but "make lint" does not include all of the same logic as "pre-commit". 

So what happens is:
```
$ git add .
stage my commit
$ make pr
passes 
$ git commit -m "..."
fail due to white spaces 
$ git add .
$ git commit -m "..."
```

michaelbrewer:
@heitorlessa maybe there is an alternative solution with the same intent :)

heitorlessa:
Gotcha, that's what I thought. Sadly there is no alternative, as the pre-commit project decided not to add them as it deems to be a non-safe operation. I was thinking on perhaps removing some of them, but there are higher priority tasks now. 

Merging this now ;) Thank you!

Issue #325, opened 2021-03-10
RFC: API Gateway Proxy Event utils


michaelbrewer:
## Key information

* RFC PR: (leave this empty)
* Related issue(s), if known:
* Area: Utilities
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-python/#tenets):  Yes

## Summary
[summary]: #summary

> One paragraph explanation of the feature.

Add the ability to help map multiple API Gateway Proxy events to a single lambda much like how Chalice or Lambda Proxy does it, but in a very simple and light weight way and still be compatible with Powertools

## Motivation
[motivation]: #motivation

Simplify the work needed to setup API Gateway Proxy Lambdas that support multiple endpoints on a single lambda

## Proposal
[proposal]: #proposal

Build something like: https://github.com/vincentsarago/lambda-proxy

But also allow the develop to keep their existing `handler` with all of the powertools decotators supported kind of like how
#324 works.

```python3

app = APIGatewayProxy()


@app.post(uri="/merchant")
def create_merchant(merchant: dict) -> dict:
   # return a 200 OK response with JSON details
    return {"id": ...}

@app.get(uri="/merchant/{uid}", include_header=True)
def get_merchant(uid: str, headers: dict):
   return {"name":...}


def handler(event, context):
    return app.resolve(event, context)


```

## Drawbacks
[drawbacks]: #drawbacks

> Why should we *not* do this?
> Do we need additional dependencies? Impact performance/package size?
No additional dependencies

## Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

* **What other designs have been considered? Why not them?**
* **What is the impact of not doing this?**

## Unresolved questions
[unresolved-questions]: #unresolved-questions
> Optional, stash area for topics that need further development e.g. TBD

heitorlessa:
Hey @michaelbrewer - Could you let us know when the RFC body is ready to review? It's missing a few sections like  drawbacks, rationale, a more complete proposal besides the link, and any open questions you might have (if any), etc.

michaelbrewer:
@heitorlessa  here is a super lightweight implementation:

```python3
from typing import Any, Dict, Tuple, Callable, List

from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEvent
from aws_lambda_powertools.utilities.typing import LambdaContext


class ApiGatewayResolver:
    def __init__(self):
        self._resolvers: List[Dict] = []

    def _register(
[...]

PR #369, opened 2021-03-28
feat(event-handler): add http ProxyEvent handler

michaelbrewer:
**Issue #, if available:**
#325

## Description of changes:

Very basic / minimal http proxy event handler (ALB Events, Http api V1 and V2)

Changes:
- feat(data-classes):Add `path` and `http_method` methods to `BaseProxyEvent`
- feat(data-classes): Add `json_body` helper property to `BaseProxyEvent` that json parses `body`
- docs: Add example `ApiGatewayResolver` event handler
- tests: Add test coverage for `ApiGatewayResolver`
- tests: Add shared test `utils`.`load_event`
- feat(context): Add `current_event` and `lambda_context`
- feat(decorators): Add shortcut decorators for `get`, `post`, `patch`, `put`, `delete`
- feat(cors): Add basic CORS flag
- feat(compress): Add gzip compression flag
- feat(binary): Add `isBase64Encoded` support for binary content
- feat(cache-control): Add `cache_control` option, which defaults to `no-cache` for non-200 responses
- feat(rest): Add simplification for functions just returning a dict
- feat(rest): Use shared JSONEncoder that can handle encoding `Decimal`
- feat(response): Add Response class for fine grained control of the http headers returned.
- feat(cors): Add CORSConfig class for higher control over access control headers
- feat(cors): Built in preflight support

**Features**:
* [x] Routes - `@app.get("/foo")`
* [x] Path expressions - `@app.delete("/delete/<uid>")`
* [x] Cors - `@app.post("/make_foo", cors=True)` or via CORSConfig
* [x] Base64 encode binary - `@app.get("/logo.png")`
* [x] Gzip Compression - `@app.get("/large-json", compression=True)`
* [x] Cache-control - `@app.get("/foo", cache_control="max-age=600")`
* [x] Rest API simplification with function returns a Dict
* [x] Support function returns a `Reponse` object which give fine grained control of the headers
* [x] JSON encoding of Decimals

```python3
import json
from typing import Dict, Tuple

from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEvent
from aws_lambda_powertools.utilities.event_handler.api_gateway import ApiGatewayResolver, CORSConfig, ProxyEventType, Response

# proxy_type: "APIGatewayProxyEvent", "APIGatewayProxyEventV2", "ALBEvent"
app = ApiGatewayResolver(
    proxy_type=ProxyEventType.http_api_v1,
    cors=CORSConfig(
        allow_origin="https://www.foo.com/",  # Whitelisted site
        expose_headers=["foo1"],
        allow_headers=["x-custom-header"],
        max_age=100,
        allow_credentials=True,
    )
)

@app.get("/foo", compress=True)
def get_foo() -> Tuple[int, str, str]:
    # Matches on http GET and proxy path "/foo"
    return 200, "text/html", "Hello"

@app.get("/logo.png")
def get_logo() -> Tuple[int, str, bytes]:
    # Base64 encode 
    logo: bytes = load_logo()
    return 200, "image/png", logo

@app.post("/make_foo", cors=True)
def make_foo() -> Tuple[int, str, str]:
    # Matches on http POST and proxy path "/make_foo"
    post_data: dict = app. current_event.json_body
    return 200, "application/json", json.dumps(post_data["value"])

@app.delete("/delete/<uid>")
def delete_foo(uid: str) -> Tuple[int, str, str]:
    # Matches on http DELETE and proxy path starting with "/delete/"
    assert isinstance(app.current_event, APIGatewayProxyEvent)
    assert app.current_event.request_context.authorizer.claims is not None
    assert app.current_event.request_context.authorizer.claims["username"] == "Mike"
    return 200, "application/json", json.dumps({"id": uid})

@app.get("/hello/<username>")
def hello_user(username: str) -> Tuple[int, str, str]:
    return 200, "text/html", f"Hello {username}!"

@app.get("/rest")
def rest_fun() -> Dict:
    # Returns a statusCode: 200, Content-Type: application/json and json.dumps dict
    return {"message": "Example"}

@app.get(“/foo3”)
def foo3() -> Response:
    return Response(
        status_code=200,
        body=json.dumps({“message”: “Foo3”}),
        headers={“custom-header”: “value”}, 
    )
```

Unlikely bonus features:
* Brotli Compression - via an extra install option 
* OpenAPI support - to generate swagger docs
* Rendering of html templates - to allow for html templates using Python 3 string templates or Jinja (extra deps)
* Multiple method mapping for a single func ie: `@app.route(methods=["GET", "POST"],...`
* Build in idempotency support

---

**Checklist**

 

* [x] Update tests
* [ ] Update docs


michaelbrewer:
@heitorlessa an ultralight handler for http proxy lambdas

michaelbrewer:
@heitorlessa  maybe i can do the same for AppSync handler and include the current event in the “app” rather than a method parameter (which might run into shadowing issues0

michaelbrewer:
@heitorlessa  - as an example i added CORS support, the only things i would need regularly would be compression (and therefore also binay support via bas64encoding)

jplock:
This is really cool! Would it make sense to make a Response class instead of needing to return the tri-tuple to give more control over the responses (I’m thinking specifically about the headers)?

michaelbrewer:
yeah, i was on the fence about that @jplock it would be simple to add.

michaelbrewer:
@jplock supporting a tuple  or dict as a response is pretty concise and does not require any special imports (albeit not explicit, especially in the case of a triple tuple):

```python
@app.get(“/foo1”)
def foo():
    return 200, “text/html”, “Foo1”

@app.get(“/foo2”)
def foo1():
    return {“message”: “Foo2”}
```


Maybe in the case of using a special `Response` class it is to either be explicit and/or provide access to the headers.

```python3
from aws_lambda_powertools.utilities.event_handler.api_gateway import Response

@app.get(“/foo3”)
def foo3() -> Response:
    return Response(
        status_code=200, 
        headers={“Content-Type”: “application/json”}, 
         body=json.dumps({“message”: “Foo3”}),
    )
```

michaelbrewer:
@jplock  i updated this handler to support returning a Response with control over which headers are returned.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,144 @@
...
+    def resolve(self, event, context) -> Dict[str, Any]:
+        self.current_event = self._as_data_class(event)
+        self.lambda_context = context
+        route, args = self._find_route(self.current_event.http_method, self.current_event.path)
+        result = route.func(**args)
+
+        if isinstance(result, Response):
+            response = result
+        elif isinstance(result, dict):
+            response = Response(status_code=200, content_type="application/json", body=json.dumps(result))
```
jplock, line 107: do you want to default `json.dumps` to `json.dumps(result, separators=(',', ':'), indent=None, sort_keys=True)` for consistent minimized outputs?

also would it make sense to set `content_type="application/json; charset=utf-8"`?

michaelbrewer: @jplock true, it could be more concise.

michaelbrewer: @jplock - seems like UTF-8 is the default encoding for JSON anyway: https://www.ietf.org/rfc/rfc4627.txt

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,149 @@
...
+        self.status_code = status_code
+        self.body = body
+        self.base64_encoded = False
+        self.headers: Dict = headers or {}
+        self.headers.setdefault("Content-Type", content_type)
+
+    def add_cors(self, method: str):
+        self.headers["Access-Control-Allow-Origin"] = "*"
+        self.headers["Access-Control-Allow-Methods"] = method
+        self.headers["Access-Control-Allow-Credentials"] = "true"
```
heitorlessa, line 44: Let's allow customers to specify a more strict CORS header as this is overly permissive. If they don't specify.

For example, `Allow-Origin` being `*`, and `Allow-Methods` isn't allowing or testing for pre-flight CORS (`OPTIONS`) and it'll fail.

heitorlessa: It's also missing [Allow-Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers) - For that, you'd want to have a default of `Content-Type`.

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -0,0 +1,340 @@
...
+    # AND the function returns a 200 status code
+    result = handler({"path": "/success", "httpMethod": "GET"}, None)
+
+    # THEN return the set Cache-Control
+    headers = result["headers"]
+    assert headers["Content-Type"] == TEXT_HTML
+    assert headers["Cache-Control"] == "max-age=600"
+
+
+def test_cache_control_non_200():
```
heitorlessa, line 280: Could you also test for CORS being set on non-200, too?

Otherwise, the browser will get a Internal Server Error due to missing some CORS Headers, despite receiving a 400/500 from Lambda in this case.

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -0,0 +1,340 @@
...
+    path = Path(str(Path(__file__).parent.parent.parent.parent) + "/docs/media/" + file_name)
+    return path.read_bytes()
+
+
+LOAD_GW_EVENT = load_event("apiGatewayProxyEvent.json")
+TEXT_HTML = "text/html"
+APPLICATION_JSON = "application/json"
+
+
+def test_alb_event():
```
heitorlessa, line 26: could you split that into a separate file? For example, [ALB has a maximum of 1MB response size](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#prepare-lambda-function) -- Do we want to take limits like that into account as ALB and API GW differ?.

ALB also supports [healthchecks](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#enable-health-checks-lambda) (disabled by default), but I think that's better to do in Dataclass so one can easily check for that in a method  by inspecting the user header (don't think is necessary but throwing it out, etc.

heitorlessa (changes requested):
Looking great and slim ;) Loved it! Added some comments on CORS not being complete, and whether we want to take into account ALB tight limits like 1MB response etc

tests/functional/event_handler/test_api_gateway.py
michaelbrewer: @heitorlessa CORS headers should always be set, but i will at a test to enforce that.

aws_lambda_powertools/event_handler/api_gateway.py
michaelbrewer: Ok @heitorlessa i will further in to that. I was thinking we could provide a simple default. And then if they want more flexibility, then just use the `Response` class.

Any ideas on the DX for this extra settings?

tests/functional/event_handler/test_api_gateway.py
michaelbrewer: @heitorlessa do you know what’s the limit for the response size in the API GW? When you setup the resolver  you do have to specify the type of integration. So we could to handle it there? Or do you want a class per integration type? It seems unneeded.

tests/functional/event_handler/test_api_gateway.py
heitorlessa: Perhaps it's more explicit to have a method for CORS where you could have optional kwargs to override origin (ala FastAPI), etc.  Reasoning is that this would be across all method routes and one can override with headers in Response when necessary (avoid having to add CORS on every method route)

tests/functional/event_handler/test_api_gateway.py
heitorlessa: API GW would be 10MB but effectively Lambda limits synchronous requests to 6MB. So 6M for API GW and 1M for Lambda.

I have second thoughts on whether we should account for these, lemme ask @pcolazurdo or @am29d on this -- Do you think we should raise an Exception if the Lambda response is bigger than the event source supports OR let it explicitly fail as it is today?

Since we're not technically a framework and don't handle infra but app code, I'm heading towards the latter - Let API GW or ALB handle that as we don't have control over these limits and will unlikely be in the known on a timely fashion if these ever get up to date.

am29d: I would let it fail as it is today, the error message is clear and it is easy to spot the source and fix it.

tests/functional/event_handler/test_api_gateway.py
michaelbrewer: @heitorlessa  ok, maybe we can flag the CORS part as `Alpha`, and get some feedback on it.

aws_lambda_powertools/event_handler/api_gateway.py
michaelbrewer: I will have a look at the DX for [Chalice and CORS](https://aws.github.io/chalice/tutorials/basicrestapi.html#cors-support) and other frameworks. Seems like the lightweight flag would just add the `OPTIONS` route and “*” defaults.

michaelbrewer: @heitorlessa I have added a more complete implementation (looks like CORS is a very large part of this which is a bummer, but for those that struggle with this, I guess this is a bonus)

tests/functional/event_handler/test_api_gateway.py
michaelbrewer: So I am leaving this as is for now

tests/functional/event_handler/test_api_gateway.py
michaelbrewer: Should I just add a test case to very this is set for non-200?

michaelbrewer: I would assume for `ValueError(f"No route found for '{method}.{path}'")` case we should rather return a 404 and include the CORS if needed

heitorlessa:
Cc @marcioemiranda your feedback on this would be great before we merge and release this week.

What are your thoughts on CORS UX?

michaelbrewer:
> Cc @marcioemiranda your feedback on this would be great before we merge and release this week.
> 
> What are your thoughts on CORS UX?

Outstanding questions i have are around exceptions. Should we allow for it to ripple up, or should we always handle them and return an appropriate http response code? And in the case of CORS, also include the relavant http headers.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _find_route(self, method: str, path: str) -> Tuple[Route, Dict]:
+        for route in self._routes:
+            if method != route.method:
+                continue
+            match: Optional[re.Match] = route.rule.match(path)
+            if match:
+                return route, match.groupdict()
+
+        if method == "OPTIONS" and self._cors is not None:
```
michaelbrewer, line 213: @heitorlessa i noticed that Chalice does not do this for the api gateway, i am not exactly sure why not.

pcolazurdo (approved):
Adding some comments about CORS best practices

michaelbrewer:
> Adding some comments about CORS best practices

Sure we can add some of these to the docs. I was aiming to all for an `cors=True` shortcut like in Chalice (and other frameworks), they got you up and running quickly. For those that want full control, they should use CORSConfig.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+    alb_event = "ALBEvent"
+    api_gateway = http_api_v1
+
+
+class CORSConfig(object):
+    _REQUIRED_HEADERS = ["Content-Type", "X-Amz-Date", "Authorization", "X-Api-Key", "X-Amz-Security-Token"]
+
+    def __init__(
+        self,
+        allow_origin: str = "*",
```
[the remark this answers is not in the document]
michaelbrewer: This can be overridden when using `CORSConfig` directly. Otherwise this is by design.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+        expose_headers: List[str] = None,
+        max_age: int = None,
+        allow_credentials: bool = True,
+    ):
+        self.allow_origin = allow_origin
+        self.allow_headers = set((allow_headers or []) + self._REQUIRED_HEADERS)
+        self.expose_headers = expose_headers or []
+        self.max_age = max_age
+        self.allow_credentials = allow_credentials
+
```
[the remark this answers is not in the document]
michaelbrewer: Allow methods it only used by PreFlight which is already handled elsewhere.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+class CORSConfig(object):
+    _REQUIRED_HEADERS = ["Content-Type", "X-Amz-Date", "Authorization", "X-Api-Key", "X-Amz-Security-Token"]
+
+    def __init__(
+        self,
+        allow_origin: str = "*",
+        allow_headers: List[str] = None,
+        expose_headers: List[str] = None,
+        max_age: int = None,
+        allow_credentials: bool = True,
```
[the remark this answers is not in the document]
michaelbrewer: I am not sure about this for `cors=True` it might be easier for the DX to turn it on. I will think about our use case where we needed it on vs off.

tests/functional/event_handler/test_api_gateway.py
heitorlessa: If CORS was configured we should return on error too, otherwise the browser will interpret as 500 due to lack of headers despite being a 404

aws_lambda_powertools/event_handler/api_gateway.py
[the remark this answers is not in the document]
heitorlessa: That's the only thing I'm nervous about it - I understand people hate CORS for not knowing it, and wanting to have a quick switch but this more often than we'd like end up in a bad place.

I'm happy with the entire implementation except these two things then we can merge, or else this will be flagged by security (and for good reasons), and we'd have to take the feature down - this will impact customers the hard way

1. Allow origin should be a required option. Docs can show * in dev and stress the danger of it. If customers want to use * in prod it's a consent rather than a suboptimal default

2. Allow credentials set to False by default

heitorlessa (changes requested):
Two small CORS changes to ensure minimum security and for package not to be later flagged for insecure practices.

These are making allow origin a required argument, and credentials set to False by default.

It's not the easiest UX but a balance we must meet

tests/functional/event_handler/test_api_gateway.py
michaelbrewer: Ok i will add that as well as 404 handling for unmapped routes. 

And a 500 error for unhandled exceptions?

aws_lambda_powertools/event_handler/api_gateway.py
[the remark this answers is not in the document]
michaelbrewer: One thing that you can do with Chalice is set the CORS behavior once, and then just have a Boolean flag for which end points you want to have enabled. Should we also allow for that? Maybe for the DX we simplify it to just be something you can set once for the whole lambda. And only toggle on or off at the endpoints

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: :sigh: that's right - I totally forgot we do this at the API Gateway resource level, so it must be configured by the customer managing the API Gateway while the app code level is only responsible to ensure the response has the right headers.

You can safely remove that, sorry for the extra work

aws_lambda_powertools/event_handler/api_gateway.py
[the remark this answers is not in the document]
heitorlessa: Happy with that - have a method to set CORS or any default headers if you'd like so you can reuse it. Then a toggle one

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+
+
+class ProxyEventType(Enum):
+    http_api_v1 = "APIGatewayProxyEvent"
+    http_api_v2 = "APIGatewayProxyEventV2"
+    alb_event = "ALBEvent"
+    api_gateway = http_api_v1
+
+
+class CORSConfig(object):
```
heitorlessa, line 21: Could you add a docstring to quickly note what each CORS option is for to help educate customers who are new to this?

I know many struggle with the origin value.

**Note**: I can do this in the docs PR too, so don't worry about it.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+        if isinstance(self.body, bytes):
+            self.base64_encoded = True
+            self.body = base64.b64encode(self.body).decode()
+        if self.body:
+            result["isBase64Encoded"] = self.base64_encoded
+            result["body"] = self.body
+        return result
+
+
+class ApiGatewayResolver:
```
heitorlessa, line 111: **Note to self**: Docstrings missing

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -0,0 +1,223 @@
...
+            self.cors = CORSConfig()
+        elif isinstance(cors, CORSConfig):
+            self.cors = cors
+        else:
+            self.cors = None
+        self.compress = compress
+        self.cache_control = cache_control
+
+
+class Response:
```
heitorlessa, line 76: **Note to self**: docstrings missing

heitorlessa:
Let me know whether you want me to merge and make these changes in a separate PR, or if you'd like to make them now so we can merge it clean:

- [ ] Set CORS once and toggle `cors=True` as on/off. Perhaps, a single method to add default headers that might be needed over and over, in which CORS could also be added works too.
    - [ ] Since CORS would have to be defined via a method, you could leave `cors=True` as the default per method, if it's not set then it wouldn't have any effect, and when it is you have one less step to do.
- [ ] Remove preflight (OPTIONS) since that will have to happen at the API Gateway/ALB resource level not at the integration
- [ ] [Optional] Docstrings are missing for Resolvers and CORS. These can be added in the next PR concerning docs too.

I'm approving now so I can either merge or wait for these two last changes.

heitorlessa (approved):

michaelbrewer:
@heitorlessa I made the code changes. Just the docstrings need to be added.

aws_lambda_powertools/event_handler/api_gateway.py
pcolazurdo: @heitorlessa Not really. According to documentation: ["For a Lambda proxy integration or HTTP proxy integration, you can still set up the required OPTIONS response headers in API Gateway. However, your backend is responsible for returning the Access-Control-Allow-Origin and Access-Control-Allow-Headers headers, because a proxy integration doesn't return an integration response."](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-cors.html) - Even testing with Chalice in local mode you get the right headers when using HTTP OPTIONS, so these are being created in the function.

michaelbrewer: haha. should i stay or should i go. 

Either way, we would need docs explaining it :)

marcioemiranda:
> Cc @marcioemiranda your feedback on this would be great before we merge and release this week.
> 
> What are your thoughts on CORS UX?

@heitorlessa yesterday I posted a feature request that is totally related to this one (Sorry, I haven't reviewed the current Pull Requests, just the issues). Please review if any of that makes sense here. 
@michaelbrewer great job on this feature. I will want to use it asap.

Anyways, my initial thought was having something like this @app.post("/make_foo", cors=True).
However, it's likely that one will enable cors for all methods, so the centralized config makes sense.

I set the following headers in my functions for all responses, both success and error:
headers = {
    "Access-Control-Allow-Origin": "*",
    "Access-Control-Expose-Headers": "x-amzn-ErrorType"
} 

Some additional comments:
- Does it make sense to assume a default 200 | application/json response and assume the returned value is the response model? I would say this would represent the majority of the usage, so writing the tuple for every method seemed cumbersome. 
- Looking at one of the examples I wonder if it would make sense to have an additional decorator for authorization.

Regarding exceptions I would say always handle them and map to proper status codes. Expose the exceptions to developers so that they can use them in the code. For example, a ResourceNotFoundError could map to status code 404. A ServiceUnavailableError, triggered when a downstream service is down, could be mapped to another status code. The CORS headers should be included in the error responses as well.

heitorlessa:
Alright... tested on CORS, here's the confusion clarification at least on my part:

1. [CORS Pre-flight can be handled by API GW at the resource level](https://github.com/aws-samples/aws-serverless-shopping-cart/blob/b87bfa8534ace0741ce97e1f8b9e9a9a63124cee/backend/product-mock.yaml#L29) by using a MOCK response on OPTIONS method. You can also forward to a Lambda function, if you'd like to handle it (/*, ANY). 
2. When running locally, pre-flight requests **might** be handled by a [local emulator like SAM CLI if CORS configuration is configured](https://github.com/aws/aws-sam-cli/issues/323#issuecomment-550090879) - This can be tricky as local endpoints will be different thus having people changing to `*` that can lead to security issues in prod

Verdict: We must handle CORS pre-flight requests because **a)** people might configure ANY method and `/*` to a single Lambda (not recommended but can happen), and **b)** [running locally can be error prone with CORS](https://github.com/aws/chalice/blob/1ee44df31a0d21d3079fafe74d66072db6d0eec4/chalice/local.py#L525) so we should handle that to improve the experience (why this lib exist).


API GW handling pre-flight CORS response at the resource level

image

image


@michaelbrewer could you add pre-flight CORS back and I'll merge it, please? It'll be easier to revert that as you have it locally than me creating another PR just to add that back in.

I'll add the docstrings, examples, and the official docs in the next PR.

Thanks a lot everyone for the help here

heitorlessa (approved):
Merging this now so we can start the docs in a separate PR - As agreed, we'll release this as Beta by Friday EOD, so we can iterate on UX.

I always find that as soon as we write docs we find some good areas to improve, then more customers using it they can give us feedback on areas to iron out before we take a 1-way door decision.

Thanks a lot again Michael, and everyone, for the help on getting this out !

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: We'll have to revert the `separator` optimization - This breaks people's unit tests, for example:

```python
assert ret["body"] == json.dumps({"message": "hello universe"})
```

```python
FAILED tests/unit/test_hello_world_function.py::test_lambda_handler - assert '{"message":"hello universe"}' == '{"message": "hello universe"}'
```

I'll have dinner and release a patch version with it

Issue #383, opened 2021-04-05
@tracer.capture_method or middleware and xray does not handle errors


michaelbrewer:
```python3
import json
from aws_lambda_powertools import Tracer
import boto3

tracer = Tracer()


@tracer.capture_method
def my_method():
    sts = boto3.client("sts")
    sts.get_caller_identity()
    raise RuntimeError("Error happens!")


@tracer.capture_lambda_handler
def lambda_handler(event, context):
    my_method()
    return {"statusCode": 200, "body": json.dumps({"hello": "world"})}

```

Without `@tracer.capture_method` 

<img width="1396" alt="Screen Shot 2021-04-04 at 10 31 01 PM" src="https://user-images.githubusercontent.com/5442469/113540661-745f4b00-9595-11eb-9186-afb141a297d2.png">

With `@tracer.capture_method`

<img width="1331" alt="Screen Shot 2021-04-04 at 10 32 12 PM" src="https://user-images.githubusercontent.com/5442469/113540735-9d7fdb80-9595-11eb-9d0f-0d3134f85c9f.png">

```logs
[ERROR]	2021-04-05T05:33:00.732Z	8a4d5027-4faa-4f4f-8d08-df63b7ea72ed	Failed to serialize ## lambda_handler
Traceback (most recent call last):
  File "/var/task/aws_lambda_powertools/tracing/tracer.py", line 313, in decorate
    response = lambda_handler(event, context, **kwargs)
  File "/var/task/app.py", line 17, in lambda_handler
    my_method()
  File "/var/task/aws_lambda_powertools/tracing/tracer.py", line 612, in decorate
    response = method(*args, **kwargs)
  File "/var/task/app.py", line 12, in my_method
    raise RuntimeError("Error happens!")
RuntimeError: Error happens!

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/var/task/aws_xray_sdk/core/models/entity.py", line 263, in serialize
    return json.dumps(self.to_dict(), default=str)
  File "/var/task/aws_xray_sdk/core/models/subsegment.py", line 157, in to_dict
    subsegment_dict = super(Subsegment, self).to_dict()
  File "/var/task/aws_xray_sdk/core/models/entity.py", line 284, in to_dict
    entity_dict[key]['working_directory'] = self.cause['working_directory']
TypeError: string indices must be integers
```

heitorlessa:
[...]

PR #384, opened 2021-04-05
fix: lock X-Ray SDK to 2.6.0 due to regression

heitorlessa:
**Issue #, if available:** #383

## Description of changes:

X-Ray SDK 2.7.0 replaced `jsonpickle` with own implementation for serializing subsegments data. This breaks when there's an exception raised as per #383. 

This PR locks X-Ray SDK to 2.6.0 as it's the last working version until there's a fix.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


michaelbrewer:
Thanks for this. I did manage to pin the older version via "pipenv" on my existing projects.

PR #403, opened 2021-04-18
feat(parser): Support for API GW v1 proxy schema & envelope

risenberg-cyberark:
This schema exists in the validator but not in the parser. It was requested in the webinar audience.
This feature will also provide an envelope class which will parse the string body parameter as a pydantic Model.

Also added new identity parameters as defined at https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html


aws_lambda_powertools/utilities/parser/envelopes/apigw.py
```diff
@@ -0,0 +1,32 @@
...
+
+from ..models import APIGatewayProxyEventModel
+from ..types import Model
+from .base import BaseEnvelope
+
+logger = logging.getLogger(__name__)
+
+
+class ApiGatewayEnvelope(BaseEnvelope):
+    """Api Gateway envelope to extract data within detail key"""
```
heitorlessa, line 12:
```suggestion
    """API Gateway envelope to extract data within body key"""
```

aws_lambda_powertools/utilities/parser/models/apigw.py
```diff
@@ -0,0 +1,109 @@
...
+
+    @root_validator()
+    def check_message_id(cls, values):
+        message_id, event_type = values.get("messageId"), values.get("eventType")
+        if message_id is not None and event_type != "MESSAGE":
+            raise TypeError("messageId is available only when the `eventType` is `MESSAGE`")
+        return values
+
+    @root_validator(pre=True)
+    def check_both_http_methods(cls, values):
```
heitorlessa, line 96: could you add a note why this check is needed?

aws_lambda_powertools/utilities/parser/models/apigw.py
```diff
@@ -0,0 +1,109 @@
...
+        http_method, req_ctx_http_method = values.get("httpMethod"), values.get("requestContext", {}).get(
+            "httpMethod", ""
+        )
+        if http_method != req_ctx_http_method:
+            raise TypeError("httpMethods and requestContext.httpMethod must be equal")
+        return values
+
+    @root_validator(pre=True)
+    def check_both_paths(cls, values):
```
heitorlessa, line 105: could you add a note why this check is needed?

aws_lambda_powertools/utilities/parser/models/apigw.py
```diff
@@ -0,0 +1,109 @@
...
+    queryStringParameters: Optional[Dict[str, str]]
+    multiValueQueryStringParameters: Optional[Dict[str, List[str]]]
+    requestContext: APIGatewayEventRequestContext
+    pathParameters: Optional[Dict[str, str]]
+    stageVariables: Optional[Dict[str, str]]
+    isBase64Encoded: bool
+    body: str
+
+    @root_validator()
+    def check_message_id(cls, values):
```
heitorlessa, line 89: could you add a note why this check is needed? e.g. spoofing?

docs/utilities/parser.md
```diff
@@ -161,6 +161,7 @@ Parser comes with the following built-in models:
 | **KinesisDataStreamModel** | Lambda Event Source payload for Amazon Kinesis Data Streams        |
 | **SesModel**               | Lambda Event Source payload for Amazon Simple Email Service        |
 | **SnsModel**               | Lambda Event Source payload for Amazon Simple Notification Service |
+| **APIGatewayProxyEvent**   | Lambda Event Source payload for Amazon API Gateway                 |
```
heitorlessa, line 164: API Gateway v1, v2, or both? e.g. REST vs HTTP API. By my first look I'd guess REST API

docs/utilities/parser.md
```diff
@@ -294,16 +295,16 @@ Here's an example of parsing a model found in an event coming from EventBridge,
...
+| **SnsEnvelope**                                                                                                                              | 1. Parses data using `SnsModel`. <br/> 2. Parses records in `body` key using your model and return them in a list.                                                                                          | `List[Model]`                      |
+| **SnsSqsEnvelope**                                                                                                                           | 1. Parses data using `SqsModel`. <br/> 2. Parses SNS records in `body` key using `SnsNotificationModel`. <br/> 3. Parses data in `Message` key using your model and return them in a list.                  | `List[Model]`                      |
+| **ApiGatewayEnvelope**         1. Parses data using `APIGatewayProxyEventModel`. <br/> 2. Parses `body` key using your model and returns it. | `Model`                                                                                                                                                                                                     |
```
heitorlessa, line 307: API Gateway v1, v2, or both? e.g. REST vs HTTP API. By my first look I'd guess REST API

heitorlessa (changes requested):
Awesome, thank you @risenberg-cyberark. A few minor questions on the additional checks, and an explicit comment to tag whether this is REST API (v1) or HTTP API (v2), or both, so it's easier to add and use HTTP API later judging by the import, naming, etc.

aws_lambda_powertools/utilities/parser/models/apigw.py
risenberg-cyberark: what would you like to write here? basically what i saw that in the v1 version there's a duplication 3 fields, so i made sure that all values are actually the same.

aws_lambda_powertools/utilities/parser/models/apigw.py
risenberg-cyberark: what would you like to write here? basically what i saw that in the v1 version there's a duplication 3 fields, so i made sure that all values are actually the same. V2 removes this duplication

aws_lambda_powertools/utilities/parser/models/apigw.py
risenberg-cyberark: what would you like to write here? basically what i saw that in the v1 version there's a duplication 3 fields, so i made sure that all values are actually the same. V2 removes this duplication

docs/utilities/parser.md
risenberg-cyberark: V1

heitorlessa:
Thanks for the clarifications @risenberg-cyberark - I'm happy to merge after we remove those root validators.

I don't see a need a benefit in keeping them if we, the provider, are the ones duplicating these possibly for backward compatible reasons (until HTTP API when they're gone).

Alternatively, I can merge and fiz that later too - up to you.

Thanks a lot for adding this support!

risenberg-cyberark:
> Thanks for the clarifications @risenberg-cyberark - I'm happy to merge after we remove those root validators.
> 
> I don't see a need a benefit in keeping them if we, the provider, are the ones duplicating these possibly for backward compatible reasons (until HTTP API when they're gone).
> 
> Alternatively, I can merge and fiz that later too - up to you.
> 
> Thanks a lot for adding this support!

I'll explain. They are duplicated in the payload. I didnt duplicate them, i just made sure that the values are the same since they clearly are supposed to be the same. Those 2 items reside on the main event and are also found in the request context part. 

You can keep them or remove them but I do think they add a value becuase if there's a discrepancy in between them, you should not use the entire payload at all ,

heitorlessa:
> > Thanks for the clarifications @risenberg-cyberark - I'm happy to merge after we remove those root validators.
> 
> > 
> 
> > I don't see a need a benefit in keeping them if we, the provider, are the ones duplicating these possibly for backward compatible reasons (until HTTP API when they're gone).
> 
> > 
> 
> > Alternatively, I can merge and fiz that later too - up to you.
> 
> > 
> 
> > Thanks a lot for adding this support!
> 
> 
> 
> @heitorlessa 
> 
> I'll explain. They are duplicated in the payload. I didnt duplicate them, i just made sure that the values are the same since they clearly are supposed to be the same. Those 2 items reside on the main event and are also found in the request context part. 
> 
> 
> 
> You remove them but I do think they add a value because if there's a discrepancy in between them, you should not use the entire payload at all ,

That's what I understood too @risenberg-cyberark ;)

Putting in another way, I wouldn't like to break customers applications because this behaviour isn't documented. For this reason, I don't trust that API Gateway Lambda integration to follow through these checks consistently.

risenberg-cyberark:
> > > Thanks for the clarifications @risenberg-cyberark - I'm happy to merge after we remove those root validators.
> > 
> > 
> > > 
> > 
> > 
> > > I don't see a need a benefit in keeping them if we, the provider, are the ones duplicating these possibly for backward compatible reasons (until HTTP API when they're gone).
> > 
> > 
> > > 
> > 
> > 
> > > Alternatively, I can merge and fiz that later too - up to you.
> > 
> > 
> > > 
> > 
> > 
> > > Thanks a lot for adding this support!
> > 
> > 
> > @heitorlessa
> > I'll explain. They are duplicated in the payload. I didnt duplicate them, i just made sure that the values are the same since they clearly are supposed to be the same. Those 2 items reside on the main event and are also found in the request context part.
> > You remove them but I do think they add a value because if there's a discrepancy in between them, you should not use the entire payload at all ,
> 
> That's what I understood too @risenberg-cyberark ;)
> 
> Putting in another way, I wouldn't like to break customers applications because this behaviour isn't documented. For this reason, I don't trust that API Gateway Lambda integration to follow through these checks consistently.

np, removed the relevant root validator. is the last root validator ok?

aws_lambda_powertools/utilities/parser/models/apigw.py
```diff
@@ -0,0 +1,93 @@
...
+    resourcePath: str
+    domainName: Optional[str]
+    domainPrefix: Optional[str]
+    extendedRequestId: Optional[str]
+    httpMethod: Literal["DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT"]
+    path: str
+    connectedAt: Optional[datetime]
+    connectionId: Optional[str]
+    eventType: Optional[Literal["CONNECT", "MESSAGE", "DISCONNECT"]]
+    eventType: Optional[str]
```
heitorlessa, line 66: Is this perhaps a typo as it's defined twice?

That's for WebSockets and the line 65 is correct

heitorlessa (changes requested):
All good on the validators except I think there's an accidental duplicates field (eventType).

Mind checking before I merge?

aws_lambda_powertools/utilities/parser/models/apigw.py
risenberg-cyberark: seems like a bad copy paste :)
removed

heitorlessa:
I bet :D Glad I checked w/ you before merging :) Merging now!

heitorlessa (approved):

Issue #387, opened 2021-04-06
Appending to entries via extra to structure_logs(append=True) entries


bml1g12:
**Is your feature request related to a problem? Please describe.**
After having set an attribute globally:

e.g.
```
LOGGER.structure_logs(append=True,
                        context={"device_name": device_name})
```

I would like to log additional items to the context attribute e.g. 
```
LOGGER.info("hello world", 
            extra={"context": {"device_location": "UK"}})
```

**Describe the solution you'd like**

LOGGER.info("hello world")  would include a JSON payload with both device_location and device_name in the above example whereas currently it overwrites the context attribute with the new entry,  

**Describe alternatives you've considered**

I could do:
```
LOGGER.info("hello world", 
            extra={"context": {"device_name": device_name, 
                                        "device_location": "UK"}})
```
But if the device_name is not globally available throughout the code base, it is not ideal.

heitorlessa:
Hey @bml1g12, have you considered adding these context keys as root keys instead?

For example: 

> new method that will be available in the next release

LOGGER.append_keys(device_name="mydevice")

LOGGER.append_keys(device_location="UK")

{
    "message": "hello",
    "device_name": "mydevice",
    "device_location": "",
    ...
}

These could happen anywhere in the codebase and whenever you'd run "LOGGER.info()", they'd appear - no need for extra.

That said, I understand there are scenarios where having a single key like context with nested content makes more sense.

Extras might not be explicit enough to whoever is reading your code to understand you wanted to append additional value in context key instead of overwriting (default behaviour in standard Logging, structlog etc) within that log statement.

For a method however it can be done, so here are a few questions on behaviour to help flesh this out:

- Would you want to append in non-dictionaries too like lists and tuples?
- Are there scenarios where you'd want to update or remove a given nested key? E.g context["device_location"]

Re-reading this I'm on the fence between 1/ an override flag in the new append_keys, or 2/ expose the structures log so you can mutate as you please as an escape hatch to balance maintenance and unknown requirements

Option 2 would help with both questions and more that I asked, and it would look like more or less like this:

> Dict

LOGGER.log_structure["context"].update(...)

> List/SequenceMapping
LOGGER.log_structure["context"]["devices"].append("something in my list")

Let me know :)

Issue #381, opened 2021-04-04
Timezone aware timestamp logging


bml1g12:
**Describe the solution you'd like**

I would like to be able to easily log in a timezone aware manner; a typical use case being to always log in UTC time, regardless of where the code is run. At present, the Powertools Logger() uses localtime for the timestamp field as under the hood it uses the `asctime()` method.

**Describe alternatives you've considered**
I raised this issue as a documentation query at first https://github.com/awslabs/aws-lambda-powertools-python/issues/379 and @heitorlessa kindly helped me find a workaround to this there using monkeypatching or overwritting the formatter.

It can be a new feature for Logger to have a flag to convert to UTC or a given offset, which I agree would be a useful feature as I imagine many users would desire the ability to ensure the timestamp is UTC. For example, this is actually a requirement if one wished to log strictly according to the  ISO 8601 standard.

heitorlessa:
Thanks again for creating the feature request @bml1g12 - In the meantime, I've updated the docs with that fix until we decide on the best way to implement this for everyone ;)

https://awslabs.github.io/aws-lambda-powertools-python/develop/core/logger/#setting-timestamp-to-utc

saragerion:
@bml1g12 I completely agree with your proposal. Local time is logged based on the context of the current machine and that might be different from the developer local time. 
It's important that all logs generate use the same common "timezone".

The TypeScript Powertool is supporting ISO 8601 by default, and the timezone is always zero UTC offset. Developers can choose to customize the format if they want.
What do you think?

heitorlessa:
This is now out in 1.15.0: https://github.com/awslabs/aws-lambda-powertools-python/discussions/423.

```logger = Logger(service="example", utc=True)``` ;)

Issue #215, opened 2020-11-22
'sampling_rate' and 'xray_trace_id' cannot be suppressed


papi-tokei:
Thanks for the useful tools.
I use aws-lambda-powertools-python in Lambda python 3.8.

for document

> Logger allows you to either change the format or suppress the following keys altogether at the initialization: location, timestamp, level, xray_trace_id, and datefmt

but I can't suppress the 'sampling_rate' and 'xray_trace_id'
(Is it a specification that 'sampling_rate' cannot be suppressed?)

I want to suppress the 'sampling_rate' and 'xray_trace_id'.

## Expected Behavior

```
logger = Logger(
    service="payment",
    level="INFO",
    sampling_rate=None,
    xray_trace_id=None,
    timestamp=None,
    location=None,
)
```

Expected output is below.

```
{"level": "INFO", "message": {"test": 123}, "service": "payment"}
```

## Current Behavior

I write below code.

```
logger = Logger(
    service="payment",
    level="INFO",
    sampling_rate=None,
    xray_trace_id=None,
    timestamp=None,
    location=None,
)
```

The actual output is below.
(xray_trace_id is masked)

```
{"level": "INFO", "message": {"test": 123}, "service": "payment", "sampling_rate": 0.0, "xray_trace_id": "1-xxxxxxx"}
```

## Steps to Reproduce (for bugs)

1. I write code for Lambda function.
2. I invoke Lambda function.
3. Display the output.(but not the output I expected)

## Environment

* **Powertools version used**: 1.8.0
* **Packaging format (Layers, PyPi)**: Layers
* **AWS Lambda function runtime:** Python 3.8
* **Debugging logs**

heitorlessa:
Hey Hiroya, thanks for opening up an issue.

Sampling rate is part of the spec and as you noticed cannot be suppressed.

For X-Ray, it is automatically suppressed if X-Ray isn't enabled for the function. When X-Ray is enabled, this log entry allows CloudWatch Service Lens to combine Logs, Metrics and Traces in one page.

Aside for sampling rate, may I ask why you'd like to suppress X-Ray Trace ID field despite enabling it in the function?

I believe the docs can be clearer on that, so I'll amend it later :)

Thanks

papi-tokei:
Hello Heitor.
Thanks for replying in detail.

> Sampling rate is part of the spec and as you noticed cannot be suppressed.
[...]

Issue #98, opened 2020-08-06
Logger: overwrite keys while logging new stuff


alvaropc:
Hi!

I am here to ask you about the Keys of the logger.

I see that you can update the keys with *structure_logs()* method, but once you put a new key and assign a value, you won't be able to change it without calling *structure_logs()* again. 

It would be interesting to be able to overwrite some key values at the same time you  are logging, think about some KPI you would like to log as a key for instance, it would be nice to easily overwrite the value.

Although I know it is posible to write keys under "message" key it feels like not the best option for me, the main keys seem to be the place for this kind of information in order to catch it easily with tools like datadog or something like that.

I don't know if I am explaining it correctly but I hope you consider this as a new feature, I would also love to know your opinion about it as well as possible alternative aproaches, maybe I am just missunderstanding the way how you want the logger to be used and wirting more keys under messages is the correct way to do it!

Thank you very much in advance!
And congrats for such an amazing tool!

heitorlessa:
Buenas @alvaropc, and thanks for opening an issue.

 I understood you perfectly :) Allow me to break the explanation down to each point you raised:

* **structure_logs behaviour**: This is by design, as customers wanted to either override everything previously created, or append additional keys
* **add/override top keys while you log**: This is the first time we're hearing this, and I'd like to hear more use cases from other customers before considering it. We support adding any object as part of the message being logged to allow for additional keys, but not top keys as that is a different behaviour of structured logging.
* **difficult to catch keys under `message`**: As we're logging as JSON, [all these tools you mentioned should be able to easily access anything inside `message` ](https://docs.datadoghq.com/logs/explorer/analytics/?tab=timeseries). 
    - We recommend customers to either add keys they know will be constant across their invocations via `structure_logs`, or to agree on a contract when logging to ensure it is well understood and always available during visualization
    - For example: `log.info( { "operations": "collect_payment": "details": { "some_key": "some_value"...} })`
    - Using Datadog Log Explorer or any Log Analytics tool, you can access it via `@message.operations, @message.details.some_key`

Hope that helps :) 

We'll reopen if we hear other customers wanting this behaviour to be different while maintaining backward compatibility.

Muchas gracias Alvaro

PR #404, opened 2021-04-19
refactor(logger): BYOFormatter and Handler, UTC support, and more

heitorlessa:
**Issue #, if available:** #381 #387 #215 #98

## Description of changes:

Major refactor of Logger to **a)** make it more flexible for those coming from other loggers, **b)** bring your own formatter for those with a common structure that differs from our opinionated one, **c)** UX enhancements to add or remove keys from the logger, **d)** address long standing tech debts on  `LambdaPowertoolsFormatter`.

### UTC support

> Note: By default, Python standard logging use local time. 

```python

logger = Logger(utc=True)

logger.info("Hello")  # "timestamp":"2021-04-19 12:58:46.635+0100", dst
```

### New convenience property to change logger handler, if necessary

```python

logger = Logger()

handler = logger.registered_handler  # takes into account parent or child loggers
```

### Bring your own handler

> Note: By default, we log to standard output due to Lambda integration. This unlocks using Logger in places like Glue, or locally when you want to log into a filename for example.

```python

handler = logging.FileHandler(filename="log_file.json")
logger = Logger(logger_handler=handler)

logger.info("Hello")  # populatest `log_file.json` in your filesystem
```

### Bring your own formatter

```python
from aws_lambda_powertools.logging.formatter import BasePowertoolsFormatter

class CustomFormatter(BasePowertoolsFormatter):
    custom_format = {}

    def append_keys(self, **additional_keys):
        # used by inject_lambda_context decorator, Logger initialization, and structure_log
        self.custom_format.update(additional_keys)

    def remove_keys(self, keys: Iterable[str]):
        # used by Logger public method
        for key in keys:
            self.custom_format.pop(key, None)

    def format(self, record: logging.LogRecord) -> str:  # noqa: A003
        return json.dumps(
            {
                "message": super().format(record),
                "timestamp": self.formatTime(record),
                "my_default_key": "test",
                **self.custom_format,
            }
        )

custom_formatter = CustomFormatter()
logger = Logger(logger_formatter=custom_formatter)
```

### Updating formatter made easier

```python

logger = Logger()

logger.append_keys(request_id="id", order_id="id")
logger.info("with both request and order id keys")

logger.remove_keys(["request_id"])
logger.info("with request id key only")
```

**Checklist**

 

* [x] Update tests
* [ ] Update docs
* [x] **New features**
  * [x] Bring your own formatter via `Logger(log_formatter=...)`
    * [x] Create ABC to support custom formatters to enforce `append_keys` and `remove_keys`
    * [x] Update Logger to use `append_keys` for Lambda decorator and `structure_log` calls
  * [x] Add support for `utc=True`
  	* [x] Include timezone in `timestamp` key
  	* [x] Create custom msec string interpolation to not break previous format (RFC 3339), since std logging doesn't support msec when built-in `formatTime` is overriden
* [x] **Tech debt**
  * [x] Introduce `append_keys`, `remove_keys` over `structure_log(append=True)` to align with other runtimes
  * [x] Allow customers to set bare minimum (only message, and xray_trace_id)
    * [x] Allow to suppress sampling_rate
    * [x] Remove sampling_rate when not in use (set to None by default over 0.0)
  * [x] Allow custom JSON serializer/deserializer for those seeking faster performance
  * [x] Compact logging to save on ingestion & storage cost (`separators[]` to remove whitespace)
  * [x] [Internal] Refactor standalone methods into staticmethods
  * [x] Rename testing to powertools formatter given we have vastly deviated from the original implementation
    * [x] Create alias for JsonFormatter to prevent breaking existing customers
  * [x] Add convenience properties to access handler and formatter
  * [x] Support all standard log attrs (%()s) if given via kwargs
* [x] Support custom handler
* [x] Update docstring parameters in Logger
* [x] Update docstring parameters in Formatter

---

**Create separate PR for docs to include enhancements**:

* [ ] Document examples
  * [ ] Default JSON fn for unserializable (json.dumps(data, default=json_default))
  * [ ] Std log attrs %()s support
  * [ ] Bring your own formatter
  * [ ] Bare minimal attrs (level, message, service)
  * [ ] Update all docs with additional TZ in timestamp
  * [ ] Bringing `orjson` for those seeking to gain every possible microsecond
* [ ] Update `Testing your code` section to include how to use `stream=sys.stdout`

## Breaking change checklist

**RFC issue #**:


michaelbrewer:
@heitorlessa nice, this would help alot when you have company mandated structured logging formats.

heitorlessa:
@bml1g12 - what are your thoughts on the UTC UX for your use case?

Anything else you need while I'm at it?

bml1g12:
Currently, I am declaring in each handler 
```
LOGGER = Logger(datefmt="%Y-%m-%dT%H:%M:%SZ")
LOGGER._logger.handlers[0].formatter.converter = time.gmtime  #pylint: disable = protected-access
```
So in this new UX I would instead simply do
```
LOGGER = Logger(datefmt="%Y-%m-%dT%H:%M:%SZ", utc=True)
```
In each handler and the child would inherit this right?

If so, looks great to me! Much appreciated addition.

heitorlessa:
Yes exactly. I noticed you’re using RFC 8601 in your datefmt, I’ll make
sure this will be the default timestamp in v2, as changing now would break
customers.

bml1g12:
Yeah I chose this as I saw a few blogs recommending it as a standard across languages for structured logging; I think its the default for javascript logging library maybe

michaelbrewer:
@heitorlessa I am liking these changes so long, I am glad to see that now even `sampling_rate` can be suppressed :)

aws_lambda_powertools/logging/formatter.py
```diff
@@ -27,63 +31,140 @@
...
+        # so we create a custom one (%F) and replace logging record ts
+        # Reason 2 is that std logging doesn't support msec after TZ
+        record_ts = self.converter(record.created)
+        if datefmt:
+            ts_formatted = time.strftime(datefmt, record_ts)
+        else:
+            msec = "%03d" % record.msecs
+            custom_fmt = self.default_time_format.replace(self.custom_ms_time_directive, msec)
+            ts_formatted = time.strftime(custom_fmt, record_ts)
+        return ts_formatted
```
michaelbrewer, line 144:
```suggestion
        record_ts = self.converter(record.created)
        if datefmt:
            return time.strftime(datefmt, record_ts)

        # NOTE: Python `time.strftime` doesn't provide msec directives
        # so we create a custom one (%F) and replace logging record ts
        # Reason 2 is that std logging doesn't support msec after TZ
        msecs = "%03d" % record.msecs
        custom_fmt = self.default_time_format.replace(self.custom_ms_time_directive, msecs)
        return time.strftime(custom_fmt, record_ts)
```

michaelbrewer (changes requested):
LGTM, just a minor tweak and typo

Issue #406, opened 2021-04-21
Metrics not appearing at 1 minute resolution as expected


bml1g12:
I am publishing metrics like this

```
"""Logging metedata utils"""
import os
import time

from aws_lambda_powertools import Logger, Metrics
from aws_lambda_powertools.metrics import MetricUnit

METRICS = Metrics()
METRICS.add_dimension(name="environment", value=os.environ["NODE_ENV"])

LOGGER = Logger(child=True)

def log_processing_statistics(start_processing_timestamp, annotation_metadata):
    """
    Logs some metrics on how long it took to process or receive items.

    :param start_processing_timestamp: when we started processing an annotation
    :param dict annotation_metadata: keys include end_timestamp and queue_arrival_timestamp
    """
    now = time.time()
    queue_arrival_timestamp = annotation_metadata["queue_arrival_timestamp"]
    queue_to_processed_time = now - queue_arrival_timestamp
    LOGGER.info("[META]:queue_to_processed_time: %.2f", queue_to_processed_time)
    METRICS.add_metric(name="entrance_counter_queue_to_processed_time",
                       unit=MetricUnit.Seconds,
                       value=queue_to_processed_time)
```

Where my lambda handler top level function has been decorated like this:
```
METRICS = Metrics()
METRICS.add_dimension(name="environment", value=os.environ["NODE_ENV"])
LOGGER = Logger(datefmt="%Y-%m-%dT%H:%M:%SZ")
LOGGER._logger.handlers[0].formatter.converter = time.gmtime  #pylint: disable = protected-access
log.initialize_structured_logging()
TRACER = Tracer()  # Sets service via env var

@METRICS.log_metrics(capture_cold_start_metric=
                     os.environ.get("POWERTOOLS_METRICS_CAPTURE_COLD_START",
                                    "false").lower() == "true")
@TRACER.capture_lambda_handler(capture_error=True)
```
I expect to see cloudwatch metrics here being logged every few minutes; but instead, I see them every 40 minutes or so; or at least at erratic times with a much larger interval than they are being logged. 

e.g. 
I have checked my cloudwatch logs and found the following two entries suggesting that on some level things are working as expected:

```
{
    "_aws": {
        "Timestamp": 1618965743152,
        "CloudWatchMetrics": [
            {
                "Namespace": "ThriEntranceCounterLambdaGenesis",
                "Dimensions": [
                    [
                        "service"
                    ]
                ],
                "Metrics": [
                    {
[...]

PR #410, opened 2021-04-22
feat: add support to persist default dimensions

heitorlessa:
**Issue #, if available:** #406

## Description of changes:

**Checklist**

 

* [x] Update tests
* [x] Update docs
* [x] Add new `set_default_dimensions` method
* [x] Add new `clear_default_dimensions` method
* [x] Docs: Update testing section to inspect Metrics created
* [x] Docs: Update testing section to reset Default Dimensions
* [x] Docs: Banner about adding metrics outside the handler
* [x] Docs: Banner about sparse cold start metrics for cost
* [x] Docs: Add Functional testing section

## UX

Persists two dimensions across all metrics, every Metrics instance throughout the codebase, and across all Lambda invocations.

```python

metrics = Metrics(namespace="ExampleApplication", service="booking")
metrics.set_default_dimensions(environment="prod", another="one")

@metrics.log_metrics
def lambda_handler(evt, ctx):
```

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/metrics/metrics.py
+++ b/aws_lambda_powertools/metrics/metrics.py
@@ -128,6 +128,7 @@ class Metrics(MetricManager):
         lambda_handler: Callable[[Any, Any], Any] = None,
         capture_cold_start_metric: bool = False,
         raise_on_empty_metrics: bool = False,
+        default_dimensions: Dict[str, str] = None,
     ):
         """Decorator to serialize and publish metrics at the end of a function execution.
 
@@ -150,11 +151,13 @@ class Metrics(MetricManager):
         Parameters
         ----------
         lambda_handler : Callable[[Any, Any], Any], optional
-            Lambda function handler, by default None
+            lambda function handler, by default None
         capture_cold_start_metric : bool, optional
-            Captures cold start metric, by default False
+            captures cold start metric, by default False
         raise_on_empty_metrics : bool, optional
-            Raise exception if no metrics are emitted, by default False
+            raise exception if no metrics are emitted, by default False
+        default_dimensions: Dict[str, str], optional
+            metric dimensions as key=value that will always be present
 
         Raises
         ------
@@ -170,11 +173,14 @@ class Metrics(MetricManager):
                 self.log_metrics,
                 capture_cold_start_metric=capture_cold_start_metric,
                 raise_on_empty_metrics=raise_on_empty_metrics,
+                default_dimensions=default_dimensions,
             )
 
         @functools.wraps(lambda_handler)
         def decorate(event, context):
             try:
+                if default_dimensions:
+                    self.set_default_dimensions(**default_dimensions)
                 response = lambda_handler(event, context)
                 if capture_cold_start_metric:
                     self.__add_cold_start_metric(context=context)
--- a/aws_lambda_powertools/metrics/metrics.py
+++ b/aws_lambda_powertools/metrics/metrics.py
@@ -108,9 +108,10 @@ class Metrics(MetricManager):
             def lambda_handler():
                    return True
         """
-        self.default_dimensions.update(**dimensions)
         for name, value in dimensions.items():
             self.add_dimension(name, value)
+
+        self.default_dimensions.update(**dimensions)
 
     def clear_default_dimensions(self):
         self.default_dimensions.clear()
@@ -120,8 +121,7 @@ class Metrics(MetricManager):
         self.metric_set.clear()
         self.dimension_set.clear()
         self.metadata_set.clear()
-        # re-add default dimensions
-        self.dimension_set.update(**self._default_dimensions)
+        self.set_default_dimensions(**self.default_dimensions)  # re-add default dimensions
 
     def log_metrics(
         self,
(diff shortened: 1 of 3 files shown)


docs/core/metrics.md
```diff
@@ -345,5 +375,41 @@ If you prefer setting environment variable for specific tests, and are using Pyt
         metrics = Metrics()
         metrics.clear_metrics()
         metrics_global.is_cold_start = True  # ensure each test has cold start
+		metrics.clear_default_dimensions()   # remove persisted default dimensions, if any
```
pcolazurdo, line 378: remove extra indentation

docs/core/metrics.md
```diff
@@ -98,6 +104,28 @@ You can create metrics using `add_metric`, and you can create dimensions for all
 !!! note "Metrics overflow"
 	CloudWatch EMF supports a max of 100 metrics per batch. Metrics utility will flush all metrics when adding the 100th metric. Subsequent metrics, e.g. 101th, will be aggregated into a new EMF object, for your convenience.
 
+
+### Adding default dimensions
+
+You can add default metric dimensions to ensure they are persisted across Lambda invocations using `set_default_dimenions`.
+
+!!! info "If you'd like to remove them at some point, you can use `clear_default_dimensions` method"
+	Note that they continue to count against the maximum of 9 dimensions.
```
pcolazurdo, line 113: This sounds like even if you clear them, they still count. Is that the case?

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -90,11 +87,41 @@ def __init__(self, service: str = None, namespace: str = None):
...
+    def clear_default_dimensions(self):
+        self.default_dimensions.clear()
+
     def clear_metrics(self):
         logger.debug("Clearing out existing metric set from memory")
         self.metric_set.clear()
         self.dimension_set.clear()
         self.metadata_set.clear()
+        # re-add default dimensions
+        self.dimension_set.update(**self._default_dimensions)
```
pcolazurdo, line 124: It looks like _default_dimensions is not updated (in 79 you are updating self.default_dimensions, not the other way around), so it looks like it may be empty at this point? - Also, in L111-L113 you are calling the self.add_dimensions() inside the set_default_dimensions() but once dimension_set is cleared if clear_metrics() is called you are not calling add_dimensions again, so it would look like this will be empty on a second invocation. Can't test at the moment, but that what it looks like from a quick overview of the code.

pcolazurdo (commented):
Only a few documentation comments and a question around the behavior of clear_metrics()

pcolazurdo:
I'd like this approach for some use cases but I think it may add a "burden" on the dev UX to remember where the defaults where set. I'd prefer an approach where the decorator has something like: 
```
@metrics.log_metrics(default_dimensions=**dimensions)
```
Which keeps the defaults very close to the function. I wouldn't expect this defaults to be shared across other decorators in my opinion - i.e. if the code is reusing a library that has it's own defaults

docs/core/metrics.md
heitorlessa: If you clear them, nope. The limit counts when we serialize and dump the JSON to stdout. If they are cleared, there will be no default dimensions in the EMF blob

heitorlessa:
> I'd like this approach for some use cases but I think it may add a "burden" on the dev UX to remember where the defaults where set. I'd prefer an approach where the decorator has something like:
> 
> ```
> @metrics.log_metrics(default_dimensions=**dimensions)
> ```
> 
> Which keeps the defaults very close to the function. I wouldn't expect this defaults to be shared across other decorators in my opinion - i.e. if the code is reusing a library that has it's own defaults

Yeah, I'd have loved to go down that route too, however `param=` doesn't accept unpacking a dictionary, it'd have to be `@metrics.log_metrics(**dimensions)` in which case makes it less explicit to whoever reading the code what this does. 

I also experimented with:

```python
dimensions = [{"name": "dimension_1"}, "value": "value"}, {"name": "dimension_2"}, "value": "value"}]
@metrics.log_metrics(default_dimensions=dimensions)
```

But this was more explicit and can work outside Lambda handler like in Lambda layers code, other modules, etc.

```python
metrics = Metrics()

metrics.set_default_dimensions(environment="prod", dimension_one="value", ...)
```

aws_lambda_powertools/metrics/metrics.py
heitorlessa: Yeah, it's a bug as dimensions aren't added in the right format - Thanks for catching it!

Everything else isn't empty because we're "abusing" class instance vars similarly to what the Borg pattern does.

L72 `_default_dimensions: Dict[str, Any] = {}`, here we initialize it as an empty dict and as class instance so if we were to initialize a new instance next time it'll keep the state.

pcolazurdo:
Wouldn't something like this below help? 

```
@my_metrics.log_metrics(default_dimensions={"environment": "dev"})
def lambda_handler(evt, context):
  return True
```
I have a sample version here: https://github.com/heitorlessa/aws-lambda-powertools-python/compare/feat/metrics-default-dimensions...pcolazurdo:pr_410?expand=1

heitorlessa:
This does @pcolazurdo - I was initially thinking the name=, value=, which didn't fit.

I'll include that in today now that the implementation is already done

pcolazurdo (approved):
Looks great

PR #419, opened 2021-05-04
fix(parser): Improve types for parser.py

carlos-alberto:
When attempting to use the parse function we noticed we were getting
error messages like:

    Value of type variable "Model" of "parse" cannot be "Type[UserModel]"

Further investigation (using the example shown at
https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/parser/#envelopes)
showed that the types were incorrect for the example. The function
signatures were expecting an instance of `Model`, however the example
(and code) seemeed to expect a `Type[Model]`.

Similarly the signatures were expecting an optional instance of an
Envelope, but the code allows for either an instance or a Type.

Finally, added a `TypeVar` for the decorator, specifying that the
event_parser` decorator returns the same type as the handler used.

## Description of changes:

Fix type signatures of parser.py

**Checklist**

 

* [ ] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/parser/parser.py
+++ b/aws_lambda_powertools/utilities/parser/parser.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Any, Callable, Dict, Optional
+from typing import Any, Callable, Dict, Optional, Type, TypeVar, Union
 
 from ...middleware_factory import lambda_handler_decorator
 from ..typing import LambdaContext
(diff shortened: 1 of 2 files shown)


heitorlessa:
This is awesome @carlos-alberto!

Thanks a lot for taking time to make your first contribution.

@pcolazurdo in case you have some cycles to review.

I'm working on the new event handler docs and can look into this next. Either way, this can make into this week's release

pcolazurdo:
As @heitorlessa said, thank you for your contribution @carlos-alberto - The code changes seems ok to me, thanks for catching these, including the mistake in the documentation. 
Reflecting on this, it looks like we should add some mypy tests coverage too, to capture these in the future. Thoughts?

heitorlessa:
> Reflecting on this, it looks like we should add some mypy tests coverage too, to capture these in the future. Thoughts?

@pcolazurdo 100%. It's something we wanted to do, including fixing all other MyPy issues across the code base. I hadn't done that for largely 2 reasons: 1/ Bandwidth, 2/ MyPy has some false positives and need to wrestle with config and learn how to "tame" it before adding in the CI, since that would block some contributions for the unfamiliar with it.

Merging this now, thanks a lot to both of you!

PR #426, opened 2021-05-07
fix(logger): Support .info("foo %s", "bar") formatting

huonw:
**Issue #, if available:** N/A

## Description of changes:

This adjusts the default logging formatter (`LambdaPowertoolsFormatter`)
to handle non-structured logs with formatting, like:

```python
logger.info("foo %s", "bar")
```

This allows the formatter to be used in more situations, such as the
universal formatter for all log messages coming out of a system, even
libraries using the normal logging functionality, without explicit
support for aws-lambda-powertools.

If formatting is performed, this is taken as a hint that the value
probably isn't semantically JSON, even if it's syntactically valid as
JSON.

(I've called this a bug fix in the title but could be a 'feat' too, I guess?)

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**: N/A

### DIFF
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -176,6 +176,9 @@ class LambdaPowertoolsFormatter(BasePowertoolsFormatter):
         if isinstance(message, dict):
             return message
 
+        if log_record.args:  # logger.info("foo %s", "bar") requires formatting
+            return log_record.getMessage()
+
         if isinstance(message, str):  # could be a JSON string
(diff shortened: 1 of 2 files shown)


heitorlessa (approved):

heitorlessa:
Hey @huonw thank you so much for this ;)

huonw:
Thanks for merging so quickly! (And thanks for your efforts on this library, it makes our app noticeably smoother.)

Issue #436, opened 2021-05-23
Security issue in Pydanic


risenberg-cyberark:
Upgrade Pydantic to 1.8.2 to resolve the security issue. 
See
https://pydantic-docs.helpmanual.io/changelog/
and 
https://github.com/samuelcolvin/pydantic/security/advisories/GHSA-5jqp-qgf6-3pvh

heitorlessa:
Released as part of 1.16.1

PR #437, opened 2021-05-23
feat(parser): security issue in Pydantic #436

risenberg-cyberark:
upgrade pydantic to 1.8.2
see issue at #436


heitorlessa (approved):
Thank you for that @risenberg-cyberark - As it's a CVE, I'll make a patch release now.

Issue #434, opened 2021-05-18
Add APIGateway*EventModelV2 to parser


xpayn:
I think that being able to use the parser module with APIGatewayV2 would be a nice little ergonomic improvement.
today my code looks like:
```Python
def lambda_handler(event: Dict[str, Any], context: LambdaContext) -> Dict[str, Any]:
    event: APIGatewayProxyEventV2 = APIGatewayProxyEventV2(event)
    ...
```

It's not a big diffence but I think it looks better like this
```Python
@event_parser(model=APIGatewayProxyEventModelV2)
def lambda_handler(event: APIGatewayProxyEventModelV2, context: LambdaContext) -> Dict[str, Any]:
    ...
```

If you accept PR and I have some spare time, I'll try to make something.
Thx

heitorlessa:
Hey @xpayn - Absolutely, I'd love to help with any guidance if you can make a PR.

heitorlessa:
> @heitorlessa note, we would have to ensure the any of the data classes are json serializable (ie: Idempotent handler expects a dict or something that can be serialized to json)

Sure, it'd need an additional check if it's a Pyndantic Model, then you can call `.json()` to do the serialization of all of their custom types and such: https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeljson

risenberg-cyberark:
@xpayn @heitorlessa  if you havent started it yet, i can do it this week.

risenberg-cyberark:
see https://github.com/awslabs/aws-lambda-powertools-python/pull/441

xpayn:
I was planning to make the PR today, I just wanted to review my code before, but i'll read yours instead

michaelbrewer:
> see #441

Seems like we are mixing things here. I think @xpayn  is referring to the Event Source Data Classes.

michaelbrewer:
Note, @xpayn  there is also https://awslabs.github.io/aws-lambda-powertools-python/latest/core/event_handler/api_gateway/

michaelbrewer:
@xpayn @risenberg-cyberark 

Below is what i mean as an equivalent implementation for event source data classes (all of the in fact), and you would also
benefit from the utility methods like  `get_header_value`

```python3
@lambda_handler_decorator
def event_source(
  handler: Callable[[Any, LambdaContext], Any], 
  event: Dict[str, Any], context: LambdaContext, 
  model
):
   # Where `model` is the passed in event source data class.
[...]

PR #442, opened 2021-05-26
feat(data-classes): decorator to instantiate data_classes and docs updates

michaelbrewer:
**Issue #, if available:**
#434
https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/19

## Description of changes:

Changes:
- feat: Create a convenient decorator to instantiate `data_classes` instances
- docs: Update docs to include `@event_source` example
- docs: Add missing docs for `ALBEvent` and format examples to work with PyCharm
- fix: Fix `_generate_hash` to handle the `data` being an instance of `DictWrapper`

```
TypeError: Object of type APIGatewayProxyEventV2 is not JSON serializable
```

### Example usage

```python3
from typing import Dict, Any

from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEventV2, event_source


@event_source(data_class=APIGatewayProxyEventV2)
def lambda_handler(event: APIGatewayProxyEventV2, context: LambdaContext) -> Dict[str, Any]:
    assert event.get_header_value("x-foo") == "Foo"
```

### And can be used with any of the existing data classes like:

```python3
from aws_lambda_powertools.utilities.data_classes.connect_contact_flow_event import (
    ConnectContactFlowChannel,
    ConnectContactFlowEndpointType,
    ConnectContactFlowEvent,
    ConnectContactFlowInitiationMethod,
)


@event_source(data_class=ConnectContactFlowEvent)
def lambda_handler(event: ConnectContactFlowEvent, context):
    assert event.contact_data.attributes == {"Language": "en-US"}
    assert event.contact_data.channel == ConnectContactFlowChannel.VOICE
    assert event.contact_data.customer_endpoint.endpoint_type == ConnectContactFlowEndpointType.TELEPHONE_NUMBER
    assert event.contact_data.initiation_method == ConnectContactFlowInitiationMethod.API
```

### And this still works with the existing idempotent handler

```python3
persistence_layer = DynamoDBPersistenceLayer(table_name="IdempotencyTable")

@event_source(data_class=APIGatewayProxyEventV2)
@idempotent(persistence_store=persistence_layer)
def lambda_handler(event: APIGatewayProxyEventV2, context):
    assert isinstance(event, APIGatewayProxyEventV2)
    ...
```

**Checklist**

 

* [x] Update tests
* [x] Update docs


michaelbrewer:
Should i update the docs then to include this?

heitorlessa:
Confirming here that I have blocked time to review and get this incorporated tomorrow.

Just been abnormally busy in the last few days with a customer project

tests/functional/idempotency/test_idempotency.py
```diff
@@ -845,3 +848,41 @@ def handler(event, context):
...
+        assert data_record.idempotency_key == self.expected_idempotency_key
+
+    def _update_record(self, data_record: DataRecord) -> None:
+        assert data_record.idempotency_key == self.expected_idempotency_key
+
+    def _get_record(self, idempotency_key) -> DataRecord:
+        ...
+
+    def _delete_record(self, data_record: DataRecord) -> None:
+        ...
```
heitorlessa, line 868: I'm on the fence about this one - I might remove it in the future if we end up not reusing it elsewhere in the tests.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -224,6 +224,8 @@ def _generate_hash(self, data: Any) -> str:
             Hashed representation of the provided data
 
         """
+        if hasattr(data, "raw_event"):
+            data = data.raw_event
```
heitorlessa, line 228: `hasattr` doesn't work in the way you might think it does - it's a misnomer in Python, really. It'll effectively do a blanket try/except, throw the results away, and specially in this case it'll access all getters from `data` -- the more getters the worse the performance.

Here's a safer alternative which also saves us a branching:

```suggestion
data = getattr(data, "raw_event", data)  # could be a data class depending on decorator order
```

heitorlessa (changes requested):
One minor change to make it safer and more performant for the future, all rest LGTM :)

I don't have permission to push to your fork so will push the next release to early next week to ensure we incorporate these in time

tests/functional/idempotency/test_idempotency.py
michaelbrewer: We could use it to verify if the parser works with idempotent decorator.

michaelbrewer:
@heitorlessa  does this look go to you now?

heitorlessa (approved):

heitorlessa:
Yes, thanks a lot Mike!

PR #462, opened 2021-06-06
feat(data-classes): add AttributeValueType to DynamoDBStreamEvent

michaelbrewer:
**Issue #, if available:**

https://github.com/awslabs/aws-lambda-powertools-python/issues/461

## Description of changes:

Changes:
- [x] Add new enum AttributeValueType for the type of AttributeValue
- [x] Add new property `get_type` to return said enum
- [x] Add new property `get_value` which returns converted value
- [x] Update docs with any revisited changes in the type conversion process

Example usage:

```python3
from aws_lambda_powertools.utilities.data_classes import event_source, DynamoDBStreamEvent
from aws_lambda_powertools.utilities.data_classes.dynamo_db_stream_event import AttributeValueType, AttributeValue


def lambda_handler(event: DynamoDBStreamEvent, context: LambdaContext):
    for record in event.records:
        key: AttributeValue = record.dynamodb.keys["id"]
        if key == AttributeValueType.Number:
            assert key.get_value == key.n_value
            print(key.get_value)
        elif key == AttributeValueType.Map:
            assert key.get_value == key.map_value
            print(key.get_value)
```

Where `AttributeValueType` is:

```python3
class AttributeValueType(Enum):
    Binary = "B"
    BinarySet = "BS"
    Boolean = "BOOL"
    List = "L"
    Map = "M"
    Number = "N"
    NumberSet = "NS"
    Null = "NULL"
    String = "S"
    StringSet = "SS"
```

## How AttributeValue does the raw values

Example Values:

* b_value:
```{"B": "dGhpcyB0ZXh0IGlzIGJhc2U2NC1lbmNvZGVk"}```
* bs_value
```{"BS": ["U3Vubnk=", "UmFpbnk=", "U25vd3k="]}```
* bool_value
```{"BOOL": True}```
* list_value/l_value
```{"L": [ {"S": "Cookies"} , {"S": "Coffee"}, {"N": "3.14159"}]}```
* map_value/m_value
```{"M": {"Name": {"S": "Joe"}, "Age": {"N": "35"}}}```
* n_value
```{"N": "123.45"}```
* ns_value
```{"NS": ["42.2", "-19", "7.5", "3.14"]}```
* null_value
```{"NULL": True}```
* s_value
```{"S": "Hello"}```
* ss_value
```{"SS": ["Giraffe", "Hippo" ,"Zebra"]}```

> NOTE: All return types are `Optional`

Return types

| Property           | Description                                                      | Raw Return Type             |
|--------------------|------------------------------------------------------------------|-----------------------------|
| b_value            | An attribute of type Base64-encoded binary data object           | str                         |
| bs_value           | An attribute of type Array of Base64-encoded binary data objects | List[str]                   |
| bool_value         | An attribute of type Boolean                                     | bool                        |
| list_value l_value | An attribute of type Array of AttributeValue objects             | List["AttributeValue"]      |
| map_value m_value  | An attribute of type String to AttributeValue object map         | Dict[str, "AttributeValue"] |
| n_value            | An attribute of type Number                                      | str                         |
| ns_value           | An attribute of type Number Set                                  | List[str]                   |
| null_value         | An attribute of type Null                                        | bool                  |
| s_value            | An attribute of type String                                      | str                         |
| ss_value           | An attribute of type Array of strings                            | List[str]                   |

## How Dynamodb Type does deserialization

| DynamoDB	|	Python |
|-----------|----------|
| {'NULL': True}	|	None |
| {'BOOL': True/False}	|	True/False |
| {'N': str(value)}	|	Decimal(str(value)) |
| {'S': string}	|	string |
| {'B': bytes}	|	Binary(bytes) |
| {'NS': [str(value)]}	|	set([Decimal(str(value))]) |
| {'SS': [string]}	|	set([string]) |
| {'BS': [bytes]}	|	set([bytes]) |
| {'L': list}	|	list |
| {'M': dict}	|	dict |

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
+++ b/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from typing import Dict, Iterator, List, Optional
+from typing import Any, Dict, Iterator, List, Optional, Union
 
 from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
 
@@ -23,10 +23,16 @@ class AttributeValue(DictWrapper):
     Documentation: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_AttributeValue.html
     """
 
-    @property
-    def get_type(self) -> AttributeValueType:
-        """Get the attribute value type based on the contained data"""
-        return AttributeValueType(list(self.raw_event.keys())[0])
+    def __init__(self, data: Dict[str, Any]):
+        """AttributeValue constructor
+
+        Parameters
+        ----------
+        data: Dict[str, Any]
+            Raw lambda event dict
+        """
+        super().__init__(data)
+        self.dynamodb_type = list(data.keys())[0]
 
     @property
     def b_value(self) -> Optional[str]:
@@ -123,6 +129,29 @@ class AttributeValue(DictWrapper):
             >>> {"SS": ["Giraffe", "Hippo" ,"Zebra"]}
         """
         return self.get("SS")
+
+    @property
+    def get_type(self) -> AttributeValueType:
+        """Get the attribute value type based on the contained data"""
+        return AttributeValueType(self.dynamodb_type)
+
+    @property
+    def value(self) -> Union[Optional[bool], Optional[str], Optional[List], Optional[Dict]]:
+        """Get the attribute value"""
+        try:
+            return getattr(self, f"{self.dynamodb_type.lower()}_value")
+        except AttributeError:
+            raise TypeError(f"Dynamodb type {self.dynamodb_type} is not supported")
+
+    @property
+    def l_value(self) -> Optional[List["AttributeValue"]]:
+        """Alias of list_value"""
+        return self.list_value
+
+    @property
+    def m_value(self) -> Optional[Dict[str, "AttributeValue"]]:
+        """Alias of map_value"""
+        return self.map_value
 
 
 def _attribute_value_dict(attr_values: Dict[str, dict], key: str) -> Optional[Dict[str, AttributeValue]]:
--- a/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
+++ b/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
@@ -4,11 +4,29 @@ from typing import Dict, Iterator, List, Optional
 from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
 
 
+class AttributeValueType(Enum):
+    Binary = "B"
+    BinarySet = "BS"
+    Boolean = "BOOL"
+    List = "L"
+    Map = "M"
+    Number = "N"
+    NumberSet = "NS"
+    Null = "NULL"
+    String = "S"
+    StringSet = "SS"
+
+
 class AttributeValue(DictWrapper):
     """Represents the data for an attribute
 
     Documentation: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_AttributeValue.html
     """
+
+    @property
+    def get_type(self) -> AttributeValueType:
+        """Get the attribute value type based on the contained data"""
+        return AttributeValueType(list(self.raw_event.keys())[0])
 
     @property
     def b_value(self) -> Optional[str]:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by michaelbrewer** [comment]
@gwlester what do you think?

**Comment by gwlester** [comment]
Looks great!

Thanks for the fast work!

**Review by gwlester — APPROVED** [review]

File: aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py [hunk 1]
```diff
@@ -4,12 +4,30 @@
...
+
+
 class AttributeValue(DictWrapper):
     """Represents the data for an attribute
 
     Documentation: https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_AttributeValue.html
     """
 
+    @property
+    def get_type(self) -> AttributeValueType:
```
jplock, line 27: What about a generic “get_value” method that can inspect the type and return the proper value?

michaelbrewer: @jplock -  But then i would need to type hint as a Union? Coz it is kind of nice having the type hinting

jplock: That’s true. I end up using the boto3.dynamodb.types TypeDeserializer directly anyway

gwlester: I happen to need the **get_type**, but a **get_value** is an obvious idea, should have also suggested that.

michaelbrewer: Thanks @jplock for pointing me to https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/types.html

I am thinking of a type hinted version of this.

michaelbrewer: @jplock @gwlester @heitorlessa 

I added "value" implementation, as it is not quite a deserialize. But maybe it should be and do the binary handling to? However that would be a breaking change.

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -1,15 +1,39 @@
 from enum import Enum
-from typing import Dict, Iterator, List, Optional
+from typing import Any, Dict, Iterator, List, Optional, Union
 
 from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
 
 
+class AttributeValueType(Enum):
```
heitorlessa, line 7: Could you update the DynamoDB Stream data class doc to include an example usage?

heitorlessa (changes requested):
I'll release 1.17.0 tomorrow and I could get this included too, if we could have an example usage in the docs. Thanks a lot!

gwlester:
Will this be in the next Release?

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
heitorlessa: I'm fine either way. 'get_value' would be more consistent and inline with the operation it does here than 'value'

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
michaelbrewer: Ok sure.

michaelbrewer: @heitorlessa CI process seems to be broken

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
michaelbrewer: @heitorlessa renamed

heitorlessa (approved):

heitorlessa:
Thanks @michaelbrewer for the changes! 

I'm gonna look into the CI failures

michaelbrewer:
> Thanks @michaelbrewer for the changes!
> 
> I'm gonna look into the CI failures

Failures seem to be intermittent getting the codecov token

heitorlessa:
Yeah I raised a ticket back then and seemed resolved for a while.

I also noticed concurrent doc builds fail due to git divergence when using
Mike

gwlester:
FYI, added it into the code (Dynamo Stream -> Lambda -> TimeStream -- for audit trail) and it works great!!!

Thanks again!

heitorlessa:
Fantastic to hear!

PR #504, opened 2021-07-02
fix(data-classes): include milliseconds in scalar types

### LINKED ISSUE #503: bug: aws scalar type for AWSDateTime should include milliseconds
**What were you trying to accomplish?**

## Expected Behavior

AWSDateTime - An extended ISO 8601 date and time string in the format YYYY-MM-DDThh:mm:ss.sssZ.

## Current Behavior

AWSDateTime is not including the millisecond part `sss`.

## Possible Solution

Generate timestamps to include the milliseconds

## Steps to Reproduce (for bugs)

```python3
> print(aws_datetime())
2021-07-02T17:09:47Z
````

## Environment

* **Powertools version used**: 1.17.0
* **Packaging format (Layers, PyPi)**: PyPi
* **AWS Lambda function runtime:** Python 3.8

michaelbrewer:
**Issue #, if available:**

#503

## Description of changes:

Set the formatted output to include milliseconds

fixes #503

**Checklist**

 

* [X] Update tests
* [X] Update docs

## Breaking change checklist

Potentially a breaking change if expecting only seconds.

---

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/appsync/scalar_types_utils.py
+++ b/aws_lambda_powertools/utilities/data_classes/appsync/scalar_types_utils.py
@@ -19,15 +19,21 @@ def _formatted_time(now: datetime.date, fmt: str, timezone_offset: int) -> str:
     str
         Returns string formatted time with optional timezone offset
     """
+    if timezone_offset != 0:
+        now = now + datetime.timedelta(hours=timezone_offset)
+
+    datetime_str = now.strftime(fmt)
+    if fmt.endswith(".%f"):
+        datetime_str = datetime_str[:-3]
+
     if timezone_offset == 0:
-        return now.strftime(fmt + "Z")
+        postfix = "Z"
+    else:
+        postfix = "+" if timezone_offset > 0 else "-"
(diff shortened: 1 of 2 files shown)


michaelbrewer:
@heitorlessa  I think that `%f` might cause issues with AppSync dates as they are expecting milliseconds.

michaelbrewer:
@heitorlessa seems to work as microseconds, so should i still trim off the extra 3 positions? or leave it

heitorlessa:
thanks for the fix!! Let's trim off to stay as milliseconds (`[:-3]`) to stay compliant whether validation catches server-side - Depending on how much data one pumps in, this should also save on costs

heitorlessa (changes requested):
Trim last three to stick as ms `[:-3]`

heitorlessa (approved):
Thanks for the changes, Mike!

Note to self: break down `test_data_classes` as it's 1.3K LOC atm

michaelbrewer:
> Thanks for the changes, Mike!
> 
> Note to self: break down `test_data_classes` as it's 1.3K LOC atm

It would be great if pytest can run parts in parallel :). Then i would definitely help split up things.

PR #506, opened 2021-07-04
feat(api-gateway): add common service errors

michaelbrewer:
**Issue #, if available:**

- https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/8

Pairs with https://github.com/awslabs/aws-lambda-powertools-python/pull/507

## Description of changes:

Changes:
- Add base ServiceError exception for rest api service errors
- BaseRequestError for 400s
- UnauthorizedError for 401s
- NotFoundError for 404s
- InternalServerError for 500s

Example usage for a unauthorized error

```python3
@app.route(method="GET", rule="/unauthorized-error")
def unauthorized_error():
    raise UnauthorizedError("Unauthorized")
```

will return:

```json5
{
	"statusCode": 401,
	"headers": {
		"Content-Type": "application/json"
	},
	"body: "{'statusCode':401,''message':'Unauthorized'}"
}
```

> NOTE: This is related to https://github.com/awslabs/aws-lambda-powertools-python/pull/507

**Checklist**

 

* [X] Update tests
* [ ] Update docs

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -4,6 +4,7 @@ import logging
 import re
 import zlib
 from enum import Enum
+from http import HTTPStatus
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 
 from aws_lambda_powertools.shared.json_encoder import Encoder
@@ -35,28 +36,28 @@ class BadRequestError(ServiceError):
     """Bad Request Error"""
 
     def __init__(self, msg: str):
-        super().__init__(400, msg)
+        super().__init__(HTTPStatus.BAD_REQUEST.value, msg)
 
 
 class UnauthorizedError(ServiceError):
     """Unauthorized Error"""
 
     def __init__(self, msg: str):
-        super().__init__(401, msg)
+        super().__init__(HTTPStatus.UNAUTHORIZED.value, msg)
 
 
 class NotFoundError(ServiceError):
     """Not Found Error"""
 
     def __init__(self, msg: str = "Not found"):
-        super().__init__(404, msg)
+        super().__init__(HTTPStatus.NOT_FOUND.value, msg)
 
 
 class InternalServerError(ServiceError):
     """Internal Server Error"""
 
     def __init__(self, message: str):
-        super().__init__(500, message)
+        super().__init__(HTTPStatus.INTERNAL_SERVER_ERROR.value, message)
 
 
 class ProxyEventType(Enum):
@@ -511,10 +512,10 @@ class ApiGatewayResolver:
 
         return ResponseBuilder(
             Response(
-                status_code=404,
+                status_code=HTTPStatus.NOT_FOUND.value,
                 content_type=APPLICATION_JSON,
                 headers=headers,
-                body=self._json_dump({"code": 404, "message": "Not found"}),
+                body=self._json_dump({"statusCode": HTTPStatus.NOT_FOUND.value, "message": "Not found"}),
             )
         )
 
@@ -527,7 +528,7 @@ class ApiGatewayResolver:
                 Response(
                     status_code=e.status_code,
                     content_type=APPLICATION_JSON,
-                    body=self._json_dump({"code": e.status_code, "message": e.msg}),
+                    body=self._json_dump({"statusCode": e.status_code, "message": e.msg}),
                 ),
                 route,
             )
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -7,57 +7,14 @@ from enum import Enum
 from http import HTTPStatus
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 
+from aws_lambda_powertools.event_handler import content_types
+from aws_lambda_powertools.event_handler.exceptions import ServiceError
 from aws_lambda_powertools.shared.json_encoder import Encoder
 from aws_lambda_powertools.utilities.data_classes import ALBEvent, APIGatewayProxyEvent, APIGatewayProxyEventV2
 from aws_lambda_powertools.utilities.data_classes.common import BaseProxyEvent
 from aws_lambda_powertools.utilities.typing import LambdaContext
 
 logger = logging.getLogger(__name__)
-APPLICATION_JSON = "application/json"
-
-
-class ServiceError(Exception):
-    """Service Error"""
-
-    def __init__(self, status_code: int, msg: str):
-        """
-        Parameters
-        ----------
-        status_code: int
-            Http status code
-        msg: str
-            Error message
-        """
-        self.status_code = status_code
-        self.msg = msg
-
-
-class BadRequestError(ServiceError):
-    """Bad Request Error"""
-
-    def __init__(self, msg: str):
-        super().__init__(HTTPStatus.BAD_REQUEST.value, msg)
-
-
-class UnauthorizedError(ServiceError):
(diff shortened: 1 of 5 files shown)


michaelbrewer:
@heitorlessa I am not sure which of the top service errors to include. So if there are more or less I can update it. Otherwise I can add some docs on usage.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -467,15 +540,25 @@ def _not_found(self, method: str) -> ResponseBuilder:
...
         """Actually call the matching route with any provided keyword arguments."""
-        return ResponseBuilder(self._to_response(route.func(**args)), route)
+        try:
+            return ResponseBuilder(self._to_response(route.func(**args)), route)
+        except ServiceError as e:
+            return ResponseBuilder(
+                Response(
+                    status_code=e.status_code,
+                    content_type=APPLICATION_JSON,
+                    body=json.dumps({"message": str(e)}),
```
jplock, line 558: Thoughts on including the status code in a “code” property as well?

michaelbrewer: @jplock - in the json response? That could work. I was looking at the standard errors coming back from the api gateway

jplock: Yea putting the status code also in the json response for clients

michaelbrewer: Cool. @jplock do you think it should be the exception class name (like in Chalice?) or that http code? Or something custom that you can define?

jplock: I thought the value of “e.status_code” for now

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -12,6 +12,51 @@
...
+
+    def __init__(self, msg: str):
+        super().__init__(401, msg)
+
+
+class NotFoundError(ServiceError):
+    """Not Found Error"""
+
+    def __init__(self, msg: str = "Not found"):
+        super().__init__(404, msg)
```
heitorlessa, line 52: Could you move these to an `exceptions.py`?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -12,6 +12,51 @@
 from aws_lambda_powertools.utilities.typing import LambdaContext
 
 logger = logging.getLogger(__name__)
+APPLICATION_JSON = "application/json"
```
heitorlessa, line 15: Because we're adding APPLICATION_JSON as part of this PR, what are your thoughts in creating a `content_types.py` should customers ask for binary mime types, for example?

That'd be easier to reason for newcomers and less options in the autocomplete that matters (subjective but sure).

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -467,18 +512,27 @@ def _not_found(self, method: str) -> ResponseBuilder:
...
         """Actually call the matching route with any provided keyword arguments."""
-        return ResponseBuilder(self._to_response(route.func(**args)), route)
+        try:
+            return ResponseBuilder(self._to_response(route.func(**args)), route)
+        except ServiceError as e:
+            return ResponseBuilder(
+                Response(
+                    status_code=e.status_code,
+                    content_type=APPLICATION_JSON,
+                    body=self._json_dump({"code": e.status_code, "message": e.msg}),
```
heitorlessa, line 530: for consistency, may we please use `statusCode` instead?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -467,18 +512,27 @@ def _not_found(self, method: str) -> ResponseBuilder:
         return ResponseBuilder(
             Response(
                 status_code=404,
-                content_type="application/json",
+                content_type=APPLICATION_JSON,
                 headers=headers,
-                body=json.dumps({"message": "Not found"}),
+                body=self._json_dump({"code": 404, "message": "Not found"}),
```
heitorlessa, line 517: for consistency, may we please use `statusCode` instead?

heitorlessa (changes requested):
Looks great! Thanks for adding this one in, Mike! Two minors only - **1**/ Moving exceptions to `exceptions.py`, and **2/** Whether we want to have `content_types.py` in anticipation that binary mime types will be asked

heitorlessa:
Fun fact I saw this week which I didn't know is that Python has HTTP Status as a built-in enum. I think the way it's now suffice tbh but thought you'd want to know:

```python
from http import HTTPStatus

HTTPStatus.BAD_REQUEST.value
```

aws_lambda_powertools/event_handler/api_gateway.py
michaelbrewer: @heitorlessa  But what about if appsync has it's own exceptions? Or do we reused them?

aws_lambda_powertools/event_handler/api_gateway.py
michaelbrewer: @heitorlessa same as the exceptions, i wonder if we should then make `api_gateway` into a "real" module?

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: AppSync uses GraphQL and has a different mechanism to handle exceptions. These errors, for example, wouldn't make back to the client as it'd in API Gateway unless customers implement them in VTL.

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: Exceptions is different. Could you expand on what "real" module means? Do you mean a separate utility altogether, a separate package in a different namespace, a framework?

heitorlessa (changes requested):
Explained why exceptions wouldn't be reused in AppSync. Exceptions is the only change pending

aws_lambda_powertools/event_handler/api_gateway.py
michaelbrewer: @heitorlessa  just mean a sub folder for api gatway.

heitorlessa:
To make things easier in terms of import, I'm totally fine with having the localized `exceptions.py` as well as exported in the API Gateway module itself, for example:

```python
from aws_lambda_powertools.event_handler.api_gateway import ApiGatewayResolver, NotFoundError
```

For those wanting more explicitness, they could:

```python
from aws_lambda_powertools.event_handler.api_gateway.exceptions import NotFoundError
```

Alternatively, if you want to keep exceptions within the utility namespace, then we'd need more explicit names to not confuse customers when to use them in AppSync, API Gateway, `anotherEventHandler` in the future:

```python
from aws_lambda_powertools.event_handler.exceptions import ApiGatewayNotFoundError
```

heitorlessa:
Merging as-is and fixing docstrings to not confuse AppSync customers accidentally using these exceptions meant for API GW/ALB in `develop`

aws_lambda_powertools/event_handler/content_types.py
```diff
@@ -0,0 +1,2 @@
+APPLICATION_JSON = "application/json"
+PLAIN_TEXT = "plain/text"
```
jplock, line 2: This should be “text/plain”

heitorlessa: Thanks for catching it, fixed now in develop w/ a note on `mimetypes` lib.

PR #516, opened 2021-07-09
feat(logger): add get_correlation_id method

michaelbrewer:
**Issue #, if available:**

- https://github.com/awslabs/aws-lambda-powertools-python/issues/518

## Description of changes:

- Call `remove_keys["correlation_id"]` when the correlation_id is set to None.
- Add `get_correlation_id` call

```python

logger = Logger()
logger.set_correlation_id("Foo")
assert "Foo" == logger.get_correlation_id()
logger.set_correlation_id(None)
assert logger.get_correlation_id() is None
```

**Checklist**

 

* [X] Update tests
* [X] Update docs


Nr18 (approved):
LGTM ;-)

michaelbrewer:
@heitorlessa should we go ahead with this feature?

heitorlessa:
I've just added a comment in the RFC.

TL;DR: Definitely yes for Getter, perhaps even a generic one to grab any key. Not yet for the remove correlation ID method.

aws_lambda_powertools/logging/logger.py
```diff
@@ -387,15 +387,28 @@ def structure_logs(self, append: bool = False, **keys):
...
         Parameters
         ----------
-        value : str
+        value : str, optional
+            Value for the correlation id. None will remove the correlation_id
+        """
+        if value is None:
+            self.remove_keys(["correlation_id"])
+        else:
+            self.append_keys(correlation_id=value)
```
heitorlessa, line 401: Keys with `None` value get removed by default when emitting a log message, so this wouldn't be necessary: https://github.com/awslabs/aws-lambda-powertools-python/blob/89e8151b5c930ba79d685d05a83ae362cf94839d/aws_lambda_powertools/logging/formatter.py#L126

heitorlessa (changes requested):
By default, Logger [removes any key with `None` value before emitting a log entry](https://github.com/awslabs/aws-lambda-powertools-python/blob/89e8151b5c930ba79d685d05a83ae362cf94839d/aws_lambda_powertools/logging/formatter.py#L126).

tests/functional/test_logger.py
```diff
@@ -460,6 +460,32 @@ def handler(event, _):
...
+    logger = Logger(service=service_name, stream=stdout)
+    logger.set_correlation_id("foo")
+
+    # WHEN calling set_correlation_id with None
+    logger.set_correlation_id(None)
+
+    # THEN there should be no correlation_id
+    logger.info("Foo")
+    log = capture_logging_output(stdout)
+    assert "correlation_id" not in log
```
heitorlessa, line 474: That's not needed ;) https://github.com/awslabs/aws-lambda-powertools-python/blob/89e8151b5c930ba79d685d05a83ae362cf94839d/tests/functional/test_logger_powertools_formatter.py#L175

aws_lambda_powertools/logging/logger.py
michaelbrewer: resolved.

tests/functional/test_logger.py
michaelbrewer: done

michaelbrewer:
@heitorlessa  made the changes. the diff is much more minimal, but still meet the needs of the original RFC

heitorlessa (approved):

heitorlessa:
Thanks a lot again @michaelbrewer !

PR #521, opened 2021-07-12
fix(mypy): fixes to resolve no implicit optional errors

michaelbrewer:
**Issue #, if available:**

## Description of changes:

Resolve mypy errors, down from 124 to 30 :-) .

Changes:
- Resolve errors related to the "no_implicit_optional" flag in mypy.ini
- Fix typing for `resolve_truthy_env_var_choice`

**Checklist**

 

* [X] Update tests
* [X] Update docs


michaelbrewer:
@heitorlessa  with `no_implicit_optional=True` we have alot more errors coming through in PyCharm.

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -244,7 +244,7 @@ def patch(self, modules: Tuple[str] = None):
 
     def capture_lambda_handler(
         self,
-        lambda_handler: Union[Callable[[Dict, Any], Any], Callable[[Dict, Any, Optional[Dict]], Any]] = None,
+        lambda_handler: Union[Callable[[Dict, Any], Any], Optional[Callable[[Dict, Any, Optional[Dict]], Any]]] = None,
```
michaelbrewer, line 247: @heitorlessa haha, quiet a handful :)

heitorlessa: This one is massive. I can't remember why Mypy asked us to have the second Callable w/ 3 arguments - it was back when you made the first PR on Mypy. I **think** it was a customer from the Nordics who injects additional kwargs in the handler.

This deserves its own type alias OR a generic altogether: https://github.com/awslabs/aws-lambda-powertools-python/pull/517/files#diff-ac2e7c32fce52e9e17243aafe7f69a62615e44d9d3a15dabd164d16b476dc5daR22

aws_lambda_powertools/shared/functions.py
```diff
@@ -2,14 +2,14 @@
...
-def resolve_truthy_env_var_choice(env: Any, choice: bool = None) -> bool:
+def resolve_truthy_env_var_choice(env: str, choice: Optional[bool] = None) -> bool:
     """Pick explicit choice over truthy env value, if available, otherwise return truthy env value
 
     NOTE: Environment variable should be resolved by the caller.
 
     Parameters
     ----------
-    env : Any
+    env : str
```
heitorlessa, line 12: it surprises me it didn't suggest Optional[str] as `env` could be None if we don't call `getenv` with a default. I suspect it scanned the code to see how it was called dunno

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -714,11 +714,11 @@ def _is_tracer_disabled() -> Union[bool, str]:
...
         self,
-        service: str = None,
-        disabled: bool = None,
-        auto_patch: bool = None,
-        patch_modules: Union[List, Tuple] = None,
-        provider: BaseProvider = None,
+        service: Optional[str] = None,
+        disabled: Optional[bool] = None,
+        auto_patch: Optional[bool] = None,
+        patch_modules: Union[List, Tuple, None] = None,
```
heitorlessa, line 720: is this correct? Shouldn't that be Optional[Sequence]?

heitorlessa (changes requested):
only one question on whether `Optional[Sequence]` is the correct type for `__build_config` as opposed to `Union[List, Tuple, None]`

aws_lambda_powertools/shared/functions.py
michaelbrewer: `strtobool(env)` would fail if `env` is not set (val.lower())

```python
def strtobool (val):
    """Convert a string representation of truth to true (1) or false (0).
    True values are 'y', 'yes', 't', 'true', 'on', and '1'; false values
    are 'n', 'no', 'f', 'false', 'off', and '0'.  Raises ValueError if
    'val' is anything else.
    """
    val = val.lower()
    if val in ('y', 'yes', 't', 'true', 'on', '1'):
        return 1
    elif val in ('n', 'no', 'f', 'false', 'off', '0'):
        return 0
    else:
        raise ValueError("invalid truth value %r" % (val,))
```

michaelbrewer: @heitorlessa see comment

aws_lambda_powertools/tracing/tracer.py
michaelbrewer: @heitorlessa  i will try that. There are a number of other places like this.

aws_lambda_powertools/shared/functions.py
heitorlessa: Ah yes! That's why I had set as Any at first I think as I was playing with having error handling or good defaults

aws_lambda_powertools/tracing/tracer.py
michaelbrewer: mypy did not like this, so i change it to be Optional  like the others.

heitorlessa:
I've just tried fixing the Union[Tuple, List] with a Sequence and it worked w/ MyPy -- perhaps I have MyPy misconfigured in PyCharm?

I'll look into other bug now then merge as-is if you're still seeing the issue.

```python
from typing import Any, Callable, Dict, List, Optional, Tuple, Union, Sequence

    def __init__(
        self,
        service: Optional[str] = None,
        disabled: Optional[bool] = None,
        auto_patch: Optional[bool] = None,
        patch_modules: Optional[Sequence[str]] = None,
        provider: Optional[BaseProvider] = None,
    ):

    def patch(self, modules: Optional[Sequence[str]] = None):
      
    def __build_config(
        self,
        service: Optional[str] = None,
        disabled: Optional[bool] = None,
        auto_patch: Optional[bool] = None,
        patch_modules: Optional[Sequence[str]] = None,
        provider: Optional[BaseProvider] = None,
    ):
```

I've also added the following modules to ignore their lack of typehints, which makes me wonder how reliable this is: https://pypi.org/project/mypy-boto3/

```ini
[mypy-boto3]
ignore_missing_imports = True

[mypy-boto3.dynamodb.conditions]
ignore_missing_imports = True

[mypy-botocore.config]
ignore_missing_imports = True

[mypy-botocore.exceptions]
ignore_missing_imports = True

[mypy-aws_xray_sdk.ext.aiohttp.client]
ignore_missing_imports = True
```

michaelbrewer:
@heitorlessa ok your suggestions work prefectly, i have pushed the latest

heitorlessa (approved):

Issue #481, opened 2021-06-14
Dataclasses are not... actual dataclasses (?!)


LouisAmon:
**What were you initially searching for in the docs?**

I'm using the @event_source decorator and I was looking for a way to print my event in my Cloudwatch logs.

When I simply `print(event)`, I just get a class instance, but I want to be able to see the actual keys and values :)

Being used to working with Python dataclasses, I naturally tried [asdict](https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict):
```python
print(event.asdict())
```

Or even...
```python
import dataclasses

dataclasses.asdict(event)
```
...and I was very surprised by the result:
```
asdict() should be called on dataclass instances
```
**Is this related to an existing part of the documentation? Please share a link**

https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/data_classes/

**Describe how we could make it clearer**

Dataclasses are meant to be a Pythonic way to define a data model.

If you define your own class (`DictWrapper` in your case) which inherits from nothing then all the APIs and principles behind dataclasses are lost.

Dataclasses are used in libraries such as sqlalchemy or strawberry to define a unified way to communicate with a database or build a GraphQL API... dataclasses provides unity between all these, so why not AWS Lambda ? :)

**If you have a proposed update, please share it here**

I think we should either revamp the code to actually define proper data classes, or stop calling it as such : it's just confusing.

What do you think ?

heitorlessa:
Hey @LouisAmon thanks for raising this. We have created #346 to rewrite the entire documentation on dataclass utility - naming is hard. This next release is 80% dedicated to that doc.

This Friday, I'll start revamping that doc and lay down ambitions for v2 - that includes changing `dataclasses` to <something> more prescriptive so we don't break anyone in v1.

If you have ideas for naming I'd love to hear them so we can improve everyone's experience. If we could have a better name by the time the new docs are completely rewritten, then I could easily create code aliases and redirect docs altogether

Thank you!

LouisAmon:
Well to be honest I think `dataclasses` are indeed a best practice in Python, especially when it comes to data modeling and typing

May I ask why the project didn't go into actually using the `@dataclass` decorator ?

I guess now that a huge amount of code is in place, you can't just add `@dataclass` on top of `DictWrapper`... but still, have you tried ?

[...]

PR #526, opened 2021-07-15
feat(event-handler): Support AppSyncResolverEvent subclassing

whardier:
…vent

No outsstanding issue

## Description of changes:

Allows for custom subclass of AppSyncResolverEvent data_class to be offered to AppSyncResolver.resolve

**Checklist**

* [x] Update tests
* [x] Update docs

## Breaking change checklist

Not breaking

**RFC issue #**:


whardier:
Alternatively the annotation 'current_event: AppSyncResolverEvent' could simply be overridden and then utilized via self.__annotations__['current_event'](event) .. but I'm not sure if __annotations__ is optimized out or how recognized that pattern is.

heitorlessa:
Hey @whardier, thanks for the contribution!!

Could you share a sample subclass that you created for AppSync event?

This will help me understand the use case, and what's potentially missing from the AppSyncEvent class before we add tests for this PR.

Thank you

whardier:
Sure.. the following is a non-patched solution to provide the custom properties we are accessing from the underlaying _data:

```python
from aws_lambda_powertools.utilities.data_classes.appsync_resolver_event import (
    AppSyncIdentityCognito,
    AppSyncResolverEvent,
)

class CustomAppSyncResolverEvent(AppSyncResolverEvent):
    @cached_property
    def auth_client_id(self) -> Optional[str]:
        auth_client_id: Optional[str] = None
        if isinstance(self.identity, AppSyncIdentityCognito):
            auth_client_id = self.identity.claims.get("custom:client_id")
        return auth_client_id

    @cached_property
    def auth_user_profile_id(self) -> Optional[str]:
        auth_user_profile_id: Optional[str] = None
        if isinstance(self.identity, AppSyncIdentityCognito):
            auth_user_profile_id = self.identity.claims.get("custom:user_profile_id") or self.identity.sub
        return auth_user_profile_id

# ... below to assist with hints

class CustomAppSyncResolver(AppSyncResolver):
    current_event: CustomAppSyncResolverEvent
    
# ... specific enough sample appsync lambda event handler

@appsync.resolver(type_name="Mutation", field_name="logThingUserEvent")
@validate_arguments
@typechecked
def log_thing_user_event(
    event_type: ThingUserEventTypeEnum,
    measure_reference: Annotated[str, Field(max_length=256)],
    measure_source: Annotated[str, Field(max_length=32)] = DEFAULT_THING_USER_EVENT_APPSYNC_METRIC_SOURCE,
):

    # ... sample code
    
    auth_client_id: Optional[str] = appsync.current_event.auth_client_id
    auth_user_profile_id: Optional[str] = appsync.current_event.auth_user_profile_id
    
    # ... other junk

@core_tracer.capture_lambda_handler
@core_logger.inject_lambda_context(correlation_id_path=correlation_paths.APPSYNC_RESOLVER)
@event_source(data_class=CustomAppSyncResolverEvent)
@typechecked
def appsync_handler(event: CustomAppSyncResolverEvent, context: AnyLambdaContext) -> dict:

    # NOTE: Until aws lambda powertools allows for hooks/class references to be part of the resolve method to set up a
    # proper current_event with a subclassed dataclass and repeat the innards of the resolve method here (SRS).

    appsync.current_event = event
    appsync.lambda_context = context
    resolver = appsync._get_resolver(event.type_name, event.field_name)
    return resolver(**event.arguments)
    
```

The proposed patch mimics a bit of the model_class approach seen elsewhere in this project.  However a sneaky pattern to do this is as follows and seems to be gaining in popularity:

```python
class LibraryDataClass:
    pass


class CustomDataClass(LibraryDataClass):
    @property
    def custom_property(self):
        return "custom property data"


class LibraryExecutionClass:
    data: LibraryDataClass

    def execute(self):
        data_class = self.__annotations__["data"]
        self.data = data_class()


class CustomExecutionClass(LibraryExecutionClass):
    data: CustomDataClass


# >>> custom_executor = CustomExecutionClass()
# >>> custom_executor.execute()
# >>> custom_executor.data.custom_property
# 'custom property data'
```

whardier:
I concerned that using model as a kwarg here would lead to thoughts that this should be a parser.model rather than a dictwrap data_class.  I know I would have been confused seeing model.

heitorlessa:
> I concerned that using model as a kwarg here would lead to thoughts that this should be a parser.model rather than a dictwrap data_class. I know I would have been confused seeing model.

That's a great point. I find naming hard because even `data_class` could lead to discussions like #481.

How does `data_model` sound? I've made all changes necessary - docs, tests, mypy, agreeing on that and we can merge and release ;)

nmoutschen:
> > I concerned that using model as a kwarg here would lead to thoughts that this should be a parser.model rather than a dictwrap data_class. I know I would have been confused seeing model.
> 
> That's a great point. I find naming hard because even `data_class` could lead to discussions like #481.
> 
> How does `data_model` sound? I've made all changes necessary - docs, tests, mypy, agreeing on that and we can merge and release ;)

Sharing my 2 cents here, I'd personally prefer `data_model` as `model` is quite overloaded in different Python frameworks and libraries.

heitorlessa:
Waiting for @whardier before we merge it

whardier:
I'm still down with something as simple as data_class (as seen here: `@event_source(data_class=CustomAppSyncResolverEvent)
`) but I'm not as behind the scenes on this project ;) - I can't find any other examples on where passing the data_class class name is done any other way.  I will have to catch up on #481.  I really dig the dict wrapper and feel it is appropriately named and unfortunately overlapping other useful naming.  For continuity sake I would prefer to see data_class used until a breaking change can be made to ultimately address any concerns surrounding it.

My preferences aside.  I approve this and thank you for catching the __call__ case and handling the typing the proper way.  It was nice being able to submit some thoughts this way and see changes come back as code.

heitorlessa:
We'll keep it as `data_model` as the amount of energy spent on #481 was abnormal on naming things - I'm merging and release it this week ;)

Thanks a lot again @whardier

Issue #465, opened 2021-06-07
Tracer.capture_method decorator does not preserve function type hints


zlalvani:
**What were you trying to accomplish?**
I want to use the `@tracer.capture_method` decorator for detailed function level tracing however it's currently unusable because it disables all errors when linting with `mypy`.

## Expected Behavior
See code example below

## Current Behavior
Function signatures (argument and return types) are not preserved when decorated with `@tracer.capture_method` and linting with `mypy`. 

## Possible Solution
Mypy can be forced to look at the decorated function signature: https://mypy.readthedocs.io/en/stable/generics.html#declaring-decorators

## Steps to Reproduce (for bugs)

1. Use code sample below
2. Run mypy (tested on `mypy==0.782`)

## Environment

* **Powertools version used**: `aws-lambda-powertools==1.10.2`
* **Packaging format (Layers, PyPi)**: PyPi

> [How to enable debug mode](https://awslabs.github.io/aws-lambda-powertools-python/#debug-mode)**

```python
tracer = Tracer()

@tracer.capture_method
def foo(bar: str) -> str:
    return bar

x = foo('baz')
x.get('this is not a dict')  # should cause mypy error
y = foo(123)  # should cause mypy error
```

heitorlessa:
Hey @zlalvani thanks a lot for raising this. While we've had numerous MyPy issues fixed we don't yet have full support for it -- Roadmap item that you can help us +1 to prioritize: https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/3

I'm unsure how to fix this atm as I had thought `functools.wraps` would ensure some level of correctness here (not at wrapt level, tho) -- This is that part of the code: https://github.com/awslabs/aws-lambda-powertools-python/blob/8724294013e3157de9478f90e7c1283a952de54f/aws_lambda_powertools/tracing/tracer.py#L607

I've been meaning to refactor `capture_method` as it's the most complex logic we have in the codebase, so it'd be a good exercise nevertheless to get this improved.

I'd appreciate if you have any guidance on how we can satisfy MyPy in a backwards compatible way - happy to provide any guidance if you'd like to try contributing it.

Thank you!

heitorlessa:
Updating here to let you know the fix will be available in the next release

heitorlessa:
Just merged the fix for synchronous functions - Async however requires the new ParamSpec only available in 3.10, so I'll wait for more customers using mypy + async fn as it'll increase the package size by 240K.

[...]

PR #529, opened 2021-07-16
fix(tracer): mypy generic to preserve decorated method signature

heitorlessa:
**Issue #, if available:** #465

## Description of changes:

**Checklist**

 

* [x] Use generic callable upper bound to preserve signature for type checker (mypy)
* [ ] ~~Use generic awaitable upper bound to preserve signature for type checker (mypy)~~
    - [See limitation](https://github.com/awslabs/aws-lambda-powertools-python/pull/529#issuecomment-881891898)

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/tracing/tracer.py
+++ b/aws_lambda_powertools/tracing/tracer.py
@@ -5,7 +5,7 @@ import inspect
 import logging
 import numbers
 import os
-from typing import Any, Callable, Dict, Optional, Sequence, Union
+from typing import Any, Awaitable, Callable, Dict, Optional, Sequence, TypeVar, Union, cast, overload
 
 from ..shared import constants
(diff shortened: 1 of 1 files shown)


heitorlessa:
**Limitation:** Async functions signature for type checking can only be preserved by using ParamSpec introduced in 3.10 - [PEP 612](https://www.python.org/dev/peps/pep-0612/)

We could bring typing_extensions module (240K) which also brings other fixes we'll likely encounter as we adopt mypy more broadly in this lib, including nice features like [Protocol for interfaces](https://www.python.org/dev/peps/pep-0544/).

I'll postpone bringing `typing_extensions` until we either **a)** hear directly from customers using mypy with async fn as this will manifest elsewhere too not just Powertools, or **b)** find that we need `typing_extensions` in order to resolve current mypy findings someplace else.

cc @michaelbrewer @pcolazurdo @cakepietoast

Issue #520, opened 2021-07-11
Error resolving path in Api Gateway event handler


marcioemiranda:
Suppose one has two routes to delete resources like the following:

Use Case 1:

Route in handler

```
@app.delete("/accounts/<account_id>")
def delete_account(account_id: str) -> None:
```

Request

```
{
        "path": "/accounts/123",
        "httpMethod": "DELETE",
        "requestContext": {"requestId": "c6af9ac6-7b61-11e6-9a41-93e8deadbeef"},  # correlation ID
}
```

Use Case 2:

```
@app.delete("/accounts/<account_id>/source_networks")
def delete_source_networks(account_id: str) -> Account:
```

Request

```
{
        "path": "/accounts/123/source_networks"
        "httpMethod": "DELETE",
        "requestContext": {"requestId": "c6af9ac6-7b61-11e6-9a41-93e8deadbeef"},  # correlation ID
}
```

## Expected Behavior

Expected behavior is that account_id equals 123 in both use cases

## Current Behavior

For use case 1 resolver calls the right function and sets account_id = 123
For use case 2, resolver calls the function of use case 1 and sets accound_id = 123/source_networks

## Possible Solution

Not sure if this is actually the expected behavior of proxy integration with its greedy regexp or a bug in Lambda Power Tools

## Environment

Power Tools version 1.17.1

```python
delete_source_networks_event = {'body': '{"networks": ["172.217.0.0/16", "8.8.8.8"]}', 'httpMethod': 'DELETE', 'path': **'/accounts/123/source_networks'**, ...}
{"level":"INFO","location":"delete:209","message":"Account deleted","timestamp":"2021-07-11 15:12:42,205-0300","service":"admin_api","cold_start":false,"function_name":"FUNCTION_NAME","function_memory_size":1024,"function_arn":"INVOKED_FUNCTION_ARN","function_request_id":"AWS_REQUEST_ID","correlation_id":"c6af9ac6-7b61-11e6-9a41-93e8deadbeef",**"account_id":"123/source_networks"**,"tax_id":"123","service_id":3}
```

heitorlessa:
Hey @marcioemiranda , thanks for raising it! Could you share how you're configuring API Gateway?

At a first glance this looks like API GW route matching, so I've created this additional test for use case 2 and it works as expected --- I'm creating an API Gateway with the same path to be on the safe side and report back with a repro repo.

```python
def test_nested_dynamic_routes():
    # GIVEN
[...]

PR #533, opened 2021-07-18
fix(api-gateway): non-greedy route pattern regex

heitorlessa:
**Issue #, if available:** #520

## Description of changes:

This fixes a regex bug that used a greedy pattern ending with incorrect path resolution, as any path after a pattern would be included in the argument.

Excerpt:

```python
@app.get("/accounts/<account_id>")
def get_account(account_id: str):
    print(f"Account ID ({account_id}) would be 123")

# Greedy regex would inject the incorrect function parameter
@app.get("/accounts/<account_id>/source_networks")
def get_account_networks(account_id: str):
    # re.compile()
    print(f"Account ID ({account_id}) would be 123/source_networks")
```

In this example, say we have a GET request as `/accounts/123` and another as `/accounts/123/source_networks`, we'd have the following effect prior to this fix:

Function | Regex | Effective account_id value
------------------------------------------------- | --------------------------------------------------------------------------------- | -------------------------------------------------
get_account | `r'^/accounts/(?P<account_id>.+)$'` | `123`
get_account_networks | `r'^/accounts/(?P<account_id>.+)/source_networks$'` |  `123/source_networks`

With this fix, `account_id` parameter would be 123 in both occasions due to [word boundary](https://www.regular-expressions.info/wordboundaries.html) not being non-greedy. This also allows an arbitrary number of dynamic route paths and static route paths.

Function | Regex | Effective account_id value
------------------------------------------------- | --------------------------------------------------------------------------------- | -------------------------------------------------
get_account | `r'^/accounts/(?P<account_id>\\w+\\b)$'` | `123`
get_account_networks | `r'^/accounts/(?P<account_id>\\w+\\b)/source_networks$'` |  `123`

**Checklist**

 

* [x] Update tests

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -20,6 +20,9 @@ from aws_lambda_powertools.utilities.typing import LambdaContext
 
 logger = logging.getLogger(__name__)
 
+_DYNAMIC_ROUTE_PATTERN = r"(<\w+>)"
+_NAMED_GROUP_BOUNDARY_PATTERN = r"(?P\1\\w+\\b)"
+
 
 class ProxyEventType(Enum):
     """An enumerations of the supported proxy event types."""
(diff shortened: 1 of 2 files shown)


heitorlessa:
Note: one other bug and potential minor issue to address as part of 1.18 release or the next one after that

1. Raise an exception if the same function has been registered for a separate route rule

2. App.route decorator doesn't preserve fn signature and will likely fail Mypy like tracer.capture_method older bug

PR #553, opened 2021-07-22
fix(parser): apigw wss validation check_message_id; housekeeping

michaelbrewer:
**Issue #, if available:**

## Description of changes:

Various housekeeping, code coverage and mypy tasks

Changes:
- Fix bug with `check_message_id` not validating against the RequestContext
- Add some more unit tests for `check_message_id`
- Add more mypy coverage
- Add missing `pragma: no cover`
- Simplify logic in `idempotency.py` and remove unused `max_handler_retries` property

**Checklist**

 

* [X] Update tests
* [X] Update docs


michaelbrewer:
@risenberg-cyberark  - can you check the apigw fix I made for the validation.

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -183,7 +178,7 @@ def _get(self, name: str, **sdk_options) -> str:
         # Explicit arguments will take precedence over keyword arguments
         sdk_options["Key"] = {self.key_attr: name}
 
-        return self.table.get_item(**sdk_options)["Item"][self.value_attr]
+        return str(self.table.get_item(**sdk_options)["Item"][self.value_attr])
```
risenberg-cyberark, line 181: why always str? dynamodb can store all sorts of data types

aws_lambda_powertools/utilities/parameters/dynamodb.py
```diff
@@ -209,4 +204,4 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
             response = self.table.query(**sdk_options)
             items.extend(response.get("Items", []))
 
-        return {item[self.sort_attr]: item[self.value_attr] for item in items}
+        return {str(item[self.sort_attr]): str(item[self.value_attr]) for item in items}
```
risenberg-cyberark, line 207: why always str? dynamodb can store all sorts of data types

aws_lambda_powertools/utilities/feature_toggles/appconfig_fetcher.py
```diff
@@ -56,11 +56,15 @@ def get_json_configuration(self) -> Dict[str, Any]:
             parsed JSON dictionary
         """
         try:
-            return self._conf_store.get(
-                name=self.configuration_name,
-                transform=TRANSFORM_TYPE,
-                max_age=self._cache_seconds,
-            )  # parse result conf as JSON, keep in cache for self.max_age seconds
+            # parse result conf as JSON, keep in cache for self.max_age seconds
+            return cast(
```
risenberg-cyberark, line 60: if for somereason the cast fails, you dont catch the exception.

aws_lambda_powertools/utilities/parser/models/apigw.py
```diff
@@ -68,6 +68,13 @@ class APIGatewayEventRequestContext(BaseModel):
     routeKey: Optional[str]
     operationName: Optional[str]
 
+    @root_validator()
+    def check_message_id(cls, values):
```
risenberg-cyberark, line 72: you cant define a root validator like that . when you use @root_validator it needs to be inside the class that uses it. Usually you define it outside when you want to reuse it at several places but then you define it differetly inside the classes which use it.

risenberg-cyberark (changes requested):

aws_lambda_powertools/utilities/parameters/dynamodb.py
michaelbrewer: @risenberg-cyberark  - this is typing related, the outer type is excepting a str.

aws_lambda_powertools/utilities/parameters/dynamodb.py
michaelbrewer: @risenberg-cyberark  the outer typing is excepting a `Dict[str, str]`

aws_lambda_powertools/utilities/parser/models/apigw.py
michaelbrewer: @risenberg-cyberark - ok, interesting. Coz this passes the unit tests now. And before it was not actually catching invalid events.

aws_lambda_powertools/utilities/parameters/dynamodb.py
michaelbrewer: So, if this is practically true for parameters, we should then update the outer type to `Dict[str, Any]`, but i am not sure this would work correctly with the right of the parameters features.

aws_lambda_powertools/utilities/parser/models/apigw.py
```diff
@@ -83,10 +90,3 @@ class APIGatewayProxyEventModel(BaseModel):
     stageVariables: Optional[Dict[str, str]]
     isBase64Encoded: bool
     body: str
-
-    @root_validator()
-    def check_message_id(cls, values):
-        message_id, event_type = values.get("messageId"), values.get("eventType")
```
michaelbrewer, line 89: @risenberg-cyberark  - tried to get this to be triggered, but `values` was for the parent and so the `messageId` was actually inside of the `requestContext`, but referencing that here was causing other issues.

Another solution would be just to remove this check, if AWS is doing this for us?

aws_lambda_powertools/utilities/parser/models/apigw.py
risenberg-cyberark: https://pydantic-docs.helpmanual.io/usage/validators/#root-validators
that's very odd, i'm using the documented approach of root validators

risenberg-cyberark: it's because it has an extra () by mistake in the decoator

risenberg-cyberark: can you remove the () from the decorator?

tests/functional/parser/test_apigw.py
```diff
@@ -100,3 +103,20 @@ def test_apigw_event():
     assert request_context.operationName is None
     assert identity.apiKey is None
     assert identity.apiKeyId is None
+
+
+def test_apigw_event_with_invalid_websocket_request():
+    # GIVEN an event with an eventType != MESSAGE and has  a messageId
```
michaelbrewer, line 109: @risenberg-cyberark - this is how i am validating this?

aws_lambda_powertools/utilities/parser/models/apigw.py
risenberg-cyberark: you can remove it, it's not that critical

tests/functional/parser/test_apigw.py
risenberg-cyberark: you should provide a full valid event with just event type being not aligned otherwise it would fail way before it gets here

michaelbrewer: @risenberg-cyberark  - i thought of doing that, but this is just a minimal example. In the assert i do check if the specific message is retuned. Which it is after the changes

risenberg-cyberark: I dont know how it affects the order the validation since it's not documented in pydantic. The root validator is supposed to run after the entire item has passed validation, which with your input, it doesnt as it's a partial event.
that's why in docs you can have root_validator(pre=True)

risenberg-cyberark: anyway, it's not the expected behavior of the validator here, i'd suggest we either move it back and fix it or remove it completly if you dont think it's valuable enough

tests/functional/parser/test_apigw.py
```diff
@@ -100,3 +103,20 @@ def test_apigw_event():
...
+            "messageId": "messageId",
+        },
+    }
+
+    # WHEN calling event_parser with APIGatewayProxyEventModel
+    with pytest.raises(ValidationError) as err:
+        handle_apigw_event(event, LambdaContext())
+
+    # THEN raise TypeError for invalid event
+    assert "messageId is available only when the `eventType` is `MESSAGE`" in str(err.value)
```
michaelbrewer, line 122: @risenberg-cyberark - coz it is definitely being triggered. 

But i will try to build up a full event regardless.

tests/functional/parser/test_apigw.py
michaelbrewer: I will tweak it and see if it makes sense. Otherwise we can remove it, as it is definitely is not working as is. Interesting placing with it none the less.

risenberg-cyberark:
btw, the code coverage is insane :) @michaelbrewer

aws_lambda_powertools/utilities/parser/models/apigw.py
michaelbrewer: @risenberg-cyberark  - did this, but this does not change much

tests/functional/parser/test_apigw.py
michaelbrewer: @risenberg-cyberark  i have asserted that there is only one error now.

aws_lambda_powertools/utilities/parameters/dynamodb.py
michaelbrewer: @heitorlessa do you think this type casting is the right idea? Or should we ignore these kind of Mypy issues?

aws_lambda_powertools/utilities/parameters/dynamodb.py
heitorlessa: I'd agree with @risenberg-cyberark and this points out that we might have incorrect return typing annotations, and that tests aren't covering negatives

michaelbrewer: @heitorlessa ok then in some cases we expect a `str`, so we should maybe add some type checking there.

michaelbrewer:
@heitorlessa  i will leave out the other possible type checking and validation checks for idempotent for another time

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -78,9 +78,7 @@ def idempotent(
         try:
             return idempotency_handler.handle()
         except IdempotencyInconsistentStateError:
-            if i < max_handler_retries:
-                continue
-            else:
+            if i == max_handler_retries:
```
heitorlessa, line 81: nice one!!

heitorlessa (approved):

Issue #544, opened 2021-07-20
BaseProxyEvent: This should draw from self.decoded_body()


whardier:
It seems as though .json_body() would also attempt to decode the body of a proxyevent if the body is marked as base64 encoded.

https://github.com/awslabs/aws-lambda-powertools-python/blob/cb0d54bc36513961422fdebd5953bc11453b4651/aws_lambda_powertools/utilities/data_classes/common.py#L68

michaelbrewer:
@whardier  i have a patch for that. Going to add an example test case to

heitorlessa:
Thanks @michaelbrewer and thanks @whardier for raising it.

@whardier we'd appreciate if you use the issue templates we have available next time you open an issue.

Tagging this for 1.19.0 in early August

heitorlessa:
Now released as part of 1.19.0: https://github.com/awslabs/aws-lambda-powertools-python/releases/tag/v1.19.0

whardier:
YAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAY

PR #560, opened 2021-07-23
feat(data-classes): decode json_body if based64 encoded

michaelbrewer:
**Issue #, if available:**

- #544

## Description of changes:

Use `decoded_body` for the body in `json_body`. This will allow for a json string that is base64 encoded to be decoded and parsed in one go.

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -65,7 +65,7 @@ class BaseProxyEvent(DictWrapper):
     @property
     def json_body(self) -> Any:
         """Parses the submitted body as json"""
-        return json.loads(self["body"])
+        return json.loads(self.decoded_body)
 
     @property
     def decoded_body(self) -> str:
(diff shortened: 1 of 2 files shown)


michaelbrewer:
@heitorlessa it is a pretty small change, if you want to have a look

heitorlessa:
Oh, I thought I did!! Thanks for the nudge

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -65,7 +65,7 @@ def body(self) -> Optional[str]:
     @property
     def json_body(self) -> Any:
```
heitorlessa, line 66: Should we fix the older type annotation you Dict[str, Any] here too?

heitorlessa (approved):

aws_lambda_powertools/utilities/data_classes/common.py
michaelbrewer: It could be a list too :)

heitorlessa:
Eer, true

PR #579, opened 2021-08-01
feat(event-handler): prefixes to strip for custom mappings

### LINKED ISSUE #576: Allow APIGatewayResolver to handle Custom Domain with Mapping and still work
Have opened as a feature request as I believe this is not exactly a bug, perhaps an oversight in the ApiGatewayResolver design (happy to discuss more).  
A couple of months ago I developed an API for a project whose API gateway was associated with a custom domain. 
This week I built a new API extension (new gateway) using AWS Lambda Powertools for Python and have applied several routes into the one lambda using the API Gateway Resolver with the intention of adding to the same custom domain since want the ApiGateway to be hosted on the one common DNS domain.  When associating a second gateway to a custom domain you must associate a  mapping for the additional gateways so the API paths do not collide and everything works.

Within my lambda resolver setup for this second gateway if I have a resolver route of "/status" setup this works fine if it is mounted as-is on the root of the domain.  If I add to a custom domain as a second gateway I need to add in a "mapping", for example, sake let's say I choose "unique".  The AWS API Gateway event "path" for the API when I call it as "https://mycustomdomain.com/unique/status" is set to "/unique/status" which means the power tools Resolver will respond with a "404, NOT FOUND" since that path is not setup within the Resolver routes.

I have noticed that the "resource" path in the event correctly holds the route as "/status" but the path holds the route as "/unique/status"

This is a complicated one - the ApiGatewayResolver does not allow me to mount the gateways developed with this component on any custom domain with a mapping and have the lambda API actually work - I actually kind of think this is a bug but not raising as such since the implementation seems perfectly reasonable.

What I would ideally like is the freedom to be able to mount my Python API to a custom domain using any mapping I choose and still have the API resolver find my routes within the lambda code correctly.

The current implementation uses the "path" of the ApiGateway Lambda event which houses the complete API path including the mapping which breaks the resolver.

As a work around I can simply change my routes to include the proposed mapping but then this stops me from being able to use Api Gateway configuration to remap an API in the future and actually have it work without changing my code which is not ideal given this is a feature of using Services like the AWS Api gateway.

**Additional context**
This kind of also brings into question how the Event content is generated and passed to lambda by the ApiGateway since one could argue it makes no sense that the "path" also includes the logical "mapping" from the API gateway Custom Domain configuration (not an argument I want to start but a consideration given the logical config nature of this scenario).

I have taken a look at the event structure from this configuration and notice the "resource" contains a correct path that matches the route I have in my Lambda Resolver routes in python code but is possibly not ideal.

michaelbrewer:
**Issue #, if available:** https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/34

- closes #576

## Description of changes:

Changes:
- Add `prefix` to be stripped from the event path, default is `None` which does not modify the path
- Add docstring for `serializer`

Example usage:

- Passing in multiple `strip_prefixes` "/unique/v1" and "/foo/v1"
- Route `/status` will then match on `/unique/v1/status` (and `/status`)
- By default `strip_prefixes` is None, and therefore does not touch the event path
- Note the prefix should not end with a `/`

```python3
import os

app = ApiGatewayResolver(strip_prefixes=["/unique/v1", "/foo/v1"])

@app.get("/status")
def foo():
    return {"status": "OK"}

# All calls should return a 200
response = app({"httpMethod": "GET", "path": "/unique/v1/status"}, None)
response = app({"httpMethod": "GET", "path": "/foo/v1/status"}, None)
response = app({"httpMethod": "GET", "path": "/status"}, None)
```

**Checklist**

 

* [X] Update tests
* [X] Update docs


michaelbrewer:
@walmsles  this might help solve the issue you raised.

walmsles:
Hi @michaelbrewer it kind of does in a limited way but would think there should be a no code/config option of doing this to use APIGW fully.

API GW allows me to create a lambda that can be mapped to multiple domains/mappings and I should be free to do that to mount an API with resolver to multiple domains with potentially different mapping paths without need integrated to consider environment setup as well.  

Does the solution allow for multiple mappings? It seems singular.

Let me add some more comments/data to the original request to help more.  I feel a design change on the actual router implementation would be simpler, less code and more flexible (although might feel like a rewrite so needs careful consideration)

michaelbrewer:
> Hi @michaelbrewer it kind of does in a limited way but would think there should be a no code/config option of doing this to use APIGW fully.
> 
> API GW allows me to create a lambda that can be mapped to multiple domains/mappings and I should be free to do that to mount an API with resolver to multiple domains with potentially different mapping paths without needing to consider environment setup as well.
> 
> Does the solution allow for multiple mappings? It seems singular.
> 
> Let me add some more comments/data to the original request to help more. I feel a design change on the actual router implementation would be simpler, less code and more flexible (although might feel like a rewrite so needs careful consideration)

Ok interesting so you have the same "status" lambda for many different api gateways / mappings and it needs to automagically find this "status" route?

With some test events maybe there is a way, but I would not to result in strange breaking behavior.

michaelbrewer:
@walmsles  also note, this would have to work for Api Gateway V1, V2 and ALB lambdas.

walmsles:
@michaelbrewer  understand - not familiar with all the event models for these other API components so maybe there is no nice way - Have added a sample test event to the original request for discussion :-)

walmsles:
@michaelbrewer fair on support for ALL event formats.  What you have proposed is useful - the use case of same lambda to many mappings is an unlikely use case but APIGW does allow/support that.  

For my exact use case this week  - it was the need to having my ApiGatewayResolver configured differently to my APIGW config which felt weird and wrong :-)  The prefix idea will assist and resolve this particular use case so happy if that's the outcome.

I don't see where I would actually normally use the same APIGW mapped to many custom domains - the use case exists and is where my Software engineering brain went thinking about how APIGW works in this case - but in reality question is probably a weird edge case :-)

michaelbrewer:
@walmsles - i will run some test cases on a deployed lambda to see if there is a method to determine the internally mapped lambda vs the custom mapped lambda

heitorlessa:
I'd like us to think how we could solve this on our side as a fallback mechanism without having to ask customers to change their code - whether they use custom domain mappings or not.

michaelbrewer:
@heitorlessa - any ideas would be great, based on what we get back from the aws events.

heitorlessa:
I totally missed option 3, since that seems to be always true in REST API v1 payload: `requestContext. path`

Since we know upfront whether customers are using it for REST API v1, ALB, or HTTP API, what's the downside of using `requestContext.path` for REST API v1 only besides breaking customers unit tests?

heitorlessa:
> **UPDATE**: Option 3 won't work as REST API v1 also includes the stage name (see new comment)
> 
> I totally missed option 3, since that seems to be always true in REST API v1 payload: `requestContext. path`
> 
> Since we know upfront whether customers are using it for REST API v1, ALB, or HTTP API, what's the downside of using `requestContext.path` for REST API v1 only besides breaking customers unit tests?
> 
> Replaced `_remove_prefix` with `_extract_path`:
> 
> ```python
>     def _resolve(self) -> ResponseBuilder:
>         """Resolves the response or return the not found response"""
>         method = self.current_event.http_method.upper()
>         path = self._extract_path()   # NEW
>         for route in self._routes:
>             if method != route.method:
>                 continue
>             match: Optional[re.Match] = route.rule.match(path)
>             if match:
>                 logger.debug("Found a registered route. Calling function")
>                 return self._call_route(route, match.groupdict())  # pass fn args
> 
>         logger.debug(f"No match found for path {path} and method {method}")
>         return self._not_found(method)
> 
>     # NEW
>     def _extract_path(self) -> str:
>         """Extracts correct path regardless of custom domain mapping setting"""
>         resolver = self._proxy_type
>         if self._proxy_type == ProxyEventType.APIGatewayProxyEvent:
>             return self.current_event.request_context.path
> ```

It sounded too good to be true on only relying on `requestContext.path`...

This problem gets further complicated when you have nested mapping paths of variable lengths.This alone will be difficult to keep track in the environment variable as you might have countless of these, if you follow this edge case of having multiple APIs with different mappings pointing to the same Lambda fn.

We would to have to either compromise by stating what we won't support nested paths and allow for a list of one level mapping paths, or something else like maybe regex, jmespath. Either will make it complex on customers going this route, even if we drop it and suggest a regex for a route using lookahead.

what do you reckon @walmsles @michaelbrewer?  

---

Single custom domain mapping path - **`v1`**

```json
{
    "path": "/v1/payment",
    "resource": "/payment",
    "requestContext": {
        "resource": "/payment",
        "path": "/v1/payment",
        "httpMethod": "GET",
        "requestContext": {
            "resourceId": "j9knhf",
            "resourcePath": "/payment",
            "httpMethod": "GET",
            "path": "/v1/payment",
            "stage": "default",
            "domainPrefix": "api",
            "domainName": "api.serverlessa.dev",
        },
    }
}
```

Nested custom domain mapping path - **`v1/nested`**

```json
Custom domain mapping - proxy resource

{
    "path": "/v1/nested/payment/123456789-afekl-13456/",
    "resource": "/payment/{invoice+}",
    "requestContext": {
        "resource": "/payment/{invoice+}",
        "path": "/v1/nested/payment/123456789-afekl-13456/",
        "httpMethod": "GET",
        "requestContext": {
            "resourceId": "8em8dt",
            "resourcePath": "/payment/{invoice+}",
            "httpMethod": "GET",
            "path": "/v1/nested/payment/123456789-afekl-13456/",
            "stage": "default",
            "domainPrefix": "api",
            "domainName": "api.serverlessa.dev",
        }
    }
}
```

heitorlessa:
@michaelbrewer lemme know what types of events you want me to generate to help with tests. I can also help write these too.

These events above are actual events coming from API GW hence why my surprise to `requestContext.path` idea not being an actual solution after testing - I trimmed it down to easily focus on the necessary bits.

Your initial solution on `self._prefix` and such continues to work, it just needs to be a list of str. We can figure out a better name for the env var later

michaelbrewer:
@heitorlessa  - i am sure with more test cases we can come up with a cleaner solutions. Test cases wise:

#### Rest API

This is the problematic one when mapping the routes against the path, as custom mappings or a direct mapping via Route 53 gives different outcomes.

1) Test directly via stage
2) Test via Route 53 mapping
3) Test via 2 different custom mappings

michaelbrewer:
### Http API (V1)

About the same as the Rest API (although, i think are some differences). So similar test cases as above

## HTTP API (V2)

Does not seem to need any changes

heitorlessa:
I'll set them up tomorrow and dump them here

walmsles:
@heitorlessa  and @michaelbrewer Apologies for being quiet for a while.  I think the solution above is workable, I feel it is an edge case (multiple API mappings) but awesome that you have arrived at a solution - let me know if I can assist.

heitorlessa:
Dropping most of the events here except Route53 (need to set that up). One issue I noticed on HTTP API too is that the `stage` is also part of the `rawPath` so this needs changing too.

I created the same API in all three versions (REST, HTTP v1, HTTP v2 payload) using `/payment` and `/payment/{invoice+}` routes. 

---

## REST API

### Direct call using stage

```json
{
    "resource": "/payment",
    "path": "/payment/",
    "httpMethod": "GET",
    "requestContext": {
        "resourcePath": "/payment",
        "path": "/default/payment/",
        "stage": "default",
        "domainPrefix": "qeo6tqt75i",
        "domainName": "qeo6tqt75i.execute-api.eu-west-1.amazonaws.com",
    }
}
```


### Single custom mapping

```json
{
    "path": "/v1/payment",
    "resource": "/payment",
    "httpMethod": "GET",
    "requestContext": {
        "resourcePath": "/payment",
        "path": "/v1/payment",
        "stage": "default",
        "domainPrefix": "api",
        "domainName": "api.serverlessa.dev",
    },
}
```


### Nested custom mapping


```json
{
    "path": "/v1/nested/payment/123456789-afekl-13456/",
    "resource": "/payment/{invoice+}",
    "httpMethod": "GET",
    "requestContext": {
        "resourcePath": "/payment/{invoice+}",
        "path": "/v1/nested/payment/123456789-afekl-13456/",
        "stage": "default",
        "domainName": "api.serverlessa.dev",
    }
}
```

### Route 53

TBC


## HTTP API v1

### Direct call using stage

```json
{
    "resource": "/payment",
    "path": "/default/payment",
    "httpMethod": "GET",
    "requestContext": {
        "resourcePath": "/payment",
        "path": "/default/payment",
        "stage": "default",
        "domainPrefix": "yx3odp7ks2",
        "domainName": "yx3odp7ks2.execute-api.eu-west-1.amazonaws.com"
    }
}
```

### Single custom mapping

```json
{
    "resource": "/payment/{invoice+}",
    "path": "/v2/payment/",
    "httpMethod": "GET",
    "requestContext": {
        "resourcePath": "/payment/{invoice+}",
        "path": "/default/payment/",
        "stage": "default",
        "domainName": "api.serverlessa.dev",
        "domainPrefix": "api"
    }
}
```


### Nested custom mapping

```json
{
    "resource": "/payment/{invoice+}",
    "path": "/v2/payment/123456789-afekl-13456",
    "httpMethod": "GET",
    "requestContext": {
        "resourcePath": "/payment/{invoice+}",
        "path": "/default/payment/123456789-afekl-13456",
        "stage": "default",
        "domainName": "api.serverlessa.dev",
        "domainPrefix": "api"
    }
}
```

### Route53

TBC


## HTTP API v2

### Direct call using stage

```json
{
    "message": {
        "routeKey": "ANY /payment",
        "rawPath": "/default/payment",
        "requestContext": {
          "apiId": "swgeupeic9",
          "domainName": "swgeupeic9.execute-api.eu-west-1.amazonaws.com",
          "domainPrefix": "swgeupeic9",
          "http": {
            "method": "GET",
            "path": "/default/payment"
          },
          "routeKey": "ANY /payment",
          "stage": "default"
        }
      }
}
```


### Single custom mapping

```json
{
    "routeKey": "ANY /payment",
    "rawPath": "/default/payment",
    "requestContext": {
        "domainName": "api.serverlessa.dev",
        "domainPrefix": "api",
        "http": {
            "method": "GET",
            "path": "/default/payment"
        },
        "routeKey": "ANY /payment",
        "stage": "default"
    }
}
```

### Nested custom mapping

```json
{
    "message": {
      "routeKey": "ANY /payment/{invoice+}",
      "rawPath": "/default/payment/123456789-afekl-13456",
      "requestContext": {
        "domainName": "api.serverlessa.dev",
        "domainPrefix": "api",
        "http": {
          "method": "GET",
          "path": "/default/payment/123456789-afekl-13456"
        },
        "routeKey": "ANY /payment/{invoice+}",
        "stage": "default"
      }
    }
  }
```

### Route53

TBC

heitorlessa:
Just added Route 53 event to complete the list of scenarios @michaelbrewer lemme know if there's any other data missing

michaelbrewer:
@heitorlessa - ok i have added support for using a list of prefixes. also it will auto add the last `/`.

ie: path `/fooBar/status` with strip prefix `/foo` should not match. but `/fooBar` would.

heitorlessa:
Great, makes sense. Could you update the PR description using a list of strings instead of the env var?

We also don't know the env var values based on the description

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -533,6 +540,15 @@ def _resolve(self) -> ResponseBuilder:
         logger.debug(f"No match found for path {path} and method {method}")
         return self._not_found(method)
 
+    def _remove_prefix(self, path: str) -> str:
+        """Remove the configured prefix from the path"""
+        if self._strip_prefixes:
```
heitorlessa, line 545: that's gonna fail if someone passes a `""` or some other type

heitorlessa: correction, `[""]` works because of the way you're handling the slicing. It wouldn't if another type is sent, for example:

app = ApiGatewayResolver(strip_prefixes=[os.getenv("ENV_VAR_I_DIDNT_SET")]) # [None]

heitorlessa (changes requested):

aws_lambda_powertools/event_handler/api_gateway.py
michaelbrewer: ok i added many different invalid variations :)

heitorlessa:
LGTM - @walmsles are you happy with it? Or do we must support environment variable too?

If we do need an environment variable, we would need the values to be as comma-separated values: "`/v1, /v1/payment, /v1/order`"

If not, then we can merge it and move to a few last improvements and prepare a release

walmsles:
@heitorlessa I think no env var support is good - means AppConfig or Parameters can be used to dynamically control the prefix configuration which IMO is cleaner and what I would look to do to achieve the outcome from this feature.  Working out complex Env Var values on LAMBDA is hard!

heitorlessa:
> @heitorlessa I think no env var support is good - means AppConfig or Parameters can be used to dynamically control the prefix configuration which IMO is cleaner and what I would look to do to achieve the outcome from this feature. Working out complex Env Var values on LAMBDA is hard!

Amazing - merging it then ;) Thanks for the help flagging it @walmsles, and obviously a huge help as always to @michaelbrewer

heitorlessa (approved):

walmsles:
Love your work @heitorlessa @michaelbrewer - thankyou!

Issue #572, opened 2021-07-29
Enable tracing only when running on AWS Lambda


Dunedan:
**Is your feature request related to a problem? Please describe.**

When running code related to an AWS Lambda function outside of AWS Lambda (e.g. when running unit tests or when re-using a function for something else like a CLI script),  `Tracer` with its default configuration fails, as it expects an open X-Ray segment, which usually isn't available in these cases.

That default behavior requires explicitly disabling tracing when not running on AWS Lambda, which is quite annoying and AWS Lambda Powertools could do a better job to simplify things for its users there.

This problem got discussed in detail in #515.

**Describe the solution you'd like**

I'd like to see a default behavior where tracing is only enabled if the code is running on AWS Lambda.

**Describe alternatives you've considered**

An alternative to enable tracing only on AWS Lambda could be to let `Tracer` create a segment if there isn't already one, however I'm not sure about possible side effects.

heitorlessa:
Thanks a lot Daniel! We’ll work on it

michaelbrewer:
@Dunedan - might be fixed in this PR (https://github.com/awslabs/aws-lambda-powertools-python/pull/598)

heitorlessa:
Now released as part of 1.19.0: https://github.com/awslabs/aws-lambda-powertools-python/releases/tag/v1.19.0

PR #598, opened 2021-08-10
feat(tracer): disable tracer when for non-Lambda envs

michaelbrewer:
**Issue #, if available:**

- #572

## Description of changes:

Disables tracer when not running as a lambda ie: `LAMBDA_TASK_ROOT` env is not set

Changes:
- If `LAMBDA_TASK_ROOT` is not set then we assume tracer should be disabled
- Remove `SAM_LOCAL_ENV` and `CHALICE_LOCAL_ENV` checks
- Update tests

**Checklist**

 

* [x] Update tests
* [ ] Update docs


michaelbrewer:
@heitorlessa  - I am not sure if we need to modify the docs for this. But it should help simplify some people's life.

heitorlessa (approved):

PR #610, opened 2021-08-14
feat(data-classes): AppSync Lambda authorizer event

michaelbrewer:
**Issue #, if available:**
- Related to https://github.com/awslabs/aws-lambda-powertools-python/issues/543

## Description of changes:

Change:
- [X] Add test fixtures for request and response
- [X] Add Request event source
- [X] Add correlation_id `APPSYNC_AUTHORIZER`
- [x] Add Response helper
- [x] Add docs

Example usage:

```python3
from typing import Dict

from aws_lambda_powertools.logging.logger import Logger
from aws_lambda_powertools.utilities.data_classes.appsync_authorizer_event import (
    AppSyncAuthorizerEvent,
    AppSyncAuthorizerResponse,
)
from aws_lambda_powertools.utilities.data_classes.event_source import event_source

logger = Logger()


def get_user_by_token(token: str):
    """Look a user by token"""


@logger.inject_lambda_context(correlation_id_path=correlation_paths.APPSYNC_AUTHORIZER)
@event_source(data_class=AppSyncAuthorizerEvent)
def lambda_handler(event: AppSyncAuthorizerEvent, context) -> Dict:
    user = get_user_by_token(event.authorization_token)

    if not user:
        # No user found, return not authorized
        return AppSyncAuthorizerResponse().to_dict()

    return AppSyncAuthorizerResponse(
        authorize=True,
        resolver_context={"id": user.id},
        # Only allow admins to delete events
        deny_fields=None if user.is_admin else ["Mutation.deleteEvent"],
    ).asdict()
```

**Checklist**

 

* [x] Update tests
* [x] Update docs


michaelbrewer:
@heitorlessa - i am not sure if fluent apis is a thing in Python? Maybe just optionals?

gwlester:
How does this tie in with 
<div class="tabbed-set" data-tabs="3:1" style="box-sizing: inherit; border-radius: 0.1rem; display: flex; flex-wrap: wrap; margin: 1em 0px; position: relative; color: rgba(0, 0, 0, 0.87); font-family: Roboto, -apple-system, BlinkMacSystemFont, Helvetica, Arial, sans-serif; font-size: 17.6px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div class="tabbed-content" style="box-sizing: inherit; box-shadow: 0 -0.05rem var(--md-default-fg-color--lightest); display: block; order: 99; width: 918.203px;">

app.resolve(event, context)
--


</div></div><br class="Apple-interchange-newline">

michaelbrewer:
> How does this tie in with 
> 
> <div class="tabbed-set" data-tabs="3:1" style="box-sizing: inherit; border-radius: 0.1rem; display: flex; flex-wrap: wrap; margin: 1em 0px; position: relative; color: rgba(0, 0, 0, 0.87); font-family: Roboto, -apple-system, BlinkMacSystemFont, Helvetica, Arial, sans-serif; font-size: 17.6px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial;"><div class="tabbed-content" style="box-sizing: inherit; box-shadow: 0 -0.05rem var(--md-default-fg-color--lightest); display: block; order: 99; width: 918.203px;">
> 
> 
> 
> app.resolve(event, context)
> 
> --
> 
> 
> 
> 
> 
> </div></div><br class="Apple-interchange-newline">


@gwlester - this applies to AppSync only. API gateway use case would be based on how this is implemented.

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+
+from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
+
+
+class AppSyncAuthorizerEventRequestContext(DictWrapper):
+    """Request context"""
+
+    @property
+    def api_id(self) -> str:
+        """AppSync api id"""
```
heitorlessa, line 11:
```suggestion
        """AppSync API ID"""
```

Consistency

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+        return self["requestContext"]["accountId"]
+
+    @property
+    def request_id(self) -> str:
+        """Requestt ID"""
+        return self["requestContext"]["requestId"]
+
+    @property
+    def query_string(self) -> str:
+        """Graphql query string"""
```
heitorlessa, line 26:
```suggestion
        """GraphQL query string"""
```

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+        return self["requestContext"]["queryString"]
+
+    @property
+    def operation_name(self) -> Optional[str]:
+        """Graphql operation name, optional"""
```
heitorlessa, line 31:
```suggestion
        """GraphQL operation name, optional"""
```

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+        return self["requestContext"].get("operationName")
+
+    @property
+    def variables(self) -> Dict:
+        """Graphql variables"""
```
heitorlessa, line 36:
```suggestion
        """GraphQL variables"""
```

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+    resolver_context: Optional[Dict[str, Any]]
+        A JSON object visible as `$ctx.identity.resolverContext` in resolver templates
+        Warning: The total size of this JSON object must not exceed 5MB.
+    denied_fields: Optional[List[str]]
+        A list of which are forcibly changed to null, even if a value was returned from a resolver.
+        Each item is either a fully qualified field ARN in the form of
+        `arn:aws:appsync:us-east-1:111122223333:apis/GraphQLApiId/types/TypeName/fields/FieldName`
+        or a short form of TypeName.FieldName. The full ARN form should be used when two APIs
+        share a lambda function authorizer and there might be ambiguity between common types
+        and fields between the two APIs.
```
heitorlessa, line 84:
```suggestion
        A list of fields that will be set to `null` regardless of the resolver's return.
        
        A field is either `TypeName.FieldName`, or an ARN such as
        `arn:aws:appsync:us-east-1:111122223333:apis/GraphQLApiId/types/TypeName/fields/FieldName`
        
        Use the full ARN for correctness when sharing a Lambda function authorizer between APIs.
```

Made it slightly more succinct to ease parsing in the IDE

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    def __init__(
+        self,
+        authorize: bool = False,
+        ttl: Optional[int] = None,
```
heitorlessa, line 90:
```suggestion
        max_age: Optional[int] = 300,
```

This will save us another if block.

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+        resolver_context: Optional[Dict[str, Any]] = None,
+        denied_fields: Optional[List[str]] = None,
+    ):
+        self._data: Dict = {"isAuthorized": authorize}
+
+        if ttl is not None:
+            self._data["ttlOverride"] = ttl
```
heitorlessa, line 97:
```suggestion
        self._data: Dict = {"isAuthorized": authorize, "ttlOverride": ttl}
```

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if denied_fields:
+            self._data["deniedFields"] = denied_fields
```
heitorlessa, line 100: what happens if we return an empty list to AppSync? That way we can set a default upfront at the class initialization, then construct `self._data` only once.

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+
+        if resolver_context:
+            self._data["resolverContext"] = resolver_context
```
heitorlessa, line 103: Docs says JSON Object, does it work if we pass `{"something": True}`? If not, we should be serializing it.

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def to_dict(self) -> dict:
+        """Return the response as a dict"""
+        return self._data
```
heitorlessa, line 107: Because we're mimicking `dataclasses` API, let's move the building of the dictionary here instead of the constructor.

This will also allow us to correctly instantiate the class as `Instance.authorize` doesn't work, for example. 


```python
    def __init__(
        self,
        authorize: bool = False,
        max_age: int = 300,
        resolver_context: Optional[Dict[str, Any]] = None,
        deny_fields: Optional[List[str]] = None,
    ):
        self.authorize = authorize
    	self.max_age = max_age
        self.resolver_context = resolver_context or {}
        self.deny_fields = deny_fields or []
        
   def to_dict(self) -> dict:
       """Return the response as dict"""
       return {
           "isAuthorized": self.authorize,
           "resolverContext": json.dumps(self.resolver_context),
           "ttlOverride": self.max_age,
           "deniedFields": self.deny_fields
       }
```

heitorlessa (changes requested):
Looks good! Some minor changes. Should we have a method to ease adding a fully qualified field name?

Because we're mimicking the data classes, I think we can set some good defaults, correct the constructor, and simply get rid of `self._data`.

If we do add a method to add a fully qualified field, we can make a nice experience in the AppSync Handler since we would know both API ID and Region to construct the ARN while also making it standalone ;)

```python
    def __init__(
        self,
        authorize: bool = False,
        max_age: int = 300,
        resolver_context: Optional[Dict[str, Any]] = None,
        deny_fields: Optional[List[str]] = None,
    ):
        self.authorize = authorize
    	self.max_age = max_age
        self.resolver_context = resolver_context or {}
        self.deny_fields = deny_fields or []
        
   def to_dict(self) -> dict:
       """Return the response as dict"""
       return {
           "isAuthorized": self.authorize,
           "resolverContext": json.dumps(self.resolver_context),
           "ttlOverride": self.max_age,
           "deniedFields": self.deny_fields
       }

```

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
```diff
@@ -0,0 +1,107 @@
...
+    """AppSync Lambda authorizer response helper
+
+    Parameters
+    ----------
+    authorize: bool
+        authorize is a boolean value indicating if the value in authorizationToken
+        is authorized to make calls to the GraphQL API. If this value is
+        true, execution of the GraphQL API continues. If this value is false,
+        an UnauthorizedException is raised
+    ttl: Optional[int]
```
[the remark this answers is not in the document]
michaelbrewer: it is an optional override so i am calling it `ttl_override` to remove confusion

aws_lambda_powertools/utilities/data_classes/appsync_authorizer_event.py
michaelbrewer: I will test this.

michaelbrewer: @heitorlessa 

`{"something": True}` does work.

However what does not work, no matter how you encode it
`{"something": [True]}`
`{"something": None}`
`{"something": {"else": "Foo"}}`

I did update the docs to include that nested values are not supported

heitorlessa (approved):

PR #612, opened 2021-08-15
chore(api-docs): enable allow_reuse to fix the docs

michaelbrewer:
**Issue #, if available:**

- https://github.com/awslabs/aws-lambda-powertools-python/issues/611

## Description of changes:

Set `allow_reuse=True` for all of the pydantic validators, so the docs can render.

**Checklist**

 

* [x] Update tests
* [x] Update docs


heitorlessa:
Just hit the issue - it happens only when I open `utilities`.

heitorlessa:
turns out it's due to the import system as Pydantic somehow believes it's a duplicate. Let's keep as is, if it happens again then we take time to investigate down the stack.

PR #625, opened 2021-08-20
feat(idempotency): support for any synchronous function

heitorlessa:
**Issue #, if available:** https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/32

## Description of changes:

* [x] Remove dependency on context
* [x] Use `data` instead of `event` for params
* [x] Make IdempotencyHandler generic for any function
* [x] Update docstring that refers to Lambda event only
* [x] New decorator that takes a data_argument
* [x] Update docs with new UX for any function
* [x] Preserve decorated function argument types w/ generic callable

## UX

```python
from aws_lambda_powertools.utilities.batch import sqs_batch_processor
from aws_lambda_powertools.utilities.idempotency import idempotent_function, DynamoDBPersistenceLayer, IdempotencyConfig


dynamodb = DynamoDBPersistenceLayer(table_name="idem")
config =  IdempotencyConfig(
    event_key_jmespath="messageId",
    use_local_cache=True,
)


@idempotent_function(data_keyword_argument="record", persistence_layer=dynamodb, config=config)
def record_handler(record):
    return do_something_with(record["body"])


@sqs_batch_processor(record_handler=record_handler)
    return {"statusCode": 200}
```

**Checklist**

 

* [x] Update tests
* [x] Update docs


heitorlessa:
@michaelbrewer @cakepietoast could either of you have a look just in case? tks!

aws_lambda_powertools/utilities/idempotency/base.py
```diff
@@ -0,0 +1,181 @@
...
+class IdempotencyHandler:
+    """
+    Base class to orchestrate calls to persistence layer.
+    """
+
+    def __init__(
+        self,
+        function: Callable,
+        function_payload: Any,
+        idempotency_config: IdempotencyConfig,
```
michaelbrewer, line 33: @heitorlessa  sometimes this is `config` (existing code) and sometimes it is `idempotentcy_config`. Should we consistent? Or is this a matter of a better style?

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -66,174 +57,88 @@ def idempotent(
...
-                # Allow the exception to bubble up after max retries exceeded
-                raise
+    return idempotency_handler.handle()
 
 
-class IdempotencyHandler:
+def idempotent_function(
+    function: Optional[AnyCallableT] = None,
+    *,
+    data_keyword_argument: str,
```
michaelbrewer, line 76: @heitorlessa  - i added `data_keyword_argument` is not part of `ImdempotencyConfig` as it only applies to function decorators? Although even a lambda handler (via a decorator have additional parameters)

michaelbrewer (commented):
looking good.

aws_lambda_powertools/utilities/idempotency/idempotency.py
heitorlessa: yes, that's right @michaelbrewer. Because Lambda has a known signature where the `data_keyword_argument` would always be event, so we don't need it.

aws_lambda_powertools/utilities/idempotency/base.py
heitorlessa: Good catch. I think this is a left over from the many refactoring attempts I had at it, lemme correct it to `config`

PR #643, opened 2021-08-22
refactor(data-classes): clean up internal logic for APIGatewayAuthorizerResponse

michaelbrewer:
**Issue #, if available:**

## Description of changes:

> **NOTE:** We might want to explain "This will also pollute Lambda metrics with errors, also causing cold starts since it's not handled." in the docs to.

Changes:
- Clean up the internal logic for `APIGatewayAuthorizerResponse` 
- Update the internal docs to include a new `DENY_ALL_RESPONSE`
- Add missing property `usageIdentifierKey` for Rest API responses only
- Add `from_route_arn` to build APIGatewayAuthorizerResponse from an `arn`
- Add a relaxed pylintrc configuration for those use have this installed

### Update Example

```python
from aws_lambda_powertools.utilities.data_classes.api_gateway_authorizer_event import (
    DENY_ALL_RESPONSE,
    APIGatewayAuthorizerRequestEvent,
    APIGatewayAuthorizerResponse,
    HttpVerb,
)
from secrets import compare_digest


def get_user_by_token(token):
    if compare_digest(token, "admin-foo"):
        return {"id": 0, "name": "Admin", "isAdmin": True}
    elif compare_digest(token, "regular-foo"):
        return {"id": 1, "name": "Joe"}
    else:
        return None


    user = get_user_by_token(event.get_header_value("Authorization"))

    if user is None:
        # No user was found
        # to return 401 - `{"message":"Unauthorized"}`, but pollutes lambda metrics
        # raise Exception("Unauthorized")
        # to return 403 - `{"message":"Forbidden"}`
        return DENY_ALL_RESPONSE

    # Create the response builder from parts of the `methodArn`
    # and set the logged in user id and context
    policy = APIGatewayAuthorizerResponse.from_route_arn(
        arn=event.method_arn, principal_id=user["id"], context=user
    )

    # Conditional IAM Policy
    if user.get("isAdmin", False):
        policy.allow_all_routes()
    else:
        policy.allow_route(HttpVerb.GET, "/user-profile")

    return policy.asdict()
```

**Checklist**

 

* [x] Update tests
* [x] Update docs

-----
[View rendered docs/utilities/data_classes.md](https://github.com/gyft/aws-lambda-powertools-python/blob/refactor/api-authorizer/docs/utilities/data_classes.md)

### DIFF
--- a/docs/utilities/data_classes.md
+++ b/docs/utilities/data_classes.md
@@ -96,18 +96,15 @@ Use **`APIGatewayAuthorizerRequestEvent`** for type `REQUEST` and **`APIGatewayA
 
     When the user is found, it includes the user details in the request context that will be available to the back-end, and returns a full access policy for admin users.
 
-    ```python hl_lines="2-5 29 36-42 47 49"
+    ```python hl_lines="2-6 29 36-42 47 49"
     from aws_lambda_powertools.utilities.data_classes import event_source
     from aws_lambda_powertools.utilities.data_classes.api_gateway_authorizer_event import (
+        DENY_ALL_RESPONSE,
         APIGatewayAuthorizerRequestEvent,
         APIGatewayAuthorizerResponse,
         HttpVerb,
     )
     from secrets import compare_digest
-
-
-    class UnAuthorizedError(Exception):
-        ...
 
 
     def get_user_by_token(token):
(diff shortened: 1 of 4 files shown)


docs/utilities/data_classes.md
```diff
@@ -119,25 +123,24 @@ Use **`APIGatewayAuthorizerRequestEvent`** for type `REQUEST` and **`APIGatewayA
     def handler(event: APIGatewayAuthorizerRequestEvent, context):
         user = get_user_by_token(event.get_header_value("Authorization"))
 
+        if user is None:
+            # No user was found, so we raised a not authorized error
+            raise UnAuthorizedError("Not authorized to perform this action")
```
heitorlessa, line 128: We need to test this with API GW. I don't think API GW is smart enough to recognise and return 403 - it'll throw a 500 back IIRC.

heitorlessa (changes requested):

docs/utilities/data_classes.md
heitorlessa: This will also pollute Lambda metrics with errors, also causing cold starts since it's not handled.

We can make this work with event handler by catching it but not here I'm afraid

michaelbrewer:
@heitorlessa - i addressed your feedback in this PR.

heitorlessa (approved):

Issue #646, opened 2021-08-23
ApiGatewayResolver root routes and strip_prefixes


BVMiko:
I ran into an issue when using a route with the rule `"/"` along with `strip_prefixes` feature.  This rule is unable to trigger due to the `path` variable being reduced down to an empty string instead of a forward slash. 

```python
from aws_lambda_powertools.event_handler.api_gateway import ApiGatewayResolver
app = ApiGatewayResolver(strip_prefixes=["/some-path"])
@app.get('/')
def my_func():
    return {'execution_status':'success'}
```

## Expected Behavior
Accessing `domain.com/some-path` would execute `my_func()`.

## Current Behavior
Accessing `domain.com/some-path` returns a 404.  Of note though, accessing `domain.com/some-path/` (note the trailing slash) *will* execute `my_func()`.

## Possible Solution
Add an conditional [here](https://github.com/awslabs/aws-lambda-powertools-python/blob/c8cf3ba2020e034e48543f2295a1aaf927d7177b/aws_lambda_powertools/event_handler/api_gateway.py#L549-L550) which will map an empty string to `"/"`.

## Steps to Reproduce (for bugs)
1. Set up the code example above
2. Set up ApiGateway to use a custom domain, and bind the Lambda to a custom path of `some-path`
3. Visit `https://CUSTOMDOMAIN/some-path` and `https://CUSTOMDOMAIN/some-path/` and note the different results

michaelbrewer:
Nice one @BVMiko , i put in a review comment for adding a test case. 

Otherwise, i was thinking whether it should be `""` instead of `"/"`? Then this would work when the request is `/some-path`

```python
from aws_lambda_powertools.event_handler.api_gateway import ApiGatewayResolver
app = ApiGatewayResolver(strip_prefixes=["/some-path"])
@app.get("")
def my_func():
    return {'execution_status':'success'}
```

As the original request path of `/some-path` does different from `/some-path/`. Although `@app.get("")` does like strange.

BVMiko:
[...]

PR #647, opened 2021-08-23
fix: Fix issue with strip_prefixes

BVMiko:
**Issue #, if available: #646**

## Description of changes:

After removing path prefixes, allow an empty path to return `"/"` so that it will match a normal routing rule.

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -546,6 +546,8 @@ class ApiGatewayResolver:
             return path
 
         for prefix in self._strip_prefixes:
+            if path == prefix:
+                return "/"
             if self._path_starts_with(path, prefix):
                 return path[len(prefix) :]
 
--- a/docs/core/event_handler/api_gateway.md
+++ b/docs/core/event_handler/api_gateway.md
@@ -476,6 +476,8 @@ This will lead to a HTTP 404 despite having your Lambda configured correctly. Se
     }
     ```
 
+Note: After removing a path prefix with `strip_prefixes`, the new root path will automatically be mapped to the path argument of `/`. For example, when using `strip_prefixes` value of `/pay`, there is no difference between a request path of `/pay` and `/pay/`; and the path argument would be defined as `/`.
+
 ## Advanced
 
 ### CORS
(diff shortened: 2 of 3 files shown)


michaelbrewer:
@BVMiko -  Nice catch!

Could you add a test case?  Something like:

```python
def test_api_gateway_request_path_equals_strip_prefix():
    # GIVEN a strip_prefix matches the request path
    app = ApiGatewayResolver(strip_prefixes=["/foo"])
    event = {"httpMethod": "GET", "path": "/foo"}

    @app.get("/")
    def base():
        return {}

    # WHEN calling the event handler
    # WITH a route "/"
    result = app(event, {})

    # THEN process event correctly
    assert result["headers"]["Content-Type"] == content_types.APPLICATION_JSON
```

Regarding for the fix, i guess it could also default to `""` over `"/"`, but i would defer that to @heitorlessa

michaelbrewer:
@BVMiko - FYI, Heitor will be about for 5 weeks. How critical would a fix for this be?

BVMiko:
For now I have a workaround in place which can remain until this issue is resolved:
```python
app = ApiGatewayResolver(strip_prefixes=["/foo"])

@app.get("/")
def base():
    return {}

    if event['path'] == '/foo':
        event['path'] = '/foo/'
```

michaelbrewer:
> @BVMiko - Nice catch!
> 
> Could you add a test case? Something like:
> 
> ```python
> def test_api_gateway_request_path_equals_strip_prefix():
>     # GIVEN a strip_prefix matches the request path
>     app = ApiGatewayResolver(strip_prefixes=["/foo"])
>     event = {"httpMethod": "GET", "path": "/foo"}
> 
>     @app.get("/")
>     def base():
>         return {}
> 
>     # WHEN calling the event handler
>     # WITH a route "/"
>     result = app(event, {})
> 
>     # THEN process event correctly
>     assert result["statusCode"] == 200
>     assert result["headers"]["Content-Type"] == content_types.APPLICATION_JSON
> ```
> 
> Regarding for the fix, i guess it could also default to `""` over `"/"`, but i would defer that to @heitorlessa

@BVMiko - would you be able to include a test like this into the api gateway handler test suite?

BVMiko:
@michaelbrewer: The test has been added

michaelbrewer (approved):
LGTM thanks @BVMiko

Issue #714, opened 2021-09-30
Feature Flag only evaluates first rule in a feature


gwlester:
**What were you trying to accomplish?**
Having a feature flag with multiple rules.

## Expected Behavior

Rules should be evaluated for a feature until one is true or all rules have been evaluated.  It appears that test test_flags_conditions_rule_match_multiple_actions_multiple_rules_multiple_conditions does not work as expected.

## Current Behavior

The default is returned if the first rule is false.

## Possible Solution

I will submit a pull request to correct the code.
I have not debugged the test.

## Steps to Reproduce (for bugs)

1. See test_flags_conditions_rule_match_multiple_actions_multiple_rules_multiple_conditions
2.

## Environment

* **Powertools version used**: 1.20.2
* **Packaging format (Layers, PyPi)**: Layers
* **AWS Lambda function runtime:** 3.8
* **Debugging logs**

> [How to enable debug mode](https://awslabs.github.io/aws-lambda-powertools-python/#debug-mode)**

```python
# paste logs here
```

gwlester:
@heitorlessa -- I think this should have the label: **status/staged-next-release**

heitorlessa:
This is now available as part of today's 1.21.0 release -- Thank you for the contribution, credits given at the top ;)

https://github.com/awslabs/aws-lambda-powertools-python/releases/tag/v1.21.0

PR #715, opened 2021-09-30
fix(feature_flags): Fix for but 714 -- only evaluting first rule

gwlester:
**Issue #, if available:** #714

## Description of changes:

Corrected indentation in evaluate rules so that it will only return the default after all rules are evaluated.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


heitorlessa:
cc @risenberg-cyberark I believe that's the expected behaviour in the original schema, no?

from the docs:

> You can have multiple rules with different names. **The rule engine will return the first result when_match of the matching rule configuration**, or default value when none of the rules apply.

heitorlessa:
I'd take that back - Another part of the docs also says:

> **For multiple conditions**, we will evaluate the **list of conditions as a logical AND**, so all conditions needs to match to return when_match value.

I'd defer to @risenberg-cyberark who did the original schema to confirm before proceeding here, as we'll need remove ambiguity from docs too thereafter.

risenberg-cyberark:
I remembered i solved a similar bug in the conditions about 1.5 months ago with the same indendation in the conditions checks: _evaluate_conditions. 
Yep, it's the same silly indentation issue.
The docs seem ok to me.
Anyways, thx for the fix. It's correct.

risenberg-cyberark:
I think the pr should also fix a test for multiple rules where the second rule matches, not the first.

heitorlessa:
> I remembered i solved a similar bug in the conditions about 1.5 months ago with the same indendation in the conditions checks: _evaluate_conditions. Yep, it's the same silly indentation issue. The docs seem ok to me. Anyways, thx for the fix. It's correct.

I remembered that but wasn't sure if was fixed by pushing identation or pulling it. Either way, if you have the bandwidth for fixing the test and merge conflict by Monday, I can include it in Tue/Wed release as it's an important one

gwlester:
> > I remembered i solved a similar bug in the conditions about 1.5 months ago with the same indentation in the conditions checks: _evaluate_conditions. Yep, it's the same silly indentation issue. The docs seem ok to me. Anyways, thx for the fix. It's correct.
> 
> I remembered that but wasn't sure if it was fixed by pushing indentation or pulling it. Either way, if you have the bandwidth for fixing the test and merge conflict by Monday, I can include it in Tue/Wed release as it's an important one

The code I pushed allows multi-rules to work as well as multi-conditions.  
multi-conditions are **and**.  multi-rules are **or**

Personally, I think this is critical.

risenberg-cyberark:
> > > I remembered i solved a similar bug in the conditions about 1.5 months ago with the same indentation in the conditions checks: _evaluate_conditions. Yep, it's the same silly indentation issue. The docs seem ok to me. Anyways, thx for the fix. It's correct.
> > 
> > 
> > I remembered that but wasn't sure if it was fixed by pushing indentation or pulling it. Either way, if you have the bandwidth for fixing the test and merge conflict by Monday, I can include it in Tue/Wed release as it's an important one
> 
> The code I pushed allows multi-rules to work as well as multi-conditions. multi-conditions are **and**. multi-rules are **or**
> 
> Personally, I think this is critical.

I agree

risenberg-cyberark:
I'm on it

gwlester:
Ok, a dumb question -- the coverage report says this decreased the coverage by 0.04%-- but the only file touched (...owertools/utilities/feature_flags/feature_flags.py) is at 100% coverage!

What gives?

gwlester:
> I think the pr should also fix a test for multiple rules where the second rule matches, not the first.

Unless I'm misreading it, that is what **test_flags_conditions_rule_match_multiple_actions_multiple_rules_multiple_conditions** is supposed to do with:
```python
    # match second rule
    toggle = feature_flags.evaluate(name="my_feature", context={"tenant_id": "4446", "username": "az"}, default=False)
```

I have no idea how this test was passing before.

risenberg-cyberark:
it was working because the match value of the rule was the same as the default value of the feature.
When i change expected_value_second_check to be True and run the test. it fails, because it doesnt match on the second rule.
After your fix, the test works with expected_value_second_check = True or False, because it returns the correct value.
I'm creating a PR. 
see https://github.com/awslabs/aws-lambda-powertools-python/pull/724

risenberg-cyberark:
@gwlester  I think you pushed the commits to the wrong branch

gwlester:
> @gwlester I think you pushed the commits to the wrong branch
No, pulled the changes into here when attempting to resolve conflicts -- so we close this PR and just go with yours since it fixes the test as well as the code, right?

risenberg-cyberark:
> > @gwlester I think you pushed the commits to the wrong branch
> > No, pulled the changes into here when attempting to resolve conflicts -- so we close this PR and just go with yours since it fixes the test as well as the code, right?

indeed.
I meant that you added the other new actions here.

PR #724, opened 2021-10-01
fix(feature-flags): rules should evaluate with an AND op

risenberg-cyberark:
**Issue #, if available:**

## Description of changes:

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


risenberg-cyberark:
The test i changed  was working because the match value of the rule was the same as the default value of the feature.
When i change expected_value_second_check to be True and run the test it fails, because it doesnt match on the second rule.
After the fix, the test works with expected_value_second_check = True or False, because it returns the correct value.

heitorlessa (approved):

heitorlessa:
Thank you and specially @gwlester for catching it

Issue #694, opened 2021-09-23
DynamoDBPersistenceLayer should support Tables with Range Keys


Tankanow:
**Is your feature request related to a problem? Please describe.**

I cannot use the `DynamoDBPersistenceLayer` class when my table has HASH and RANGE keys. I do not want to create a separate table for tracking lambda idempotence. [channeling my inner Houlihan] One table to rule them all!

**Describe the solution you'd like**

I would like the `DynamoDBPersistenceLayer` to accept a configuration such that the HASH key is set to a known value, e.g. `idempotence#<feature-name>`, and the RANGE key contains the current values managed by the `data_record.idempotency_key`.

**Describe alternatives you've considered**

There are of course other HASH/RANGE key designs. Another possible design is to split the current `data_record.idempotency_key` value, which is `<function name>#<hash>`, to maybe be HASH key = `<function name>` and RANGE key = `<hash>`.

**Additional context**

I would like this to be a non-breaking change for current consumers. If we agree on the first HASH/RANGE key design proposal, the client could configure a `DynamoDBPersistenceLayer` with these extra parameters:
```
        key_attr_value: str = "idempotence",  # this is just a hardcoded value
        range_attr: str = "SK",                         # this is the range key to store the current `data_record.idempotency_key`
```
If these are set, then `DynamoDBPersistenceLayer` uses the HASH/RANGE key implementation, else it reverts to HASH key implementation.

Tankanow:
I'm happy to submit a PR for this upon API agreement.

heitorlessa:
Hey Adam, thanks a lot for raising this! It was definitely an oversight on
our part not supporting a range key — I’m happy with the first design by
creating a new constructor attribute in the Persistence layer.

I’m returning from vacation next week and can help review and provide any
guidance if you need, if that’s okay.

Your PR is super super appreciated ;-)

Tankanow:
@heitorlessa, wanted to keep you in the loop. I'll have a PR this week.

cakepietoast:
PR is merged. I'll update the docs in the coming days so we can get this released. Thanks for the contribution @Tankanow!

Tankanow:
Thank you, Tom! And thanks for your patience with this one.

heitorlessa:
awesome, thank you so much @Tankanow for going through all of this ;)

PR #740, opened 2021-10-06
improv: Support a composite primary key in DynamoDBPersistenceLayer

Tankanow:
**Issue #694**

## Description of changes:

Add new optional parameters, `key_attr_value` and `sort_key_attr`, to DynamoDBPersistenceLayer. If `sort_key_attr` is set, then DynamoDBPersistenceLayer will use a composite primary key with the _partition key_ set to a static value, defaults to `powertools#idempotency`, and the _sort key_ set to the hashed idempotency key.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
@@ -22,7 +22,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
         self,
         table_name: str,
         key_attr: str = "id",
-        key_attr_value: str = f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, 'test-func')}",
+        key_attr_value: str = f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, '')}",
         sort_key_attr: Optional[str] = None,
         expiry_attr: str = "expiration",
         status_attr: str = "status",
--- a/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
@@ -22,7 +22,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
         self,
         table_name: str,
         key_attr: str = "id",
-        key_attr_value: str = f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, '')}",
+        static_pk_value: str = f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, '')}",
         sort_key_attr: Optional[str] = None,
         expiry_attr: str = "expiration",
         status_attr: str = "status",
@@ -40,7 +40,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
             Name of the table to use for storing execution records
         key_attr: str, optional
             DynamoDB attribute name for partition key, by default "id"
-        key_attr_value: str, optional
+        static_pk_value: str, optional
             DynamoDB attribute value for partition key, by default "idempotency#<function-name>".
             This will be used if the sort_key_attr is set.
         sort_key_attr: str, optional
@@ -79,7 +79,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
         self._table = None
         self.table_name = table_name
         self.key_attr = key_attr
-        self.key_attr_value = key_attr_value
+        self.static_pk_value = static_pk_value
         self.sort_key_attr = sort_key_attr
         self.expiry_attr = expiry_attr
         self.status_attr = status_attr
@@ -108,7 +108,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
 
     def _get_key(self, idempotency_key: str) -> dict:
         if self.sort_key_attr:
-            return {self.key_attr: self.key_attr_value, self.sort_key_attr: idempotency_key}
+            return {self.key_attr: self.static_pk_value, self.sort_key_attr: idempotency_key}
         return {self.key_attr: idempotency_key}
 
     def _item_to_data_record(self, item: Dict[str, Any]) -> DataRecord:
--- a/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
@@ -20,6 +20,8 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
         self,
         table_name: str,
         key_attr: str = "id",
+        key_attr_value: str = "powertools#idempotency",
+        sort_key_attr: Optional[str] = None,
         expiry_attr: str = "expiration",
         status_attr: str = "status",
         data_attr: str = "data",
@@ -35,7 +37,12 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
         table_name: str
             Name of the table to use for storing execution records
         key_attr: str, optional
-            DynamoDB attribute name for key, by default "id"
+            DynamoDB attribute name for partition key, by default "id"
+        key_attr_value: str, optional
+            DynamoDB attribute value for partition key, by default "powertools#idempotency".
+            This will be used if the sort_key_attr is set.
+        sort_key_attr: str, optional
+            DynamoDB attribute name for the sort key
         expiry_attr: str, optional
             DynamoDB attribute name for expiry timestamp, by default "expiration"
         status_attr: str, optional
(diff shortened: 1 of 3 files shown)


aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -20,6 +20,8 @@ def __init__(
         self,
         table_name: str,
         key_attr: str = "id",
+        key_attr_value: str = "powertools#idempotency",
```
Tankanow, line 23: @heitorlessa, I created this as a Draft PR to get your early feedback on the interface. Once we agree on the ergonomics, I'm happy to update documentation and add examples.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -93,6 +102,11 @@ def table(self, table):
         """
         self._table = table
 
+    def _get_key(self, idempotency_key: str) -> dict:
```
Tankanow, line 105: NOTE: the implementation was straightforward

tests/functional/idempotency/conftest.py
```diff
@@ -2,7 +2,9 @@
 import hashlib
 import json
 from collections import namedtuple
+from dataclasses import dataclass
```
Tankanow, line 5: NOTE: The testing changes were a bit more in-depth. My goal was to parametrize the `persistence_store` fixture, that way every test using that fixture would test with both simple and composite primary keys. All of the other changes were required to support that idea. Overall, I think the code is a _bit_ more complicated, b/c fixtures now return functions in some cases; however, that complication comes with the benefit of fewer code locations of hardcoding the keys.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
Tankanow: @heitorlessa, I didn't see any feedback so I moved this from Draft to Ready. I would still like to update documentation and examples before merging this PR, and would love some feedback on the API before I take the time to do that.

Tankanow:
Sorry for poking Heitor. I wasn't sure if the Draft status of the PR kept
it from your busy queue.

tests/functional/idempotency/conftest.py
heitorlessa: Let's do this in a separate PR so we can focus on the feature at hand - there are more opportunities to clean up and make it easier.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
heitorlessa: Because DynamoDB IAM resource conditions don't allow you to restrict row level access based on SK, where the function name is known, I'd suggest we change the primary key slightly to: `idempotency#function-name`.

If they want, for example, to restrict anyone from ever modifying idempotency tokens for X function that has a **prod** stage or something more specific to their business in the name they now can.

Here are the scenarios I understood by going through the PR, as **C** is new - please correct me if I misunderstood them.

---

## Scenarios

**A) No hash key set.:** Use default hash key and value

```python
from aws_lambda_powertools.utilities.idempotency import DynamoDBPersistenceLayer

# KEY_ATTR="id"
# KEY_ATTR_VALUE="test-func#ec534ff6b4746107270b412333b59f52"
persistence_layer = DynamoDBPersistenceLayer(table_name="IdempotencyTable")

```

**B) Hash key set.:** Use new hash key and default value

```python
# KEY_ATTR="idempotency_key"
# KEY_ATTR_VALUE="test-func#ec534ff6b4746107270b412333b59f52"
persistence_layer = DynamoDBPersistenceLayer(table_name="IdempotencyTable", key_attr="idempotency_key")
```

**C) Hash and Sort key set.:** Use new hash and sort key, and default value as sort key 

```python
# KEY_ATTR="pk"
# SORT_KEY_ATTR="sk"
# KEY_ATTR_VALUE="idempotency#test-func"
# SORT_KEY_ATTR_VALUE="test-func#ec534ff6b4746107270b412333b59f52"
persistence_layer = DynamoDBPersistenceLayer(
    table_name="IdempotencyTable", 
    key_attr="idempotency_key",
    sort_key_attr="idempotency_key")
```

heitorlessa (dismissed):
Two small changes: **1/** Change PK to something IAM can restrict later, **2/** rollback test fixtures changes so we can do in a separate PR.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
Tankanow: @heitorlessa, re the scenarios: yes, C is new. I think there is some copy/pasta in the code, the `key_attr` and `sort_key_attr` should match the comments, as in:

```python
# KEY_ATTR="pk"
# SORT_KEY_ATTR="sk"
# KEY_ATTR_VALUE="idempotency#test-func"
# SORT_KEY_ATTR_VALUE="test-func#ec534ff6b4746107270b412333b59f52"
persistence_layer = DynamoDBPersistenceLayer(
    table_name="IdempotencyTable", 
    key_attr="pk",       # <- changed to match comment
    sort_key_attr="sk")  # <- changed to match comment
```

There is also a related scenario in which the `DynamoDBPersistenceLayer` passes the desired value for `key_attr_value`, e.g.:

```python
# KEY_ATTR="pk"
# SORT_KEY_ATTR="sk"
# KEY_ATTR_VALUE="known-static-partition-key-value"
# SORT_KEY_ATTR_VALUE="test-func#ec534ff6b4746107270b412333b59f52"
persistence_layer = DynamoDBPersistenceLayer(
    table_name="IdempotencyTable", 
    key_attr="pk",
    key_attr_value="known-static-partition-key-value",       
    sort_key_attr="sk")
```

tests/functional/idempotency/conftest.py
Tankanow: Reverted tests for now.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -64,10 +73,14 @@ def __init__(
 
         self._boto_config = boto_config or Config()
         self._boto3_session = boto3_session or boto3.session.Session()
+        if sort_key_attr == key_attr:
```
Tankanow, line 76: NOTE: I snuck in this extra check =).

heitorlessa:
I'll look into this right after merging the new Event Handle Router, so we can merge this later ;)

cakepietoast:
This looks good to me - small nitpicks only! Could we possibly change the name of `key_attr_value` to something more descriptive? Perhaps `static_key_value`? This could just be a symptom of Friday afternoon, but it took me a while to understand from the code that this was just allowing users to set a static value for the partition key when using a composite primary key. 

Also, can we change the variable default to remove the "test-func" part. I understand that code is probably only exercised when testing (outside of a Lambda environment), but it feels wrong to have that in the "real" class. How about letting it be an empty string: 

`f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, '')}",`

Tankanow:
Thanks for you patience @heitorlessa and @cakepietoast. 

I made the `f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, '')}",` change. 

Re `static_key_value` vs `key_attr_value`, I'm not sure the former is more descriptive; I like the latter because it mirrors the `key_attr` argument.

Tankanow:
> > Thanks for you patience @heitorlessa and @cakepietoast.
> > I made the `f"idempotency#{os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, '')}",` change.
> > Re `static_key_value` vs `key_attr_value`, I'm not sure the former is more descriptive; I like the latter because it mirrors the `key_attr` argument.
> 
> Apologies for the late response on this one, I missed your reply. I understand your point that `key_attr_value` looks nice next to `key_attr`, but I have a hard time with the semantics. `key_attr` is the name of attribute to use as the partition key. `partition_key_attr` would have been a much better name, but its hard to change now without breaking compatibility - my mistake there is partly why I'm being a bit persistent on this topic 😄
> 
> "Key attribute value" sounds like we're setting the value of the "key attribute", which to my mind is saying the same thing as setting the "key attribute". What we're actually doing is setting a static value to use for the partition key, which is where my suggestion came from.
> 
> Do you feel any better about `static_pk_value`? If you don't mind changing that we'll get this merged and documented asap.

I like `static_pk_value` better. Making the change ASAP.

cakepietoast:
> I like `static_pk_value` better. Making the change ASAP.

Much appreciated!

Tankanow:
> > I like `static_pk_value` better. Making the change ASAP.
> 
> Much appreciated!

Done. Ready for re-review.

cakepietoast (approved):
LGTM!

PR #757, opened 2021-10-10
refactor(apigateway): Add BaseRouter and duplicate route check

michaelbrewer:
**Issue #, if available:**

## Description of changes:

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -4,7 +4,9 @@ import logging
 import os
 import re
 import traceback
+import warnings
 import zlib
+from abc import ABC, abstractmethod
 from enum import Enum
 from functools import partial, wraps
 from http import HTTPStatus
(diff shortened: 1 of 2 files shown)


heitorlessa (changes requested):
That's a great cleanup, two thoughts:

**1/** Can we make BaseRouter an actual class to merge the routing logic between both ApiGatewayResolver and Router? 

e.g., Router registers & decorates a route, then re-registers & re-decorates again, and do extra checks on cors and such, then finally gets called by the parent. We could do it later too, as we can launch as-is, have a maintenance issue, and tackle them later as a chore item.

**2/** Let's support multiple methods for a given route in the parent so it's aligned and makes it easier to merge logic in the Router later too.

Minors as GH is giving me a hard time annotating in the right LOC:

* [x] `Router.api` -> `Router.routes`
* [x] `method: Union[str, Tuple[str], List[str]],` -> `method: Union[str, Liststr]],`
  * I'd typically use `Sequence[str]` or `Iterable[str]` but it's been deprecated in 3.9; a list of strings suffices
* [x] Docstrings for `include_router` as I'll surely forget the logic a few weeks ahead
* [x] `include_router` has the prefix logic we accidentally had a bug on "/" before, perhaps **1/** can help in case we find another one in the future.

Lemme know your thoughts on **1**, as I think we had chatted about 2 already. Happy with everything else

michaelbrewer:
@heitorlessa - i will start working on the docs PR next

PR #804, opened 2021-11-07
feat(feature_flags): support beyond boolean values (JSON values)

ran-isenberg:
**Issue #, if available:** [#777 ](https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/71)

## Description of changes:

**example schema**

```json
 app_config_schema = {
        "my_feature": {
            "default": [],
            "boolean_type": False,
            "rules": {
                "tenant id equals 345345435": {
                    "when_match": ["value1", "value2"],
                    "conditions": [
                        {
                            "action": "EQUALS",
                            "key": "tenant_id",
                            "value": "345345435",
                        }
                    ],
                }
            },
        }
    }
```

In this case, if you want to use non boolean types, you must supply boolean_type: False, otherwise True is assumed and a validation error is raised because "default" and when_match are not boolean.
If you dont write boolean_type at all, it's assumed as True and class acts as it did prior to this addition-> non breaking change for current users.

### UX

```python
feature_value: List[str] = features.evaluate(name="my_feature", context={"tenant_id": "345345435"}, default=[])
```

```python
feature_value is ["value1", "value2"] because tenant_id = 345345435 and it matches the conditions.
```

```python
feature_value: List[str] = features.evaluate(name="my_feature", context={"tenant_id": "1"}, default=[])
```
```python
feature_value is [] because tenant_id = 1 and it  does NOT match the conditions.
```

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


heitorlessa:
Quick note - As we discussed, I'm happy with the direction. Here's what I'd like to have to better understand to craft a messaging and docs to better communicate this to customers.

* **Provide an example scenario along with schema and return values**
* **Clarify what the return values should be** - Are we returning the value it matches?
* **Add return type primitives instead of ANY**

ran-isenberg:
@heitorlessa  updated description with schema example and usage

heitorlessa:
Thanks a lot @ran-isenberg - As we discussed, I'll work on fixing some bug reports on Parser for the next imminent release, then I'll get back to this PR for having a deeper look so we can find a way to communicate this feature better for customers.

If you could update the diagram, the mentions on the doc that is no longer a boolean only, and a concrete use case we could use to communicate the value that would help prioritize it for the 1.23.0 release (the one after next).

michaelbrewer:
@ran-isenberg @heitorlessa - i see we are missing a test for `not boolean_feature` cases where we skip over?

ran-isenberg:
> @ran-isenberg @heitorlessa - i see we are missing a test for `not boolean_feature` cases where we skip over?

Didnt push all tests, waited for an approval for the idea. I'm working on it ;)

ran-isenberg:
added a bunch of tests

ran-isenberg:
potential uses case:
user permitted actions: for admins users you return a list of actions ['read', 'write']
but default users, feature toggle value is ['read']

```
 app_config_schema = {
        "my_feature": {
            "default": ['read'],
            "boolean_type": False,
            "rules": {
                "username startswith with admin": {
                    "when_match": ["read", "write"],
                    "conditions": [
                        {
                            "action":"STARTSWITH",
                            "key": "user",
                            "value": "admin",
                        }
                    ],
                }
            },
        }
    }
```

heitorlessa:
Looking to merge this one today

ran-isenberg:
@heitorlessa  can't wait :)

heitorlessa:
[The docs PR on consistency for all pages](https://github.com/awslabs/aws-lambda-powertools-python/pull/919) took way more hours than anticipated, I'm making a minor change on this tomorrow and merging before release (also tomorrow). Apologies for the delay

https://github.com/awslabs/aws-lambda-powertools-python/pull/919

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -97,21 +97,29 @@ def _evaluate_conditions(
...
             rule_match_value = rule.get(schema.RULE_MATCH_VALUE)
 
             # Context might contain PII data; do not log its value
             self.logger.debug(
-                f"Evaluating rule matching, rule={rule_name}, feature={feature_name}, default={feat_default}"
+                f"Evaluating rule matching, rule={rule_name}, feature={feature_name}, default={str(feat_default)}, boolean_feature={boolean_feature}"  # type: ignore # noqa: E501
             )
             if self._evaluate_conditions(rule_name=rule_name, feature_name=feature_name, rule=rule, context=context):
-                return bool(rule_match_value)
+                return bool(rule_match_value) if boolean_feature else rule_match_value
```
heitorlessa, line 117: note to self: add comment so we don't forget in the future while bool and non-bool here

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -181,14 +189,15 @@ def evaluate(self, *, name: str, context: Optional[Dict[str, Any]] = None, defau
             Attributes that should be evaluated against the stored schema.
 
             for example: `{"tenant_id": "X", "username": "Y", "region": "Z"}`
-        default: bool
+        default: Any
```
heitorlessa, line 192: note to self: since it's now Any, double check type annotation in docs

heitorlessa:
@ran-isenberg as we're in beta, does it worth making a note that this feature will be the default once/right before GA?

As in, it'd be great to not require this additional field in their schema -- return whatever the value when match or default value.

Keen to hear your thoughts as I know you're using in prod regardless of the label beta.

ran-isenberg:
I dont mind keeping the new field because that way we do get the boolean schema validation instead of allowing blindly everything. There's also the get all enabled flags API which I think should return only the boolean rules that match (it returns a list of string - name of features that are True).

heitorlessa:
Gotcha - for the get all enabled features I was thinking in keeping it as
is, and create a new get all matching features to cater for the new returns.

As I want to release it today, I’ll mark to revisit before we go GA - my
only concern is the complexity of newcomers into the feature, not so much
about maintenance as we can make some minor changes.

I’ll go ahead and make the minimal changes to ease maintenance, help Mypy
inform issues, write the docs, and add a note to revisit this feature on
whether we can make it simpler

ran-isenberg:
you mean a function  that returns a Dict[ str, Any] with feature name and  the evaluated value?
that could work too

heitorlessa:
Yeah, just like we did for Parameters when you wanna get multiple
parameters at once (recursively for SSM, or as a Query for the same pk in
DynamoDB)

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -259,11 +277,22 @@ def get_enabled_features(self, *, context: Optional[Dict[str, Any]] = None) -> L
         for name, feature in features.items():
             rules = feature.get(schema.RULES_KEY, {})
             feature_default_value = feature.get(schema.FEATURE_DEFAULT_VAL_KEY)
+            boolean_feature = feature.get(
+                schema.FEATURE_DEFAULT_VAL_TYPE_KEY, True
+            )  # backwards compatability ,assume feature flag
+            if not boolean_feature:
+                self.logger.debug(f"skipping feature because it is not a boolean feature flag, name={name}")
+                continue
```
heitorlessa, line 285:
```suggestion
```

Just found this piece when fixing Mypy and adding GA notices and I'm removing it -- Let me know if my reading is incorrect.

In the docs, we call out `get_all_enabled_features` will return [any features that match the input context](https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/feature_flags/#getting-all-enabled-features) - This code block however ignores non-boolean and doesn't honour our docs, additional tests failed too.

Looking up the code to confirm whether the docs was accidentally incorrect, docstrings confirm it:

```python
def get_enabled_features(self, *, context: Optional[Dict[str, Any]] = None) -> List[str]:
		...
		
        Returns
        ----------
        List[str]
            list of all feature names that either matches context or have True as default
```

heitorlessa:
Thanks again Ran for implementing this feature!

I've made the following changes to ease maintenance, plus docs. I'm merging, working on one last PR then publishing 1.24 release with it ;-)

- [x] Bugfix: 
  - [x] Allow non-boolean values to be returned upon rule match
  - [x] 5 Mypy issues on types
  - [x] Replaced `Any` with `JSONType` to ease Mypy and IDE findings for non-JSON primitives
- [x] Tests: 
  - [x] Merged tests into a single file, removed one/two duplicate
  - [x] Renamed tests to be more explicit to the behaviour they're testing
  - [x] Tests for truthy default non-boolean default values
- [x] Docs:
  - [x] Updated Schema Validation API docs
  - [x] Updated Schema section
  - [x] New section `Beyond boolean feature flags`

image
image
image

PR #818, opened 2021-11-12
fix(batch): delete >10 messages in legacy sqs processor

whardier:
**No related issue**

## Description of changes:

SQS/Batch - SImple addition to process chunks of records that need removed in 10's.  Counterintuitive and migrating this behaviour to the batch processor scope itself would be too disruptive.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

No breaking changes.  Introduces no new exception handling and would fix batch handling of more than 10 records.

**No RFC**:


whardier:
Working on updating the tests now that tests are actually looking at the mocked values.. sigh.

heitorlessa:
Hey @whardier could you create an issue first detailing the bug please?

For example, what configuration are you using so that your function receives more than 10 messages per invocation where this is necessary? e.g. large batch size, longer batch window?

If a single function receives a few thousand records with 50% partial failure, this has the side effect of increasing function duration and potentially timing out if one is not careful.

An issue helps us think about potential side effects like these, compromises, what information we need to provide in the docs to help other customers unfamiliar with this scenario, etc.

Thanks a lot!

whardier:
https://github.com/awslabs/aws-lambda-powertools-python/issues/824

heitorlessa:
@mploski before you tackle the redis one, could you help revamp this PR so we can merge it?

Ask. Move to concurrent futures and pass the client to make it thread safe as explained in the review

Background. This is the legacy batch we keep around that we're only deleting 10 messages instead of all failed messages in a batch (could be hundreds).

mploski:
> @mploski before you tackle the redis one, could you help revamp this PR so we can merge it?
> 
> Ask. Move to concurrent futures and pass the client to make it thread safe as explained in the review
> 
> Background. This is the legacy batch we keep around that we're only deleting 10 messages instead of all failed messages in a batch (could be hundreds).

Sure @heitorlessa will look at this at the end of the next week.

mploski:
@whardier @heitorlessa  I will make changes on top of this PR's branch today.

mploski:
@whardier Could you fetch latest changes from upstream into your clone? You're lagging behind a bit :-)

mploski:
@whardier @heitorlessa  Applied suggestions, please review.

whardier:
This looks clean.  I can't test it right away however.  I like it.

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -120,23 +123,39 @@ def _prepare(self):
         self.success_messages.clear()
         self.fail_messages.clear()
 
-    def _clean(self):
+    def _clean(self) -> Optional[List]:
```
heitorlessa, line 126: niiiice one! Thanks

heitorlessa (approved):
Looks awesome, thank you so much @mploski and also @whardier for kick starting it.

We can now make moves to release several papercuts and the mypy official support now.

Thank you!

Issue #834, opened 2021-11-18
Provide access to ApiGatewayResolver object from Router object


gwlester:
**Is your feature request related to a problem? Please describe.**

When attempting to use the, introduced in 1.22.0, Split routes with Router methods, we found that many of our end point functions needed access to the ApiGatewayResolver object.  This is due to the fact that we have subclassed the ApiGatewayResolver to add some application-specific functionality.

**Describe the solution you'd like**

Please provide, either in the Router object or in the BaseProxyEvent (referenced by the current_event property) access to the ApiGatewayResolver object.  Ideally, the BaseProxyEvent could provide access to both the ApiGatewayResolver object and Router object.

**Describe alternatives you've considered**

Not using the new router functionality.

**Additional context**

cakepietoast:
While this is a corner case, we should be able to implement it easily and without affecting other functionality. I'll put a PR in soon and ask for you to validate that it solves your use case.

gwlester:
Thanks!

gwlester:
Reviewed the code, it looks good.

cakepietoast:
Thanks @gwlester - we'll get it merged soon. Out of curiosity - are you able to share some details about your use case? I'd love to understand the need for subclassing`ApiGatewayResolver` - maybe its something we can help make more convenient.

gwlester:
@cakepietoast , yes (and maybe we did things wrong but...)

We subclassed the resolver to add an attribute about the authenticated user (including what feature flags they had access to) and also the logger (to add information about the session and transaction required by our company from the headers to the log attributes) to use.  We then pass down the app to all the lower-level routines to use app.logger and app.user.

heitorlessa:
It'll be available as part of today's release!

PR #842, opened 2021-11-22
feat(apigateway): access parent api resolver from router

cakepietoast:
**Issue #, if available:** #834

## Description of changes:

Enable access to parent api resolver class from child router classes. Example:

app.py
```python
...
app = ApiGatewayResolver()
app.include_router(users.router, prefix="/users") # prefix 

def lambda_handler(event: Dict, context: LambdaContext):

```

users.py
```python
...

router = Router()

@router.get("/users/")
def get_users() -> Dict:
    router.api_resolver.custom_functionality()
    return {"message": "success"}

```

**Checklist**

 

* [X] Update tests
* [X] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -380,7 +380,7 @@ class BaseRouter(ABC):
 
 
 class ApiGatewayResolver(BaseRouter):
-    """API Gateway and ALB proxy api_resolver
+    """API Gateway and ALB proxy resolver
 
     Examples
     --------
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -688,6 +688,7 @@ class Router(BaseRouter):
         cache_control: Optional[str] = None,
     ):
         def register_route(func: Callable):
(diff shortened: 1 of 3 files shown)


heitorlessa:
Should we add a note in the docs? E.g. subclassing ApiGatewayResolver etc

cakepietoast:
> Should we add a note in the docs? E.g. subclassing ApiGatewayResolver etc

I think if we add documentation here it needs to be a more widely scoped section explaining how and why you might want to subclass `ApiGatewayResolver` as opposed to just the particular detail this PR implements. I'd like to try and understand how common this is, and what some of the use cases are, before expanding on the docs there.

cakepietoast:
> Should we add a note in the docs? E.g. subclassing ApiGatewayResolver etc

I think if we add documentation here it needs to be a more widely scoped section explaining how and why you might want to subclass `ApiGatewayResolver`, as opposed to just the particular detail this PR implements. I'd like to try and understand how common this is, and what some of the use cases are, before expanding on the docs there.

heitorlessa:
Agree - that’s what I meant by whether it should be in the docs (explained
with an example).

LGTM

heitorlessa (approved):

PR #864, opened 2021-12-01
docs(apigateway): fix sample layout provided

michaelbrewer:
Current version does not work when deployed

**Issue #, if available:**

- https://github.com/awslabs/aws-lambda-powertools-python/issues/835

## Description of changes:

Fixes the sample layout for multiple file project

**Checklist**

 

* [X] Update tests
* [X] Update docs

-----
[View rendered docs/core/event_handler/api_gateway.md](https://github.com/gyft/aws-lambda-powertools-python/blob/doc-updated-layout/docs/core/event_handler/api_gateway.md)

### DIFF
--- a/docs/core/event_handler/api_gateway.md
+++ b/docs/core/event_handler/api_gateway.md
@@ -1027,43 +1027,42 @@ When necessary, you can set a prefix when including a router object. This means 
 
 #### Sample layout
 
-!!! info "We use ALB to demonstrate that the UX remains the same"
-
-This sample project contains an Users function with two distinct set of routes, `/users` and `/health`. The layout optimizes for code sharing, no custom build tooling, and it uses [Lambda Layers](../../index.md#lambda-layer) to install Lambda Powertools.
+This sample project contains a Users function with two distinct set of routes, `/users` and `/health`. The layout optimizes for code sharing, no custom build tooling, and it uses [Lambda Layers](../../index.md#lambda-layer) to install Lambda Powertools.
 
 === "Project layout"
 
 
-    ```python hl_lines="6 8 10-13"
+    ```python hl_lines="1 8 10 12-15"
     .
-    ├── Pipfile                   # project app & dev dependencies; poetry, pipenv, etc.
+    ├── Pipfile                    # project app & dev dependencies; poetry, pipenv, etc.
     ├── Pipfile.lock
-    ├── mypy.ini                  # namespace_packages = True
-    ├── .env                      # VSCode only. PYTHONPATH="users:${PYTHONPATH}"
-    ├── users
-    │   ├── requirements.txt      # sam build detect it automatically due to CodeUri: users, e.g. pipenv lock -r > users/requirements.txt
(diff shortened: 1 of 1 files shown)


docs/core/event_handler/api_gateway.md
```diff
@@ -1073,6 +1072,8 @@ This sample project contains an Users function with two distinct set of routes,
             MemorySize: 512
             Runtime: python3.9
             Tracing: Active
+            Architectures:
+                - arm64
```
heitorlessa, line 1076: We build Layer for x86 only atm

docs/core/event_handler/api_gateway.md
```diff
@@ -1027,43 +1027,42 @@ When necessary, you can set a prefix when including a router object. This means
...
+This sample project contains a Users function with two distinct set of routes, `/users` and `/health`. The layout optimizes for code sharing, no custom build tooling, and it uses [Lambda Layers](../../index.md#lambda-layer) to install Lambda Powertools.
 
 === "Project layout"
 
 
-    ```python hl_lines="6 8 10-13"
+    ```python hl_lines="1 8 10 12-15"
     .
-    ├── Pipfile                   # project app & dev dependencies; poetry, pipenv, etc.
+    ├── Pipfile                    # project app & dev dependencies; poetry, pipenv, etc.
```
heitorlessa, line 1037: Shall we update to poetry pyproject.tomo since we're gonna update the sample cookiecutter already?

heitorlessa (changes requested):
One minor change as we don't build w Lambda Layer for ARM yet, and one question

heitorlessa:
Thanks a lot for helping revert to the previous sample

michaelbrewer:
But it works. ;-)

> One minor change as we don't build w Lambda Layer for ARM yet, and one question

heitorlessa:
This is what puzzles me hahaha

docs/core/event_handler/api_gateway.md
michaelbrewer: making it `x86_64` for now :)

heitorlessa:
On a more serious note though, Tracer uses X-Ray SDK which depends on wrapt that uses C bindings.

Unless wrapt is giving a wheel for all platforms - like Pydantic - instead of the platform in question, then in theory Tracer shouldn't work.

All else should work as we don't depend on C bindings but pure Python.

docs/core/event_handler/api_gateway.md
michaelbrewer: @heitorlessa - currently I like `Pipefile` as this is the default for SAM cli and Amplify templates. However I am cool to switch to Poetry

michaelbrewer:
@heitorlessa - updated.

docs/core/event_handler/api_gateway.md
heitorlessa: So let's keep as-is until we migrate others.

heitorlessa (approved):

Issue #857, opened 2021-11-30
Papercuts we're currently prioritizing


whardier:
Need to add proper ContextManager typing via typing.ContrxtManager in several places.

Currently overriding single_metric (-> ContextManager[SingleMetric]) and noticed several other areas where this needs done as well.

Also be nice if there was a proper DummySubSegment type returned for tracing provider.in_subsegment to avoid including types from aws modules to cast.

## Expected Behavior
Have rich type annotations along for the ride when using context managers.

## Current Behavior
Need to self annotate.

## Possible Solution
Add in proper type annotation to some or all of the context manager decorated functions.

heitorlessa:
Thanks a lot for raising this @whardier ! I was about to create a list of papercuts we'd like to address, these being two of them.

The SubSegment typing is the most elaborate one, I need one full day to create and refactor a few things, including having an easy way to unit test your application with a Fake Tracer Provider.

heitorlessa:
When you have a moment, could you help us by enumerating some of them like the SingleMetric ctx manager?

I'm gonna rename this to top papercuts so we can tackle them accordingly as many as possible 

![image](https://user-images.githubusercontent.com/3340292/144075378-1ae6f0f4-2425-42a0-85a3-95ded0342892.png)

heitorlessa:
Top papercuts we'd like to address:

- [ ] [Docs: API Gateway sample layout](https://github.com/awslabs/aws-lambda-powertools-python/issues/835)
- [ ] [Parser: Typing on model subclassing](https://github.com/awslabs/aws-lambda-powertools-python/issues/849)
- [ ] [Parser: Typing when using envelope](https://github.com/awslabs/aws-lambda-powertools-python/issues/802)
- [ ] [Event Handler: BaseProxyEvent missing overload to match Proxy Type Request Context](https://github.com/awslabs/aws-lambda-powertools-python/issues/761)
- [x] [Tracer: Typing provider when none is passed](https://github.com/awslabs/aws-lambda-powertools-python/pull/861)
- [ ] Tracer: Typing SubSegment and Context Manager
- [ ] [Metrics: Typing SingleMetric context manager](https://github.com/awslabs/aws-lambda-powertools-python/issues/857)

elliottyates:
[...]

PR #865, opened 2021-12-02
fix(metrics): explicit type to single_metric ctx manager

whardier:
**Issue #, if available:**

* [Maintenance: address top papercuts #857](https://github.com/awslabs/aws-lambda-powertools-python/issues/857)

## Description of changes:

Simply update type return for `metrics.metric.single_metric`.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/metrics/metric.py
+++ b/aws_lambda_powertools/metrics/metric.py
@@ -1,7 +1,7 @@
 import json
 import logging
 from contextlib import contextmanager
-from typing import Dict, Optional, Union
+from typing import Dict, Iterator, Optional, Union
 
 from .base import MetricManager, MetricUnit
 
@@ -61,7 +61,7 @@ class SingleMetric(MetricManager):
 
 
 @contextmanager
-def single_metric(name: str, unit: MetricUnit, value: float, namespace: Optional[str] = None):
+def single_metric(name: str, unit: MetricUnit, value: float, namespace: Optional[str] = None) -> Iterator[SingleMetric]:
     """Context manager to simplify creation of a single metric
 
     Example
--- a/aws_lambda_powertools/metrics/metric.py
+++ b/aws_lambda_powertools/metrics/metric.py
@@ -1,7 +1,7 @@
 import json
 import logging
 from contextlib import contextmanager
-from typing import Dict, Iterator, Optional, Union
+from typing import Dict, Optional, Union, Generator
 
 from .base import MetricManager, MetricUnit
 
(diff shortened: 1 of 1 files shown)


aws_lambda_powertools/metrics/metric.py
```diff
@@ -61,7 +61,7 @@ def add_metric(self, name: str, unit: Union[MetricUnit, str], value: float) -> N
 
 
 @contextmanager
-def single_metric(name: str, unit: MetricUnit, value: float, namespace: Optional[str] = None):
+def single_metric(name: str, unit: MetricUnit, value: float, namespace: Optional[str] = None) -> Iterator[SingleMetric]:
```
heitorlessa, line 64: TIL: An Iterator *could* be problematic if someone tries to iterate it: https://github.com/python/typeshed/issues/2772

What we could do instead is using `Generator[yield_value, NoSendType, NoReturnType]` resulting in `Generator[SingleMetric, None, None]`.

```suggestion
def single_metric(name: str, unit: MetricUnit, value: float, namespace: Optional[str] = None) -> Generator[SingleMetric, None, None]:
```

aws_lambda_powertools/metrics/metric.py
```diff
@@ -1,7 +1,7 @@
 import json
 import logging
 from contextlib import contextmanager
-from typing import Dict, Optional, Union
+from typing import Dict, Iterator, Optional, Union
```
heitorlessa, line 4: See reason at the bottom

```suggestion
from typing import Dict, Optional, Union, Generator
```

heitorlessa (changes requested):
One minor change to use `Generator[SingleMetric, None, None]` due to a new issue discovered in Python typeshed when using `Iterator[]` in this context.

aws_lambda_powertools/metrics/metric.py
heitorlessa: On a practical level, mypy wise, it leads to the same type result:

image

image

heitorlessa:
I created a new project using VSCode and PyLance and I couldn't reproduce this issue, same with PyCharm so perhaps their InteliSense is smart by inspecting the yield type `SingleMetric`.

Nevertheless, [TIL why](https://github.com/python/typeshed/issues/2772) we can't use `Iterator[SingleMetric]` and instead should use `Generator[SingleMetric, None, None`.

For the sake of correctness regardless of what IDE one might use, I'm gonna commit the Generator suggestion to avoid issues and merge it.

heitorlessa (approved):

whardier:
Interesting.  I poured over a ton of python code and standard libs and Iterator seemed like the best fit.  I will read up on this.

heitorlessa:
Yep same. It seems to be a “new” thing

PR #869, opened 2021-12-04
fix(idempotency): include decorated fn name in hash

### LINKED ISSUE #855: Incomplete input for hash function for @idempotent_function
When annotating functions via @idempotent_function, the wrapped function's name will not go into the hash computation; instead, only the payload and an environment-specific (but not function-specific) function name is assumed. 
This causes idempotency to break in certain cases. 

## Expected Behavior
When two distinct functions which accept the very same payload as parameters are wrapped using @idempotent_function, invocation of one function should not affect invocation of the other function. 

## Current Behavior
Only the payload is considered. 
The culprit appears to be 
```python
generated_hash = self._generate_hash(data=data)
function_name = os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, "test-func")
```
(lines 180-181) in `persistence/base.py`. One can clearly see that only the environment variable or "test-func" is assumed - but that is - of course - not specific to the actual function being wrapped.
So in short, the library currently only really distinguishes by payload and not be function name, hence causing the aforementioned behaviour. 

## Possible Solution
Pass on the wrapped function's name and use it for the return value of `_get_hashed_idempotency_key`. 

## Steps to Reproduce (for bugs)
```python
def test():
    myIdempotentFunc(product = "bananas")
    anotherIdempotentFunc(product="bananas")
    return {}

@idempotent_function(data_keyword_argument="product", persistence_store=dynamodb)
def myIdempotentFunc(product):
    print("I got invoked with {product}".format(product = product))
    return None

@idempotent_function(data_keyword_argument="product", persistence_store=dynamodb)
def anotherIdempotentFunc(product):
    print("I will never get invoked; someone stole my payload :(".format(product = product))
    return None
```
This only prints myIdempotentFunc's output, so anotherIdempotentFunc never gets invoked. 

## Environment

Using "aws-lambda-powertools==1.22.0" on Python 3.9.7

Thank you in advance!

michaelbrewer:
**Issue #, if available:**

- #855

## Description of changes:

Include the callable function name in the generated idempotent key

**Checklist**

 

* [x] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -124,7 +124,7 @@ class BasePersistenceLayer(ABC):
         self._cache: Optional[LRUDict] = None
         self.hash_function = None
 
-    def configure(self, config: IdempotencyConfig) -> None:
+    def configure(self, config: IdempotencyConfig, function_name: str) -> None:
         """
         Initialize the base persistence layer from the configuration settings
 
@@ -133,6 +133,7 @@ class BasePersistenceLayer(ABC):
         config: IdempotencyConfig
             Idempotency configuration settings
         """
+        self.function_name = function_name
         if self.configured:
             # Prevent being reconfigured multiple times
             return
@@ -178,7 +179,7 @@ class BasePersistenceLayer(ABC):
             warnings.warn(f"No value found for idempotency_key. jmespath: {self.event_key_jmespath}")
 
         generated_hash = self._generate_hash(data=data)
-        function_name = os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, "test-func")
+        function_name = os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, "test-func") + "." + self.function_name
         return f"{function_name}#{generated_hash}"
 
     @staticmethod
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -112,6 +112,7 @@ class BasePersistenceLayer(ABC):
 
     def __init__(self):
         """Initialize the defaults"""
+        self.function_name = None
         self.configured = False
         self.event_key_jmespath: Optional[str] = None
         self.event_key_compiled_jmespath = None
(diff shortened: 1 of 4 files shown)


aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -124,7 +125,7 @@ def __init__(self):
         self._cache: Optional[LRUDict] = None
         self.hash_function = None
 
-    def configure(self, config: IdempotencyConfig) -> None:
+    def configure(self, config: IdempotencyConfig, function_name: str) -> None:
```
michaelbrewer, line 128: @heitorlessa - should this be optional? or required?

heitorlessa: Optional. Only because it's a public method and we have no way of knowing if someone took a dependency on it via tests or something - if they did, we will break them and take another release to fix it

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -178,7 +180,7 @@ def _get_hashed_idempotency_key(self, data: Dict[str, Any]) -> str:
             warnings.warn(f"No value found for idempotency_key. jmespath: {self.event_key_jmespath}")
 
         generated_hash = self._generate_hash(data=data)
-        function_name = os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, "test-func")
+        function_name = os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, "test-func") + "." + self.function_name
```
heitorlessa, line 183: I was gonna call out that this will impact customers also using `@idempotency` instead of `@idempotency_function` only as reported... but then it ocurred to me customers using a single function with multiple adapters **and** idempotency will end up in the same bug.

suggestion to use f-string instead of multiple concatenation for performance and nitpick late on Monday evening

```suggestion
	lambda_fn_name = os.getenv(constants.LAMBDA_FUNCTION_NAME_ENV, "test-func")
	function_name = f"{lambda_fn_name}.{self.function_name}"
```

cc @cakepietoast for awareness

heitorlessa (changes requested):
two minor changes: **1/** make `function_name` optional as we don't know who ended up taking a dependency on that public method (Hyrum's Law), **2/** use f-string to optimize memory and a CPU cycle as `str` is more expensive in Python than usual

michaelbrewer:
@heitorlessa @cakepietoast - i have pushed recommended changes.

heitorlessa (approved):
Thanks a lot Mike!

PR #877, opened 2021-12-08
fix(parameters): appconfig transform and return types

ran-isenberg:
**Issue #, if available:**: 

## Description of changes:

Fix return type and handling transform value as str instead of bytes

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

No breaking changes.  Introduces no new exception handling and would fix batch handling of more than 10 records.

**No RFC**:

### DIFF
--- a/aws_lambda_powertools/utilities/parameters/appconfig.py
+++ b/aws_lambda_powertools/utilities/parameters/appconfig.py
@@ -84,7 +84,7 @@ class AppConfigProvider(BaseProvider):
 
         super().__init__()
 
-    def _get(self, name: str, **sdk_options) -> bytes:
+    def _get(self, name: str, **sdk_options) -> str:
         """
         Retrieve a parameter value from AWS App config.
 
@@ -104,7 +104,7 @@ class AppConfigProvider(BaseProvider):
         sdk_options["ClientId"] = CLIENT_ID
 
         response = self.client.get_configuration(**sdk_options)
-        return response["Content"].read()  # read() of botocore.response.StreamingBody
+        return response["Content"].read().decode("utf-8")  # read() of botocore.response.StreamingBody
 
     def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
         """
@@ -121,7 +121,7 @@ def get_app_config(
     force_fetch: bool = False,
     max_age: int = DEFAULT_MAX_AGE_SECS,
     **sdk_options
-) -> Union[str, list, dict, bytes]:
+) -> Union[str, list, dict]:
     """
     Retrieve a configuration value from AWS App Config.
 
--- a/aws_lambda_powertools/utilities/parameters/appconfig.py
+++ b/aws_lambda_powertools/utilities/parameters/appconfig.py
@@ -84,7 +84,7 @@ class AppConfigProvider(BaseProvider):
 
         super().__init__()
 
-    def _get(self, name: str, **sdk_options) -> str:
+    def _get(self, name: str, **sdk_options) -> bytes:
         """
         Retrieve a parameter value from AWS App config.
 
--- a/aws_lambda_powertools/utilities/parameters/appconfig.py
+++ b/aws_lambda_powertools/utilities/parameters/appconfig.py
@@ -104,7 +104,7 @@ class AppConfigProvider(BaseProvider):
         sdk_options["ClientId"] = CLIENT_ID
 
         response = self.client.get_configuration(**sdk_options)
-        return response["Content"].read().decode("utf-8")  # read() of botocore.response.StreamingBody
+        return response["Content"].read()  # read() of botocore.response.StreamingBody
 
     def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
         """
@@ -121,7 +121,7 @@ def get_app_config(
     force_fetch: bool = False,
     max_age: int = DEFAULT_MAX_AGE_SECS,
     **sdk_options
-) -> Union[str, list, dict]:
+) -> Union[str, list, dict, bytes]:
     """
     Retrieve a configuration value from AWS App Config.
 
--- a/aws_lambda_powertools/utilities/parameters/base.py
+++ b/aws_lambda_powertools/utilities/parameters/base.py
@@ -93,6 +93,8 @@ class BaseProvider(ABC):
             raise GetParameterError(str(exc))
 
         if transform is not None:
+            if isinstance(value, bytes):
+                value = value.decode("utf-8")
             value = transform_value(value, transform)
 
         self.store[key] = ExpirableValue(value, datetime.now() + timedelta(seconds=max_age))
@@ -100,7 +102,7 @@ class BaseProvider(ABC):
         return value
 
     @abstractmethod
-    def _get(self, name: str, **sdk_options) -> str:
+    def _get(self, name: str, **sdk_options) -> bytes:
         """
         Retrieve parameter value from the underlying parameter store
         """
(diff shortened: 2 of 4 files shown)


heitorlessa:
Thanks a lot Ran! I'll look into it tomorrow morning

aws_lambda_powertools/utilities/parameters/appconfig.py
```diff
@@ -84,7 +84,7 @@ def __init__(
 
         super().__init__()
 
-    def _get(self, name: str, **sdk_options) -> str:
+    def _get(self, name: str, **sdk_options) -> bytes:
```
heitorlessa: This makes MyPy fail in a different way, since this overrides an abstract method return type previously defined as `str` in `BaseProvider` -- I'm gonna add to the papercut list to ensure we run MyPy checks as part of CI; it's gonna increase the bar for contribution but reduce these type of issues as they seem to be growing elsewhere too.

`error: Return type "bytes" of "_get" incompatible with return type "str" in supertype BaseProvider`

There's also another problem - `get_app_config()` returns `Union[str, list, dict]` when it's in fact `bytes`, unless Transform is used. 

A quick non-breaking change fix for now might be having the base `_get` and `get_app_config` to accept multiple return types.

A proper solution would involve [`@overload` type when using specific transform](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading) (json, binary), and an explicit cast to the expected type in certain areas (needs evaluation w/ MyPy and tests).

heitorlessa (changes requested):
Thanks a lot for the quick fix, Ran!

We will need to also update the base `_get` from `BaseProvider` to accept multiple types or else this will make MyPy fail with invalid override, and more importantly `get_app_config` also has the incorrect return type (bytes is missing).

I'm gonna update the list of papercuts to ensure we run MyPy at CI to catch these

aws_lambda_powertools/utilities/parameters/appconfig.py
```diff
@@ -104,7 +104,7 @@ def _get(self, name: str, **sdk_options) -> str:
         sdk_options["ClientId"] = CLIENT_ID
 
         response = self.client.get_configuration(**sdk_options)
-        return response["Content"].read()  # read() of botocore.response.StreamingBody
+        return response["Content"].read().decode("utf-8")  # read() of botocore.response.StreamingBody
```
heitorlessa, line 107: oh wait, I missed this part as I just noticed the PR was showing outdated. Doesn't this break what consumers are already expecting despite the incorrect type?

Our tests were expecting bytes before.

heitorlessa:
Bringing the discussion we're having on Slack so we can track the history afterwards @whardier, and to make sure the problem is well understood and we don't accidentally create another one.

Based on your transform issue report, we seem to have two issues not one -- please correct me if I misunderstood it:

## Transform

**Problem**. By default AppConfig returns `bytes` not `str` as the interfaces suggest.

**Fix**. We need to update [transform_value function](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parameters/base.py#L218:L218) signature to also accept `bytes` value and handle it accordingly (avoid double decoding if it's bytes) - This function isn't documented despite being public (it shouldn't, oversight during the review), so it will not be a breaking change either way.

## Return types

**Problem**. AppConfig implementation of `_get` interface from BaseProvider uses a different return type which creates subtle bugs like these. Not surprisingly in the public documentation, we [provide an example and explicitly call out `get_app_config` return type as bytes](https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/parameters/#fetching-app-configurations).


**Fix**. We will need `_get` to also return `bytes`. For another PR, we should do a proper solution in doing function overload to handle `transform` return types correctly.

aws_lambda_powertools/utilities/parameters/appconfig.py
ran-isenberg: yes, you are correct.

aws_lambda_powertools/utilities/parameters/appconfig.py
heitorlessa: summary comment updated to reflect and remove the ambiguous part now that we have a better understanding

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -93,14 +93,16 @@ def get(
...
+                value = value.decode("utf-8")
             value = transform_value(value, transform)
 
         self.store[key] = ExpirableValue(value, datetime.now() + timedelta(seconds=max_age))
 
         return value
 
     @abstractmethod
-    def _get(self, name: str, **sdk_options) -> str:
+    def _get(self, name: str, **sdk_options) -> bytes:
```
heitorlessa, line 105:
```suggestion
    def _get(self, name: str, **sdk_options) -> Union[str, bytes]:
```

heitorlessa:
Leaving other issues for posterity as there could be some other subtle bugs

---

```python
aws_lambda_powertools/utilities/parameters/base.py: note: In member "get_multiple" of class "BaseProvider":
aws_lambda_powertools/utilities/parameters/base.py:158:64: error: Incompatible types in assignment (expression has type "Dict[str, str]", variable has type "Dict[str, Union[str, bytes, Dict[Any, Any], None]]")  [assignment]
                values: Dict[str, Union[str, bytes, dict, None]] = self._get_multiple(path, **sdk_options)
                                                                   ^
aws_lambda_powertools/utilities/parameters/base.py:158:64: note: "Dict" is invariant -- see https://mypy.readthedocs.io/en/stable/common_issues.html#variance
aws_lambda_powertools/utilities/parameters/base.py:158:64: note: Consider using "Mapping" instead, which is covariant in the value type
aws_lambda_powertools/utilities/parameters/base.py:164:13: error: Incompatible types in assignment (expression has type "str", variable has type "Tuple[str, Optional[str]]")  [assignment]
                for (key, value) in values.items():
                ^
aws_lambda_powertools/utilities/parameters/base.py:165:51: error: Argument 1 to "get_transform_method" has incompatible type "Tuple[str, Optional[str]]"; expected "str"  [arg-type]
                    _transform = get_transform_method(key, transform)
                                                      ^
aws_lambda_powertools/utilities/parameters/base.py:169:24: error: Invalid index type "Tuple[str, Optional[str]]" for "Dict[str, Union[str, bytes, Dict[Any, Any], None]]"; expected type "str"  [index]
                    values[key] = transform_value(value, _transform, raise_on_transform_error)
                           ^
aws_lambda_powertools/utilities/parameters/base.py:169:47: error: Argument 1 to "transform_value" has incompatible type "Union[str, bytes, Dict[Any, Any], None]"; expected "str"  [arg-type]
                    values[key] = transform_value(value, _transform, raise_on_transform_error)
                                                  ^
aws_lambda_powertools/utilities/parameters/base.py:173:16: error: Incompatible return value type (got "Dict[str, Union[str, bytes, Dict[Any, Any], None]]", expected
"Union[Dict[str, str], Dict[str, Dict[Any, Any]], Dict[str, bytes]]")  [return-value]
            return values
                   ^
aws_lambda_powertools/utilities/parameters/ssm.py: note: In member "get" of class "SSMProvider":
aws_lambda_powertools/utilities/parameters/ssm.py:90:5: error: Signature of "get" incompatible with supertype "BaseProvider"  [override]
        def get(
        ^
aws_lambda_powertools/utilities/parameters/ssm.py:131:16: error: Incompatible return value type (got "Union[str, List[Any], Dict[Any, Any], bytes, None]", expected "Union[str, List[Any], Dict[Any, Any], bytes]")  [return-value]
            return super().get(name, max_age, transform, force_fetch, **sdk_options)
                   ^
```

heitorlessa (approved):

PR #883, opened 2021-12-10
fix(parser): mypy support for payload type override as models

heitorlessa:
**Issue #, if available:** https://github.com/awslabs/aws-lambda-powertools-python/issues/857

## Description of changes:

This PR allows MyPy to not trigger an error when overriding payload types in built-in models provided by Parser, as described: https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/parser/#extending-built-in-models

**Scenario**: There are cases where you want to provide your own Model while retaining access to metadata information about the event. Previously, when using MyPy, this would cause a type override error since payload fields would only support what's expected, not a custom model as docs suggest for non-MyPy customers.

```python
from aws_lambda_powertools.utilities.parser import parse, BaseModel
from aws_lambda_powertools.utilities.parser.models import EventBridgeModel

from typing import List, Optional


class OrderItem(BaseModel):
    id: int
    quantity: int
    description: str


class Order(BaseModel):
    id: int
    description: str
    items: List[OrderItem]


class OrderEventModel(EventBridgeModel):
    detail: Order
```

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/docs/utilities/parser.md
+++ b/docs/utilities/parser.md
@@ -170,7 +170,7 @@ Parser comes with the following built-in models:
 
 You can extend them to include your own models, and yet have all other known fields parsed along the way.
 
-!!! tip "For Mypy customers, we only allow type override for fields where payload is injected e.g. `detail`, `body`, etc."
+!!! tip "For Mypy users, we only allow type override for fields where payload is injected e.g. `detail`, `body`, etc."
 
 **EventBridge example**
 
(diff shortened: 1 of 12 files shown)


docs/utilities/parser.md
```diff
@@ -170,6 +170,8 @@ Parser comes with the following built-in models:
 
 You can extend them to include your own models, and yet have all other known fields parsed along the way.
 
+!!! tip "For Mypy customers, we only allow type override for fields where payload is injected e.g. `detail`, `body`, etc."
```
cakepietoast, line 173:
```suggestion
!!! tip "For Mypy users, we only allow type override for fields where payload is injected e.g. `detail`, `body`, etc."
```

cakepietoast (approved):

Issue #858, opened 2021-11-30
RFC: New batch processing utility


cakepietoast:
## Key information

* RFC PR: 
* Related issue(s), if known:
* Area: Batch processing util
* Meet [tenets](https://awslabs.github.io/aws-lambda-powertools-python/#tenets): Yes
* Approved by: ''
* Reviewed by: ''

## Summary
[summary]: #summary

A new generic batch processing utility, which can process records from SQS, Kinesis Data Streams, and DynamoDB streams, and handle reporting batch failures. 

## Motivation
[motivation]: #motivation

With the launch of support for [partial batch responses for Lambda/SQS](https://aws.amazon.com/about-aws/whats-new/2021/11/aws-lambda-partial-batch-response-sqs-event-source/), the event source mapping can now natively handle partial failures in a batch - removing the need for calls to the delete api. This support already exists for [Kinesis](https://docs.aws.amazon.com/lambda/latest/dg/with-kinesis.html#services-kinesis-batchfailurereporting) and [DynamoDB streams](https://docs.aws.amazon.com/lambda/latest/dg/with-ddb.html#services-ddb-batchfailurereporting).

## Proposal
[proposal]: #proposal

The new utility will be implemented in the namespace which already exists for the SQS Batch Processing utility - aws_lambda_powertools.utilities.batch. Rather than calling the SQS DeleteMessage api like the existing batch utility, this version will instead inject `BatchItemFailures` (or `batchItemFailures` for kinesis and ddb) into the Lambda response. We will expose at the least, a new decorator `batch_processor`, which will accept an event type depending on the integration (Kinesis, SQS or DDB Streams). This will look similar to the Event Handler design. There will be a boolean setting to handle partial failures, defaulting to True (users will still need to enable this in the event source mapping for it to work).

```python

from aws_lambda_powertools.batch import KinesisStreamEvent, EventType


def record_handler(record):
    return do_something_with(record["body"])


@batch_processor(record_handler=record_handler, event_type=EventType.KinesisStreamEvent)
def lambda_handler(event, context):
    return {"statusCode": 200} 
```

Proposed class/method names for the new api: 
`aws_lambda_powertools.utilities.batch.base.BatchProcessor`
`aws_lambda_powertools.utilities.batch.batch_processor`

**Any configuration or corner cases you'd expect?**
[...]

PR #886, opened 2021-12-10
feat(batch): new BatchProcessor for SQS, DynamoDB, Kinesis

heitorlessa:
**Issue #, if available:** #858

## Description of changes:

Implements new Batch Processor following the [new built-in partial response handling](https://aws.amazon.com/about-aws/whats-new/2021/11/aws-lambda-partial-batch-response-sqs-event-source/) for Lambda SQS, DynamoDB Stream, and Kinesis Data Streams. 

**Checklist**

 

* [x] Update tests
* [x] Update docs
* [x] Add DynamoDB Stream support
* [x] Add Kinesis Data Stream support
* [x] Think of a better name other than `.report()` -> `response` to match Java counterpart
* [x] Support event source data classes by default
* [x] Mypy support
* [x] Review with @cakepietoast whether we still need `suppress_exception` behaviour
* [x] Bring-your-own-model Pydantic Parser as an option
* [x] Test for incorrect data/payload that won't be parsed (either Model or Data Class)
* [x] Raise `BatchProcessingError` when all batch records fail to ensure failure arises

**Tasks to be created in separate PR to ease reviewing**

* [ ] Support for handling permanent exceptions (might be for the release after next)

### UX

**As a decorator**

```python
from aws_lambda_powertools.utilities.batch import batch_processor, BatchProcessor, EventType
from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord

processor = BatchProcessor(event_type=EventType.SQS) # or, EventType.KinesisDataStreams, EventType.DynamoDB

def record_handler(record: SQSRecord): # or DynamoDBRecord, KinesisStreamRecord
    return "success"

@batch_processor(record_handler=record_handler, processor=processor)
    return processor.response() # this is necessary to ensure Lambda returns the correct response
```

**As a context manager**, in case you need full access to the batch processed, or handle exceptions yourself

```python
from aws_lambda_powertools.utilities.batch import batch_processor, BatchProcessor, EventType

processor = BatchProcessor(event_type=EventType.SQS) 


    records = event["Records"]
    with processor(records, record_handler) as batch:
        processed_messages = batch.process()
    
    return batch.response() # this is necessary to ensure Lambda returns the correct response
```

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -172,9 +172,8 @@ class BatchProcessor(BasePartialProcessor):
 
         super().__init__()
 
-    # refactor: think of a better name
-    def report(self):
-        """Report batch items that failed processing, if any"""
+    def response(self):
+        """Response containing batch items that failed processing, if any"""
         return self.items_to_report
 
     def _prepare(self):
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -213,7 +213,6 @@ class BatchProcessor(BasePartialProcessor):
             EventType.KinesisDataStreams: KinesisStreamRecord,
             EventType.DynamoDBStreams: DynamoDBRecord,
         }
-        self._EVENT_ID_MAPPING = {}
 
         super().__init__()
 
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -16,13 +16,23 @@ from aws_lambda_powertools.utilities.data_classes.kinesis_stream_event import Ki
 from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
 
 logger = logging.getLogger(__name__)
-has_pydantic = "pydantic" in sys.modules
 
-SuccessCallback = Tuple[str, Any, dict]
-FailureCallback = Tuple[str, str, dict]
+
+class EventType(Enum):
+    SQS = "SQS"
+    KinesisDataStreams = "KinesisDataStreams"
+    DynamoDBStreams = "DynamoDBStreams"
+
+
+#
+# type specifics
+#
+has_pydantic = "pydantic" in sys.modules
 _ExcInfo = Tuple[Type[BaseException], BaseException, TracebackType]
 _OptExcInfo = Union[_ExcInfo, Tuple[None, None, None]]
 
+# For IntelliSense and Mypy to work, we need to account for possible SQS, Kinesis and DynamoDB subclasses
+# We need them as subclasses as we must access their message ID or sequence number metadata via dot notation
 if has_pydantic:
     from aws_lambda_powertools.utilities.parser.models import DynamoDBStreamRecordModel
     from aws_lambda_powertools.utilities.parser.models import KinesisDataStreamRecord as KinesisDataStreamRecordModel
@@ -32,11 +42,13 @@ if has_pydantic:
         Union[Type[SqsRecordModel], Type[DynamoDBStreamRecordModel], Type[KinesisDataStreamRecordModel]]
     ]
 
-
-class EventType(Enum):
-    SQS = "SQS"
-    KinesisDataStreams = "KinesisDataStreams"
-    DynamoDBStreams = "DynamoDBStreams"
+# When using processor with default arguments, records will carry EventSourceDataClassTypes
+# and depending on what EventType it's passed it'll correctly map to the right record
+# When using Pydantic Models, it'll accept any
+EventSourceDataClassTypes = Union[SQSRecord, KinesisStreamRecord, DynamoDBRecord]
+BatchEventTypes = Union[EventSourceDataClassTypes, "BatchTypeModels"]
+SuccessCallback = Tuple[str, Any, BatchEventTypes]
+FailureCallback = Tuple[str, str, BatchEventTypes]
 
 
 class BasePartialProcessor(ABC):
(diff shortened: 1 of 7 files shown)


heitorlessa:
@cakepietoast ready to review implementation before it gets bigger with mypy and doc stuff.

ran-isenberg:
@heitorlessa  as discussed, here's the sqs example:

@validator('body', pre=True)
    def transform_body_to_dict(cls, value):
        return json.loads(value)

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -146,3 +184,108 @@ def batch_processor(
...
+        Parameters
+        ----------
+        event_type: EventType
+            Whether this is a SQS, DynamoDB Streams, or Kinesis Data Stream event
+        model: Optional["BatchTypeModels"]
+            Parser's data model using either SqsRecordModel, DynamoDBStreamRecordModel, KinesisDataStreamRecord
+        """
+        self.event_type = event_type
+        self.model = model
+        self.batch_response = self.DEFAULT_RESPONSE
```
gmcrocetti, line 204: Hey @heitorlessa , don't know if this class is meant to be extended. Case it's not, following comment won't make much sense but anyways:
Do you think it's worth being more defensive in here by adding a deepcopy ?
```suggestion
        self.batch_response = deepcopy(self.DEFAULT_RESPONSE)
```
Just to avoid a scenario where someone extends from this class and mess with the `DEFAULT_RESPONSE` through `self.batch_response` (`self.batch_response.update({"batchItemFailures": []})` for instance ) - or any sort of changes...

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -72,7 +110,7 @@ def __call__(self, records: List[Any], handler: Callable):
         self.handler = handler
         return self
 
-    def success_handler(self, record: Any, result: Any):
+    def success_handler(self, record, result: Any) -> SuccessCallback:
```
cakepietoast, line 113: Would `SuccessResponse` be a more accurate name?

heitorlessa:
🤦‍♂️ yes! Not sure why I used Callback

heitorlessa:
Great point. It’s good to anticipate someone will subclass it sooner or
later

aws_lambda_powertools/utilities/batch/base.py
heitorlessa: YES! Good catch -- will change that

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -146,3 +185,120 @@ def batch_processor(
         processor.process()
 
     return handler(event, context)
+
+
+class BatchProcessor(BasePartialProcessor):
```
jplock, line 190: would it make sense to have separate processors for each event type (SQS, DynamoDB or Kinesis) instead of growing the complexity of this class?  Then you could encapsulate the failure collection in the specific processor.

heitorlessa: That was the initial version we wanted to implemented - a KinesisDataStreamProcessor, DynamoDB... then @cakepietoast argued that this was gonna confuse customers with other available processors (Sqs, PartialProcessor, BaseProcessor), as we can only deprecate them in v2.

I'm 50/50 here if I'm honest. I prefer a separate one but I also can see customers easily confused of which one to pick despite docs change I'm gonna make.

Implementation wise, this will remain stable. The only two changes I can anticipate is 1/ supporting the new Permanent Exception parameter, and 2/ raising a descriptive exception in case we reach an AttributeError when collecting message id/sequence number from a malformed event/model.

jplock: What if we changed the nomenclature to be “producer” and “consumer” (these processors would be consumers). I had that other idea earlier to make it easier to use the SQS and DynamoDB batch write methods taking into account their batch sizes, those could be “producers” 🤷‍♂️

heitorlessa: Where would the producer sit and what would be its responsibilities?

For that suggestion on partitioning, we **should** add it to the Event Source Data Class as it's a no brainier.

I **think** the word Consumer wouldn't be explicit enough on the capabilities Batch provide - maybe something else then?

Features

- Transform incoming records into Event Source Data Class or Pydantic Models
- Call an user defined function for each record in the batch
- Keep track of exceptions raised for any record
- For partial successes, extract message IDs (SQS) or sequence numbers (DynamoDB & Kinesis) and build a response to the new BatchItemIdentifiers contract
- Future: If a raised exception  matches the permanent_exceptions set, send these records to the configured DLQ in batches of 10

heitorlessa:
Docs finally done - @michaelbrewer @ran-isenberg could you have a quick pass before we release it tomorrow? All I need to do now is merge and start writing the release notes.

Here's what I've changed:

- [x] Updated the docs to reflect the new `BatchProcessor` in favour of `sqs_batch_processor` and `PartialSQSProcessor`
- [x] Added a Migration guide for those using `sqs_batch_processor` and `PartialSQSProcessor`
- [x] Added Caveats section in case people use Tracer response auto-capture feature with this utility in large batch sizes
- [x] Added sample event and sample response for each
- [x] Added SAM infrastructure for SQS, Kinesis and DynamoDB along with DLQ and IAM Permissions
- [x] Added an Extensibility section
- [x] Added Pydantic section
- [x] Updated Create your own processor
- [x] Updated Key features
- [x] Updated Intro
- [x] Recreated Background
- [x] Repurposed a few previous sections like Sentry integration under Legacy

michaelbrewer:
@heitorlessa  super cool. i will try to play around with it once merged

ran-isenberg:
@heitorlessa  looks great, thank you very much for putting the effort supporting both parser and data classes.

Issue #893, opened 2021-12-13
Make parameter transforms callable (and retain str capabilities)


whardier:
**Is your feature request related to a problem? Please describe.**

Allow for callable transforms to be used (as well as the str "shortcuts").

**Describe the solution you'd like**

Pass something like yaml.loads or load_my_really_cool_serialized_data callable as transform argument toward parameter store fetches.

**Describe alternatives you've considered**

Due to how feature flags works.. we are unable to use yaml or any other loader before processing happens and this needs patched.  Other approaches for parameters is to not pass a transform and then transform after the fact.

**Additional context**

PR already written up.

PR #894, opened 2021-12-13
feat(parameters): Allow callable transforms

whardier:
**Issue #, if available:**

#893 

## Description of changes:

Add in changes to base parameter classes and feature flags appconfig store to allow for callable transforms to be used as well as the available stringy references.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/utilities/feature_flags/appconfig.py
+++ b/aws_lambda_powertools/utilities/feature_flags/appconfig.py
@@ -3,12 +3,12 @@ import traceback
 from typing import Any, Dict, Optional, Union, cast
 
 from botocore.config import Config
-from aws_lambda_powertools.shared.types import AnyCallableT
 
 from aws_lambda_powertools.utilities import jmespath_utils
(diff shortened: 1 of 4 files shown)


ran-isenberg:
I'm not that having a parameter as both str and callable is the best approach. maybe have it is as string, set to custom and add another param for the callable?
also, what about error handling? what happens if the transform raises an exception?

whardier:
I don't like it either but I am working within a compatible solution.  Ideally transform would be an instance of a transform object with a .transform method and a list of compatible key suffixes.

Error handling is the same as the json and binary methods.  all exceptions lead to None unless transform errors are set to raise.

whardier:
perhaps the best approach is to register transform classes by string an key suffix and then do a lookup and execute them generically.

heitorlessa:
@whardier we appreciate the bias for action in implementing a feature you need, however, we'd like to hear from more customers that this is necessary & requirements before diving into the implementation where we can iron out maintainability, etc.

Locking this PR for now until we hear from more customers

Issue #844, opened 2021-11-22
Cache loaded JSON in event data classes


brysontyrrell:
**Is your feature request related to a problem? Please describe.**

When using the provided event data classes (mainly APIGatewayProxyEvent/V2), the `json_body` property calls `json.loads()` every time. If I am passing the event object around, and not just the body content (say between different decorated functions) that means I'm performing this parsing multiple times within the same execution.  

**Describe the solution you'd like**

I'd like to see caching introduced. Not only would this mean `json.loads()` is called only once, but in environments with high execution rates there could be significant time savings:

```python
from functools import lru_cache
import json

json_body = '{"glossary": {"title": "example glossary", "GlossDiv": {"title": "S", "GlossList": {"GlossEntry": {"ID": "SGML", "SortAs": "SGML", "GlossTerm": "Standard Generalized Markup Language", "Acronym": "SGML", "Abbrev": "ISO 8879:1986", "GlossDef": {"para": "A meta-markup language, used to create markup languages such as DocBook.", "GlossSeeAlso": ["GML", "XML"]}, "GlossSee": "markup"}}}}}'

@lru_cache()
def load_json_once():
    return json.loads(json_body)

def load_json_every():
    return json.loads(json_body)
```

The below `timeit` test is an approximation of an execution calling to load the body as JSON five times, and multiplied out across 100,000 executions. Rough, but it demonstrates the aggregate time savings off the execution's duration.

```text
>>> import timeit
>>> timeit.timeit(load_json_once, number=5)* 100000
4.252400003679213
>>> timeit.timeit(load_json_every, number=5)* 100000
9.994700008064683
>>> 
```

heitorlessa:
hey @brysontyrrell, thanks for reaching out. That's mostly feasible - Cons: 1/ memory constraints for large payloads, 2/ LRU configuration to be exposed e.g. maxsize, cache eviction, etc.

Could you share more on the use case of passing the event object around? e.g. custom middlewares? custom function coupling with lambda event? etc.

Say we we would to implement caching for certain properties, would a single object as max size be acceptable? If not, how would you like to configure it?

Thanks!!

brysontyrrell:
@heitorlessa 

Use cases are about using the object at runtime without having to reassign the content into another variable reloading the JSON again. I end up falling into the following pattern in my code (API event source) as I know calling `.json_body` is calling `json.loads()` every time.

```python
json_body = dict(event.json_body)
json_body["attr1"]
[...]

PR #909, opened 2021-12-19
feat(event-sources): cache parsed json in data class

michaelbrewer:
**Issue #, if available:**
- #844

## Description of changes:

A micro optimization to cache the parsed json within the event source data class

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/tests/functional/data_classes/test_amazon_mq.py
+++ b/tests/functional/data_classes/test_amazon_mq.py
@@ -34,6 +34,7 @@ def test_active_mq_event():
     messages = list(event.messages)
     message = messages[1]
     assert message.json_data["timeout"] == 0
+    assert message.json_data["timeout"] == 0  # cached lookup
 
 
 def test_rabbit_mq_event():
@@ -47,6 +48,7 @@ def test_rabbit_mq_event():
     assert message.data is not None
     assert message.decoded_data is not None
     assert message.json_data["timeout"] == 0
+    assert message.json_data["timeout"] == 0  # cached lookup
 
     assert isinstance(message, RabbitMessage)
     properties = message.basic_properties
--- a/tests/functional/data_classes/test_amazon_mq.py
+++ b/tests/functional/data_classes/test_amazon_mq.py
@@ -34,7 +34,7 @@ def test_active_mq_event():
     messages = list(event.messages)
     message = messages[1]
     assert message.json_data["timeout"] == 0
-    assert message.json_data["timeout"] == 0  # cached lookup
+    assert message.json_data["data"] == "CZrmf0Gw8Ov4bqLQxD4E"
 
 
 def test_rabbit_mq_event():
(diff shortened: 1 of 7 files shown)


heitorlessa (changes requested):
I'm fine with this solution - better than actual caching as UX and memory would've been tricky.

As we're gonna do it, let's revisit all other fields that do any sort of deserialisation -- Kinesis is one that does it heavy (json + b64) etc

michaelbrewer:
@heitorlessa ok. I will update others like this too.

michaelbrewer:
@heitorlessa - ok i have added various singleton caching options.

brysontyrrell:
This looks great! Looking forward to applying it.

heitorlessa:
I was gonna suggest a decorator but given the amount of fields we're fine. If we get a feature request to optimize other fields that require expensive operation like decoding & unzipping CloudWatch Logs record, or decoding and JSON parsing Kinesis records, then we should refactor this to allow for growth.

**Future note:** If more fields are required, we should convert this into a reusable one-item `@cache` class decorator. This will make it easier to maintain and support any property we might want like a base64/gzip operation, e.g. `@cache` would keep the property name as key and its value in memory, `cache["property_name"] = return_value`

tests/functional/data_classes/test_amazon_mq.py
```diff
@@ -47,6 +48,7 @@ def test_rabbit_mq_event():
     assert message.data is not None
     assert message.decoded_data is not None
     assert message.json_data["timeout"] == 0
+    assert message.json_data["timeout"] == 0  # cached lookup
```
heitorlessa, line 51: As the data is the same, what are we testing here? since that would've been the same prior to this change.

Unless you're trying to do a timer test to verify deserialization access time is different when accessing one more time

heitorlessa (changes requested):
Just one question on the usefulness of tests; happy to merge otherwise

michaelbrewer:
@heitorlessa - was thinking of a `@cache` decorator, might end up in the future or just use a builtin python3 feature in the future. it is minor right now right?

heitorlessa:
> @heitorlessa - was thinking of a `@cache` decorator, might end up in the future or just use a builtin python3 feature in the future. it is minor right now right?
> 
> 

Like this but we don't need the threading safety I believe, as it'll compute once and more lightweight and faster than the LRU cache - once we deprecate older Python versions we can use it natively.

We would need to test whether that work for our data types tho, but should be easily adjustable as the logic is simple


https://github.com/python/cpython/blob/e9a01e231aae19fd1249368e477a60bc033b2646/Lib/functools.py#L955

heitorlessa:
Merging as is - I'll look into the hash dunder and decorator when Mypy is fully supported across the codebase

heitorlessa (approved):

PR #926, opened 2021-12-23
fix(apigateway): support @app.not_found() syntax & housekeeping

michaelbrewer:
**Issue #, if available:**

- https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/65

## Description of changes:

Changes:
- Allow for `@app.not_found()` decorator
- Add docstrings and typing to `strtobool`
- Add code coverage for  `strtobool`
- Minor docstring typos
- Add test for DynamoDBPersistenceLayer with a `sort_key_attr`

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -579,7 +579,7 @@ class ApiGatewayResolver(BaseRouter):
     @staticmethod
     def _path_starts_with(path: str, prefix: str):
         """Returns true if the `path` starts with a prefix plus a `/`"""
-        if not isinstance(prefix, str) or len(prefix) == 0:
+        if not isinstance(prefix, str) or prefix == "":
             return False
 
         return path.startswith(prefix + "/")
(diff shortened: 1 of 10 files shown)


aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -633,7 +633,9 @@ def _call_route(self, route: Route, args: Dict[str, str]) -> ResponseBuilder:
 
             raise
 
-    def not_found(self, func: Callable):
+    def not_found(self, func: Optional[Callable] = None):
+        if func is None:
```
heitorlessa, line 637: for future note, if this method ever supports keyword arguments, `func` will also be None and will require a change to L638

example: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/tracing/tracer.py#L286

aws_lambda_powertools/shared/functions.py
```diff
@@ -1,14 +1,21 @@
 from typing import Any, Optional, Union
 
 
-def strtobool(value):
+def _strtobool(value: str) -> bool:
```
heitorlessa, line 4: please return to `def strtobool` instead of a private one here. This might be useful to others in the future when we have a `commons` utility in v2 or v3

heitorlessa (changes requested):
minor change in reverting `strtobool` to be a public function

heitorlessa (approved):

heitorlessa:
Thanks a lot Mike for the quick fix & the housekeeping

PR #927, opened 2021-12-23
feat(logger): clone powertools logger config to any Python logger

mploski:
**Issue #, if available: https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/40 
## Description of changes:

This feature allows customers to clone Powertools Logger configuration like level & handler to any registered Python loggers - exclusion, inclusion-only are also supported.

* [x] Update tests
* [x] Update docs
* [x] PR title follows [conventional commit semantics]

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/logging/utils.py
+++ b/aws_lambda_powertools/logging/utils.py
@@ -1,0 +1,30 @@
+import logging
+from typing import List, Optional, TypeVar
+
+from .logger import Logger
+
+PowertoolsLogger = TypeVar("PowertoolsLogger", bound=Logger)
+
+
+def copy_config_to_registered_loggers(
+    source_logger: PowertoolsLogger, exclude: Optional[List[str]] = None, include: Optional[List[str]] = None
+):
+    root_loggers = [
+        logging.getLogger(name)
+        for name in logging.root.manager.loggerDict
+        if "." not in name and name != source_logger.name
+    ]
+    source_logger.debug(f"Found registered root loggers: {root_loggers}")
+    if include and not exclude:
+        root_loggers = [logger for logger in root_loggers if logger.name in include]
+    elif not include and exclude:
+        root_loggers = [logger for logger in root_loggers if logger.name not in exclude]
+    elif include and exclude:
+        root_loggers = [logger for logger in root_loggers if logger.name in include and logger.name not in exclude]
+
+    source_logger.debug(f"Filtered root loggers: {root_loggers}")
+    for logger in root_loggers:
+        logger.handlers = []
+        logger.setLevel(source_logger.level)
+        for source_handler in source_logger.handlers:
+            logger.addHandler(source_handler)
--- a/aws_lambda_powertools/logging/utils.py
+++ b/aws_lambda_powertools/logging/utils.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List, Optional, TypeVar
+from typing import Callable, List, Optional, TypeVar
 
 from .logger import Logger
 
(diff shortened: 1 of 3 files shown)


aws_lambda_powertools/logging/utils.py
```diff
@@ -0,0 +1,30 @@
+import logging
+from typing import List, Optional, TypeVar
+
+from .logger import Logger
+
+PowertoolsLogger = TypeVar("PowertoolsLogger", bound=Logger)
+
+
+def copy_config_to_registered_loggers(
+    source_logger: PowertoolsLogger, exclude: Optional[List[str]] = None, include: Optional[List[str]] = None
```
heitorlessa, line 10: Missing the log level parameter

aws_lambda_powertools/logging/utils.py
```diff
@@ -0,0 +1,30 @@
...
+):
+    root_loggers = [
```
heitorlessa, line 12: Use a more explicit name to avoid confusion in the future with actual root logger.

```suggestion
    registered_loggers = [
```

heitorlessa (changes requested):
quick initial review as it's missing the log level parameter.

They should be able to override log level when copying the config over

aws_lambda_powertools/logging/utils.py
mploski: Good catch, missed that part.Do you think we should fallback to source_logger log level when the level parameter is not provided or we should skip modifying registered logger level fully?

aws_lambda_powertools/logging/utils.py
mploski: Done, also modified corresponding internal methods

aws_lambda_powertools/logging/utils.py
heitorlessa: > Do you think we should fall back to source_logger log level when the level parameter is not provided

Yes! That was the original behaviour :) 

```python
def copy_config_to_registered_loggers(
    source_logger: Logger,
    log_level: str = None,
    include: Optional[List[str]] = None,
    exclude: Optional[List[str]] = None,
):
    """Copies source Logger level and handler to all registered loggers for consistent formatting.

    Parameters
    ----------
    source_logger : Logger
        Powertools Logger to copy configuration from
        
    HERE
    log_level : str, optional
        Logging level to set to registered loggers, by default uses source_logger logging level
```

aws_lambda_powertools/logging/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    exclude: Optional[List[str]] = None,
+    include: Optional[List[str]] = None,
+) -> None:
+    """Enable powertools logging for imported libraries.
+
+    Attach source logger handlers to external loggers.
+    Modify logger level based on source logger attribute.
+    Ensure powertools logger itself is excluded from registered list.
+    """
+
```
heitorlessa, line 20: Could you re-add the algorithm as a comment? This will help anyone else coming back to this code to understand the multi-nested conditional

mploski: done

aws_lambda_powertools/logging/utils.py
mploski: modified code to reflect requirement

docs/core/logger.md
```diff
@@ -1062,6 +1062,26 @@ for the given name and level to the logging module. By default, this logs all bo
...
+    from aws_lambda_powertools import Logger
+    from aws_lambda_powertools.logging import utils
+
+    logger = Logger()
+
+    external_logger = logging.logger()
+
+    utils.copy_config_to_registered_loggers(source_logger=logger)
+    external_logger.info("test message")
+    ```
```
heitorlessa, line 1083:
```suggestion
```python hl_lines="10" title="Cloning Logger config to all other registered standard loggers"
import logging

from aws_lambda_powertools.logging import utils

logger = Logger()

external_logger = logging.logger()

utils.copy_config_to_registered_loggers(source_logger=logger)
external_logger.info("test message")
```
```

heitorlessa (approved):

Issue #928, opened 2021-12-26
Null type becomes BOOL


sergiors:
**Is your feature request related to a problem? Please describe.**

I did a stream to copy dynamodb to elasticsearch, I used the object `DynamoDBStreamEvent`, until that okey,
but when I try get the values, the null type becomes a boolean type.

**Describe the solution you'd like**

I read the [source-code](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py#L109) the `AttributeValue.null_value` and think the null value should back to python native value, like described in [ boto3 types ](https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/types.html#Binary)

**Describe alternatives you've considered**

**Additional context**

As alternative to fix the stream copy, the only value that I need to check is the null value.
```python
{k: v.get_value if not v.null_value else None for k, v in dct.items()}
```

michaelbrewer:
Been a while since i did this, but if this is for deserialization, then `"NULL": true` would map to Python `None`. 

I will double check @heitorlessa

michaelbrewer:
@sergiors @heitorlessa 
Ok i made a PR to fix this (https://github.com/awslabs/aws-lambda-powertools-python/pull/929). It would be a breaking change (_if you where expecting `null_value` to return a boolean_), however this was the wrong behavior.

FYI, `get_value` is a convenient way to get the value out.

PR #929, opened 2021-12-27
fix(event-sources): handle dynamodb null type as none, not bool

michaelbrewer:
**Issue #, if available:**
- #928

## Description of changes:

Return for `None` for null types:

Example: 

```json5
{
  "Records": [
    {
      ...
      "dynamodb": {
        "OldImage": {
          ...
          "Message": {
            "S": "New item!"
          },
          "Id": {
            "N": "101"
          }
        },
        "Keys": {
          "Id": {
            "N": "101"
          }
        },
        "NewImage": {
          ...
          "Message": {
            "NULL": true
          },
          "Id": {
            "N": "101"
          }
        },
      }
    }
  ]
}
```

Below `null_value` and `get_value` would have returned `True`, and now they return None.

```python3
from aws_lambda_powertools.utilities.data_classes.dynamo_db_stream_event import (
    DynamoDBStreamEvent,
)

def handler(event: DynamoDBStreamEvent, context):
    record = next(event.records)
    message_value = record.dynamodb.new_image.get("message")

    # get_value would be the preferred way to get the underlying value regardless of type
    assert message_value.get_value is None

    # Otherwise when you want to use `null_value` you should first change against `get_type`
    if message_value.get_type == AttributeValueType.Null:
        assert message_value.null_value is None
```

**Checklist**

 

* [X] Update tests
* [X] Update docs

## Breaking change checklist

Before `null_value` was returning a boolean,  `True` when `NULL` value was set to `true` and `False` when `NULL` value was not set (or was `False`)

### DIFF
--- a/tests/functional/test_data_classes.py
+++ b/tests/functional/test_data_classes.py
@@ -544,7 +544,17 @@ def test_dynamo_attribute_value_null_value():
     attribute_value = AttributeValue(example_attribute_value)
 
     assert attribute_value.get_type == AttributeValueType.Null
+    assert attribute_value.null_value is None
     assert attribute_value.null_value == attribute_value.get_value
+
+
+def test_dynamo_attribute_value_null_value_invalid():
+    example_attribute_value = {"NULL": False}
+
+    attribute_value = AttributeValue(example_attribute_value)
+
+    assert attribute_value.get_type == AttributeValueType.Null
+    assert attribute_value.null_value is None
 
 
 def test_dynamo_attribute_value_s_value():
--- a/tests/functional/test_data_classes.py
+++ b/tests/functional/test_data_classes.py
@@ -548,15 +548,6 @@ def test_dynamo_attribute_value_null_value():
     assert attribute_value.null_value == attribute_value.get_value
 
 
-def test_dynamo_attribute_value_null_value_invalid():
-    example_attribute_value = {"NULL": False}
-
-    attribute_value = AttributeValue(example_attribute_value)
-
-    assert attribute_value.get_type == AttributeValueType.Null
-    assert attribute_value.null_value is None
-
-
 def test_dynamo_attribute_value_s_value():
     example_attribute_value = {"S": "Hello"}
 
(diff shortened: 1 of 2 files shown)


tests/functional/test_data_classes.py
```diff
@@ -544,9 +544,19 @@ def test_dynamo_attribute_value_null_value():
     attribute_value = AttributeValue(example_attribute_value)
 
     assert attribute_value.get_type == AttributeValueType.Null
+    assert attribute_value.null_value is None
     assert attribute_value.null_value == attribute_value.get_value
 
 
+def test_dynamo_attribute_value_null_value_invalid():
+    example_attribute_value = {"NULL": False}
```
heitorlessa, line 552: shouldn't this be `True`?

https://github.com/boto/boto3/blob/7a10e21c452e0cb0a6e3d245f7389f1500d2abcf/tests/unit/dynamodb/test_types.py#L69

michaelbrewer: @michaelbrewer - We are deserializing,  `{"NULL": False}` should never happen (so should we raise a `ValueError` or just ignore that?)

michaelbrewer: @heitorlessa - i can remove that test, i just wanted it to be clear that we don't valid 

```
    @property
    def null_value(self) -> None:
        """An attribute of type Null.

        Example:
            >>> {"NULL": True}
        """
        return None
```

michaelbrewer: which is the same as:

```
    def _deserialize_null(self, value):
        return None
```

michaelbrewer:
> Question on whether `NULL` will ever be `False`, boto tests always caters for `True`
> 
> Type table only considers `True` so perhaps DynamoDB doesn't compute that: [boto/boto3@`cf62d26`/boto3/dynamodb/types.py#L85](https://github.com/boto/boto3/blob/cf62d26cb0f28357c935d541ed3f0dae196cd26d/boto3/dynamodb/types.py#L85)

Because boto does not allow you to call `_deserialize_null` directly but we allow for it via `null_value`

heitorlessa (approved):

PR #943, opened 2022-01-10
feat(mypy): complete mypy support for the entire codebase

mploski:
**Issue #, if available:**
https://github.com/awslabs/aws-lambda-powertools-roadmap/issues/3

## Description of changes:
* Fix all typing errors discovered by mypy
* Enable mypy in CI

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -30,7 +30,7 @@ jobs:
         run: make dev
       - name: Formatting and Linting
         run: make lint
-      - name: Typing
+      - name: Static type checking
         run: make mypy
       - name: Test with pytest
         run: make test
--- a/.github/workflows/python_build.yml
+++ b/.github/workflows/python_build.yml
@@ -30,6 +30,8 @@ jobs:
         run: make dev
       - name: Formatting and Linting
         run: make lint
+      - name: Typing
+        run: make mypy
       - name: Test with pytest
         run: make test
       - name: Security baseline
--- a/aws_lambda_powertools/utilities/batch/sqs.py
+++ b/aws_lambda_powertools/utilities/batch/sqs.py
@@ -5,10 +5,12 @@ Batch SQS utilities
 """
 import logging
 import sys
-from typing import Callable, Dict, List, Optional, Tuple
+from typing import Callable, Dict, List, Optional, Tuple, cast
 
 import boto3
 from botocore.config import Config
+
+from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
 
 from ...middleware_factory import lambda_handler_decorator
 from .base import BasePartialProcessor
@@ -84,11 +86,14 @@ class PartialSQSProcessor(BasePartialProcessor):
         *_, account_id, queue_name = self.records[0]["eventSourceARN"].split(":")
         return f"{self.client._endpoint.host}/{account_id}/{queue_name}"
 
-    def _get_entries_to_clean(self) -> List:
+    def _get_entries_to_clean(self) -> List[Dict[str, str]]:
         """
         Format messages to use in batch deletion
         """
-        return [{"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]} for msg in self.success_messages]
+        return [
+            {"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]}
+            for msg in cast(List[SQSRecord], self.success_messages)
+        ]
 
     def _process_record(self, record) -> Tuple:
         """
--- a/aws_lambda_powertools/utilities/batch/sqs.py
+++ b/aws_lambda_powertools/utilities/batch/sqs.py
@@ -90,10 +90,13 @@ class PartialSQSProcessor(BasePartialProcessor):
         """
         Format messages to use in batch deletion
         """
-        return [
-            {"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]}
-            for msg in cast(List[SQSRecord], self.success_messages)
-        ]
+        entries = []
+        # success_messages has generic type of union of SQS, Dynamodb and Kinesis Streams records or Pydantic models.
+        # Here we get SQS Record only
+        messages = cast(List[SQSRecord], self.success_messages)
+        for msg in messages:
+            entries.append({"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]})
+        return entries
 
     def _process_record(self, record) -> Tuple:
         """
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -40,7 +40,7 @@ class DataRecord:
         idempotency_key,
         status: str = "",
         expiry_timestamp: Optional[int] = None,
-        response_data: str = "",
+        response_data: Optional[str] = "",
         payload_hash: Optional[str] = None,
     ) -> None:
         """
@@ -92,16 +92,16 @@ class DataRecord:
         else:
             raise IdempotencyInvalidStatusError(self._status)
 
-    def response_json_as_dict(self) -> dict:
+    def response_json_as_dict(self) -> Optional[dict]:
         """
         Get response data deserialized to python dict
 
         Returns
         -------
-        dict
+        Optional[dict]
             previous response data deserialized
         """
-        return json.loads(self.response_data)
+        return json.loads(self.response_data) if self.response_data else None
 
 
 class BasePersistenceLayer(ABC):
@@ -121,7 +121,6 @@ class BasePersistenceLayer(ABC):
         self.raise_on_no_idempotency_key = False
         self.expires_after_seconds: int = 60 * 60  # 1 hour default
         self.use_local_cache = False
-        self._cache: Optional[LRUDict] = None
         self.hash_function = None
 
     def configure(self, config: IdempotencyConfig, function_name: Optional[str] = None) -> None:
@@ -279,14 +278,14 @@ class BasePersistenceLayer(ABC):
         -------
 
         """
-        if not self.use_local_cache or self._cache is None:
+        if not self.use_local_cache:
             return
         if data_record.status == STATUS_CONSTANTS["INPROGRESS"]:
             return
         self._cache[data_record.idempotency_key] = data_record
 
     def _retrieve_from_cache(self, idempotency_key: str):
-        if not self.use_local_cache or self._cache is None:
+        if not self.use_local_cache:
             return
         cached_record = self._cache.get(key=idempotency_key)
         if cached_record:
@@ -296,7 +295,7 @@ class BasePersistenceLayer(ABC):
             self._delete_from_cache(idempotency_key=idempotency_key)
 
     def _delete_from_cache(self, idempotency_key: str):
-        if not self.use_local_cache or self._cache is None:
+        if not self.use_local_cache:
             return
         if idempotency_key in self._cache:
             del self._cache[idempotency_key]
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -40,7 +40,7 @@ class DataRecord:
         idempotency_key,
         status: str = "",
         expiry_timestamp: Optional[int] = None,
-        response_data: Optional[str] = "",
+        response_data: str = "",
         payload_hash: Optional[str] = None,
     ) -> None:
         """
@@ -279,14 +279,14 @@ class BasePersistenceLayer(ABC):
         -------
 
         """
-        if not self.use_local_cache:
+        if not self.use_local_cache or self._cache is None:
             return
         if data_record.status == STATUS_CONSTANTS["INPROGRESS"]:
             return
         self._cache[data_record.idempotency_key] = data_record
 
     def _retrieve_from_cache(self, idempotency_key: str):
-        if not self.use_local_cache:
+        if not self.use_local_cache or self._cache is None:
             return
         cached_record = self._cache.get(key=idempotency_key)
         if cached_record:
@@ -296,7 +296,7 @@ class BasePersistenceLayer(ABC):
             self._delete_from_cache(idempotency_key=idempotency_key)
 
     def _delete_from_cache(self, idempotency_key: str):
-        if not self.use_local_cache:
+        if not self.use_local_cache or self._cache is None:
             return
         if idempotency_key in self._cache:
             del self._cache[idempotency_key]
--- a/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
@@ -130,7 +130,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
             idempotency_key=item[self.key_attr],
             status=item[self.status_attr],
             expiry_timestamp=item[self.expiry_attr],
-            response_data=item.get(self.data_attr),
+            response_data=item.get(self.data_attr, ""),
             payload_hash=item.get(self.validation_key_attr),
         )
 
--- a/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
@@ -130,7 +130,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
             idempotency_key=item[self.key_attr],
             status=item[self.status_attr],
             expiry_timestamp=item[self.expiry_attr],
-            response_data=item.get(self.data_attr, ""),
+            response_data=item.get(self.data_attr),
             payload_hash=item.get(self.validation_key_attr),
         )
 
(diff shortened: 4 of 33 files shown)


mploski:
@heitorlessa github ci failing on the last error: https://github.com/mploski/aws-lambda-powertools-python/runs/4759867427?check_suite_focus=true.

I’d like to discuss a potential solution to this. In general mypy complains on subclass overwriting superclass get method with different signatures, which I believe breaks Liskov Substitution Principle.  Other errors were solved already.

mploski:
> @heitorlessa github ci failing on the last error: https://github.com/mploski/aws-lambda-powertools-python/runs/4759867427?check_suite_focus=true.
> 
> I’d like to discuss a potential solution to this. In general mypy complains on subclass overwriting superclass get method with different signatures, which I believe breaks Liskov Substitution Principle. Other errors were solved already.

Since any change to this method's spec may be breaking to users and we cannot also move get method declaration in Base class to subclass for the same reason decided to silent it and accept potential risk

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -279,14 +279,14 @@ def _save_to_cache(self, data_record: DataRecord):
         -------
 
         """
-        if not self.use_local_cache:
+        if not self.use_local_cache or self._cache is None:
```
heitorlessa, line 282: That doesn't seem right. This is a function that saves it to the cache. By fixing Mypy hint, you're introducing a new bug. 

This line reads as: If customer enables cache BUT the cache is empty, just return.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -130,7 +130,7 @@ def _item_to_data_record(self, item: Dict[str, Any]) -> DataRecord:
             idempotency_key=item[self.key_attr],
             status=item[self.status_attr],
             expiry_timestamp=item[self.expiry_attr],
-            response_data=item.get(self.data_attr),
+            response_data=item.get(self.data_attr, ""),
```
heitorlessa, line 133: Could you document what this addresses? Concern: besides type annotations, we're changing code that might have been a subtle bug OR introducing a bug we haven't covered in tests either -- it's hard to know and review.

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -87,15 +87,15 @@ def __init__(self, config: Optional[Config] = None, boto3_session: Optional[boto
 
         super().__init__()
 
-    def get(
+    def get(  # type: ignore[override]
```
heitorlessa, line 90: would be great to have a comment on the need for override so anyone can understand when coming back to the code.

aws_lambda_powertools/utilities/parser/envelopes/kinesis.py
```diff
@@ -38,5 +38,6 @@ def parse(self, data: Optional[Union[Dict[str, Any], Any]], model: Type[Model])
         parsed_envelope: KinesisDataStreamModel = KinesisDataStreamModel.parse_obj(data)
         logger.debug(f"Parsing Kinesis records in `body` with {model}")
         return [
-            self._parse(data=record.kinesis.data.decode("utf-8"), model=model) for record in parsed_envelope.Records
+            self._parse(data=cast(bytes, record.kinesis.data).decode("utf-8"), model=model)
```
heitorlessa, line 41: can we solve without casting?

aws_lambda_powertools/utilities/parser/envelopes/sns.py
```diff
@@ -69,6 +69,6 @@ def parse(self, data: Optional[Union[Dict[str, Any], Any]], model: Type[Model])
         parsed_envelope = SqsModel.parse_obj(data)
         output = []
         for record in parsed_envelope.Records:
-            sns_notification = SnsNotificationModel.parse_raw(record.body)
+            sns_notification: SnsNotificationModel = SnsNotificationModel.parse_raw(cast(str, record.body))
```
heitorlessa, line 72: can we solve without casting?

pyproject.toml
```diff
@@ -54,6 +54,7 @@ mkdocs-material = "^7.3.6"
 mkdocs-git-revision-date-plugin = "^0.3.1"
 mike = "^0.6.0"
 mypy = "^0.931"
+types-dataclasses = "^0.6.4"
```
heitorlessa, line 57: could you explain why this is needed? If it's needed, are we lazy importing only during TYPE_CHECK phase?

heitorlessa (changes requested):
Thanks a lot @mploski - it covers a lot of ground and tech debt.

I'd like to ask you to consider two key aspects:

* **Consider `cast` your last resource**. `cast` is gonna increase the execution time for end-customers, and it slowly litters the code (similar to type coercion in other languages). Sometimes propagating the type, a generic, or creating another variable and an explicit type is a solution.
* **Consider documenting beyond typing changes**. I've noticed that we seem to be introducing a new bug in Idempotency by fixing a type, as we're changing a branching logic too. There are major block changes in Parameters too. For quick notes, use the PR description, and for larger ones like parameters comment on why this is necessary. Mypy can't see beyond the types, and while we have multiple checks in place we might slip a regression here. By changing that code, it'd be great to check whether additional tests are necessary too as we might have not covered that.

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -92,12 +94,12 @@ def get(
...
 
-        if transform is not None:
+        if transform:
             if isinstance(value, bytes):
                 value = value.decode("utf-8")
             value = transform_value(value, transform)
-
-        self.store[key] = ExpirableValue(value, datetime.now() + timedelta(seconds=max_age))
+            if value:
+                self.store[key] = ExpirableValue(value, datetime.now() + timedelta(seconds=max_age))
```
heitorlessa, line 102: Recommendation generated by Amazon CodeGuru Reviewer. Leave feedback on this recommendation by replying to the comment or by reacting to the comment using emoji.

The naive datetime objects are treated by many datetime methods as local times, it is preferred to use aware datetimes to represent times in UTC. The recommended way to create an aware datetime object representing a specific timestamp in UTC is by passing `tzinfo` as an argument to the method.

[Learn more](https://docs.python.org/3/library/datetime.html#aware-and-naive-objects)

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -84,11 +86,14 @@ def _get_queue_url(self) -> Optional[str]:
...
 
-    def _get_entries_to_clean(self) -> List:
+    def _get_entries_to_clean(self) -> List[Dict[str, str]]:
         """
         Format messages to use in batch deletion
         """
-        return [{"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]} for msg in self.success_messages]
+        return [
+            {"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]}
+            for msg in cast(List[SQSRecord], self.success_messages)
```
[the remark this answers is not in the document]
mploski: Yes, by removing dict comprehension I can set expected type. Will work on it

aws_lambda_powertools/utilities/idempotency/persistence/base.py
mploski: This validation checks if self_cache is None not empty dict. 
The dict is empty when you set `use_local_cache`, so  there is no chance of having use_local_cache variable set and self_cache set to None (https://github.com/awslabs/aws-lambda-powertools-python/blob/0196883410bf4fadeb1ca27f3f331ec6fbaffbe4/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L157). Mypy doesn't know it and raise an issue that you cannot add dict key to None value (not dict). That's why I added this check to make it explicit. 
Anyway, if `_self_cache` was None code would fail with `TypeError: 'NoneType' object does not support item assignment` because we would try to add key to None value here (https://github.com/awslabs/aws-lambda-powertools-python/blob/0196883410bf4fadeb1ca27f3f331ec6fbaffbe4/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L286) which is not correct.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
mploski: So the reasoning is as following:
Previously we allowed `response_data` to be None. In class itself we define response data as
`response_data: Optional[str] = ""`. Which means that if we don't provide this value we get empty string, but if we provide a None( as with line 133) it will be accepted. This will break if user use `response_json_as_dict` method https://github.com/awslabs/aws-lambda-powertools-python/blob/0196883410bf4fadeb1ca27f3f331ec6fbaffbe4/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L95 with `TypeError: the JSON object must be str, bytes or bytearray, not NoneType` . Because of that I decided to set response_data to empty string instead of None here.

Different approach I can take is to change behaviour of `response_json_as_dict` and validate if input is not None and only then try to create json from it. What do you think @heitorlessa ?

aws_lambda_powertools/utilities/parameters/ssm.py
mploski: Ok!

aws_lambda_powertools/utilities/parser/envelopes/kinesis.py
mploski: Yes, I can explicitly declare local variable and check if record.kinesis.data is byte type and only then do this translation - otherwise do something else. What should we do in case that user pass another data  model in data key as allowed here: 
https://github.com/awslabs/aws-lambda-powertools-python/blob/0196883410bf4fadeb1ca27f3f331ec6fbaffbe4/aws_lambda_powertools/utilities/parser/models/kinesis.py#L17

pyproject.toml
mploski: This package is needed by mypy to work correctly with python 3.6 due to:
```
Run make mypy
poetry run mypy --pretty aws_lambda_powertools
aws_lambda_powertools/utilities/idempotency/base.py:31:1: error: Library stubs
not installed for "dataclasses" (or incompatible with Python 3.6)  [import]
            import dataclasses
    ^
aws_lambda_powertools/utilities/idempotency/base.py:31:1: note: Hint: "python3 -m pip install types-dataclasses"
aws_lambda_powertools/utilities/idempotency/base.py:31:1: note: (or run "mypy --install-types" to install all missing stub packages)
aws_lambda_powertools/utilities/idempotency/base.py:31:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 125 source files)

make: *** [Makefile:91: mypy] Error 1
Error: Process completed with exit code 2.
```
Great idea with Lazy loading will add it.

aws_lambda_powertools/utilities/parser/envelopes/sns.py
mploski: Sure, I can explicitly check if body is of type str and do this parsing then and fail if its of type `Type[BaseModel]`

aws_lambda_powertools/utilities/parameters/base.py
mploski: I don't think it is relevant here. We use naive datetime object but still we compare those dates in the same timezone (region) so there is no risk of comparing two diferent datetime object with different timezones leading to wrong ttl's values.
@heitorlessa Please advise if you see any other problems here and what is your recommandation.

heitorlessa: What!?!? I didn't post this HAHA. ignore it.

pyproject.toml
heitorlessa: I see! Because it's Py3.6, it's EOL, and it's conditional import we can simply ignore the type `mypy.ini` 

```ini
[mypy-dataclasses]
ignore_missing_imports = True
```

aws_lambda_powertools/utilities/parser/envelopes/sns.py
heitorlessa: Yeah I see why! Let's add a temporary variable with casting. I was wrong about `cast` as it's [only returning the value](https://github.com/python/cpython/blob/main/Lib/typing.py#L1712) now not like `str()`. 

```python
            # because we allow either AWS expected contract (str) or a custom Model, see #943
            d = cast(str, record.body)
            sns_notification = SnsNotificationModel.parse_raw(d)
```

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
heitorlessa: Gotcha!! It'd be best to handle that within `response_json_as_dict` because if we add another persistence layer like Redis then we're solved ;)

aws_lambda_powertools/utilities/idempotency/persistence/base.py
heitorlessa: I see, lemme have a second look at this and I'll come back to you. We're doing this within `configure()` but MyPy isn't smart enough to check this

pyproject.toml
mploski: Ok, lets remove this package and ignore dataclassess errors

aws_lambda_powertools/utilities/parser/envelopes/sns.py
mploski: ok, lets create temp variable. Thx!

aws_lambda_powertools/utilities/parser/envelopes/kinesis.py
mploski: Refactor code to use temp variable with comment:

```
        for record in parsed_envelope.Records:
            # We allow either AWS expected contract (str) or a custom Model, see #943
            body = cast(str, record.body)
            sns_notification = SnsNotificationModel.parse_raw(body)
            output.append(self._parse(data=sns_notification.Message, model=model))
```

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
mploski: Yep, make sense to solve it once and for all :-)

```
    def response_json_as_dict(self) -> Optional[dict]:
        """
        Get response data deserialized to python dict

        Returns
        -------
        Optional[dict]
            previous response data deserialized
        """
        return json.loads(self.response_data) if self.response_data else None
```

aws_lambda_powertools/utilities/batch/sqs.py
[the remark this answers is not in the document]
mploski: Based on discussion rewrote to have explicit cast on temp variable:
```
    def _get_entries_to_clean(self) -> List[Dict[str, str]]:
        """
        Format messages to use in batch deletion
        """
        entries = []
        # We allow success_messages to be union of SQS, Dynamodb and Kinesis Streams records or corresponding Pydantic models. 
        # Here we get SQS Record only
        messages = cast(List[SQSRecord], self.success_messages)
        for msg in messages:
            entries.append( {"Id": msg["messageId"], "ReceiptHandle": msg["receiptHandle"]})
        return entries
```

aws_lambda_powertools/utilities/idempotency/persistence/base.py
mploski: Hmmm, I think mypy has this problem since we explicitly set this internal variable in init method:
```
self._cache: Optional[LRUDict] = None
```
Because of it mypy will not allow ungarded None in methods that use it.
https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type

Since we rely on `self.use_local_cache` that is always initialised to False and modified only when Configure() method is called with specific config AND we always check against it's value in every method that also would try to use _cache we can safely remove `self._cache: Optional[LRUDict] = None` from __init__ method.

This way mypy stops complaining about it and we don't need to add ignore in 4 places. WDYT @heitorlessa ?

aws_lambda_powertools/utilities/parameters/ssm.py
mploski: Added this comments:
```
    # We break Liskov substitution principle due to differences between signature of this method and corresponding method in superclass.
    # We ignore mypy error, as change to the signature is a breaking change to users.
```

mploski:
@heitorlessa  Added few tests to fix tests coverage

tests/functional/idempotency/test_idempotency.py
```diff
@@ -648,6 +648,29 @@ def test_data_record_invalid_status_value():
     assert e.value.args[0] == "UNSUPPORTED_STATUS"
 
 
+def test_data_record_json_to_dict_mapping():
+    # GIVEN a data record with status "COMPLETED" and provided response data
```
heitorlessa, line 652:
```suggestion
    # GIVEN a data record with status "INPROGRESS" and provided response data
```

.github/workflows/python_build.yml
```diff
@@ -30,6 +30,8 @@ jobs:
         run: make dev
       - name: Formatting and Linting
         run: make lint
+      - name: Typing
```
heitorlessa, line 33:
```suggestion
      - name: Static type checking
```

heitorlessa (approved):

heitorlessa:
Added the last bits - thank you so so much again @mploski ;) A new era for Powertools. We must also keep a close eye after the release to see if this accidentally introduced any new bug that our checks + peer review wasn't able to detect.

Issue #953, opened 2022-01-12
Feature request: Add support for middlewares in event handler


BVMiko:
**Is your feature request related to a problem? Please describe.**
Event handlers (specifically ApiGatewayResolver) are incompatible with middleware decorators.  The `lambda_handler_decorator` is almost exactly what would be needed, except for the function signature.

**Describe the solution you'd like**
I'd like to be able to add middleware when using a router, and for the middleware to have access to the processed event:

```
@router_decorator
def db_middleware(handler, router, **kwargs):
    db = router.current_event.get_query_string_value(name="dbname", default_value='default_value')
    return handler(db=db, **kwargs)
    
@router.get("/users")
@db_middleware(router=router)
def get_users(db, foo) -> Dict:
    pagination_limit = router.current_event.get_query_string_value(name="limit", default_value=10)
    return {"pagination_limit": pagination_limit, "db": db}
```

**Describe alternatives you've considered**
Using a regular `lambda_handler_decorator` outside of the `router.get` is executed even when the router isn't called.  The decorator wrapper could be stored in individual project code, but since it's so similar to the existing one and would be useful for any router usage, it makes sense to consider it for the `aws-lambda-powertools` library. 

**Additional context**

heitorlessa:
Hey @BVMiko - thanks for opening this feature request.

Perhaps a generic middleware factory decorator for any Python function with an arbitrary number of args, kwargs?

That would help make it more explicit so we have one for Lambda and for any function - it also makes it easier to preserve types and function signature (ParamSpec is only supported in later Python versions)

It's something I wanted to do for a long time but couldn't prioritise earlier 

Thanks a lot btw!

theNoobExpert:
@heitorlessa Here is what would make sense for middleware : 
Defining middleware : 

```python
        def auth_middleware(app, handler):

        ### auth logic here ###

        setattr(app.current_event, "user", user_obj) # or app.current_event.extras["user"] = user_obj - a well defined place through which extra information can be passed on to routes.

        return handler()  # I haven't seen the code base so I don't know how difficult/easy it would be to implement this`
```

Registering middleware : 

app.register_middleware(auth_middleware)

heitorlessa:
Thanks a lot @karanhere3962 -- that's similar to FastAPI.

As this would run for every request, what's the UX you have in mind for a middleware that only runs on a given path?

Example:

- Register as part of route definition: `@app.get("/path", middlewares=[]`
[...]

PR #954, opened 2022-01-12
feat(middleware_factory): support router

BVMiko:
**Issue #, if available:**
#953 

## Description of changes:
The changes needed for `lambda_handler_decorator` to support routers are only to modify the arguments of the inner wrapper and its invocation, so I modified it to accept arbitrary arguments.

Due to the difference in signature (and function name), it may be preferable to instead move this to a clone named `router_decorator` instead of modifying the original `lambda_handler_decorator`.  If you would like me to do so, let me know and I can make the change.

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist
There may be potential issues with positional arguments if we reuse the same decorator.

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/middleware_factory/factory.py
+++ b/aws_lambda_powertools/middleware_factory/factory.py
@@ -122,9 +122,9 @@ def lambda_handler_decorator(decorator: Optional[Callable] = None, trace_executi
             )
 
         @functools.wraps(func)
-        def wrapper(event, context):
+        def wrapper(**kwargs_wrapper):
             try:
-                middleware = functools.partial(decorator, func, event, context, **kwargs)
+                middleware = functools.partial(decorator, func, **(kwargs_wrapper | kwargs))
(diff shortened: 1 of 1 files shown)


heitorlessa (changes requested):
Could you duplicate the decorator factory into a new function named `decorator_factory`?

That will prevent confusing anyone reading someone's else coding that uses both of them in different ways. It's also easier to update the docs with a new section as the more explicit we are.

Thank you so much.

heitorlessa:
following up in case you missed the comment above

leandrodamascena:
Hi everyone! We are working on issue #1955 to add router support to Middleware Factory and other changes.

As this PR has not been updated for some time and issue #1955 will add this support and some other improvements, so we are closing this PR.

Thank you all and if you think that something can be done in this PR, feel free to reopen it.

PR #978, opened 2022-01-26
feat(event-handler): new resolvers to fix current_event typing

### LINKED ISSUE #761: Event Handler: API Gateway current_event doesn't show all properties
When trying to access `request_context` when using either API Gateway v1 or v2 we can't have type hinting due to ALB being different - both inherit from [BaseProxyEvent](https://github.com/awslabs/aws-lambda-powertools-python/blob/d02dd2eed07f3d67ca0c02437c6a31b3360b5b41/aws_lambda_powertools/utilities/data_classes/common.py#L47).

Overloading the internal method [_to_proxy_event](https://github.com/awslabs/aws-lambda-powertools-python/blob/6fcedeca10e54312ddc9e00eb1ca8fe27064f7dc/aws_lambda_powertools/event_handler/api_gateway.py#L517) with all potential returns, even with `Literal[ProxyEventType.APIGatewayProxyEvent]` didn't seem to yield any result, as MyPy and Intelisense continues to think it's a `BaseProxyEvent`.

Need a deeper understanding of MyPy and type invariants to potentially solve this.

```python

app = ApiGatewayResolver()


@app.get("/test")
def test_it():
	# this should autocomplete when not using ALB as the ProxyType argument in ApiGatewayResolver
	app.current_event.request_context.identity.account_id 
    ...


def handler(event, context):
    app.resolve(event, context)


app({}, None)
```

michaelbrewer:
**Issue #, if available:**

closes #761

## Description of changes:

Add new specialized resolvers `APIGatewayRestResolver`, `APIGatewayHttpResolver` and `ALBResolver`

Example of an Application Load Balancer handler (ALBResolver)  where `current_event` can only be an `ALBEvent`

```python3

app = ALBResolver()

@app.get("/lambda")
def foo():
    # Now you can access ALBEvent specific attributes
    assert app.current_event.request_context.elb_target_group_arn is not None
    return Response(200, content_types.TEXT_HTML, "foo")
```

Example of an REST API handler (APIGatewayRestResolver) where `current_event` can only be an `APIGatewayProxyEvent`

```python3

@app.get("/my/path")
def get_lambda() -> Response:
    # Now you can access APIGatewayProxyEvent specific attributes
    assert app.current_event.request_context.domain_name == "id.execute-api.us-east-1.amazonaws.com"
    return Response(200, content_types.APPLICATION_JSON, json.dumps({"foo": "value"}))
```

Example of an Http API handler (APIGatewayHttpResolver) where `current_event` can only be an `APIGatewayProxyEventV2`

```python3

@app.post("/my/path")
def my_path() -> Response:
    # Now you can access APIGatewayProxyEventV2 specific attributes
    assert app.current_event.cookies[0] == "cookie1"
    return Response(200, content_types.TEXT_PLAIN, "Foo")
```

Works with VSCode and PyCharm:

Screen Shot 2022-01-25 at 11 28 36 PM

**Checklist**

 

* [X] Update tests
* [x] Update docs


aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -731,3 +731,84 @@ def register_route(func: Callable):
...
+            Optionally configure and enabled CORS. Not each route will need to have to cors=True
+        debug: Optional[bool]
+            Enables debug mode, by default False. Can be also be enabled by "POWERTOOLS_EVENT_HANDLER_DEBUG"
+            environment variable
+        serializer : Callable, optional
+            function to serialize `obj` to a JSON formatted `str`, by default json.dumps
+        strip_prefixes: List[str], optional
+            list of prefixes to be removed from the request path before doing the routing. This is often used
+            with api gateways with multiple custom mappings.
+        """
```
heitorlessa, line 759:
```suggestion
```

These will be propagated from the superclass (granted, proxy_type as a helper will too but constructor will tell it's wrong) -- this helps us prevent multiple docstrings being accidentally outdated.

Tested on both VSCode and PyCharm (PyCharm is more accurate as it shows the constructor signature)

image

image

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -731,3 +731,84 @@ def register_route(func: Callable):
...
```
heitorlessa, line 786:
```suggestion
```

Same as the other one

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -731,3 +731,84 @@ def register_route(func: Callable):
...
```
heitorlessa, line 813:
```suggestion
```

Same as above

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -731,3 +731,84 @@ def register_route(func: Callable):
...
+        super().__init__(ProxyEventType.APIGatewayProxyEventV2, cors, debug, serializer, strip_prefixes)
+
+
+class ALBEventResolver(ApiGatewayResolver):
```
heitorlessa, line 790:
```suggestion
class ALBResolver(ApiGatewayResolver):
```

Shorter and more explicit name (discussion on Slack too).

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -731,3 +731,84 @@ def register_route(func: Callable):
...
+        serializer : Callable, optional
+            function to serialize `obj` to a JSON formatted `str`, by default json.dumps
+        strip_prefixes: List[str], optional
+            list of prefixes to be removed from the request path before doing the routing. This is often used
+            with api gateways with multiple custom mappings.
+        """
+        super().__init__(ProxyEventType.APIGatewayProxyEvent, cors, debug, serializer, strip_prefixes)
+
+
+class APIGatewayProxyEventV2Resolver(ApiGatewayResolver):
```
heitorlessa, line 763:
```suggestion
class APIGatewayHttpResolver(ApiGatewayResolver):
```

Shorter and more explicit name (discussion on Slack too). This opens the door to have a WebSocket afterwards.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -731,3 +731,84 @@ def register_route(func: Callable):
             self._routes[(rule, methods, cors, compress, cache_control)] = func
 
         return register_route
+
+
+class APIGatewayProxyEventResolver(ApiGatewayResolver):
```
heitorlessa, line 736:
```suggestion
class APIGatewayRestResolver(ApiGatewayResolver):
```

Shorter and more explicit name (discussion on Slack too).

heitorlessa (changes requested):
Suggestion to make names shorter, explicit, and remove the `Event` redundant part. Also to remove duplicate docstrings as it's inherited from super class (despite the proxy_type showing in VSCode) to prevent what happened to Parameters utilities docstrings being accidentally left behind too

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: In hindsight, this one is special and deserves a short docstring to say this is for **HTTP API payload v2**.

heitorlessa (approved):

PR #1007, opened 2022-02-06
feat(data_classes): introduce event size in EventBridgeEvent

stephenbawks:
**Issue #, if available:**

## Description of changes:

Adding a new property to the EventBridgeEvent class.  Wanted to see if there was any major concerns with instituting this way and if I should make changes.


aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
```diff
@@ -67,3 +68,22 @@ def detail(self) -> Dict[str, Any]:
...
+
+    @property
+    def event_size(self) -> int:
+        """Returns the EventBridge Event Size.
+
+        URLs:
+            https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-putevent-size.html
+        """
+
+        size = 0
```
heitorlessa, line 80: Consider making use of caching.

Due to string conversion being a relatively expensive operation, check if event size was previously calculated before running through.

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
```diff
@@ -67,3 +68,22 @@ def detail(self) -> Dict[str, Any]:
...
+        if self["time"] is not None:
```
heitorlessa, line 81: Can't recall if DictWrapper will give you None or raise KeyError if not present -- an unit test here can verify

heitorlessa (changes requested):
Hey @stephenbawks thanks a lot for this contribution - quick summary of the asks before we merge

1. Link the roadmap issue you created in the PR description
2. Unit tests (missing time, missing resources)
3. Use caching (check out API Gateway Event or Kinesis data properties as examples)
4. Python3 uses utf-8 by default ;-)

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
michaelbrewer: @heitorlessa  - it would produce a key error. However `.get("time")` also works.

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
stephenbawks: I guess that being said, does it make more sense to move it to its own class?

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
stephenbawks: Funny enough get("time") was how I originally had it.

```
if self.get("time") is not None:
```

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
heitorlessa: Could you share how do you plan on using it? I suspect this would be best utilized as a function that receives an EventBridge event, because this dataclass is "immutable" (no setters). 

The usefulness of checking the event size is to prevent sending an event that will be rejected by EventBridge. This data class would highly likely contain an event that is already under 256K, unless you're using it differently hence the question ;)

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
heitorlessa: Thanks @michaelbrewer ! `.get` is the way to go then

michaelbrewer:
@stephenbawks if you have limited time, i don't might helping to complete this and @heitorlessa  and yourself can review :)

stephenbawks:
> @stephenbawks if you have limited time, i don't might helping to complete this and @heitorlessa  and yourself can review :) 

Yah git distracted and forgot to get back to this.

I would appreciate any help.

If not, I can get back to this this week.

aws_lambda_powertools/utilities/data_classes/event_bridge_event.py
```diff
@@ -67,3 +68,22 @@ def detail(self) -> Dict[str, Any]:
...
+            size += 14
+        size += len(self["source"].encode("utf-8"))
+        size += len(self["detail-type"].encode("utf-8"))
+        size += len(json.dumps(self["detail"]).encode("utf-8"))
```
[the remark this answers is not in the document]
stephenbawks: detail is required however it can be "empty"
detail – A JSON object that contains information about the event. It can be "{}".

I will update.

stephenbawks:
I am starting to think/agree that this would probably be better served as a callable function outside the `event_bridge_event.py` file. It was mentioned before but it would be doing this calculation on every event.  There is obviously a time overhead to that.  Thoughts? 

If it gets moved, any suggestions on where it should be placed? 

The original intent for this was to calculate the event size and if needed trim the event and drop that as an s3 object instead.  Trim off the actual message, append a s3 metadata attribute that points to the s3 URL, and add that back into the message.

stephenbawks:
> I am starting to think/agree that this would probably be better served as a callable function outside the `event_bridge_event.py` file. It was mentioned before but it would be doing this calculation on every event. There is obviously a time overhead to that. Thoughts?
> 
> If it gets moved, any suggestions on where it should be placed?
> 
> The original intent for this was to calculate the event size and if needed trim the event and drop that as an s3 object instead. Trim off the actual message, append a s3 metadata attribute that points to the s3 URL, and add that back into the message.

Actually being that this is a property decorator/function in the class it should not be doing the calculation on the event size unless it's being called?

leandrodamascena:
Hi @stephenbawks! I was looking at this PR and there are some unanswered comments. Also, I think we need to understand a little bit more about the use case here and how we solve the problem.

I think we can start a Discord thread to talk about it and maybe have this PR merged soon. Please let me know about this idea.

Thanks

leandrodamascena:
Closing this PR as issue #1168 has been closed.

Issue #1006, opened 2022-02-05
Support multiple Allowed Origins in CORS


straygar:
**Is your feature request related to a problem? Please describe.**
In some cases, we would like to support multiple origins in CORS when defining an API Gateway/ALB event handler (e.g. support calling an API from "localhost" in dev, or webapps living in different origins calling the same API).

**Describe the solution you'd like**
Allow providing a list of origins `CORSConfig(allow_origins=[])` ([source](https://awslabs.github.io/aws-lambda-powertools-python/latest/core/event_handler/api_gateway/#cors)). If the client's `Origin` is in that list, that value is returned in the response header. Otherwise - an arbitrary (e.g. 1st) value from the list can be returned.

michaelbrewer:
This will rely on using the `origin` to determine which one of the possible `allow_origins` to use. So a little thought will be needed in how to implement this, this article covers this fairly well: https://crashtest-security.com/multiple-values-access-control-allow-origin/

FYI @heitorlessa

heitorlessa:
Hey @straygar thanks a lot for raising this idea with us!

I'd be careful on relying on `Origin` to make that decision for two reasons: **1/** It's an attack vector as this can be easily spoofed, and **2/** API Gateway doesn't support multiple origins by default, which means you need to create a `OPTIONS` method for **each path of your API definition** pointing to your Lambda function handling it.

As 2 can become tedious and error-prone, it opens the door for these sort of attacks that are quite common: https://we45.com/blog/3-ways-to-exploit-cors-misconfiguration

I'd like to wait for more customers demand before we look into it, so we can think of additional factors besides `Origin` and make a good documentation to reduce chances of misconfiguration.

For now, you can manually do it by **a)** configuring API Gateway to invoke your Lambda function for pre-flight requests (`OPTIONS`), **b)** handling pre-flight using `@app.route(rule="*", method="OPTIONS")`, and **c)** dynamically updating `app.cors` with a `CorsConfig` object prior to return (it'll be easier once [middleware feature is finished](#954)

michaelbrewer:
Optimal CORS configuration is always tricky. Classical solutions are a plenty via nginx configurations at allow for a set of regex origins. 

Not sure why there is a gap here for API GW. But it sounds like spinning up multiple endpoints is what is recommended. 

Better to be safe.

michaelbrewer:
@heitorlessa this is what the linked article does. I am just saying we could also do it in behave of the client.

[...]

PR #1010, opened 2022-02-07
feat(event_handler): support multiple origins for cors

michaelbrewer:
**Issue #, if available:**
- #1006

## Description of changes:

> WARNING: FOR RFC PURPOSES ONLY

**Checklist**

 

* [ ] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:


walmsles:
The PR allows for multiple statically defined Origins.  Some use cases may have more dynamic requirements, e.g. An API allowing only subscribed websites (origins) to consume the API from their front-end.  This may be a more obscure use case and not the norm though - so need to apply the tenets to this comment as appropriate.

michaelbrewer:
> have more dynamic requirements, e.g. An API allowing only subscribed websites (origins) to consume the API from

This is where people might shoot themselves in the foot ? :)

walmsles:
Agree.  Need to temper feature offered to close the loop on people doing silly things.  Because CORS is generally well misunderstood 🙂

michaelbrewer:
> Agree. Need to temper feature offered to close the loop on people doing silly things. Because CORS is generally well misunderstood 🙂

And universally hated

michaelbrewer:
(closing for now and will reopening when capacity is available for this)

major-mayer:
I would relly love to see this getting merged soon.
We are using AWS Lambda Powertools with the Application Load Balancer (because of timeout limitation of the API Gateway), so there is no CORS configuration and we have to handle this ourselves.

The Lambda function is called from multiple origins, but it's a fixed list, so no dynamic matching needed.
This PR would be a very nice improvement, so that i don't have to manually check if the origin in the header is in my list of allowed origins.

michaelbrewer:
@major-mayer - possibly August this might get some eyes.

Issue #1009, opened 2022-02-07
Papercuts priorization before we consider v2


heitorlessa:
> **UPDATE Apr '22**: We are not looking for external contributions to address these.

Original: #857. Locked to ensure we don't lose focus a second time.

List of papercuts we're currently prioritizing based on their impact. Once these are completed, we will start drafting what v2 looks like as we aim towards modularization.

> NOTE: Any item missing link means we must create an issue yet. This is in anticipation of moving the central roadmap back to each respective repository - just haven't had the bandwidth to do any of that.

## High impact

- [ ] [**Code quality**: Integ test with external dependencies](https://github.com/awslabs/aws-lambda-powertools-python/issues/1251) #1226
- [ ] [**Docs**: Move code snippets into dedicated python files so they can be linted/formatted then embedded in the docs. It makes highlighting slightly harder but we can achieve consistency and prevent mistakes we keep on making](https://github.com/awslabs/aws-lambda-powertools-python/issues/1364). See #1004, #1061 -- @heitorlessa 
- [ ] **Idempotency**: [Remove data record lock in case of timeout](https://github.com/awslabs/aws-lambda-powertools-python/issues/1038) -- @rubenfonseca 

## Low impact

- [ ] [**Logger**: Provide autocomplete custom date formats (ISO8601) for `datefmt` field](https://github.com/awslabs/aws-lambda-powertools-python/issues/946)
- [ ] [**Tracer**: Typing SubSegment and Context Manager](https://github.com/awslabs/aws-lambda-powertools-python/issues/1375)
- [ ] [**Batch**: Improve maintainability of batch processing functional tests](https://github.com/awslabs/aws-lambda-powertools-python/issues/1363)
- [ ] [**Idempotency**: Consider refactoring test setup, potentially DynamoDB Local for functional](https://github.com/awslabs/aws-lambda-powertools-python/issues/1360) (not integration)
- [ ] [**Contributions**: Pre-configure GitPod and update Contributing guidelines](https://github.com/awslabs/aws-lambda-powertools-python/issues/1361) (Poetry didn't work previously; it works now)

## Maintenance automation

- [ ] [**Actions**: React to labels `size/XL`, `size/XXL` by adding `triage` label for maintainers attention, and suggest breaking the PR to reduce the chances of bugs being introduced due to cognitive load](https://github.com/awslabs/aws-lambda-powertools-python/issues/1331)
[...]

PR #1017, opened 2022-02-09
docs(parameters): add testing your code section

cakepietoast:
**Issue #, if available:** #1009

## Description of changes:

Add "Testing your code" section to the documentation for parameters utility.

**Checklist**

 

* [ ] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

-----
[View rendered docs/utilities/parameters.md](https://github.com/awslabs/aws-lambda-powertools-python/blob/docs/testing_parameters/docs/utilities/parameters.md)

### DIFF
--- a/docs/utilities/parameters.md
+++ b/docs/utilities/parameters.md
@@ -515,3 +515,76 @@ The **`config`** and **`boto3_session`** parameters enable you to pass in a cust
 		value = ssm_provider.get("/my/parameter")
 		...
 	```
+
+## Testing your code
+
+For unit testing your applications, you can mock the calls to the parameters utility to avoid calling AWS APIs. This
+can be achieved in a number of ways - in this example, we use the [pytest monkeypatch fixture](https://docs.pytest.org/en/latest/how-to/monkeypatch.html)
+to patch the `parameters.get_parameter` method:
+
+=== "tests.py"
+	```python
+	from src.index import handler
+
+	def test_handler(monkeypatch):
+
+		def mockreturn(name):
+			return "mock_value"
+
+		monkeypatch.setattr(parameters, "get_parameter", mockreturn)
+		return_val = handler({}, {})
+		assert return_val.get('message') == 'mock_value'
+
+	```
+
+=== "src/index.py"
+	```python
+	from aws_lambda_powertools.utilities import parameters
+
+	def handler(event, context):
+		# Retrieve a single parameter
+		value = parameters.get_parameter("my-parameter-name")
+		return {"message": value}
+	```
+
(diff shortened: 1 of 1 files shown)


docs/utilities/parameters.md
```diff
@@ -515,3 +515,76 @@ The **`config`** and **`boto3_session`** parameters enable you to pass in a cust
...
+	import pytest
+
+	from src.index import handler
+
+	@pytest.fixture
+	def mock_parameter_response(monkeypatch):
+		def mockreturn(name):
+			return "mock_value"
+
+    	monkeypatch.setattr(parameters, "get_parameter", mockreturn)
```
heitorlessa, line 563: **Q**: Shall we return a factory so they can control the return value at each test?

e.g. `mock_parameter_response(data_i_want_to_return)`

heitorlessa (changes requested):
Looks great! One potential typo on mocking the incorrect namespace, and one question on whether a function factory would be more powerful as an example

docs/utilities/parameters.md
cakepietoast: It can be done but its not pretty! This is slightly different to the usual fixture, since we don't actually care about the fixture's value, we just need it to run so it monkeypatches the `get_parameter` method. We _could_ do something like the below example, but IMHO it is a lot more readable to just have multiple fixtures where needed:

```python
@pytest.fixture
def mock_parameter_response(monkeypatch):
    def factory_func(parameter_return_value):
        def mockreturn(name):
            return parameter_return_value
        monkeypatch.setattr(parameters, "get_parameter", mockreturn)

    return factory_func


def test_handler(mock_parameter_response):
    mock_parameter_response("aaaaaaaaargh!")
    return_val = handler({}, {})
    assert return_val.get('message') == "aaaaaaaaargh!"
```

heitorlessa: ugh yep indeed. Let's keep as-is.

docs/utilities/parameters.md
```diff
@@ -515,3 +515,76 @@ The **`config`** and **`boto3_session`** parameters enable you to pass in a cust
...
+=== "tests.py"
+	```python
+	from src.index import handler
+
+	def test_handler(monkeypatch):
+
+		def mockreturn(name):
+			return "mock_value"
+
+		monkeypatch.setattr(parameters, "get_parameter", mockreturn)
```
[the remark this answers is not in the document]
cakepietoast: Fixed

heitorlessa (approved):

Issue #1025, opened 2022-02-13
feat(data-classes): fix SES `date` field typing and add missing fields


michaelbrewer:
**What were you trying to accomplish?**

Use SESEvent with S3 

- Related docs: https://docs.aws.amazon.com/ses/latest/dg/receiving-email-notifications-examples.html#receiving-email-notifications-examples-alert
- Example event: https://github.com/aws/aws-lambda-go/blob/main/events/testdata/ses-lambda-event.json
- types/aws-lambda - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/50031
- https://docs.aws.amazon.com/ses/latest/APIReference/API_LambdaAction.html - includes references to topicArn
- https://github.com/aws/aws-lambda-go/pull/211 - go pull request for additional fields.
- https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.SimpleEmailEvents/Actions/S3ReceiptAction.cs - DotNet lambda events also includes these fields.

## Expected Behavior

- `date` field is set to a `List[str]` and not a `str`
- All fields should be included in the data classes

## Current Behavior

Missing fields needs to be accessed via keys directly

## Possible Solution

Add missing fields to SESEvent data class

## Steps to Reproduce (for bugs)

Parsing regular SES message:

```json
{
  "Records": [
    {
      "eventVersion": "1.0",
      "ses": {
        "mail": {
          "commonHeaders": {
            "from": [
              "Amazon Web Services <aws@amazon.com>"
            ],
            "to": [
              "lambda@amazon.com"
            ],
            "returnPath": "aws@amazon.com",
            "messageId": "<CAEddw6POFV_On91m+ZoL_SN8B_M2goDe_Ni355owhc7QSjPQSQ@amazon.com>",
            "date": "Mon, 5 Dec 2016 18:40:08 -0800",
            "subject": "Test Subject"
          },
          "source": "aws@amazon.com",
          "timestamp": "1970-01-01T00:00:00.123Z",
          "destination": [
            "lambda@amazon.com"
          ],
          "headers": [
            {
              "name": "Return-Path",
              "value": "<aws@amazon.com>"
            },
            {
              "name": "Received",
[...]

PR #1026, opened 2022-02-13
fix(data_classes): Add missing fields for SESEvent

michaelbrewer:
**Issue #, if available:**

-  #1025
- Also implemented in the lambda go events https://github.com/aws/aws-lambda-go/pull/211

## Description of changes:

Change:
- Fix date type to be a str and not `List[str]`
- Add missing fields from `SESMailCommonHeaders` (`bcc`, `cc`, `sender` and `replyTo`)
- Add missing `dkimVerdict` field
- Add missing `topicArn` field
- Add missing `dmarcPolicy` field
- Add missing docs for `SESReceiptStatus` status field
- Add response builder for disposition (`disposition_response`)
- Add missing fields for `Bounce`, `S3` and `WorkMail` actions

**Checklist**

 

* [X] Update tests
* [X] Update docs

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/ses_event.py
+++ b/aws_lambda_powertools/utilities/data_classes/ses_event.py
@@ -1,4 +1,5 @@
-from typing import Iterator, List, Optional
+from enum import Enum
+from typing import Dict, Iterator, List, Optional
 
 from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
 
@@ -316,3 +317,17 @@ class SESEvent(DictWrapper):
     @property
     def receipt(self) -> SESReceipt:
         return self.record.ses.receipt
+
+
+class Disposition(Enum):
+    """No further actions in the current receipt rule will be processed, but further receipt rules can be processed."""
+
+    STOP_RULE = "STOP_RULE"
+    """No further actions or receipt rules will be processed."""
+    STOP_RULE_SET = "STOP_RULE_SET"
+    """This means that further actions and receipt rules can be processed."""
+    CONTINUE = "CONTINUE"
+
+
+def disposition_response(disposition: Disposition) -> Dict[str, str]:
+    return {"disposition": disposition.value}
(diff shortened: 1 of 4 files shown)


aws_lambda_powertools/utilities/data_classes/ses_event.py
```diff
@@ -107,18 +131,78 @@ def get_type(self) -> str:
...
+
+        Present only for the Lambda action type."""
         return self["invocationType"]
 
+    @property
+    def bucket_name(self) -> str:
+        """String that contains the name of the Amazon S3 bucket to which the message was published.
+
+        Present only for the S3 action type."""
+        return str(self["bucketName"])
```
jplock, line 160: Should we use self.get() on field that may not be present so it returns None instead of a KeyError?

michaelbrewer: In general it is optional, yes. 

But if the action type is "S3", it would be required. Same for all of the other fields for Lambda, Bounce and S3 action types.

michaelbrewer:
@heitorlessa - i just want to get verification on the action types supported (aka `Stop`, `Bounce`, `S3` and `WorkMail` actions)

jogold:
@michaelbrewer you contacted me on Twitter. You mean this https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/test/ses-tests.ts?

michaelbrewer:
> @michaelbrewer you contacted me on Twitter. You mean this [https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/test/ses-tests.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/test/ses-tests.ts?rgh-link-date=2022-02-16T09%3A40%3A27Z)?

Yes @jogold both Lex V2 and SES event documentation does not include real examples with these missing fields.

michaelbrewer:
@heitorlessa - should we remove this data class ? Or at least fix the following 

- Fix date type to be a str and not List[str]
- Add missing fields from SESMailCommonHeaders (bcc, cc, sender and replyTo)
- Add missing docs for SESReceiptStatus status field
Add missing topicArn from all actions

michaelbrewer:
Superseded by #1045

PR #1046, opened 2022-02-27
fix(metrics): flush upon a single metric 100th data point

knightjoel:
## Description of changes:

Per the [EMF spec](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html#CloudWatch_Embedded_Metric_Format_Specification_structure_target), metric value arrays must not have more than 100 elements. This PR ensures that metrics are serialized when a single metric collects 100 values in addition to the existing behavior of serializing when 100 metrics are collected.

Note: as far as I can see, this same change needs to be brought to the other Powertools languages as well. I'm not equipped nor have the skill to submit a PR for any language other than Python.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/metrics/base.py
+++ b/aws_lambda_powertools/metrics/base.py
@@ -127,7 +127,7 @@ class MetricManager:
         logger.debug(f"Adding metric: {name} with {metric}")
         self.metric_set[name] = metric
 
-        if len(self.metric_set) == MAX_METRICS:
+        if len(self.metric_set) == MAX_METRICS or len(metric["Value"]) == MAX_METRICS:
             logger.debug(f"Exceeded maximum of {MAX_METRICS} metrics - Publishing existing metric set")
             metrics = self.serialize_metric_set()
             print(json.dumps(metrics))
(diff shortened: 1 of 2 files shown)


heitorlessa:
cc @pankajagrawal16 @saragerion @flochaz - this is the bug we discussed on metric **data points** having a maximum of 100.

heitorlessa (approved):

heitorlessa:
@jaredcnance thanks again for confirming this in the spec. Next release will have this fix on Lambda Powertools for Python

pankajagrawal16:
> cc @pankajagrawal16 @saragerion @flochaz - this is the bug we discussed on metric **data points** having a maximum of 100.

@heitorlessa Since Powertools for Java internally uses https://github.com/awslabs/aws-embedded-metrics-java, this is fixed already in the lib as part of this [issue](https://github.com/awslabs/aws-embedded-metrics-java/issues/72) via this [PR](https://github.com/awslabs/aws-embedded-metrics-java/pull/74). Let me know if I am missing something still that was fixed as part of this change. As far as I can see, Powertools for Java is covered.

PR #1051, opened 2022-03-01
fix(lambda-authorizer): Propertly parse resource path

michaelbrewer:
**Issue #, if available:**

- 1047

## Description of changes:

Correctly parse method arn to include the full resource path ie: `arn:aws:execute-api:us-west-2:123456789012:ymy8tbxw7b/*/GET/foo/bar`

**Checklist**

 

* [X] Update tests
* [X] Update docs


heitorlessa (commented):
adding a comment to easily reason about this change later. Then merging.

Thanks a lot for the fix @michaelbrewer. It does make me think whether we need a hard check, but am struggling with bandwidth to make a more informed decision.

Issue #1060, opened 2022-03-07
Missing return type in lambda_handler_decorator leads to 'Untyped decorator makes function "..." untyped' errors in mypy strict mode


huonw:
**What were you trying to accomplish?**

I'm trying to use `aws_lambda_powertools.utilities.data_classes.event_source` with `mypy --strict`, and receive errors about `Untyped decorator makes function "handler" untyped`. This is because `lambda_handler_decorator` is missing a return value, so mypy doesn't know anything about any function decorated with it (including `event_source`):

https://github.com/awslabs/aws-lambda-powertools-python/blob/22f8b9a7f47a2a4682c6f3f03536ac848ae7dc0d/aws_lambda_powertools/middleware_factory/factory.py#L15

## Expected Behavior
`@event_source` and other decorators using `@lambda_handler_decorator` should work in mypy strict mode.

## Current Behavior

Mypy emits errors.

## Possible Solution

1. Add an appropriate `Callable[something, here]` return value to `lambda_handler_decorator`
2. Run `mypy --strict` in this package's CI

## Steps to Reproduce (for bugs)

requirements:

```
aws-lambda-powertools==1.25.1
aws-xray-sdk==2.9.0
boto3==1.21.13
botocore==1.24.13
fastjsonschema==2.15.3
future==0.18.2
jmespath==0.10.0
mypy-extensions==0.4.3
mypy==0.931
python-dateutil==2.8.2
s3transfer==0.5.2
six==1.16.0
tomli==2.0.1
typing-extensions==4.1.1
urllib3==1.26.8
wrapt==1.13.3
```

`input.py`:

```python
from typing import Any

from aws_lambda_powertools.middleware_factory import lambda_handler_decorator

from aws_lambda_powertools.utilities.data_classes import (
    event_source,
    KinesisStreamEvent,
)


@lambda_handler_decorator
def custom_decorator() -> None:
    return


@event_source(data_class=KinesisStreamEvent)
@custom_decorator
def handler(event: KinesisStreamEvent, context: Any) -> None:
    pass
```

invocation:

```
$ mypy --strict input.py 
input.py:16: error: Untyped decorator makes function "handler" untyped
input.py:17: error: Untyped decorator makes function "handler" untyped
```

## Environment

* **Powertools version used**: 1.25.1
* **Packaging format (Layers, PyPi)**: N/A
* **AWS Lambda function runtime:** N/A
* **Debugging logs** N/A

[...]

PR #1066, opened 2022-03-07
fix(middleware_factory): ret type annotation for handler dec

huonw:
**Issue #, if available:** #1060 

## Description of changes:

This is two versions of typing `lambda_handler_decorator` more accurately for #1060, to ensure that middleware using it have a type. They end up with type `Callable[..., Any]` but that's better than nothing.

The first commit 937e6fe tries to be more accurate via overloads and protocols, but it doesn't work as written, and I cannot quite work out why. The errors below demonstrate the issues. It's potentially resolvable with 3.10's `ParamSpec` (I'm not sure `typing_extensions` depended upon for 3.8 and 3.8?):

<details><summary>Errors</summary>

```
aws_lambda_powertools/utilities/validation/validator.py:12:2: error: No overload variant of "lambda_handler_decorator" matches argument type "Callable[[Callable[..., Any], Union[Dict[Any, Any], str], Any, Optional[Dict[Any, Any]], Optional[Dict[Any, Any]], Optional[Dict[Any, Any]], Optional[Dict[Any, Any]], str, Optional[Dict[Any, Any]]], Any]"  [call-overload]
aws_lambda_powertools/utilities/validation/validator.py:12:2: note: Possible overload variants:
aws_lambda_powertools/utilities/validation/validator.py:12:2: note:     def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator
aws_lambda_powertools/utilities/validation/validator.py:12:2: note:     def lambda_handler_decorator(decorator: None = ..., trace_execution: Optional[bool] = ...) -> Callable[[_FactoryDecorator], _HandlerDecorator]
aws_lambda_powertools/utilities/parser/parser.py:14:2: error: No overload variant of "lambda_handler_decorator" matches argument type "Callable[[Callable[[Any, LambdaContext], EventParserReturnType], Dict[str, Any], LambdaContext, Type[Model], Optional[Type[Envelope]]], EventParserReturnType]"  [call-overload]
aws_lambda_powertools/utilities/parser/parser.py:14:2: note: Possible overload variants:
aws_lambda_powertools/utilities/parser/parser.py:14:2: note:     def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator
aws_lambda_powertools/utilities/parser/parser.py:14:2: note:     def lambda_handler_decorator(decorator: None = ..., trace_execution: Optional[bool] = ...) -> Callable[[_FactoryDecorator], _HandlerDecorator]
aws_lambda_powertools/utilities/idempotency/idempotency.py:20:2: error: No overload variant of "lambda_handler_decorator" matches argument type "Callable[[Callable[[Any, LambdaContext], Any], Dict[str, Any], LambdaContext, BasePersistenceLayer, Optional[IdempotencyConfig], KwArg(Any)], Any]"  [call-overload]
aws_lambda_powertools/utilities/idempotency/idempotency.py:20:2: note: Possible overload variants:
aws_lambda_powertools/utilities/idempotency/idempotency.py:20:2: note:     def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator
aws_lambda_powertools/utilities/idempotency/idempotency.py:20:2: note:     def lambda_handler_decorator(decorator: None = ..., trace_execution: Optional[bool] = ...) -> Callable[[_FactoryDecorator], _HandlerDecorator]
aws_lambda_powertools/utilities/data_classes/event_source.py:8:2: error: No overload variant of "lambda_handler_decorator" matches argument type "Callable[[Callable[[Any, LambdaContext], Any], Dict[str, Any], LambdaContext, Type[DictWrapper]], Any]"  [call-overload]
aws_lambda_powertools/utilities/data_classes/event_source.py:8:2: note: Possible overload variants:
aws_lambda_powertools/utilities/data_classes/event_source.py:8:2: note:     def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator
aws_lambda_powertools/utilities/data_classes/event_source.py:8:2: note:     def lambda_handler_decorator(decorator: None = ..., trace_execution: Optional[bool] = ...) -> Callable[[_FactoryDecorator], _HandlerDecorator]
aws_lambda_powertools/utilities/batch/base.py:156:2: error: No overload variant of "lambda_handler_decorator" matches argument type "Callable[[Callable[..., Any], Dict[Any, Any], Dict[Any, Any], Callable[..., Any], BasePartialProcessor], Any]"  [call-overload]
aws_lambda_powertools/utilities/batch/base.py:156:2: note: Possible overload variants:
aws_lambda_powertools/utilities/batch/base.py:156:2: note:     def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator
aws_lambda_powertools/utilities/batch/base.py:156:2: note:     def lambda_handler_decorator(decorator: None = ..., trace_execution: Optional[bool] = ...) -> Callable[[_FactoryDecorator], _HandlerDecorator]
aws_lambda_powertools/utilities/batch/sqs.py:179:2: error: No overload variant of "lambda_handler_decorator" matches argument type "Callable[[Callable[..., Any], Dict[Any, Any], Dict[Any, Any], Callable[..., Any], Optional[Any], bool, Optional[Any]], Any]"  [call-overload]
aws_lambda_powertools/utilities/batch/sqs.py:179:2: note: Possible overload variants:
aws_lambda_powertools/utilities/batch/sqs.py:179:2: note:     def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator
aws_lambda_powertools/utilities/batch/sqs.py:179:2: note:     def lambda_handler_decorator(decorator: None = ..., trace_execution: Optional[bool] = ...) -> Callable[[_FactoryDecorator], _HandlerDecorator]
```

</details>

**Checklist**

 

* [x] Update tests
* [x] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/aws_lambda_powertools/middleware_factory/factory.py
+++ b/aws_lambda_powertools/middleware_factory/factory.py
@@ -1,18 +1,64 @@
 import functools
 import inspect
 import logging
+import sys
 import os
-from typing import Callable, Optional
+from typing import Any, Callable, Dict, Optional, Union, cast, overload
+
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
 
 from ..shared import constants
 from ..shared.functions import resolve_truthy_env_var_choice
 from ..tracing import Tracer
+from ..utilities.typing import LambdaContext
 from .exceptions import MiddlewareInvalidArgumentError
 
 logger = logging.getLogger(__name__)
 
+# context: Any to avoid forcing users to type it as context: LambdaContext
+_Handler = Callable[[Any, LambdaContext], Any]
+_RawHandlerDecorator = Callable[[_Handler], _Handler]
 
-def lambda_handler_decorator(decorator: Optional[Callable] = None, trace_execution: Optional[bool] = None):
+
+class _FactoryDecorator(Protocol):
+    # it'd be better for this to be using ParamSpec (available from 3.10)
+    def __call__(
+        self, handler: _Handler, event: Dict[str, Any], context: LambdaContext, **kwargs: Any
+    ) -> _RawHandlerDecorator:
+        ...
+
+
+class _HandlerDecorator(Protocol):
+    @overload
+    def __call__(self, decorator: _Handler) -> _Handler:
+        ...
+
+    @overload
+    def __call__(self, decorator: None = None, **kwargs: Any) -> _RawHandlerDecorator:
+        ...
+
+    def __call__(self, decorator: Optional[_Handler] = None, **kwargs: Any) -> Union[_Handler, _RawHandlerDecorator]:
+        ...
+
+
+@overload
+def lambda_handler_decorator(decorator: _FactoryDecorator) -> _HandlerDecorator:
+    ...
+
+
+@overload
+def lambda_handler_decorator(
+    decorator: None = None, trace_execution: Optional[bool] = None
+) -> Callable[[_FactoryDecorator], _HandlerDecorator]:
+    ...
+
+
+def lambda_handler_decorator(
+    decorator: Optional[_FactoryDecorator] = None, trace_execution: Optional[bool] = None
+) -> Union[_HandlerDecorator, Callable[[_FactoryDecorator], _HandlerDecorator]]:
     """Decorator factory for decorating Lambda handlers.
 
     You can use lambda_handler_decorator to create your own middlewares,
@@ -103,19 +149,25 @@ def lambda_handler_decorator(decorator: Optional[Callable] = None, trace_executi
     """
 
     if decorator is None:
-        return functools.partial(lambda_handler_decorator, trace_execution=trace_execution)
+        return cast(
+            Callable[[_FactoryDecorator], _HandlerDecorator],
+            functools.partial(lambda_handler_decorator, trace_execution=trace_execution),
+        )
 
     trace_execution = resolve_truthy_env_var_choice(
         env=os.getenv(constants.MIDDLEWARE_FACTORY_TRACE_ENV, "false"), choice=trace_execution
     )
 
     @functools.wraps(decorator)
-    def final_decorator(func: Optional[Callable] = None, **kwargs):
+    def final_decorator(
+        func: Optional[_RawHandlerDecorator] = None, **kwargs: Any
+    ) -> Union[_Handler, _RawHandlerDecorator]:
         # If called with kwargs return new func with kwargs
         if func is None:
             return functools.partial(final_decorator, **kwargs)
 
         if not inspect.isfunction(func):
+            assert decorator is not None
             # @custom_middleware(True) vs @custom_middleware(log_event=True)
             raise MiddlewareInvalidArgumentError(
                 f"Only keyword arguments is supported for middlewares: {decorator.__qualname__} received {func}"  # type: ignore # noqa: E501
@@ -138,4 +190,4 @@ def lambda_handler_decorator(decorator: Optional[Callable] = None, trace_executi
 
         return wrapper
 
-    return final_decorator
+    return cast(_HandlerDecorator, final_decorator)
--- a/aws_lambda_powertools/middleware_factory/factory.py
+++ b/aws_lambda_powertools/middleware_factory/factory.py
@@ -1,64 +1,19 @@
 import functools
 import inspect
 import logging
-import sys
 import os
-from typing import Any, Callable, Dict, Optional, Union, cast, overload
-
-if sys.version_info >= (3, 8):
-    from typing import Protocol
-else:
-    from typing_extensions import Protocol
+from typing import Any, Callable, Optional
 
 from ..shared import constants
 from ..shared.functions import resolve_truthy_env_var_choice
 from ..tracing import Tracer
-from ..utilities.typing import LambdaContext
 from .exceptions import MiddlewareInvalidArgumentError
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


heitorlessa:
Thanks a lot @huonw !! Question: Do we lose type signature with the current solution? e.g., create a decorator with a given type annotation on a key-value param, call it using an incorrect type and let mypy detect it.

The way I solved this for Tracer (similar, not the same) on preserving the function signature types is using [AnyCallableT](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/shared/types.py#L3) - This `...` works similar to ParamSpec that's available in future version. 

Instead of using overload, could you try with `AnyCallableT` only? I can have a look at this after we fix some bugs in Logger.

huonw:
I think this is an strict improvement over the current code (i.e. nothing is being lost): the middlewares used to be completely untyped, while now they're at least `Callable[..., Any]`, and thus the handlers they're used on are typed `Any`.

Are you proposing something like `def lambda_handler_decorator(decorator: Optional[AnyCallableT] = None, trace_execution: Optional[bool] = None) -> AnyCallableT`. Unfortunately I don't think that'll work quite right, because the type of `decorator` is different to the return value: `decorator` has arguments `handler`, `event` and `context` (and kwargs), while the value returned by `lambda_handler_decorator` just has arguments `handler` (and kwargs). Let me know if I'm misunderstanding the behaviour here.

heitorlessa:
If your current Callable type is helping preserve the type signature (not being done today), then we should stop here and merge \o/!

The AnyCallableT suggestion was for the first layer (like Tracer, Logger, Metrics decorators), it wouldn't work for the second one for reasons you called out - hence the thought to experiment as that wasn't the final solution.

Let me know the answer to the types as I'm not entirely sure I understood. If types aren't being preserved with the latest commit, we can experiment other ways (typing from my phone again, need to think of other ways still)

huonw:
Ah, I see. I think that may be tricky and/or require manual casts at each use site of the decorator:

```python
from typing import Any, Callable, cast

AnyCallable = Callable[..., Any]
def base_decorator(f: AnyCallable) -> AnyCallable:
    return f

@base_decorator
def my_decorator(g: AnyCallable) -> AnyCallable:
    return g

my_decorator = cast(Callable[[AnyCallable], AnyCallable], my_decorator)
```

There's quite a few downsides of this:

1. requires manually specifying the type at each site
2. using `Callable` directly like this still doesn't get the type correct right (doesn't distinguish between `@decorator` like the above and the partially applied version with kwargs `@decorator(foo=1) def f() ...`: it'd be restricted to only one or the other), although one could probably define an appropriate protocol
3. some tools may complain about having two declarations of `my_decorator` too (one via the decorator, and one via the cast

On balance, and especially because these decorators are designed for handlers (i.e. functions that typically aren't used elsewhere in library code), I feel the minimal `Any`-based approach is the best immediate solution.

heitorlessa:
Just ran `--strict` and: `Found 842 errors in 73 files (checked 125 source files)` - will definitely look into it later. For now, I'll test a custom middleware with an additional kwarg to check whether Mypy can detect it with the new type annotation, then merge ;)

heitorlessa:
Nah, spent the last two hours and while I made headwinds towards identifying kwargs in a non-strict mode, I hit a dead end with `too few arguments` later - Adding the patch as ZIP and the sample code I used, in case you know a solution.

Otherwise, I'll take this opportunity to fix a doc typo as part of this PR and then merge.

---

**[factory.patch.zip](https://github.com/awslabs/aws-lambda-powertools-python/files/8306094/factory.patch.zip)**

**Sample code used**

```python
from typing import Any, Callable, Dict, List, Optional

@lambda_handler_decorator
def obfuscate_sensitive_data(
    handler: Callable[..., Any], event: Dict[str, Any], context: Any, fields: Optional[List[str]] = None
) -> Any:
    # Obfuscate email before calling Lambda handler
    if fields:
        for field in fields:
            if field in event:
                event[field] = "REDACTED"

@obfuscate_sensitive_data(fields=10)
def lambda_handler(event: Dict[str, Any], context: Any) -> Any:
    ...
```

**Mypy strict output against sample code**

```python
blah.py:19:2: error: Unexpected keyword argument "fields" for "obfuscate_sensitive_data"  [call-arg]
blah.py:19:2: error: Too few arguments for "obfuscate_sensitive_data"  [call-arg]
Found 2 errors in 1 file (checked 1 source file)
```

huonw:
I have no further insight on that patch 😄

aws_lambda_powertools/middleware_factory/factory.py
```diff
@@ -12,7 +12,8 @@
 logger = logging.getLogger(__name__)
 
 
-def lambda_handler_decorator(decorator: Optional[Callable] = None, trace_execution: Optional[bool] = None):
+# giving this an accurate return type is hard
```
heitorlessa, line 15:
```suggestion
# Maintenance: we can't yet provide an accurate return type without ParamSpec etc. see #1066
```

heitorlessa (changes requested):
suggesting a reword on return type comment, accepting, and merging.

heitorlessa (approved):

heitorlessa:
Thanks again @huonw for taking the time, truly appreciate that. Hopefully, we will be able to solve this and other similar situations with typing_extensions in the future

huonw:
Woohoo, thanks for the tweak and then merging 👍

PR #1090, opened 2022-04-05
fix(idempotency): pass by value on idem key to guard inadvert mutations

ojongerius:
This protects against functions that inadvertently change the incoming event data

## Description of changes:

Switch to using a deep copy of data. Without this, we've seen two documents written for one function invocation that (inadvertently) mutated the event data:

* An _initial_ document with a hashed idempotency key _based on the original payload_ that will be stuck in INPROGRESS and has no return data
* A _second_ document with a hashed idempotency key _based on the mutated payload_ with status COMPLETED and the correct return value

Happy to add a test if there is an appetite for this change.

**Checklist**

* [x] Update tests
* [ ] Update docs

### DIFF
--- a/tests/functional/idempotency/test_idempotency.py
+++ b/tests/functional/idempotency/test_idempotency.py
@@ -274,6 +274,40 @@ def test_idempotent_lambda_first_execution_cached(
     stubber.assert_no_pending_responses()
     stubber.deactivate()
 
+@pytest.mark.parametrize("idempotency_config", [{"use_local_cache": False}, {"use_local_cache": True}], indirect=True)
+def test_idempotent_lambda_first_execution_event_mutation(
+    idempotency_config: IdempotencyConfig,
+    persistence_store: DynamoDBPersistenceLayer,
+    lambda_apigw_event,
+    expected_params_update_item,
+    expected_params_put_item,
+    lambda_response,
+    serialized_lambda_response,
+    deserialized_lambda_response,
+    hashed_idempotency_key,
+    lambda_context,
+):
+    """
+    Test idempotent decorator where lambda_handler mutates the event
+    """
+
+    stubber = stub.Stubber(persistence_store.table.meta.client)
+    ddb_response = {}
+
+    stubber.add_response("put_item", ddb_response, expected_params_put_item)
+    stubber.add_response("update_item", ddb_response, expected_params_update_item)
+    stubber.activate()
+
+    @idempotent(config=idempotency_config, persistence_store=persistence_store)
+    def lambda_handler(event, context):
+        event.popitem()
+        return lambda_response
+
+    lambda_handler(lambda_apigw_event, lambda_context)
+
+    stubber.assert_no_pending_responses()
+    stubber.deactivate()
+
 
 @pytest.mark.parametrize("idempotency_config", [{"use_local_cache": False}, {"use_local_cache": True}], indirect=True)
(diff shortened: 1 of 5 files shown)


ojongerius:
FWIW some code to reproduce, the assertion that only one document is written fails without my patch.

Results:

```python
'Items': [{'id': 'test-func.handler#8339d7030a6b244b42e6bb381d2c947c',
                'expiration': Decimal('1649130578'), 'status': 'INPROGRESS'},
                {'id': 'test-func.handler#5cd779b2a9dc323203f55b7ef9a6053b',
                 'data': 'null',
                 'expiration': Decimal('1649130578'),
                 'status': 'COMPLETED'}]
```

Test code:

```python
def test_event_change_changes_key(dynamodb, aws_credentials):
    table = dynamodb.create_table(
        TableName='idempotency',
        KeySchema=[{"AttributeName": "id", "KeyType": "HASH"}],
        AttributeDefinitions=[{"AttributeName": "id", "AttributeType": "S"}],
        BillingMode="PAY_PER_REQUEST",
    )
    table.wait_until_exists()
    assert table.scan()['Count'] == 0

    event = {"detail": {"items": ["item_one", "item_two"]}}

    @idempotent(persistence_store=DynamoDBPersistenceLayer(table_name="idempotency"))
    def handler(event, context):
        items = event.get("detail").get("items")
        items.pop() # 👈 💣

    handler(event, {})

    scan_results = table.scan()
    print(f"ddb items:{scan_results}")
    assert scan_results['Count'] == 1
```

heitorlessa:
Hi @ojongerius! Could you please create a bug report and add a test for this?

I can see customers inadvertently missing the copy by reference nature leading to this issue - deep copy is the right call.

Once we have a bug report and a quick test, we should be able to merge it quickly.

Thank you 🙏

ojongerius:
Cheers @heitorlessa! I've added a test and created #1093

tests/functional/idempotency/test_idempotency.py
```diff
@@ -274,6 +274,40 @@ def lambda_handler(event, context):
     stubber.assert_no_pending_responses()
     stubber.deactivate()
 
+@pytest.mark.parametrize("idempotency_config", [{"use_local_cache": False}, {"use_local_cache": True}], indirect=True)
+def test_idempotent_lambda_first_execution_event_mutation(
```
heitorlessa, line 278: This test passes without this change, as the `expected_params_update_item` uses the exact same hash `expected_params_put_item` uses in the first operation -- unsure yet whether that's an issue on the test setup.

I'm gonna launch a fresh stack with the latest code to isolate it

heitorlessa: yep, it's the test setup. Repro code: https://github.com/heitorlessa/pt-1090

**Pass by reference (latest code)**

image

**Pass by value (with fix)**

image

heitorlessa (commented):
there's an issue with the test setup as `hashed_idempotency_key` is the same in both instances; meaning the test passes with or without the fix.

I'll push a fix today. We'll need to improve the build out of stub expected params (eventually moving to DynamoDB Local to reduce boilerplate for functional tests)

heitorlessa:
@ojongerius tests fixed - could you please have one last look before we merge?

Thanks

ojongerius:
Thanks for that 🙏 

Had one last look and verified this in my own environment 👍

Issue #1084, opened 2022-03-29
Logger clear_state flip removing structure keys


alexanderluiscampino:
Using the `clear_state` argument in the logger decorator for lambda handlers is destroying the default structured keys. 

**What were you trying to accomplish?**
Just trying to get some lambda logs with custom keys, which ideally would be cleaned up after each warm call, such that those custom key values do not trickle to the next invocation.

## Expected Behavior
Using the clear_state argument should not remove the default structured keys. Those should remain fixed. Only custom keys should be removed upon new invocation.

## Current Behavior
Currently, default structured keys are being removed from the get-go.

## Steps to Reproduce (for bugs)
First, this:
```python
from aws_lambda_powertools import logging
from aws_lambda_powertools.utilities.typing import LambdaContext


logger = logging.Logger()

@logger.inject_lambda_context(clear_state=False)
def lambda_handler(event: dict, context: LambdaContext):
    logger.info("HEREEE")
```
Output:
```json
{
    "level": "INFO",
    "location": "lambda_handler:21",
    "message": "HEREEE",
    "timestamp": "2022-03-29 17:53:56,985+0000",
    "service": "service_undefined",
    "cold_start": true,
    "function_name": "alex-test-layers",
    "function_memory_size": "1024",
    "function_arn": "arn:aws:lambda:us-west-2:111111111111:function:alex-test-layers",
    "function_request_id": "a1453466-2abe-4a55-ad9e-640e745f4133",
    "xray_trace_id": "1-624347b4-589b3226782e49797249edb1"
}
```
Now this `clear_state=True`
```python
from aws_lambda_powertools import logging
from aws_lambda_powertools.utilities.typing import LambdaContext


logger = logging.Logger()

@logger.inject_lambda_context(clear_state=True)
def lambda_handler(event: dict, context: LambdaContext):
    logger.info("HEREEE")
```
Output:
```json
{
    "message": "HEREEE",
    "service": "service_undefined",
    "cold_start": true,
    "function_name": "alex-test-layers",
    "function_memory_size": "1024",
    "function_arn": "arn:aws:lambda:us-west-2:111111111111:function:alex-test-layers",
    "function_request_id": "e3cf7717-88ea-43dc-b80c-57754bc6b8c9",
    "xray_trace_id": "1-62435f24-54e5827e68d838640494dca4"
}
```

## Environment

* **Powertools version used**: aws-lambda-powertools==1.25.5
[...]

PR #1095, opened 2022-04-06
fix(logger): clear_state should keep custom key formats

eliasbrange:
**Issue #, if available:** #1084

## Description of changes:

When using `clear_state=True`, overriden formats for standard logging keys like `location` and `timestamp` was being overriden with the default formats.  Overriding the formats are explained in the docs like this:

```python

date_format = "%m/%d/%Y %I:%M:%S %p"
location_format = "[%(funcName)s] %(module)s"

# override location and timestamp format
logger = Logger(service="payment", location=location_format, datefmt=date_format)
```

Issue was that when `clear_state=True` was used, the default keys were reset to their defaults (by the fix implemented in #1088). This fix saves the formatting keys on instantiation and uses those instead when `clear_state` is called.

**Checklist**

 

* [x] Update tests
* [ ] Update docs

## Breaking change checklist

**RFC issue #**:

### DIFF
--- a/tests/functional/test_logger.py
+++ b/tests/functional/test_logger.py
@@ -683,6 +683,29 @@ def test_clear_state_keeps_standard_keys(lambda_context, stdout, service_name):
         assert key in second_log
 
 
+def test_clear_state_keeps_custom_keys(lambda_context, stdout, service_name):
+    # GIVEN
+    date_format = "%Y"
+    location_format = "%(module)s:%(funcName)s:%(lineno)d"
+    logger = Logger(service=service_name, stream=stdout, location=location_format, datefmt=date_format)
+
+    # WHEN clear_state is set
+    @logger.inject_lambda_context(clear_state=True)
+    def handler(event, context):
+        logger.info("Foo")
+
+    # THEN all standard keys should be available as usual
+    handler({}, lambda_context)
+    handler({}, lambda_context)
+
+    first_log, second_log = capture_multiple_logging_statements_output(stdout)
+
+    assert re.fullmatch("[0-9]{4}", first_log["timestamp"])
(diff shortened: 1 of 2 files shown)


tests/functional/test_logger.py
```diff
@@ -683,6 +683,29 @@ def handler(event, context):
...
+    # THEN all standard keys should be available as usual
+    handler({}, lambda_context)
+    handler({}, lambda_context)
+
+    first_log, second_log = capture_multiple_logging_statements_output(stdout)
+
+    assert re.fullmatch("[0-9]{4}", first_log["timestamp"])
+    assert re.fullmatch("[0-9]{4}", second_log["timestamp"])
+    assert re.fullmatch(".+:.+:[0-9]+", first_log["location"])
+    assert re.fullmatch(".+:.+:[0-9]+", second_log["location"])
```
heitorlessa, line 706: Can we use an alternative to regex to confirm custom keys are being preserved?

While correct this will slow down tests as we grow them (currently 684), and it might make it difficult for the untrained eye.

For example, we could:

* replace `datefmt=date_format` with any custom key like `foo=bar` and assert on that
* replace `:%(lineno)d` with `clear_state` `location_format` so you can assert using `test_logger.handler:clear_state` - line number doesn't become that important and we continue to assert on that custom location

Thank you!

heitorlessa (changes requested):
one suggestion on tests so we reduce the use of regex to improve test perf and ease future maintenance for the untrained eye.

tests/functional/test_logger.py
eliasbrange: Smart, not a fan of regex so happy to exclude it. Pushing update!

heitorlessa (approved):

Issue #1147, opened 2022-04-21
Bug: JMESPath causes issues with updating dev deps


sthulb:
### Expected Behaviour

To be able to mange dev deps without causing issues 

### Current Behaviour

```
Installing dependencies from lock file

  SolverProblemError

  Because aws-lambda-powertools depends on jmespath (^1.0.0) which doesn't match any versions, version solving failed.

  at /opt/hostedtoolcache/Python/3.8.12/x64/lib/python3.8/site-packages/poetry/puzzle/solver.py:241 in _solve
      237│             packages = result.packages
      238│         except OverrideNeeded as e:
      239│             return self.solve_in_compatibility_mode(e.overrides, use_latest=use_latest)
      240│         except SolveFailure as e:
    → 241│             raise SolverProblemError(e)
      242│ 
      243│         results = dict(
      244│             depth_first_search(
      245│                 PackageNode(self._package, packages), aggregate_package_nodes
make: *** [Makefile:9: dev] Error 1
Error: Process completed with exit code 
```

### Code snippet

```python
N/A
```

### Possible Solution

Remove JMESPath

### Steps to Reproduce

update any dev dep

### AWS Lambda Powertools for Python version

latest

### AWS Lambda function runtime

3.7

### Packaging format used

PyPi

michaelbrewer:
don't we love poetry :(

PR #1148, opened 2022-04-21
fix(deps-dev): remove jmespath due to dev deps conflict 

sthulb:
**Issue number:**
#1147 

## Summary
Removes JMESPath to allow downstream deps to solve this.

### Changes

Removes JMESPath dependency from `pyproject.toml`

### User experience

See issue (#1147)

## Checklist

* [X] I have performed a self-review of my this change
* [X] Changes are tested
* [X] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,8 @@ pydantic = {version = "^1.8.2", optional = true }
 email-validator = {version = "*", optional = true }
 
 [tool.poetry.dev-dependencies]
+# 2022-04-21: jmespath was removed, to be re-added once we drop python 3.6.
+# issue #1148
 coverage = {extras = ["toml"], version = "^6.2"}
 pytest = "^7.0.1"
 black = "^21.12b0"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -24,10 +24,6 @@ python = "^3.6.2"
 aws-xray-sdk = "^2.8.0"
 fastjsonschema = "^2.14.5"
 boto3 = "^1.18"
```
heitorlessa, line 26: Could you add a comment here so we review jmespath state after 3.6 Lambda deprecation?

I think we can trust boto to not drop jmespath given how crucial it is for them, but I wanna be conservative as (1) it doesn't add any extra bytes at runtime, and (2) jmespath hasn't been updated in years

heitorlessa (changes requested):

pyproject.toml
heitorlessa: After this we can merge it!

heitorlessa (approved):

PR #1247, opened 2022-06-13
chore(ci): add end to end testing mechanism

mploski:
**Issue number:** https://github.com/awslabs/aws-lambda-powertools-python/issues/1226

## Summary

This PR implements proposal shared in RFC

IT add mechanism to run an end to end tests on Lambda Powertools library using real AWS Services (Lambda, DynamoDB) in real AWS Account.
Initially, tests can be run manually by maintainers on specific branch to ensure expected feature works.
Tests should be triggered in GitHub but also maintainers/contributors should be able to run them in their local environment using their own AWS Account.

In order to run them manually:
1. Export your AWS_PROFILE to point tests to your AWS Account2. 
2. Run ` make e2e-test` from project root directory

### Changes
1. Modify Makefile to add new `e2e-test` command automating test execution. pytest uses x-dist plugin to parallelize execution over test groups.
2. Add new test groups **(logger/metrics/tracer)** under **tests/e2e** directory. They contains pytest tests that calls lambda endpoint and fetch information about execution from CloudWatch Metrics/CloudWatch Logs/X-Ray API. We store Lambda handler code under **handler/** sub directory. We use this handlers to spawn new lambdas during infrastructure deployment
3. Under **tests/e2e/utils** directory we store:
- **infrastructure.py** - contains class responsible for spawning infrastructure in test account
- **helpers.py** -  stores helpers methods here that simplify fetching data from CloudWatch Metrics/CloudWatch Logs/X-Ray API
- **Dockerfile** - used by CDK code to create custom lambda powertools layer 
4. Add session fixture in **conftest.py** file responsible for creating infrastructure for every test group and removing it at the end of test group execution. Fixture uses infrastructure class methods for creating/removing infrastructure
5. Modify GitHub workflow mechanism by adding run-e2e-tests.yaml. User triggers new workflow manually from github. Workflow authorize to our test AWS account using OIDC and then run end to end tests against it

### User experience

Users are not affected by this change. It enables new testing capabilities for maintainers and contributors to ensure highest possible code quality

### Maintainer experience
Maintainer can run this job either from GitHub Actions Page

Screenshot 2022-06-14 at 08 36 50

Or from local shell
image

## Checklist

* [x] Changes are tested
* [ ] Changes are documented

<details>
No
**RFC issue number**:
https://github.com/awslabs/aws-lambda-powertools-python/issues/1226

Checklist:

</details>

## Acknowledgment


heitorlessa:
Looking into it now. Just ran the tests in my account and it failed because I didn't do `cdk bootstrap` before. Because it's CDK CLI and it's in Node, we need to document that CDK CLI is needed and that you also must run `cdk bootstrap`.

That way we don't fiddle with Python devs wanting to contribute to have Node + CDK CLI. If they want to contribute E2E tests, they can read the prerequisites in the docs - I'm sure that was your reasoning for not having it under `make dev` target :)

MAINTAINERS.md
```diff
@@ -171,7 +172,15 @@ Some examples using our initial and new RFC templates: #92, #94, #95, #991, #122
 
 Firstly, make sure you are using the `develop` branch and it is up to date with the origin.
 
-There are three main steps to release a new version: Changelog generation, version bumping, and drafting release notes.
+There are four main steps to release a new version: Run end to end tests, Changelog generation, version bumping, and drafting release notes.
+
+#### Run end to end tests
+
+You can run tests either from your local machine's shell by exporting `AWS_PROFILE` environment variable and running `make e2e tests`. 
+You may also do this directly from GitHub Console. Go to actions tab and choose `run-e2e-tests` workflow from workflow's list. Pick the branch and run tests.
```
heitorlessa, line 180:
```suggestion
You may also do this directly from GitHub Console. Run the [run-e2e-tests workflow](https://github.com/awslabs/aws-lambda-powertools-python/actions/workflows/run-e2e-tests.yml) and pick the branch you want to run tests against.
```

Tip: GH now supports deep linking a given Actions workflow :)

heitorlessa (changes requested):
Quick pass before I look at the tests - made some suggestions and a tip on the new GH Actions deep link

tests/e2e/metrics/handlers/basic_handler.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+from aws_lambda_powertools import Metrics
+from aws_lambda_powertools.metrics import MetricUnit
+
+METRIC_NAMESPACE = os.environ["METRIC_NAMESPACE"]
+METRIC_NAME = os.environ["METRIC_NAME"]
+SERVICE_NAME = os.environ["SERVICE_NAME"]
+
+metrics = Metrics(namespace=METRIC_NAMESPACE, service=SERVICE_NAME)
```
heitorlessa, line 10: both are optional if you already are setting environment variable - just make sure the env var follows our standard: `POWERTOOLS_METRICS_NAMESPACE` and `POWERTOOLS_SERVICE_NAME`.

then it becomes `metrics = Metrics()`

tests/e2e/metrics/test_metrics.py
```diff
@@ -0,0 +1,38 @@
...
+            "METRIC_NAMESPACE": f"powertools-e2e-metric-{uuid.uuid4()}",
+            "METRIC_NAME": "business-metric",
+            "SERVICE_NAME": "test-powertools-service",
+        },
+    }
+
+
+@pytest.mark.e2e
+def test_basic_lambda_metric_visible(execute_lambda: conftest.LambdaExecution, config: conftest.LambdaConfig):
+    start_date = execute_lambda["execution_time"]
```
heitorlessa, line 25: nitpick: could you add GIVEN/WHEN/THEN for consistency?

tests/e2e/tracer/handlers/basic_handler.py
```diff
@@ -0,0 +1,16 @@
+import os
+
+from aws_lambda_powertools import Tracer
+
+tracer = Tracer()
+tracer = Tracer(service="e2e-tests-app")
+
+ANNOTATION_KEY = os.environ["ANNOTATION_KEY"]
+ANNOTATION_VALUE = os.environ["ANNOTATION_VALUE"]
+
```
heitorlessa, line 10: could you also add a test with a synchronous and asynchronous function to be traced?

heitorlessa (changes requested):
Additional comments on tests and one major on splitting `execute_lambda` to separate control and data plane operations. Gonna look at helpers and Infra next

tests/e2e/utils/helpers.py
```diff
@@ -0,0 +1,97 @@
...
+    return filtered_logs
+
+
+@lru_cache(maxsize=10, typed=False)
+@retry(ValueError, delay=1, jitter=1, tries=10)
+def get_metrics(
+    namespace: str, cw_client: Any, start_date: datetime, end_date: datetime, metric_name: str, service_name: str
+):
+    response = cw_client.get_metric_data(
+        MetricDataQueries=[
```
heitorlessa, line 54: I thought you were gonna try the new shiny Metric Insights SQL :D /s

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/cloudwatch-metrics-insights-querylanguage.html

tests/e2e/utils/helpers.py
```diff
@@ -0,0 +1,97 @@
...
+            continue
+        filtered_logs.append(message)
+
+    return filtered_logs
+
+
+@lru_cache(maxsize=10, typed=False)
+@retry(ValueError, delay=1, jitter=1, tries=10)
+def get_metrics(
+    namespace: str, cw_client: Any, start_date: datetime, end_date: datetime, metric_name: str, service_name: str
```
heitorlessa, line 51: Quite possibly not now, but I can see a future where `end_date` could be optional to get as many metrics as possible for a given metric.

tests/e2e/utils/helpers.py
```diff
@@ -0,0 +1,97 @@
...
+        EndTime=end_date,
+    )
+    result = response["MetricDataResults"][0]
+    if not result["Values"]:
+        raise ValueError("Empty response from Cloudwatch. Repeating...")
+    return result
+
+
+@retry(ValueError, delay=1, jitter=1, tries=10)
+def get_traces(lambda_function_name: str, xray_client: Any, start_date: datetime, end_date: datetime):
```
heitorlessa, line 79: you can use this for typing `xray_client`: https://youtype.github.io/boto3_stubs_docs/mypy_boto3_xray/client/

heitorlessa: `lambda_function_name` will work for now, but maybe we should be thinking of a filter expression string instead. Reason being is that if you later wanted to look up for API GW/AppSync/EventBridge/SQS Traces, or if you need to be more specific like `traces by Lambda with this annotation key`, you might need to refactor it

tests/e2e/tracer/test_tracer.py
```diff
@@ -0,0 +1,46 @@
...
+    end_date = start_date + datetime.timedelta(minutes=5)
+
+    trace = helpers.get_traces(
+        start_date=start_date,
+        end_date=end_date,
+        lambda_function_name=lambda_arn.split(":")[-1],
+        xray_client=boto3.client("xray"),
+    )
+
+    for segment in trace["Traces"][0]["Segments"]:
```
heitorlessa, line 33: this might break if we add an additional test. What could be done here is to add a specific annotation per handler and fetch traces that contain that instead using filter expression.

heitorlessa (changes requested):
Some comments and refactoring ideas to prevent some gotchas with `get_traces` helper method. I'll look into the CDK code tomorrow

tests/e2e/conftest.py
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+    lambda_arns = infra.deploy()
+    execution_time = datetime.datetime.utcnow()
+
+    for _, arn in lambda_arns.items():
+        helpers.trigger_lambda(lambda_arn=arn, client=infra.lambda_client)
+    yield {"arns": lambda_arns, "execution_time": execution_time}
+    # Ensure stack deletion is triggered at the end of the test session
+    infra.delete()
```
pcolazurdo, line 47: delete should be inside a ```try/except/finally``` construct to be sure the stack is deleted even if calling ```trigger_lambda``` fails

pyproject.toml
```diff
@@ -59,7 +59,9 @@ mypy-boto3-secretsmanager = "^1.24.0"
 mypy-boto3-ssm = "^1.24.0"
 mypy-boto3-appconfig = "^1.24.0"
 mypy-boto3-dynamodb = "^1.24.0"
```
alexpulver, line 61: Looking at https://pypi.org/project/mypy-boto3/, it says _"This tool is not longer required for annotations to work. Do not use it, install [boto3-stubs](https://pypi.org/project/boto3-stubs/) instead."_

tests/e2e/metrics/test_metrics.py
```diff
@@ -0,0 +1,38 @@
...
+from .. import conftest
+from ..utils import helpers
+
+
+@pytest.fixture(scope="module")
+def config() -> conftest.LambdaConfig:
+    return {
+        "parameters": {},
+        "environment_variables": {
+            "METRIC_NAMESPACE": f"powertools-e2e-metric-{uuid.uuid4()}",
```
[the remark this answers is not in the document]
heitorlessa: Great catch!

pyproject.toml
heitorlessa: Did boto3-stubs work for you when creating interfaces and such? 

I had varying success, especially with mypy

alexpulver: I didn't use either - saw this library here, and wanted to learn more about it, then saw this notice and commented back :)

tests/e2e/utils/infrastructure.py
```diff
@@ -0,0 +1,182 @@
...
+            handlers_dir=self.handlers_dir,
+            stack_name=self.stack_name,
+            environment_variables=self.environment_variables,
+            **self.config,
+        )
+        self._upload_assets(template, asset_root_dir)
+
+        response = self._deploy_stack(self.stack_name, template)
+
+        return self._transform_output(response["Stacks"][0]["Outputs"])
```
alexpulver, line 50: That function seems to reimplement `cdk deploy` command and its use of [cdk-assets](https://www.npmjs.com/package/cdk-assets). I assume the intent is to use Python instead of shelling out to CDK CLI for deployment. It can work, but will it be as stable as the CDK CLI implementation? Does it worth forking implementation to avoid shelling out?

alexpulver: If you are worried about performance, consider using CDK CLI with --hotswap option, and disabling automatic rollbacks: https://aws.amazon.com/about-aws/whats-new/2021/10/aws-cdk-releases-hotswap-rollback-control/

MAINTAINERS.md
```diff
@@ -171,7 +172,15 @@ Some examples using our initial and new RFC templates: #92, #94, #95, #991, #122
 
 Firstly, make sure you are using the `develop` branch and it is up to date with the origin.
 
-There are three main steps to release a new version: Changelog generation, version bumping, and drafting release notes.
+There are four main steps to release a new version: Run end to end tests, Changelog generation, version bumping, and drafting release notes.
+
+#### Run end to end tests
+
```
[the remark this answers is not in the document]
mploski: Added  sentence about prerequisites in E2E section

tests/e2e/logger/test_logger.py
```diff
@@ -0,0 +1,163 @@
...
+            "ADDITIONAL_KEY": "extra_info",
+        },
+    }
+
+
+@pytest.mark.e2e
+def test_basic_lambda_logs_visible(execute_lambda: conftest.LambdaExecution, config: conftest.LambdaConfig):
+    # GIVEN
+    lambda_arn = execute_lambda["arns"]["basichandlerarn"]
+    timestamp = int(execute_lambda["execution_time"].timestamp() * 1000)
```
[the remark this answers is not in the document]
mploski: Could you elaborate a bit more? How do you imagine it look like briefly?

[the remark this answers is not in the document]
heitorlessa: yes! `execution_timestamp` could even be a property to help w/ caching

MAINTAINERS.md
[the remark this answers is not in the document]
pcolazurdo: what about adding a e2e-prereq in Makefile to install all prereqs?

tests/e2e/utils/helpers.py
```diff
@@ -0,0 +1,97 @@
+import json
+from datetime import datetime
+from functools import lru_cache
+from typing import Any, Optional, Union
```
pcolazurdo, line 4: Import ```List``` as you are using it in ```get_logs```

Makefile
```diff
@@ -17,14 +17,17 @@ lint: format
 	poetry run flake8 aws_lambda_powertools/* tests/*
 
 test:
-	poetry run pytest -m "not perf" --cov=aws_lambda_powertools --cov-report=xml
+	poetry run pytest -m "not (perf or e2e)" --cov=aws_lambda_powertools --cov-report=xml
```
pcolazurdo, line 20: In this case, ```-m "not (perf or e2e)"``` option doesn't work as expected. [Tests should be marked](https://docs.pytest.org/en/7.1.x/example/markers.html) with ```@pytest.mark.e2e```  for the ```-m``` option to work as expected. ```--ignore tests/e2e``` (leaving ```-m "not perf"``` intact) works just fine, or you need to mark all the e2e tests with the corresponding custom mark.

Makefile
```diff
@@ -17,14 +17,17 @@ lint: format
...
 
 unit-test:
 	poetry run pytest tests/unit
 
+e2e-test:
+	poetry run pytest -n 3 --dist loadscope --durations=0 --durations-min=1 tests/e2e
+
 coverage-html:
-	poetry run pytest -m "not perf" --cov=aws_lambda_powertools --cov-report=html
+	poetry run pytest -m "not (perf or e2e)" --cov=aws_lambda_powertools --cov-report=html
```
pcolazurdo, line 30: Same as above, ```-m "not (perf or e2e)"``` doesn't work as expected as the e2e tests aren't marked

MAINTAINERS.md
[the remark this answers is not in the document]
mploski: The only thing to install is CDK CLI. I don't want to install it through our makefile as we don't control nodejs  environment of the user, including package manager (npm/yarn/pnpm) so I cannot reliably pick one without enforcing additional convention.

tests/e2e/logger/test_logger.py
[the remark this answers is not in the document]
mploski: ok, added

tests/e2e/logger/test_logger.py
```diff
@@ -0,0 +1,163 @@
...
+    cw_client = boto3.client("logs")
+
+    # WHEN
+    filtered_logs = helpers.get_logs(
+        lambda_function_name=lambda_arn.split(":")[-1], start_time=timestamp, log_client=cw_client, run="first"
```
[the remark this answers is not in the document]
mploski: I removed this attribute as this is not needed now - we cache response from cw logs per lambda arn

tests/e2e/logger/test_logger.py
```diff
@@ -0,0 +1,163 @@
...
+    timestamp = int(execute_lambda["execution_time"].timestamp() * 1000)
+    cw_client = boto3.client("logs")
+
+    # WHEN
+    filtered_logs = helpers.get_logs(
+        lambda_function_name=lambda_arn.split(":")[-1], start_time=timestamp, log_client=cw_client
+    )
+
+    # THEN
+    for log in filtered_logs:
```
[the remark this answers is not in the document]
mploski: I like it, its cleaner.

tests/e2e/metrics/handlers/basic_handler.py
mploski: ok

tests/e2e/tracer/handlers/basic_handler.py
mploski: added

tests/e2e/tracer/test_tracer.py
```diff
@@ -0,0 +1,46 @@
...
+        document = json.loads(segment["Document"])
+        if document["origin"] == "AWS::Lambda::Function":
+            for subsegment in document["subsegments"]:
+                if subsegment["name"] == "Invocation":
+                    for x_subsegment in subsegment["subsegments"]:
+                        metadata = x_subsegment["metadata"]
+                        annotation = x_subsegment["annotations"]
```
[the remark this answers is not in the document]
mploski: created helper function that extracts metadata and annotations for segments and expose it to the user

tests/e2e/tracer/test_tracer.py
```diff
@@ -0,0 +1,46 @@
...
+
+    assert (
+        metadata["e2e-tests-app"][config["environment_variables"]["ANNOTATION_KEY"]]
```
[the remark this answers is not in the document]
mploski: done

tests/e2e/utils/Dockerfile
```diff
@@ -0,0 +1,14 @@
+# The correct AWS SAM build image based on the runtime of the function will be
+# passed as build arg. The default allows to do `docker build .` when testing.
```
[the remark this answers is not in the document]
mploski: ok

tests/e2e/utils/helpers.py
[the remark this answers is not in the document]
mploski: ok added types for cw, xray and lambda

tests/e2e/conftest.py
mploski: Separated infra creation from lambda execution

pyproject.toml
mploski: @heitorlessa I also used boto3-stubs in other projects. Do you have any specific example that it didn't work for you and mypy-boto3 does?

tests/e2e/utils/infrastructure.py
mploski: Yes, the idea was - since there is no possibility to run deployment from CDK SDK directly, to synthesize cdk code to cloudformation stack and deploy it from cloudformation client ( which also includes uploading assets to be able to reference it from CF stack). Do you think it may be something unstable? The only thing that is internal to CDK here is asset naming convention Im using. The question is how probable is it will be changed. I'm happy to discuss it further and change my mind :-)
--hotswap option won't change performance here as still we need to deploy CW stack. I did measurements and decided to use batch lambda deployment in CF stack instead cf stack + hotswaping

Makefile
mploski: all e2e tests are marked with `@pytest.mark.e2e`

Makefile
mploski: will change to --ignore as it seems less error prone. Thx!

tests/e2e/utils/helpers.py
mploski: Unified types to use List everywhere

tests/e2e/tracer/test_tracer.py
mploski: The idea was to fetch all new traces once and then manipulate/search them locally to avoid multiple calls to x-ray. Maybe we could simply iterate over downloaded traces and combine results in the list?

tests/e2e/utils/infrastructure.py
mploski: By SDK I specifically meant python cdk lib which doesn't expose deployment methods ( as oppose to typescript SDK where you can do deployment directly from code). Could you elaborate a bit more how would you see this deployment designed?

Using hotswap was actually my initial idea and I created POC based on this. After discussion with and @heitorlessa @am29d and measuring latency I changed the solution to use only batch deployment using CF stack. It seems the fastest way since hotswap using CDK takes around 13 seconds in average and I still need to deploy CF stack (numbers: https://github.com/awslabs/aws-lambda-powertools-python/files/8833059/CF.stack.vs.hotswap.using.pdf.)
Also, in order to use hotswap capability with proposed design I need to add more complexity to this solution to dynamically overwrite handler file in cdk.out directory.

I will look at the cdk-assets code + will do a quick test with deplyoment through CLI and compare numbers.

tests/e2e/utils/helpers.py
mploski: Wasn't aware of this! Replacing code to use SQL

tests/e2e/utils/helpers.py
mploski: I assumed that it is responsibility of user to put end_date to now if they want to get all metrics. I can move this logic to inside of the method and set end_date to datetime.datetime.now() if it is not provided by user. What do you think?

tests/e2e/utils/helpers.py
mploski: make sense, changed to accept filter_expression instead of lambda function name

tests/e2e/utils/helpers.py
mploski: modified  end_date to be optional

.github/workflows/run-e2e-tests.yml
```diff
@@ -0,0 +1,35 @@
...
+      contents: read
+    strategy:
+      matrix:
+        version: ["3.7", "3.8", "3.9"]
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v3
+      #########################
+      # Release new version
+      #########################
```
heitorlessa: Leftover? Release new version?

MAINTAINERS.md
```diff
@@ -165,13 +166,23 @@ When necessary, be upfront that the time to review, approve, and implement a RFC
 
 Some examples using our initial and new RFC templates: #92, #94, #95, #991, #1226
 
+### Run end to end tests on merges to Develop branch
```
heitorlessa:
```suggestion
### Run end to end tests
```

MAINTAINERS.md
```diff
@@ -165,13 +166,23 @@ When necessary, be upfront that the time to review, approve, and implement a RFC
...
+
+In order to run end to end tests you need to install CDK CLI first and bootstrap your account with `cdk bootstrap` command. For additional details follow [documentation](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html).
+
+You can run tests either from your local machine's shell by exporting `AWS_PROFILE` environment variable and running `make e2e tests`. 
+You may also do this directly from GitHub Console. Run the [run-e2e-tests workflow](https://github.com/awslabs/aws-lambda-powertools-python/actions/workflows/run-e2e-tests.yml) and pick the branch you want to run tests against.
+
+> **Q: What if there's an error?**
+To be agreed
```
heitorlessa: Let's move this into its own section, like `Releasing a documentation hotfix`. We should be running end-to-end tests upon merges automatically (separate PR to avoid fatigue). Maintainers can do extra due diligence in running before a release if they'd like. 

This gives us two benefits:

* Room to make E2E tests run as part of each release automatically too in the future
* Maintainers to focus on drafting a good and intentionally informative release (e.g., release notes, etc.)

MAINTAINERS.md
```diff
@@ -165,13 +166,23 @@ When necessary, be upfront that the time to review, approve, and implement a RFC
 
 Some examples using our initial and new RFC templates: #92, #94, #95, #991, #1226
 
+### Run end to end tests on merges to Develop branch
+
+In order to run end to end tests you need to install CDK CLI first and bootstrap your account with `cdk bootstrap` command. For additional details follow [documentation](https://docs.aws.amazon.com/cdk/v2/guide/bootstrapping.html).
+
+You can run tests either from your local machine's shell by exporting `AWS_PROFILE` environment variable and running `make e2e tests`. 
+You may also do this directly from GitHub Console. Run the [run-e2e-tests workflow](https://github.com/awslabs/aws-lambda-powertools-python/actions/workflows/run-e2e-tests.yml) and pick the branch you want to run tests against.
```
heitorlessa:
```suggestion
To run locally, export `AWS_PROFILE` environment variable and run `make e2e tests`. To run from GitHub Actions, use [run-e2e-tests workflow](https://github.com/awslabs/aws-lambda-powertools-python/actions/workflows/run-e2e-tests.yml) and pick the branch you want to run tests against.

**NOTE**: E2E tests are run as part of each merge to `develop` branch.
```

nitpick to make it more direct.

MAINTAINERS.md
heitorlessa: we can remove this now.

MAINTAINERS.md
```diff
@@ -165,13 +166,23 @@ When necessary, be upfront that the time to review, approve, and implement a RFC
...
+To be agreed
+
 ### Releasing a new version
 
 > TODO: This is an area we want to increase automation while keeping communication at human level.
 
 Firstly, make sure you are using the `develop` branch and it is up to date with the origin.
 
-There are three main steps to release a new version: Changelog generation, version bumping, and drafting release notes.
+There are four main steps to release a new version: Changelog generation, version bumping, and drafting release notes.
```
heitorlessa:
```suggestion
There are three main steps to release a new version: Changelog generation, version bumping, and drafting release notes.
```

Returning it. In the near future, we'll be moving to a single step: `drafting release notes`.

tests/e2e/conftest.py
```diff
@@ -0,0 +1,61 @@
...
+        return self.arns
+
+    def get_lambda_arn(self, name: str) -> Optional[str]:
+        return self.arns.get(name)
+
+    def get_lambda_execution_time(self) -> datetime.datetime:
+        return self.execution_time
+
+    def get_lambda_execution_time_timestamp(self) -> int:
+        return int(self.execution_time.timestamp() * 1000)
```
heitorlessa: Thank you. This looks much better and it makes it easier to extend, improves readability and intent.

tests/e2e/logger/test_logger.py
```diff
@@ -0,0 +1,164 @@
+import boto3
+import pytest
+
+from .. import conftest
+from ..utils import helpers
```
heitorlessa: prefer absolute imports here. Reason being is that we will likely have more than one `conftest`, for example `tests/e2e/conftest` and `tests/e2e/logger/conftest`. When that happens it'll be easier to reason which we're using.

tests/e2e/logger/test_logger.py
```diff
@@ -0,0 +1,164 @@
...
+@pytest.mark.e2e
+def test_basic_lambda_logs_visible(execute_lambda: conftest.InfrastructureOutput, config: conftest.LambdaConfig):
+    # GIVEN
+    lambda_arn = execute_lambda.get_lambda_arn(name="basichandlerarn")
+    timestamp = execute_lambda.get_lambda_execution_time_timestamp()
+    cw_client = boto3.client("logs")
+
+    # WHEN
+    filtered_logs = helpers.get_logs(
+        lambda_function_name=lambda_arn.split(":")[-1], start_time=timestamp, log_client=cw_client
```
heitorlessa: sounds like a good candidate for a new method `get_lambda_function_name`

tests/e2e/metrics/test_metrics.py
```diff
@@ -0,0 +1,42 @@
...
+    metrics = helpers.get_metrics(
+        start_date=start_date,
+        end_date=end_date,
+        namespace=config["environment_variables"]["POWERTOOLS_METRICS_NAMESPACE"],
+        metric_name=config["environment_variables"]["METRIC_NAME"],
+        service_name=config["environment_variables"]["POWERTOOLS_SERVICE_NAME"],
+        cw_client=boto3.client(service_name="cloudwatch"),
+    )
+
+    # THEN
```
heitorlessa, line 37: can we also test cold start metric is created?

tests/e2e/tracer/handlers/basic_handler.py
```diff
@@ -0,0 +1,25 @@
...
+import os
+
+from aws_lambda_powertools import Tracer
+from aws_lambda_powertools.utilities.typing import LambdaContext
+
+tracer = Tracer(service="e2e-tests-app")
+
+ANNOTATION_KEY = os.environ["ANNOTATION_KEY"]
+ANNOTATION_VALUE = os.environ["ANNOTATION_VALUE"]
+ANNOTATION_ASYNC_VALUE = os.environ["ANNOTATION_ASYNC_VALUE"]
```
heitorlessa, line 11: that's not necessary actually, since it's a blocking operation either way ;) It's more to ensure a `async` function is annotated with Tracer.

tests/e2e/tracer/test_tracer.py
```diff
@@ -0,0 +1,55 @@
...
+from .. import conftest
+from ..utils import helpers
+
+
+@pytest.fixture(scope="module")
+def config():
+    return {
+        "parameters": {"tracing": "ACTIVE"},
+        "environment_variables": {
+            "ANNOTATION_KEY": f"e2e-tracer-{uuid.uuid4()}",
```
heitorlessa: That's gonna fail as `-` isn't a valid char in X-Ray Annotation. Use `_` instead. We wouldn't need a `uuid` as each annotation is unique to the trace + segment + subsegment.

https://github.com/aws/aws-xray-sdk-python/blob/894b419518e14b5cb2a75293ed4da17f87b2a7c0/aws_xray_sdk/core/models/entity.py#L146

tests/e2e/tracer/test_tracer.py
```diff
@@ -0,0 +1,55 @@
...
+        },
+    }
+
+
+@pytest.mark.e2e
+def test_basic_lambda_async_trace_visible(execute_lambda: conftest.InfrastructureOutput, config: conftest.LambdaConfig):
+    # GIVEN
+    lambda_arn = execute_lambda.get_lambda_arn(name="basichandlerarn")
+    start_date = execute_lambda.get_lambda_execution_time()
+    end_date = start_date + datetime.timedelta(minutes=5)
```
heitorlessa: I still find `execute_lambda` isn't quite the right wording for the intent. For example, this was already executed. It's a nitpick, we can look into that later as we try to write another E2E test in a separate PR - this will help measure the cognitive load for anyone not deep into this PR.

tests/e2e/utils/helpers.py
```diff
@@ -0,0 +1,127 @@
...
+        raise ValueError("Empty response from X-RAY. Repeating...")
+
+    trace_details = xray_client.batch_get_traces(
+        TraceIds=traces,
+    )
+
+    return trace_details
+
+
+def find_trace_additional_info(trace: Dict) -> List[TraceSegment]:
```
heitorlessa, line 109: could you add a docstring? I'm not sure what the name means. I see that you're extracting metadata and annotations from a subsegment(s), so docstrings could be helpful for the untrained eye later.... and eventually even repurpose this to be generic (any origin, any subsegment name) in the future.

tests/e2e/utils/infrastructure.py
```diff
@@ -0,0 +1,197 @@
...
+PYTHON_RUNTIME_VERSION = f"V{''.join(map(str, sys.version_info[:2]))}"
+
+
+class PythonVersion(Enum):
+    V37 = {"runtime": Runtime.PYTHON_3_7, "image": Runtime.PYTHON_3_7.bundling_image.image}
+    V38 = {"runtime": Runtime.PYTHON_3_8, "image": Runtime.PYTHON_3_8.bundling_image.image}
+    V39 = {"runtime": Runtime.PYTHON_3_9, "image": Runtime.PYTHON_3_9.bundling_image.image}
+
+
+class InfrastructureStackInterface(ABC):
```
heitorlessa:
```suggestion
class BaseInfrastructureStack(ABC):
```

for consistency across the code base.

tests/e2e/utils/infrastructure.py
```diff
@@ -0,0 +1,197 @@
...
+        ...
+
+
+class InfrastructureStack(InfrastructureStackInterface):
+    def __init__(self, handlers_dir: str, stack_name: str, config: dict) -> None:
+        self.stack_name = stack_name
+        self.handlers_dir = handlers_dir
+        self.config = config
+
+    def _create_layer(self, stack: Stack):
```
heitorlessa, line 41: Does it worth having this in the `Base` class? assuming it'll be universally used for potentially new Infrastructure Stack strategies.

I couldn't quite grasp the extensibility yet, dependency inversion, naming, etc, but let's not delay this PR any further. We can dedicate focus on this in the future, as there's lots of changes already. 

This will be easier to spot as we try to create more tests, more infra, etc.

heitorlessa (commented):
made some final comments. We don't need to address any nitpick, especially the `InfrastructureStack` thing as I don't wanna delay this PR anymore.

I think we're missing some tests but I also believe we should get the mechanism done, then as we add more tests and infrastructure it'll become clearer the areas we should simplify, abstract, etc. - it'll help us stay focused as there's a lot of changes (great changes) that will clog our judgement and over-engineer things.

tests/e2e/utils/infrastructure.py
mploski: As discussed I analysed cdk-assets code and tested also npm package with standalone client for infra creation.
cdk-assets code itself handles not only zipping asset and sending to s3 bucket - the other half handles sending docker image to ecr. The second part is something we don't need. 
cdk-assets follow this design https://github.com/aws/aws-cdk-rfcs/blob/master/text/0092-asset-publishing.md. I refactored python code to follow the same principle i.e rely on asset.manifest file to get list of assets to be zipped and pushed to specific s3 bucket.
I also measured time needed for uploading one asset (simple lambda handler) to s3. For python code it takes around 600ms. For code spawning process with cdk-assets it takes around 2.5 seconds. 

Considering cdk-assets code simplicity, added dependency from npm ecosystem ( security patching etc) and slightly lower performance I will stick to refactored python code that follow mentioned design. In case of problems we can basically swap `_upload_assets` method with the one using cdk-assets.  cc @heitorlessa

.github/workflows/run-e2e-tests.yml
mploski: removing. Leftover

MAINTAINERS.md
mploski: removed

tests/e2e/logger/test_logger.py
mploski: ok

tests/e2e/logger/test_logger.py
mploski: ok refactored

tests/e2e/metrics/test_metrics.py
mploski: Yes, but I would add it in another PR where we will be adding/adjusting tests

tests/e2e/tracer/handlers/basic_handler.py
mploski: true :-)

tests/e2e/tracer/test_tracer.py
mploski: good point, lets think about proper naming later. Any suggestions highly appreciated :-)

tests/e2e/utils/infrastructure.py
mploski: ok!

tests/e2e/utils/helpers.py
mploski: Yeah, I struggle to name it correctly as I find hard to combine annotations and metadata into one meaningful word - thats where additional information comes from :-) Will add docstring and we can adjust the naming if you have suggestion around it

tests/e2e/tracer/test_tracer.py
mploski: uuid is added as we could trigger this lambda twice and fetch traces from a timeframe that would return both of them and this is a way to distinquish it in such situation. Fixe to replace - with _

tests/e2e/utils/infrastructure.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def delete(self):
+        self.cf_client.delete_stack(StackName=self.stack_name)
+
+    def _upload_assets(self, asset_root_dir: str, asset_manifest_file: str):
+        """
+        This method is drop-in replacement for cdk-assets package s3 upload part.
+        https://www.npmjs.com/package/cdk-assets.
+        We use custom solution to avoid dependencies from nodejs ecosystem.
+        We follow the same design cdk-assets follows:
```
heitorlessa:
```suggestion
        We follow the same design cdk-assets:
```

heitorlessa (approved):

Issue #998, opened 2022-02-04
Cannot import batch with pydantic already imported, without email-validator


huonw:
**What were you trying to accomplish?**

We're trying to use the batch utility to process SQS records. We're using Pydantic for various things, including the messages pushed into SQS. If we've imported pydantic before importing `aws_lambda_powertools.utilities.batch` we hit an exception due to `email-validator` not being installed.

We're not using pydantic's `EmailStr`, and so haven't installed `email-validator` and would prefer not to if possible. (NB. email-validator itself is small (< 100K), but it seems to require `dnspython` which is almost 2MB.)

## Expected Behavior
Using batch classes shouldn't require `email-validator` to be installed, even if Pydantic has already been imported. 

## Current Behavior

```
Traceback (most recent call last):
  File "pydantic/networks.py", line 409, in pydantic.networks.import_email_validator
ModuleNotFoundError: No module named 'email_validator'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File ".../site-packages/aws_lambda_powertools/utilities/batch/__init__.py", line 7, in <module>
    from aws_lambda_powertools.utilities.batch.base import (
  File ".../site-packages/aws_lambda_powertools/utilities/batch/base.py", line 36, in <module>
    from aws_lambda_powertools.utilities.parser.models import DynamoDBStreamRecordModel
  File ".../site-packages/aws_lambda_powertools/utilities/parser/__init__.py", line 3, in <module>
    from . import envelopes
  File ".../site-packages/aws_lambda_powertools/utilities/parser/envelopes/__init__.py", line 1, in <module>
    from .apigw import ApiGatewayEnvelope
  File ".../site-packages/aws_lambda_powertools/utilities/parser/envelopes/apigw.py", line 4, in <module>
    from ..models import APIGatewayProxyEventModel
  File ".../site-packages/aws_lambda_powertools/utilities/parser/models/__init__.py", line 32, in <module>
    from .ses import (
  File ".../site-packages/aws_lambda_powertools/utilities/parser/models/ses.py", line 21, in <module>
    class SesReceipt(BaseModel):
  File "pydantic/main.py", line 204, in pydantic.main.ModelMetaclass.__new__
  File "pydantic/fields.py", line 488, in pydantic.fields.ModelField.infer
  File "pydantic/fields.py", line 419, in pydantic.fields.ModelField.__init__
[...]

PR #1259, opened 2022-06-23
feature(parser): handle when email_validator package is not present for SES model

mgochoa:
**Issue number:**
#998

## Summary
When `email_validator` package is not present in the dependencies, `pydantic.networks.EmailStr` will fail not when imported, but when it's used for typing check. Because inside `pydantic.networks.EmailStr`, there is a function `import_email_validator` which will try to import `email_validator`  dinamically.

### Changes
In `aws_lambda_powertools/utilities/parser/models/ses.py` :
```python
has_email_validator = "email_validator" in sys.modules

if has_email_validator:
    from pydantic.networks import EmailStr
else:
    logging.warning("email_validator package is not installed")
    EmailStr = NewType("EmailStr", str)  # type: ignore[no-redef, misc]
```

It means, now is checking for `email_validator` package, and defines conditionally which type it should take

### User experience

Use could perform:
```python
import pydantic
from aws_lambda_powertools.utilities import batch
````

without unexpected behavior such as:
`ImportError: email-validator is not installed, run 'pip install pydantic[email]'`

## Checklist

* [ x ] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [ ] I have performed a self-review of my this change
* [ ] Changes are tested
* [ ] Changes are documented
* [ x ] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/376ec0a2ac0d2a40e0af5717bef42ff84ca0d1b9/.github/semantic.yml#L2)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### CODE REVIEW

**Comment by heitorlessa** [comment]
Hey @mgochoa thanks a lot for taking the time to add this guard. More importantly, your first PR 🎉 !

> without unexpected behavior such as:

This is actually the intended behaviour because when installing `parser` it'll install `email-validator` package. As per the issue you referenced, this only happens for customers not following our recommended installation.

That being said, I can see this will continue to happen so best to add this small fix before we tackle it long-term. I'm making two small adjustments and we should be able to merge :)

File: aws_lambda_powertools/utilities/parser/models/ses.py [hunk 1]
```diff
@@ -1,12 +1,21 @@
...
 from pydantic.types import PositiveInt
 
 from ..types import Literal
 
+has_email_validator = "email_validator" in sys.modules
+
+if has_email_validator:
+    from pydantic.networks import EmailStr
+else:
+    logging.warning("email_validator package is not installed")
```
heitorlessa, line 16: Initialize a logger and change it to debug, so we'll know where it comes from.

```

...
logger.debug("email_validator package is not installed; using str instead")
```

aws_lambda_powertools/utilities/parser/models/ses.py
```diff
@@ -1,12 +1,21 @@
...
 
 from pydantic import BaseModel, Field
-from pydantic.networks import EmailStr
 from pydantic.types import PositiveInt
 
 from ..types import Literal
 
+has_email_validator = "email_validator" in sys.modules
+
+if has_email_validator:
```
heitorlessa, line 13: we're gonna need `#pragma: nocover` too because this will decrease our tests despite being intended behaviour ;-)

heitorlessa (changes requested):
two comments on checking `email_validator` being installed vs imported and minor optimizations.

rubenfonseca:
Hi @mgochoa thank you again so much for this PR! Any change you could incorporate the review feedback so we can merge this?

heitorlessa:
@mgochoa we're going to close this PR and credit you in #1607 for kicking the tires here, as we need to make a V2 release this week.

Thank you!

Issue #1183, opened 2021-11-03
Add public lambda layer to other regions


rubenesp87:
**Is your feature request related to a problem?**
There some missing regions on the public lambda layer, ie: me-south-1, ap-east-1, af-south-1, eu-south-1...

**Describe the solution you'd like**
Add the public lambda layer to the non-existing regions

**Describe alternatives you've considered**
Use SAR to deploy a CloudFormation stack with a copy of the Lambda Layer in all the regions, this is a bit tedious having multiple applications in multiple regions and working only in some regions.

Actual public layers
<img width="1072" alt="Screen Shot 2021-11-02 at 16 25 25" src="https://user-images.githubusercontent.com/17149817/140042703-f052b7cf-af2b-4231-9dd0-51bed0c1da2b.png">
>

heitorlessa:
Update: @am29d is working on migrating our internal Lambda Layers pipeline to CodePipeline in a Powertools AWS account. Once that's done (~H2), we should be able to expand to more AWS regions.

You can follow progress here: [github.com/orgs/awslabs/projects/51/views/11](https://github.com/orgs/awslabs/projects/51/views/11)

heitorlessa:
We're almost there @rubenesp87 - we've just finished migrated Lambda Layer pipeline to GitHub Actions, and we're now going to test the rollout to all AWS regions. Once that's done, the only thing we'll need is automation to ensure Docs are always up to date with all version we're going to have.

From there, we can expand additional Layers like ARM or Extra dependencies, etc.

PR #1278, opened 2022-07-04
chore(layers): add release pipeline in GitHub Actions

am29d:
**Issue number:** #1183 

## Summary

We have been using internal pipeline to release public layers for with powertools. Because of the initial design we could not ship public layers to all commercial regions. This PR introduces a CDK project and GitHub action workflow and we will integrate it into the release process in this repository. The goal is to deploy public layers after a release and update the documentation with the latest ARNs automatically. 

### Changes

The CDK project has two stacks `LayerStack` and `CanaryStack`. We use CDK customer resource capability to run a canary function after the layer deployment so we can verify that the layer works. We also need to track layer ARNs, powertools and the available region, so we can put this information into our documentation. To achieve that, there is dedicated version tracking application. After a successful deployment the canary functions sends an event with all the details. This application is maintained separately. The only information we need is the event bus ARN to send the data. 

To deploy into an AWS account we assume a role by using [GitHub OIDC configuration](https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services). We store the target account and the role name as secrets (though the target account will be public after the public layer is released). The OIDC configuration for target accounts will be provided, no additional work is required in this project. 

Open tasks

- [x] add secrets
- [x] add python caching for workflow jobs

After merge
- [ ] remove single deployment manually, to have consistent ARNs across all regions
- [ ] uncomment to enable all region

## Checklist

* [x] Changes are tested
* [x] Changes are documented

## Acknowledgment

-----
[View rendered layer/README.md](https://github.com/am29d/aws-lambda-powertools-python/blob/chore/layer-release/layer/README.md)


.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,207 @@
...
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '16.12'
+      - name: Setup python
+        uses: actions/setup-python@v4
```
heitorlessa, line 35:
```suggestion
      - name: Setup python
        uses: actions/setup-python@v4
        python-version: '3.9'
        cache: 'pip'
```

Enabling cache and explicit Python version to prevent potential breaks

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,207 @@
...
+#          "ap-southeast-3",
+#          "ap-northeast-3",
+#          "me-south-1"
+        ]
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
```
heitorlessa: what's the intention behind it? it might not be needed

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,207 @@
...
+        with:
+          name: cdk.out
+      - name: unzip cdk.out.zip
+        run: unzip cdk.out.zip
+      - name: CDK Deploy Layer
+        run: cdk deploy --app cdk.out --context region=${{ matrix.region }} 'PowertoolsLayerStack' --require-approval never --verbose
+      - name: CDK Deploy Canary
+        run: cdk deploy --app cdk.out --context region=${{ matrix.region}} --parameters DeployStage="BETA" 'CanaryStack' --require-approval never --verbose
+
+  deploy-prod:
```
heitorlessa, line 72: you can make a reusable GitHub Action workflow using `workflow_call` - example: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/workflows/export_pr_details.yml#L2

Ignore the `outputs` dance there. You'll need to define that Layers Account secret as a `secret` in the reusable workflow, as they don't inherit inputs/secrets by default.

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,207 @@
+name: Deploy layer to all regions
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  # TODO: remove after setup is finished
+  workflow_dispatch: { }
```
heitorlessa: Unless you truly need a explicit checkout from the `pull_request.head.ref`, you can also enable manual workflow runs - example: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/workflows/rebuild_latest_docs.yml#L12

Omitting those will checkout HEAD and run its course normally. If you'd want to checkout from a specific point in time OR from a separate repository, then these would be needed.

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,207 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '16.12'
```
heitorlessa, line 32: Use `env` to make it easier to change it afterwards - example: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/workflows/on_merged_pr.yml#L10:L10

.github/workflows/publish_layer.yml
heitorlessa: Example with `env` to make it easier to switch to another version later too: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/workflows/on_merged_pr.yml#L10:L10

heitorlessa (changes requested):
quick initial review on the Workflow with some tips

layer/README.md
```diff
@@ -0,0 +1,26 @@
+<!-- markdownlint-disable MD041 MD043-->
+# CDK Powertools layer
+
+This is a CDK project to build and deploy AWS Lambda Powertools layer to multiple commercial regions.
```
heitorlessa:
```suggestion
This is a CDK project to build and deploy AWS Lambda Powertools Lambda Layer to multiple commercial regions.
```

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Build the layer
+
+To build the layer construct you need to provide the powertools version.
```
heitorlessa:
```suggestion
To build the layer construct you need to provide the Powertools version that is [available in PyPi](https://pypi.org/project/aws-lambda-powertools/).
```

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+You can pass it as a context variable when running `synth` or `deploy`,
+
+```shell
+cdk synth --context version=1.25.1
+```
+
+## Canary stack
+
+We use a canary stack to verify that the deployment is successful and we can use the layer by adding it to a newly created Lambda function.
```
heitorlessa, line 17:
```suggestion
We use a canary stack to verify that the deployment is successful and we can use the Layer by adding it to a newly created Lambda function.
```

Capital `L` for Layer.

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+The canary is deployed after the layer construct. Because the layer ARN is created during the deploy we need to pass this information async via SSM parameter.
```
heitorlessa, line 18:
```suggestion
The canary is deployed after the Layer construct. Because the Layer ARN is created during the deploy we need to pass this information async via SSM parameter.
```

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+To achieve that, we establish a naming convention for the SSM key. The layer construct knows where to write the layer ARN after the deployment and the Canary stacks know where to read this information.
```
heitorlessa:
```suggestion
To achieve that, we establish a naming convention for the SSM key. The Layer construct knows where to write the Layer ARN after the deployment and the Canary stacks know where to read this information.
```

heitorlessa: what is the naming convention?

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Version tracking
+
+AWS Lambda versions the layers by incrementing a number at the end of the arn.
```
heitorlessa:
```suggestion
AWS Lambda versions Lambda Layers by incrementing a number at the end of the ARN.
```

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+This means it's hard to tell which powertools version is inside the layer.
```
heitorlessa:
```suggestion
This means it's hard to tell which Powertools version is inside the Layer.
```

Nit: `This makes it challenging to know which Powertools version a Layer contains.`

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+For better tracking of the ARNs and the corresponding version we need to keep track which powertools version was deployed to which layer.
```
heitorlessa, line 26:
```suggestion
For better tracking of the ARNs and the corresponding version we need to keep track which Powertools version was deployed to which Layer.
```

layer/README.md
```diff
@@ -0,0 +1,26 @@
...
+To achieve that we created two components. First, we created a version tracking app which receives evnets via Event Bridge. Second, after a successful canary deployment we send the layer ARN, powertools version, and the region to this Event Bridge.
```
heitorlessa:
```suggestion
To achieve that we created two components. First, we created a version tracking app which receives events via EventBridge. Second, after a successful canary deployment we send the layer ARN, Powertools version, and the region to this EventBridge.
```

EventBridge is the branded name

heitorlessa (changes requested):
quick typos and branding

layer/app.py
```diff
@@ -0,0 +1,31 @@
...
+
+from layer.canary_stack import CanaryStack
+from layer.layer_stack import LayerStack
+
+app = cdk.App()
+
+POWERTOOLS_VERSION: str = app.node.try_get_context("version")
+SSM_PARAM_LAYER_ARN: str = "/layers/powertools-layer-arn"
+
+VERSION_TRACKING_EVENT_BUS_ARN: str = "arn:aws:events:eu-central-1:027876851704:event-bus/VersionTrackingEventBus"
```
heitorlessa: nit: should we hard code it? I doubt we will change accounts anytime in the foreseeable future

layer/source.bat
```diff
@@ -0,0 +1,13 @@
+@echo off
```
heitorlessa: do we need this file?

layer/layer/canary/app.py
```diff
@@ -0,0 +1,97 @@
...
+                "region": os.environ["AWS_REGION"],
+                "version": powertools_version,
+                "layerArn": layer_arn,
+            }
+        ),
+    }
+
+    logger.info(f"sending notification event: {event}")
+
+    client = boto3.client("events", region_name="eu-central-1")
```
heitorlessa, line 94: do we need to hard code the region?

layer/layer/canary/app.py
```diff
@@ -0,0 +1,97 @@
...
+import os
+from importlib.metadata import version
+
+import boto3
+
+from aws_lambda_powertools import Logger, Metrics, Tracer
+
+logger = Logger(service="version-track")
+tracer = Tracer()
+metrics = Metrics(namespace="powertools-layer-canary", service="PowertoolsLayerCanary")
```
heitorlessa, line 12: you're setting different services, shouldn't they be the same?

You can also use `POWERTOOLS_SERVICE_NAME` env var in the CDK Canary Stack and omit them from the constructor, including `POWERTOOLS_METRICS_NAMESPACE`

layer/layer/canary/app.py
```diff
@@ -0,0 +1,97 @@
...
+    print(event)
+    request_type = event["RequestType"]
+    # we handle only create events, because we recreate the canary on each run
+    if request_type == "Create":
+        return on_create(event)
+
+
+def on_create(event):
+    props = event["ResourceProperties"]
+    print("create new resource with properties %s" % props)
```
heitorlessa: logger.info?

layer/layer/canary/app.py
```diff
@@ -0,0 +1,97 @@
...
+    logger.info("Running checks")
+    check_envs()
+    verify_powertools_version()
+    send_notification()
+    return True
+
+
+def on_event(event, context):
+    print(event)
+    request_type = event["RequestType"]
```
heitorlessa:
```suggestion
@logger.inject_lambda_context(log_event=True)
def on_event(event, context):
    request_type = event["RequestType"]
```

No need for `print`. Logger can also inject context so we know Lambda request ID if something went wrong

layer/layer/canary/app.py
heitorlessa: Do we plan to use them within the canary or simply instantiate? If the former, we're missing a few things like `@metrics.log_metrics(capture_cold_start_metric=True`, `@tracer.capture_lambda_handler`, `@tracer.capture_method`, and `@logger.inject_lambda_context`

layer/layer/canary/app.py
```diff
@@ -0,0 +1,97 @@
...
+    # we handle only create events, because we recreate the canary on each run
+    if request_type == "Create":
+        return on_create(event)
+
```
heitorlessa, line 34:
```suggestion
    # we handle only create events, because we recreate the canary on each run
    if request_type == "Create":
        return on_create(event)

     return "Nothing to be processed"
```

heitorlessa (changes requested):
some comments and suggestions on the canary

heitorlessa:
This is so so so exciting. I can't believe we're finally at the stage we will have Layers with Extras, ARM, all commercial regions, except the special ones (**yet**), and more importantly tightly integrated with the release process now.

Can't thank you enough!

.github/workflows/publish_layer.yml
am29d: Yeah, this does not make sense.

.github/workflows/publish_layer.yml
am29d: done

.github/workflows/publish_layer.yml
am29d: Not sure there is more value to extract it to a variable. We don't do it for python 3.8 version across workflows yet. It won't change frequently though and imho helps to read it more fluently and not scroll back to the top of the variable definition. What do you think?

.github/workflows/publish_layer.yml
am29d: added cache

layer/source.bat
am29d: maybe you want to switch to windows some day 😄 , removed the file

.github/workflows/publish_layer.yml
heitorlessa: It's a nitpick (forgot to tag it accordingly). I agree. we don't do it elsewhere yet, so let's stay consistent. Resolving.

layer/README.md
am29d: Why capital L 🤔? It's in lower case in the docs.

layer/README.md
am29d: The name itself is actually irrelevant we just write and read ssm key. The word "convention" is probably misleading. I will write it.

layer/app.py
am29d: I use it only in the canary, so make sense to place it there.

layer/layer/canary/app.py
am29d: yep

layer/layer/canary/app.py
am29d: yes, this is mandatory because the target EventBridge bus is in this region. The client is not smart enough to derive this information from the ARN.

layer/README.md
heitorlessa: we should be fixing those then (rewriting examples now, then branding and doc rewrite). The reason is that `Layer` with capital L means a reference to a product or something (copywriting), while `layer` being a noun.

am29d: No I mean `layer` is used in the [official docs](https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html)

heitorlessa: Ah! that's because they introduce it with `Lambda layer` and points to a glossary which we don't have. From there they continuously refer to Layer as it belongs to Lambda. Maybe we link to it and keep as `layer` then?

https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-concepts.html#gettingstarted-concepts-layer

.github/workflows/publish_layer.yml
am29d: Adding input to `workflow_dispatch` is necessary because we need the version. Any ideas on how to assign one of the inputs to the `RELEASE_TAG_VERSION`? Need some bash wizardry without if-statements

```
RELEASE_TAG_VERSION=${{ github.event.release.tag_name }}  || ${{ inputs.latest_published_version }}
```

Not sure nested `$` works in yaml to assign default value 🤔

```
RELEASE_TAG_VERSION=${ ${{ github.event.release.tag_name }}:-${{ inputs.latest_published_version }} }
```

layer/layer/canary/app.py
am29d: No plans to use the metrics. I just want to instantiate the classes to see if it works, in some cases the import is not enough.

am29d:
I have configured the GitHub OIDC connection to AWS accounts and added the secrets used in for layer deployments.

.github/workflows/delploy_cdk_stack.yml
```diff
@@ -0,0 +1,86 @@
...
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - name: aws credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-region: ${{ matrix.region }}
+          role-to-assume: arn:aws:iam::${{ secrets.target-account }}:role/${{ secrets.AWS_GITHUB_OIDC_ROLE }}
+      - name: Print assumed role
+        run: aws sts get-caller-identity
```
heitorlessa: debugging purposes?

.github/workflows/delploy_cdk_stack.yml
```diff
@@ -0,0 +1,86 @@
...
+          python-version: '3.9'
+          cache: 'pip'
+      - name: install cdk and deps
+        run: |
+          npm install -g aws-cdk@2.29.0
+          cdk --version
+      - name: install deps
+        run: |
+          cd layer
+          pip install -r requirements.txt
```
heitorlessa: nit

```suggestion
        run: pip install -r layer/requirements.txt
```

Should we also install `layer/requirements-dev.txt`?

.github/workflows/delploy_cdk_stack.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Download a single artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: cdk.out
+      - name: unzip cdk.out.zip
+        run: unzip cdk.out.zip
+      - name: CDK Deploy Layer
+        run: cdk deploy --app cdk.out --context region=${{ matrix.region }} 'PowertoolsLayerStack' --require-approval never --verbose
```
heitorlessa: you meant `LayerStack`? I'm not well versed in CDK as you're deploying from `cdk.out` (synthetized version), so I dunno if you're deploying the same set of resources from `layer/app.py` or something else

.github/workflows/delploy_cdk_stack.yml
```diff
@@ -0,0 +1,86 @@
...
+        type: string
+    secrets:
+      target-account:
+        required: true
+
+jobs:
+  deploy-cdk-stack:
+    runs-on: ubuntu-latest
+    needs:
+      - build-layer
```
heitorlessa: left over? I believe this will fail because there isn't a job `build-layer` defined within this workflow to depend on

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,77 @@
...
+          path: cdk.out.zip
+
+  deploy-beta:
+    needs:
+      - build-layer
+    uses: ./.github/workflows/deploy_cdk_stack.yml
+    with:
+      stage: "BETA"
+    secrets:
+      target-account: ${{ secrets.LAYERS_BETA_ACCOUNT }}
```
heitorlessa: this looks so clean ;) !

layer/requirements.txt
```diff
@@ -0,0 +1,3 @@
+aws-cdk-lib==2.29.0
+constructs>=10.0.0,<11.0.0
+cdk-lambda-powertools-python-layer==2.0.48
```
heitorlessa: for deterministic behaviour, shouldn't we pin `constructs` from CDK too?

for security, we might want to use their hash. Try this https://github.com/jazzband/pip-tools

heitorlessa (commented):
looks awesome! loved how reusable workflow turned out, so clean! Made some questions as I think you accidentally left a workflow dependency which might fail in the first run.

The only thing missing is the ability to add custom builds for when we introduce additional layers like Extra deps, and when we optimize boto size after our E2E are done -- we can do that later as part of introducing them.

.github/workflows/delploy_cdk_stack.yml
am29d: si, will remove

.github/workflows/delploy_cdk_stack.yml
am29d: yes, it should be outside when we call the step.

.github/workflows/delploy_cdk_stack.yml
am29d: `requirements-dev.txt` is not ... required 😅

am29d: Hmm the weird part is that everything needs to run inside the `layer` folder, since it's an isolated project, therefore change dir to `layer` and run everything else there, including cdk commands, and unzip.

heitorlessa: yeah it's just that I didn't see it being used, so I thought: "Hmmm? Lemme ask Alex"

.github/workflows/delploy_cdk_stack.yml
am29d: You are right, it used be called `PowertoolsLayerStack`, will adjust.

layer/requirements.txt
am29d: Oh absolutely

am29d: Added pinned versions and hashed to `requirements.txt` with `pip-compile`.

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,79 @@
...
+      - name: Archive CDK artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: cdk.out
+          path: cdk.out.zip
+
+  deploy-beta:
+    needs:
+      - build-layer
+    uses: ./.github/workflows/deploy_cdk_stack.yml
```
heitorlessa: maybe `reusable_deploy_layer_stack`?

.github/workflows/delploy_cdk_stack.yml
```diff
@@ -0,0 +1,84 @@
...
+        run: |
+          npm install -g aws-cdk@2.29.0
+          cdk --version
+      - name: install deps
+        run: |
+          pip install -r requirements.txt
+      - name: Download a single artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: cdk.out
```
heitorlessa: This could cause problems IF we have another workflow uploading `cdk.out` with the same name. Two solutions here:

1. Securely download an artifact generated by a trusted workflow run, example we have elsewhere: https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/workflows/export_pr_details.yml#L41:L41
2. Use an explicit name so we don't accidentally download `cdk.out` that isn't related to Layers

heitorlessa (changes requested):
two final comments: workflow naming, and address a potential security vulnerability when downloading `cdk.out`

.github/workflows/publish_layer.yml
```diff
@@ -0,0 +1,79 @@
+name: Deploy layer to all regions
+
+permissions:
+  id-token: write
+  contents: read
+
+on:
+  # TODO: remove after setup is finished
```
heitorlessa:
```suggestion
```

Removing as I actually had misread by in the other workflow. There could be scenarios where we might want to trigger it manually (hotfix) in the far future.

.github/workflows/delploy_cdk_stack.yml
am29d: Passing artefact name explicitly sounds good. ~~I was thinking about cleaning up the artefacts after the run with something like this: https://github.com/marketplace/actions/delete-artifact We don't need to keep the assets after a successful deployment. 🤔~~

heitorlessa (approved):

heitorlessa:
I can't believe it's finally here for the next release ;) In a separate PR, we should include a new section in the [MAINTAINERS playbook](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/MAINTAINERS.md) to explain a bit more on Lambda Layers - how to check it worked, areas in the docs to update (until we have automation), etc.

PR #1299, opened 2022-07-14
fix(parser): raise ValidationError when SNS->SQS keys are intentionally missing

rubenfonseca:
**Issue number:** #1009 

### Changes

When `UnsubscribeURL` or `SigningCertURL` were intentionally missing from the SNS -> SQS payload, `KeyError` was thrown.

Since the objective of the parser is to validate the payload, we changed the code so it returns
an expected [ValidationError](https://prompt-build--pydantic-docs.netlify.app/#error-handling) instead.

## Checklist

* [x] Changes are tested
* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/tests/functional/parser/test_sns.py
+++ b/tests/functional/parser/test_sns.py
@@ -1,3 +1,4 @@
+import json
 from typing import Any, List
 
 import pytest
@@ -103,3 +104,29 @@ def handle_sns_sqs_json_body(event: List[MySnsBusiness], _: LambdaContext):
 def test_handle_sns_sqs_trigger_event_json_body():  # noqa: F811
     event_dict = load_event("snsSqsEvent.json")
     handle_sns_sqs_json_body(event_dict, LambdaContext())
+
+
+def test_handle_sns_sqs_trigger_event_json_body_missing_signing_cert_url():
+    # GIVEN an event with a missing SigningCertURL
+    event_dict = load_event("snsSqsEvent.json")
+    payload = json.loads(event_dict["Records"][0]["body"])
+    assert payload.pop("SigningCertURL") is not None
+    event_dict["Records"][0]["body"] = json.dumps(payload)
+
+    # WHEN parsing the payload
+    # THEN raise a ValidationError error
+    with pytest.raises(ValidationError):
+        handle_sns_sqs_json_body(event_dict, LambdaContext())
+
+
+def test_handle_sns_sqs_trigger_event_json_body_missing_unsubscribe_url():
+    # GIVEN an event with a missing UnsubscribeURL
+    event_dict = load_event("snsSqsEvent.json")
+    payload = json.loads(event_dict["Records"][0]["body"])
+    assert payload.pop("UnsubscribeURL") is not None
+    event_dict["Records"][0]["body"] = json.dumps(payload)
+
+    # WHEN parsing the payload
+    # THEN raise a ValidationError error
+    with pytest.raises(ValidationError):
+        handle_sns_sqs_json_body(event_dict, LambdaContext())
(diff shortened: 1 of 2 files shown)


tests/functional/parser/test_sns.py
```diff
@@ -103,3 +104,29 @@ def handle_sns_sqs_json_body(event: List[MySnsBusiness], _: LambdaContext):
...
+    # GIVEN an event with a missing SigningCertURL
+    event_dict = load_event("snsSqsEvent.json")
+    payload = json.loads(event_dict["Records"][0]["body"])
+    assert payload.pop("SigningCertURL") is not None
+    event_dict["Records"][0]["body"] = json.dumps(payload)
+
+    # WHEN parsing the payload
+    # THEN raise a ValidationError error
+    with pytest.raises(ValidationError):
+        handle_sns_sqs_json_body(event_dict, LambdaContext())
```
heitorlessa:
```suggestion
def test_handle_sns_sqs_trigger_event_json_body_missing_signing_cert_url():
    # GIVEN an event is tampered with a missing SigningCertURL
    event_dict = load_event("snsSqsEvent.json")
    payload = json.loads(event_dict["Records"][0]["body"])
    payload.pop("SigningCertURL")
    event_dict["Records"][0]["body"] = json.dumps(payload)

    # WHEN parsing the payload
    # THEN raise a ValidationError error
    with pytest.raises(ValidationError):
        handle_sns_sqs_json_body(event_dict, LambdaContext())
```

Added an explicit description on GIVEN that this is tampered, not the behaviour we'd expect from SNS. Also removed the `assert` as it'll only test whether we're changing the underlying fake event.

tests/functional/parser/test_sns.py
```diff
@@ -103,3 +104,29 @@ def handle_sns_sqs_json_body(event: List[MySnsBusiness], _: LambdaContext):
...
+    # GIVEN an event with a missing UnsubscribeURL
+    event_dict = load_event("snsSqsEvent.json")
+    payload = json.loads(event_dict["Records"][0]["body"])
+    assert payload.pop("UnsubscribeURL") is not None
+    event_dict["Records"][0]["body"] = json.dumps(payload)
+
+    # WHEN parsing the payload
+    # THEN raise a ValidationError error
+    with pytest.raises(ValidationError):
+        handle_sns_sqs_json_body(event_dict, LambdaContext())
```
heitorlessa:
```suggestion
def test_handle_sns_sqs_trigger_event_json_body_missing_unsubscribe_url():
    # GIVEN an event is tampered with a missing UnsubscribeURL
    event_dict = load_event("snsSqsEvent.json")
    payload = json.loads(event_dict["Records"][0]["body"])
    payload.pop("UnsubscribeURL")
    event_dict["Records"][0]["body"] = json.dumps(payload)

    # WHEN parsing the payload
    # THEN raise a ValidationError error
    with pytest.raises(ValidationError):
        handle_sns_sqs_json_body(event_dict, LambdaContext())
```

Same as earlier

heitorlessa (changes requested):
Looks great, a minor suggestion to make the tests more explicit. Super excited to see your first PR ;-)

rubenfonseca:
@heitorlessa thank you for the feedback, it makes 100% sense! 😎

tests/functional/parser/test_sns.py
leandrodamascena: @heitorlessa this is an interesting approach. Technically I think that Assert makes sense but the reason you ask to remove it is to make the code cleaner and without deviation from what should really be tested. We are informing the payload statically (snsSqsEvent.json) and we are not previously changing it. So, in this case we are sure that the command (payload.pop) will work.

Good catch and new thing learned! Btw, I would have done exactly like you @rubenfonseca lol.

Congrats for your first PR @rubenfonseca :rocket: :+1:

heitorlessa (approved):

heitorlessa:
Thanks a lot @rubenfonseca ;)

PR #1316, opened 2022-07-19
fix(ci): lambda layer workflow release version and conditionals

am29d:
**Issue number:**

## Summary

In this PR we fix few issues with layers we have discovered in last release. 

First, we need to make sure to keep the old layer versions, something I have missed during the migration. 

Second, there is a case when the release upstream workflow can fail but it will still trigger the layer deployment because the condition is set to `complete`, and it can be `success` or `failure`. Thus, we need a job conditional to proceed only upon success OR a dispatch event.

### Changes

* set removal policy of the layer to `RETAIN` so we can keep the old layer versions
* add `if` block so we can trigger the workflow manually as well

### User experience

## Checklist

* [x] Changes are tested
* [ ] Changes are documented

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


.github/workflows/publish_layer.yml
```diff
@@ -16,9 +16,24 @@ on:
...
+  on-failure:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'failure' }}
+    steps:
+      - run: echo 'The triggering workflow failed, aborting'
+
+
   build-layer:
     runs-on: ubuntu-latest
+    needs: on-success
```
heitorlessa: why not a conditional `if` instead of dummy jobs, so that this job and any dependency are skipped in the event of failure? I might be missing something

For example:

```suggestion
  build-layer:
    runs-on: ubuntu-latest
    if: ${{ (github.event.workflow_run.conclusion == 'success') || (github.event_name == 'workflow_dispatch') }}
```

layer/layer/layer_stack.py
```diff
@@ -15,5 +15,6 @@ def __init__(
         )
 
         layer.add_permission("PublicLayerAccess", account_id="*")
+        layer.apply_removal_policy(RemovalPolicy.RETAIN)
```
heitorlessa, line 18: I'm so glad you caught this up early

heitorlessa (changes requested):
looks great! Suggested a job conditional instead of dummy jobs, but I might be missing something

.github/workflows/publish_layer.yml
am29d: Yep, you are right, much better solution 💯

heitorlessa (approved):

rubenfonseca (approved):

PR #1325, opened 2022-07-20
fix(event_handlers): handle lack of headers when using auto-compression feature

tbuatois:
**Issue number**: #1327 

## Summary

This PR add the ability use compress=True  in decorator without throwing an error for a request containing null headers

### Changes

* Add check for NoneType / null header in `get_header_value` function and return default if so
* Add corresponding unit test `test_compress_no_accept_encoding_null_headers`

### User experience

* Before
Error for a request with empty/null headers to a method with `compress=True`:
`'NoneType' object has no attribute 'items'", "errorType": "AttributeError"`  

That was due to an error from api_gateway.py at line 212:

```python
 if self.route.compress and "gzip" in (event.get_header_value("accept-encoding", "") or ""):
```

* After 

No more error, return default value as expected. 

## Checklist

* [ ] Changes are documented
* [x ] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -32,13 +32,12 @@ def get_header_value(
     headers: Dict[str, str], name: str, default_value: Optional[str], case_sensitive: Optional[bool]
 ) -> Optional[str]:
     """Get header value by name"""
-    if case_sensitive:
-        return headers.get(name, default_value)
-
     # If headers is NoneType, return default value
     if not headers:
         return default_value
 
+    if case_sensitive:
+        return headers.get(name, default_value)
     name_lower = name.lower()
 
     return next(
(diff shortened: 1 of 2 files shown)


aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -35,6 +35,10 @@ def get_header_value(
     if case_sensitive:
         return headers.get(name, default_value)
 
+    # If headers is NoneType, return default value
+    if not headers:
+        return default_value
+
```
heitorlessa:
```suggestion
    # If headers is NoneType, return default value
    if not headers:
        return default_value

    if case_sensitive:
        return headers.get(name, default_value)
```

Changing the order to prevent `case_sensitive=True` logic to kick in and fail on `AttributeError`, since `None` wouldn't have `.get`

heitorlessa (changes requested):
tiny change to address another potential error

heitorlessa (approved):

Issue #1038, opened 2021-10-05
Idempotency and Lambda timeouts


n2N8Z:
**Is your feature request related to a problem? Please describe.**
Currently if a lambda timesout the idempotency item is left in the 'INPROGRESS' status and subsequent lambda retries fail with IdempotencyAlreadyInProgressError.

**Describe the solution you'd like**
Add an additional field to the persistence record e.g. function_timeout and store in it the timestamp when the lambda will timeout: now().timestamp() + int(context.get_remaining_time_in_millis() / 1000)
Expand the DDB condition: OR function_timeout < :now

**Describe alternatives you've considered**
Create a signal handler in the lambda that raises an exception just before the lambda is about to timeout.

**Additional context**
N/A

heitorlessa:
@n2N8Z and everyone - We have recently found a better way to solve this using an internal no-op extension.

POC below - This allows us to receive a `SIGTERM` in a background thread to do any cleanup, like changing the status of the idempotency data record -- we let it takes its course if we don't receive a `SIGTERM` signal.

https://gist.github.com/heitorlessa/4aad06c39a1d520ff8c42adc72b0bcd5

heitorlessa:
Yes, it is the Lambda scheduler who orchestrates sandboxes that are recycled when a time out happens. A SIGTERM is only sent if an internal or external extension is registered -- the former is what the POC above does using a no-op internal extension -- The lambda scheduler guarantees 500ms for this logic to run.

SIGALRM, in your experience, is 
not guaranteed as you discovered. For cleanup activities whether a timeout happens or not (clean exit,
Idle for X time), an external Lambda extension is more appropriate since you not only have 2s to handle it but it doesn't interfere in the request path (e.g., early return is possible while cleanup happens in the background).

https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html

Hopefully that explains it.

@cakepietoast had thoughts about both solutions not being good enough

cakepietoast:
Fundamentally, I'm still of the mind that deleting INPROGRESS records violates idempotency. When this occurs and there are "stuck" INPROGRESS records, it is because there's no way for the utility to know whether the Lambda failed before, during, or after the unit of work its being asked to perform. As such, in the rare cases where this occurs, cases would need to be evaluated to understand the best approach. 

[...]

PR #1387, opened 2022-07-27
feat(idempotency): handle lambda timeout scenarios for INPROGRESS records

rubenfonseca:
**Issue number:** #1038

## Summary

Expire in-progress invocations caused by Lambda time outs. 

### Changes

During an invocation, we try to figure out how much time the Lambda invocation has before it expires, and save it on the idempotency record together will all the other fields.

When a new invocation with the same key arrives, we check against this timestamp to see if there's any `INPROGRESS` invocation past the timeout timestamp. If there is, we allow the invocation to be retried.

We also moved the squence diagrams of the idempotency module in documentation to [mermaid.js](https://mermaid-js.github.io/mermaid/#/).

#### Caveats

- Only works out of the box for the `@idempotent` decorator, because right now it's the only way to access the Lambda context to get the remaining time left. I've added a `register_lambda_context` to the `IdempotencyConfig` class as a workaround

carbon (2)

### User experience

Before: if a Lambda invocation times out, the invocation will get stuck until the idempotency record expires (default = 1 hour).

After: if a Lambda invocation times out, the invocation will be tried again on the next retry.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered docs/utilities/idempotency.md](https://github.com/rubenfonseca/aws-lambda-powertools-python/blob/feat/expire-inprogress/docs/utilities/idempotency.md)


aws_lambda_powertools/utilities/idempotency/config.py
```diff
@@ -41,3 +43,7 @@ def __init__(
         self.use_local_cache = use_local_cache
         self.local_cache_max_items = local_cache_max_items
         self.hash_function = hash_function
+        self.lambda_context: Optional[LambdaContext] = None
```
heitorlessa: missing doc string. 

Future note: We should automate this with a docstring linter https://github.com/terrencepreilly/darglint

tests/functional/idempotency/conftest.py
```diff
@@ -32,14 +31,17 @@ def lambda_apigw_event():
...
+            self.function_name = "test-func"
+            self.memory_limit_in_mb = 128
+            self.invoked_function_arn = "arn:aws:lambda:eu-west-1:809313241234:function:test-func"
+            self.aws_request_id = "52fdfc07-2182-154f-163f-5f0f9a621d72"
+
+        def get_remaining_time_in_millis(self) -> int:
+            return 1000
 
-    return namedtuple("LambdaContext", lambda_context.keys())(*lambda_context.values())
+    return LambdaContext()
```
heitorlessa, line 44: Nice, must needed cleanup!

heitorlessa (changes requested):
early comments before I resume reading it after lunch. At a quick glance, the pieces that would benefit the most now are: 

* **1/** Comments on the formula you're using since I saw Lambda context giving us milliseconds, then epoch (sensible for comparison), then type coercion to string (?) - sounds like we could reuse the milliseconds instead of wall clock time. Also, `datetime.datetime.now()` uses timezone offset too, instead of a safe non-backward timestamp like [time.monotonic](https://www.geeksforgeeks.org/python-time-monotonic-method/)
* **2/** I think the Update Expression would benefit to having the new one as a separate variable, then use f-string to combine them and make it more readable (which conditional is for which use case)

aws_lambda_powertools/utilities/idempotency/config.py
```diff
@@ -41,3 +46,7 @@ def __init__(
         self.use_local_cache = use_local_cache
         self.local_cache_max_items = local_cache_max_items
         self.hash_function = hash_function
+        self.lambda_context: Optional[LambdaContext] = lambda_context
+
+    def register_lambda_context(self, lambda_context: LambdaContext):
+        self.lambda_context = lambda_context
```
heitorlessa:
```suggestion
    def register_lambda_context(self, lambda_context: LambdaContext):
        """Include Lambda context to calculate remaining time before it times out"""
        self.lambda_context = lambda_context
```

Docstring for public method. There's gotta be a better wording but can't think of one right now

aws_lambda_powertools/utilities/idempotency/base.py
```diff
@@ -167,6 +190,13 @@ def _handle_for_status(self, data_record: DataRecord) -> Optional[Dict[Any, Any]
             raise IdempotencyInconsistentStateError("save_inprogress and get_record return inconsistent results.")
 
         if data_record.status == STATUS_CONSTANTS["INPROGRESS"]:
+            if data_record.in_progress_expiry_timestamp is not None and data_record.in_progress_expiry_timestamp < int(
+                datetime.datetime.now().timestamp()
```
heitorlessa: * 1000 for ms? tests for that too as it should've failed (as per your own word :D)

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -344,6 +350,14 @@ def save_inprogress(self, data: Dict[str, Any]) -> None:
             payload_hash=self._get_hashed_payload(data=data),
         )
 
+        if remaining_time_in_millis:
+            now = datetime.datetime.now()
+            period = datetime.timedelta(milliseconds=remaining_time_in_millis)
+
+            data_record.in_progress_expiry_timestamp = int((now + period).timestamp() * 1000)
```
heitorlessa: maybe split into a separate variable `(now + period).timestamp` so it's easier to read that you're using the adding the delta and multiplying by 1000 for ms

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -344,6 +350,14 @@ def save_inprogress(self, data: Dict[str, Any]) -> None:
...
+        else:
+            warnings.warn("Expires in progress is enabled but we couldn't determine the remaining time left")
```
heitorlessa: left over? Give an actionable feedback on what they can do when reading the warning (if they do)

docs/utilities/idempotency.md
```diff
@@ -22,6 +22,7 @@ times with the same parameters**. This makes idempotent operations safe to retry
 * Ensure Lambda handler returns the same result when called with the same payload
 * Select a subset of the event as the idempotency key using JMESPath expressions
 * Set a time window in which records with the same payload should be considered duplicates
+* Expires in-progress executions when the Lambda handler times out
```
heitorlessa:
```suggestion
* Expires in-progress executions if the Lambda function times out halfway through
```

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -153,34 +158,51 @@ def _put_record(self, data_record: DataRecord) -> None:
...
         except self.table.meta.client.exceptions.ConditionalCheckFailedException:
             logger.debug(f"Failed to put record for already existing idempotency key: {data_record.idempotency_key}")
             raise IdempotencyItemAlreadyExistsError
 
     def _update_record(self, data_record: DataRecord):
         logger.debug(f"Updating record for idempotency key: {data_record.idempotency_key}")
-        update_expression = "SET #response_data = :response_data, #expiry = :expiry, #status = :status"
+        update_expression = (
+            "SET #response_data = :response_data, #expiry = :expiry, "
+            "#status = :status, #in_progress_expiry = :in_progress_expiry"
```
jeromevdl: The record (from save_success) doesn't contain the in_progress_expiry. This won't have any effect... Also do we actually need to add this field when the status is COMPLETED ?

rubenfonseca: Thanks! Great comment, we really don't need it in this case, so I've removed it.

heitorlessa:
I've pushed an updated doc with the following changes:

* Reordered the wording so we give the positive news first (what exactly we do we do?)
* Give comfort to those using `@idempotent` (When do they need to care?)
* Used banners to emphasize when they need to take action (what should I do?)
* Moved sequence diagram for latter so they're informed first, then they can dig deeper into the mechanics


image

heitorlessa:
Shortening the wording on sequence diagram to fit in the final SVG

New (pushing shortly)

image

heitorlessa (approved):

Issue #1435, opened 2022-08-11
Maintenance: Refactor E2E mechanism for extensibility, ease test writing, and unblock integ tests


heitorlessa:
### Summary

We defined the E2E mechanism on #1226 and recently implemented a POC. This issue tracks the extensibility work required to allow E2E to own infrastructure. This includes making it easier for tests to be able to control the payload each E2E test uses to invoke their respective Lambda function.

Tasks

* [x] #1436
* [x] #1437
* [x] #1438
* [x] #1439
* [x] #1440
* [x] #1441
* [x] #1464

### Why is this needed?

When trying to increase E2E coverage, we identified a gap in the mechanism as it doesn't allow each feature group (_e.g., Metrics_) to customize their own infrastructure. 

This made it difficult to add E2E tests for Idempotency where we wanted to **(1)** create a DynamoDB table and reference it in Lambda functions env var, and **(2)** send different payloads to test idempotency results.

Once this is complete, we can resume increasing coverage for other features, including defining test strategies (when to use which) and resume integration tests.

### Which area does this relate to?

Tests

### Solution

This is how it looks like on my fork using the new refactoring (one or more PRs depending on final size).

## Metrics infrastructure

`tests/e2e/metrics/infrastructure.py` fully controls what infrastructure Metrics should have. It keeps the original mechanism of creating Lambda functions within the handlers directory but now more explicit. It also allows to override any CDK Lambda function prop.

```python
from pathlib import Path

from tests.e2e.utils.infrastructure import BaseInfrastructureV2


class MetricsStack(BaseInfrastructureV2):
    def __init__(self, handlers_dir: Path, feature_name: str = "metrics") -> None:
        super().__init__(feature_name, handlers_dir)

    def create_resources(self):
        self.create_lambda_functions()

```

## Metrics infrastructure parallelization

`tests/e2e/metrics/conftest.py` handles infrastructure deployment so tests can run in parallel after deployment. It now handles stack deletion in case of failures too.

```python
import json
from pathlib import Path
from typing import Dict

import pytest
from _pytest import fixtures
from filelock import FileLock

from tests.e2e.metrics.infrastructure import MetricsStack


@pytest.fixture(autouse=True, scope="module")
def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
    """Setup and teardown logic for E2E test infrastructure

    Parameters
    ----------
[...]

PR #1444, opened 2022-08-11
chore(tests): refactor E2E test mechanics to ease maintenance, writing tests and parallelization

heitorlessa:
**Issue number:** #1435

## Summary

This PR refactors the E2E test mechanism to make tests more explicit, easier to reason and write, and improves maintenance aspects of the CDK internal logic.

I’ll rename the `V2` and delete the older one after all tests are rewritten (future PRs)

### Changes

See new comments to prevent PR body become too big.

**Remaining tasks**

- [x] Update cold start test: Ensure cold start function is executed twice and a single metric is created
- [ ] Encapsulate parallelization logic

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
## Encapsulates CDK Assets logic

I had challenges understanding our CDK drop-in replacement to CDK largely due to not understanding CDK Assets structure. For that reason, I've created:

* **Asset models**: `tests/e2e/utils/models.py` contains Pydantic models with the internal structure of an Asset. This makes it easier to understand what's inside and also prevent common mistakes of accessing deeply nested dictionary keys manually
* **Encapsulate assets logic**: `tests/e2e/utils/asset.py` contains all drop-in replacement logic under `Assets` and `Asset`. I simplified the code to foster readability (`Path` over `os`) and composition for maintenance.

### Result

#### BEFORE

```python
    def deploy(self, Stack: Type[BaseInfrastructureStack]) -> Dict[str, str]:

        stack = Stack(handlers_dir=self.handlers_dir, stack_name=self.stack_name, config=self.config)
        template, asset_root_dir, asset_manifest_file = stack()
        self._upload_assets(asset_root_dir, asset_manifest_file)

        response = self._deploy_stack(self.stack_name, template)

        return self._transform_output(response["Stacks"][0]["Outputs"])
```


#### AFTER

```python
    def deploy(self) -> Dict[str, str]:
        """Creates CloudFormation Stack and return stack outputs as dict

        Returns
        -------
        Dict[str, str]
            CloudFormation Stack Outputs with output key and value
        """
        template, asset_manifest_file = self._synthesize()
        assets = Assets(cfn_template=asset_manifest_file, account_id=self.account_id, region=self.region)
        assets.upload()
        return self._deploy_stack(self.stack_name, template)
```

heitorlessa:
## Give infrastructure control to each feature group

This allows each feature group to control their infrastructure within `tests/e2e/<feature-group>/infrastructure.py`. For this PR, I've changed `metrics` so it's now independent from the stack resources that Tracer and Logger would create.

### Result

#### BEFORE

> tests/e2e/conftest.py

```python
@dataclass
class InfrastructureOutput:
    arns: Dict[str, str]
    execution_time: datetime.datetime

    def get_lambda_arns(self) -> Dict[str, str]:
        return self.arns

    def get_lambda_function_arn(self, cf_output_name: str) -> Optional[str]:
        return self.arns.get(cf_output_name)

    def get_lambda_function_name(self, cf_output_name: str) -> Optional[str]:
        lambda_arn = self.get_lambda_function_arn(cf_output_name=cf_output_name)
        return lambda_arn.split(":")[-1] if lambda_arn else None

    def get_lambda_execution_time(self) -> datetime.datetime:
        return self.execution_time

    def get_lambda_execution_time_timestamp(self) -> int:
        return int(self.execution_time.timestamp() * 1000)


@pytest.fixture(scope="module")
def create_infrastructure(config, request) -> Generator[Dict[str, str], None, None]:
    stack_name = f"test-lambda-{uuid.uuid4()}"
    test_dir = request.fspath.dirname
    handlers_dir = f"{test_dir}/handlers/"

    infra = infrastructure.Infrastructure(stack_name=stack_name, handlers_dir=handlers_dir, config=config)
    yield infra.deploy(Stack=infrastructure.InfrastructureStack)
    infra.delete()
```


#### AFTER

> tests/e2e/metrics/infrastructure.py

```python
from pathlib import Path

from tests.e2e.utils.infrastructure import BaseInfrastructureV2


class MetricsStack(BaseInfrastructureV2):
    def __init__(self, handlers_dir: Path, feature_name: str = "metrics") -> None:
        super().__init__(feature_name, handlers_dir)

    def create_resources(self):
        self.create_lambda_functions()
```

This is now deployed within `conftest.py` for each feature. This fixture `infrastructure` now conveys a more explicit message that it contains the output of that infrastructure to be used within tests.

> tests/e2e/metrics/conftest.py

```python
import json
from pathlib import Path
from typing import Dict

import pytest
from _pytest import fixtures
from filelock import FileLock

from tests.e2e.metrics.infrastructure import MetricsStack


@pytest.fixture(autouse=True, scope="module")
def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
    """Setup and teardown logic for E2E test infrastructure

    Parameters
    ----------
    request : fixtures.SubRequest
        test fixture containing metadata about test execution

    Returns
    -------
    MetricsStack
        Metrics Stack to deploy infrastructure

    Yields
    ------
    Iterator[MetricsStack]
        Deployed Infrastructure
    """
    stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
    try:
        yield stack.deploy()
    finally:
        stack.delete()
```

heitorlessa:
## Make tests more explicit and easier to write

This change gives control back to each test definition to decide how they want to invoke their functions. It also makes it more explicit what information needs to be created as part of the test - explicit functions over fixtures to make it easier to trace the logic. 

To confirm this was easier, I've also increased coverage to ensure Lambda cold starts are captured.

### Result


#### BEFORE

```python
import datetime
import uuid

import boto3
import pytest
from e2e import conftest
from e2e.utils import helpers


@pytest.fixture(scope="module")
def config() -> conftest.LambdaConfig:
    return {
        "parameters": {},
        "environment_variables": {
            "POWERTOOLS_METRICS_NAMESPACE": "powertools-e2e-metric",
            "POWERTOOLS_SERVICE_NAME": "test-powertools-service",
            "METRIC_NAME": f"business-metric-{str(uuid.uuid4()).replace('-','_')}",
        },
    }


def test_basic_lambda_metric_visible(execute_lambda: conftest.InfrastructureOutput, config: conftest.LambdaConfig):
    # GIVEN
    start_date = execute_lambda.get_lambda_execution_time()
    end_date = start_date + datetime.timedelta(minutes=5)

    # WHEN
    metrics = helpers.get_metrics(
        start_date=start_date,
        end_date=end_date,
        namespace=config["environment_variables"]["POWERTOOLS_METRICS_NAMESPACE"],
        metric_name=config["environment_variables"]["METRIC_NAME"],
        service_name=config["environment_variables"]["POWERTOOLS_SERVICE_NAME"],
        cw_client=boto3.client(service_name="cloudwatch"),
    )

    # THEN
    assert metrics.get("Timestamps") and len(metrics.get("Timestamps")) == 1
    assert metrics.get("Values") and len(metrics.get("Values")) == 1
    assert metrics.get("Values") and metrics.get("Values")[0] == 1
```


#### AFTER

```python
import json

import pytest

from tests.e2e.utils import helpers


@pytest.fixture
def basic_handler_fn(infrastructure: dict) -> str:
    return infrastructure.get("BasicHandler", "")


@pytest.fixture
def basic_handler_fn_arn(infrastructure: dict) -> str:
    return infrastructure.get("BasicHandlerArn", "")


@pytest.fixture
def cold_start_fn(infrastructure: dict) -> str:
    return infrastructure.get("ColdStart", "")


@pytest.fixture
def cold_start_fn_arn(infrastructure: dict) -> str:
    return infrastructure.get("ColdStartArn", "")


METRIC_NAMESPACE = "powertools-e2e-metric"


def test_basic_lambda_metric_is_visible(basic_handler_fn: str, basic_handler_fn_arn: str):
    # GIVEN
    metric_name = helpers.build_metric_name()
    service = helpers.build_service_name()
    dimensions = helpers.build_add_dimensions_input(service=service)
    metrics = helpers.build_multiple_add_metric_input(metric_name=metric_name, value=1, quantity=3)

    # WHEN
    event = json.dumps({"metrics": metrics, "service": service, "namespace": METRIC_NAMESPACE})
    _, execution_time = helpers.trigger_lambda(lambda_arn=basic_handler_fn_arn, payload=event)

    metrics = helpers.get_metrics(
        namespace=METRIC_NAMESPACE, start_date=execution_time, metric_name=metric_name, dimensions=dimensions
    )

    # THEN
    metric_data = metrics.get("Values", [])
    assert metric_data and metric_data[0] == 3.0


def test_cold_start_metric(cold_start_fn_arn: str, cold_start_fn: str):
    # GIVEN
    metric_name = "ColdStart"
    service = helpers.build_service_name()
    dimensions = helpers.build_add_dimensions_input(function_name=cold_start_fn, service=service)

    # WHEN
    event = json.dumps({"service": service, "namespace": METRIC_NAMESPACE})
    _, execution_time = helpers.trigger_lambda(lambda_arn=cold_start_fn_arn, payload=event)

    metrics = helpers.get_metrics(
        namespace=METRIC_NAMESPACE, start_date=execution_time, metric_name=metric_name, dimensions=dimensions
    )

    # THEN
    metric_data = metrics.get("Values", [])
    assert metric_data and metric_data[0] == 1.0

```

heitorlessa:
## Parallelize tests

As of today, we parallelize CFN Stack creation but tests are run sequentially. Now that each feature group can control their infrastructure, this change parallelize tests too in a safely manner. It also handles a leftover to delete stacks in the case of failure (e.g, test runner, exceptions, etc.). 

This also takes into account Step Through debugging in any IDE allowing to run one test at a time or in parallel.

**Better yet**, our tests were taking 175s before and now it takes ~94s even with an additional Lambda function.

### Result

#### BEFORE

```python
import json
from pathlib import Path
from typing import Dict

import pytest
from _pytest import fixtures
from filelock import FileLock

from tests.e2e.metrics.infrastructure import MetricsStack


@pytest.fixture(autouse=True, scope="module")
def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
    """Setup and teardown logic for E2E test infrastructure

    Parameters
    ----------
    request : fixtures.SubRequest
        test fixture containing metadata about test execution

    Returns
    -------
    MetricsStack
        Metrics Stack to deploy infrastructure

    Yields
    ------
    Iterator[MetricsStack]
        Deployed Infrastructure
    """
    stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
    try:
        yield stack.deploy()
    finally:
        stack.delete()
```


#### AFTER


```python
import json
from pathlib import Path
from typing import Dict

import pytest
from _pytest import fixtures
from filelock import FileLock

from tests.e2e.metrics.infrastructure import MetricsStack


@pytest.fixture(autouse=True, scope="module")
def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
    """Setup and teardown logic for E2E test infrastructure

    Parameters
    ----------
    request : fixtures.SubRequest
        test fixture containing metadata about test execution

    Returns
    -------
    MetricsStack
        Metrics Stack to deploy infrastructure

    Yields
    ------
    Iterator[MetricsStack]
        Deployed Infrastructure
    """
    stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
    try:
        if worker_id == "master":
            # no parallelization, deploy stack and let fixture be cached
            yield stack.deploy()
        else:
            # tmp dir shared by all workers
            root_tmp_dir = tmp_path_factory.getbasetemp().parent

            cache = root_tmp_dir / "cache.json"
            with FileLock(f"{cache}.lock"):
                # If cache exists, return stack outputs back
                # otherwise it's the first run by the main worker
                # deploy and return stack outputs so subsequent workers can reuse
                if cache.is_file():
                    stack_outputs = json.loads(cache.read_text())
                else:
                    stack_outputs: Dict = stack.deploy()
                    cache.write_text(json.dumps(stack_outputs))
            yield stack_outputs
    finally:
        stack.delete()
```

tests/e2e/metrics/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+
+import pytest
+from _pytest import fixtures
+from filelock import FileLock
+
+from tests.e2e.metrics.infrastructure import MetricsStack
+
+
+@pytest.fixture(autouse=True, scope="module")
+def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
```
mploski: should worker_id has a type?

tests/e2e/metrics/infrastructure.py
```diff
@@ -0,0 +1,11 @@
+from pathlib import Path
+
+from tests.e2e.utils.infrastructure import BaseInfrastructureV2
+
+
+class MetricsStack(BaseInfrastructureV2):
```
mploski, line 6: My first idea was to use strategy pattern and simply inject Resources class as dependency injection in every test group and only use one core infrastructure class without need to subclassing. Your approach is more clean and easier to comprehend though!

tests/e2e/metrics/test_metrics.py
```diff
@@ -1,40 +1,69 @@
...
-    return {
-        "parameters": {},
-        "environment_variables": {
-            "POWERTOOLS_METRICS_NAMESPACE": "powertools-e2e-metric",
-            "POWERTOOLS_SERVICE_NAME": "test-powertools-service",
-            "METRIC_NAME": f"business-metric-{str(uuid.uuid4()).replace('-','_')}",
-        },
-    }
 
+@pytest.fixture
```
mploski, line 8: Shouldn’t this fixtures go to shared pytest config file? I can easily imagine we gonna reuse them in every test group we have so far (metric, tracing, logging)

tests/e2e/metrics/test_metrics.py
```diff
@@ -1,40 +1,69 @@
...
+    metric_name = helpers.build_metric_name()
+    service = helpers.build_service_name()
+    dimensions = helpers.build_add_dimensions_input(service=service)
+    metrics = helpers.build_multiple_add_metric_input(metric_name=metric_name, value=1, quantity=3)
 
     # WHEN
+    event = json.dumps({"metrics": metrics, "service": service, "namespace": METRIC_NAMESPACE})
+    _, execution_time = helpers.trigger_lambda(lambda_arn=basic_handler_fn_arn, payload=event)
+
+    metrics = helpers.get_metrics(
```
mploski, line 42: I like the idea of having fully independent tests that calls lambda with dynamic input. Not using cache simplifies things, but also I see a risk here. If we imagine we have 20-30 or even more tests in total this will lead to having the same number of parallelized processes calling lambdas and waiting in worst case scenario 1,2 minutes on a data being present in AWS Service ( and retrying for a result every few seconds). It may cause resource saturation issues, especially on limited CI environment + we may hit API throttling issue for CW in get_metric or other methods that fetch the result . Using cache, gave us the same performance but without hard parallelization needed. Do we have any data validating/invalidating my theory?

tests/e2e/utils/helpers.py
```diff
@@ -55,28 +63,61 @@ def get_logs(lambda_function_name: str, log_client: CloudWatchClient, start_time
     return filtered_logs
 
 
-@lru_cache(maxsize=10, typed=False)
-@retry(ValueError, delay=1, jitter=1, tries=20)
+@retry(ValueError, delay=1, jitter=1, tries=10)
```
mploski: In some rare occasions the 10 tries was not enough - thats why I raised this number to 20

tests/e2e/utils/infrastructure.py
```diff
@@ -210,3 +214,173 @@ def _find_assets(self, asset_template: str, account_id: str, region: str):
...
+    def deploy(self) -> Dict[str, str]:
+        """Creates CloudFormation Stack and return stack outputs as dict
+
+        Returns
+        -------
+        Dict[str, str]
+            CloudFormation Stack Outputs with output key and value
+        """
+        template, asset_manifest_file = self._synthesize()
+        assets = Assets(cfn_template=asset_manifest_file, account_id=self.account_id, region=self.region)
```
mploski: I think that asset_manifest_file is not a cfn template. It is separate entity. I would suggest renaming this variable to `asset_manifest_file` or something similar

mploski:
> ## Parallelize tests
> 
> As of today, we parallelize CFN Stack creation but tests are run sequentially. Now that each feature group can control their infrastructure, this change parallelize tests too in a safely manner. It also handles a leftover to delete stacks in the case of failure (e.g, test runner, exceptions, etc.).
> 
> This also takes into account Step Through debugging in any IDE allowing to run one test at a time or in parallel.
> 
> **Better yet**, our tests were taking 175s before and now it takes ~94s even with an additional Lambda function.
> ### Result
> #### BEFORE
> 
> ```python
> import json
> from pathlib import Path
> from typing import Dict
> 
> import pytest
> from _pytest import fixtures
> from filelock import FileLock
> 
> from tests.e2e.metrics.infrastructure import MetricsStack
> 
> 
> @pytest.fixture(autouse=True, scope="module")
> def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
>     """Setup and teardown logic for E2E test infrastructure
> 
>     Parameters
>     ----------
>     request : fixtures.SubRequest
>         test fixture containing metadata about test execution
> 
>     Returns
>     -------
>     MetricsStack
>         Metrics Stack to deploy infrastructure
> 
>     Yields
>     ------
>     Iterator[MetricsStack]
>         Deployed Infrastructure
>     """
>     stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
>     try:
>         yield stack.deploy()
>     finally:
>         stack.delete()
> ```
> 
> #### AFTER
> 
> ```python
> import json
> from pathlib import Path
> from typing import Dict
> 
> import pytest
> from _pytest import fixtures
> from filelock import FileLock
> 
> from tests.e2e.metrics.infrastructure import MetricsStack
> 
> 
> @pytest.fixture(autouse=True, scope="module")
> def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
>     """Setup and teardown logic for E2E test infrastructure
> 
>     Parameters
>     ----------
>     request : fixtures.SubRequest
>         test fixture containing metadata about test execution
> 
>     Returns
>     -------
>     MetricsStack
>         Metrics Stack to deploy infrastructure
> 
>     Yields
>     ------
>     Iterator[MetricsStack]
>         Deployed Infrastructure
>     """
>     stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
>     try:
>         if worker_id == "master":
>             # no parallelization, deploy stack and let fixture be cached
>             yield stack.deploy()
>         else:
>             # tmp dir shared by all workers
>             root_tmp_dir = tmp_path_factory.getbasetemp().parent
> 
>             cache = root_tmp_dir / "cache.json"
>             with FileLock(f"{cache}.lock"):
>                 # If cache exists, return stack outputs back
>                 # otherwise it's the first run by the main worker
>                 # deploy and return stack outputs so subsequent workers can reuse
>                 if cache.is_file():
>                     stack_outputs = json.loads(cache.read_text())
>                 else:
>                     stack_outputs: Dict = stack.deploy()
>                     cache.write_text(json.dumps(stack_outputs))
>             yield stack_outputs
>     finally:
>         stack.delete()
> ```

One remark here :-) In E2E  tests version 1, tests took 175 seconds because we are building powertools layer at the test beginning. This step is heavy and takes visible portion of overall test latency. In this PR you use already deployed layer so those numbers are not comparable. My assumption is that overall numbers will be similar at the end as you compensated lack of lambda result caching with parallelization

tests/e2e/utils/helpers.py
heitorlessa: good point, I decreased because it took forever to fail. Let's meet in the middle by increasing the jitter to 1.5 and `delay` to 2 - metrics will never be available in 1s, and half a second more for jitter gives a chance for concurrent processes to adapt

mploski (approved):

tests/e2e/metrics/test_metrics.py
heitorlessa: Nope, because we learned the hard way how [this grows](https://github.com/heitorlessa/aws-lambda-powertools-python/blob/chore/refactor-e2e/tests/functional/idempotency/test_idempotency.py#L47) and Pytest fixtures dependency injection makes it hard to trace where a fixture comes from.

This is a trade-off where we're keeping the absolutely minimum number of fixtures as close to the tests as possible (I can CTRL+F).

tests/e2e/metrics/test_metrics.py
heitorlessa: I totally see your concern and all great points - allow me to expand on why our risk is minimized as we spoke earlier. 

Caching wouldn't help as the payload is now dynamic to ease writing handlers and more explicit tests, it could cause the opposite problem now where we're filling out memory (not really a problem, just a data point) and no cache hit.

For the resource exhaustion, we have a **controllable demand** where we own when those tests will run (on merge/push and on label add). This reduces the risk. We also have jitter and delays that can adapt retries based on their APIs they might be exhausting - this means we would even fail the test before we are able to succeed OR we succeed eventually within our max retry. 

For customers' laptop resource exhaustion, it is also minimized because **(1)** we control the recommendation on how they should run (one vs one module vs the entire test suite), **(2)** E2E will cost customers money thus giving them an incentive to run at our CI instead, and **(3)** E2E tests will run on a single runtime compared to CI (3.7, 3.8, 3.9 etc.).

tests/e2e/utils/infrastructure.py
heitorlessa: Great catch!! Renaming to `asset_manifest`.

heitorlessa:
> > ## Parallelize tests
> > As of today, we parallelize CFN Stack creation but tests are run sequentially. Now that each feature group can control their infrastructure, this change parallelize tests too in a safely manner. It also handles a leftover to delete stacks in the case of failure (e.g, test runner, exceptions, etc.).
> > This also takes into account Step Through debugging in any IDE allowing to run one test at a time or in parallel.
> > **Better yet**, our tests were taking 175s before and now it takes ~94s even with an additional Lambda function.
> > ### Result
> > #### BEFORE
> > ```python
> > import json
> > from pathlib import Path
> > from typing import Dict
> > 
> > import pytest
> > from _pytest import fixtures
> > from filelock import FileLock
> > 
> > from tests.e2e.metrics.infrastructure import MetricsStack
> > 
> > 
> > @pytest.fixture(autouse=True, scope="module")
> > def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
> >     """Setup and teardown logic for E2E test infrastructure
> > 
> >     Parameters
> >     ----------
> >     request : fixtures.SubRequest
> >         test fixture containing metadata about test execution
> > 
> >     Returns
> >     -------
> >     MetricsStack
> >         Metrics Stack to deploy infrastructure
> > 
> >     Yields
> >     ------
> >     Iterator[MetricsStack]
> >         Deployed Infrastructure
> >     """
> >     stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
> >     try:
> >         yield stack.deploy()
> >     finally:
> >         stack.delete()
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > #### AFTER
> > ```python
> > import json
> > from pathlib import Path
> > from typing import Dict
> > 
> > import pytest
> > from _pytest import fixtures
> > from filelock import FileLock
> > 
> > from tests.e2e.metrics.infrastructure import MetricsStack
> > 
> > 
> > @pytest.fixture(autouse=True, scope="module")
> > def infrastructure(request: fixtures.SubRequest, tmp_path_factory: pytest.TempPathFactory, worker_id) -> MetricsStack:
> >     """Setup and teardown logic for E2E test infrastructure
> > 
> >     Parameters
> >     ----------
> >     request : fixtures.SubRequest
> >         test fixture containing metadata about test execution
> > 
> >     Returns
> >     -------
> >     MetricsStack
> >         Metrics Stack to deploy infrastructure
> > 
> >     Yields
> >     ------
> >     Iterator[MetricsStack]
> >         Deployed Infrastructure
> >     """
> >     stack = MetricsStack(handlers_dir=Path(f"{request.fspath.dirname}/handlers"))
> >     try:
> >         if worker_id == "master":
> >             # no parallelization, deploy stack and let fixture be cached
> >             yield stack.deploy()
> >         else:
> >             # tmp dir shared by all workers
> >             root_tmp_dir = tmp_path_factory.getbasetemp().parent
> > 
> >             cache = root_tmp_dir / "cache.json"
> >             with FileLock(f"{cache}.lock"):
> >                 # If cache exists, return stack outputs back
> >                 # otherwise it's the first run by the main worker
> >                 # deploy and return stack outputs so subsequent workers can reuse
> >                 if cache.is_file():
> >                     stack_outputs = json.loads(cache.read_text())
> >                 else:
> >                     stack_outputs: Dict = stack.deploy()
> >                     cache.write_text(json.dumps(stack_outputs))
> >             yield stack_outputs
> >     finally:
> >         stack.delete()
> > ```
> 
> One remark here :-) In E2E tests version 1, tests took 175 seconds because we are building powertools layer at the test beginning. This step is heavy and takes visible portion of overall test latency. In this PR you use already deployed layer so those numbers are not comparable. My assumption is that overall numbers will be similar at the end as you compensated lack of lambda result caching with parallelization

So true, I missed that. Fat Layer eh? :D

tests/e2e/metrics/conftest.py
heitorlessa: fixed!

mploski (approved):

PR #1457, opened 2022-08-16
chore(tests): refactor E2E tracer to ease maintenance, writing tests and parallelization

heitorlessa:
**Issue number:** #1440

## Summary

This PR refactors Tracer E2E test following the new Infrastructure and explicit tests. Unrelated to Tracer, it also breaks up the various helpers into their own respective modules (`data_fetcher/traces.py` and `data_builder/metrics.py`).

**Tasks**

* [x] Refactor tracer fetching recursive logic into `TraceFetcher`
* [x] Update `get_traces` factory to return `TraceFetcher`
* [x] Self-document trace segments, trace documents, etc. to ease maintenance
* [x] Create `TracerStack` and use new infrastructure
* [x] Pass `POWERTOOLS_SERVICE_NAME` as Env Var; `Tracer` [needs changing to allow dynamic service injection](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/tracing/tracer.py#L310)
* [x] Test capture lambda handler and cold start annotation
* [x] Test methods (sync, async)
* [x] Break data builders into their own module
* [x] Ensure minimum number of traces are found before returning to address flaky tests 

### Changes

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
**Test differences**

Besides data builders, I've encapsulated trace fetching and manipulation of different views in its own class. The logic is largely the same, what changed is that we have more flexibility in the type of trace we want (origin, segment name, etc.) and have getters to ease assertion.

## Before

```python
def test_basic_lambda_async_trace_visible(execute_lambda: conftest.InfrastructureOutput, config: conftest.LambdaConfig):
    # GIVEN
    lambda_name = execute_lambda.get_lambda_function_name(cf_output_name="basichandlerarn")
    start_date = execute_lambda.get_lambda_execution_time()
    end_date = start_date + datetime.timedelta(minutes=5)
    trace_filter_exporession = f'service("{lambda_name}")'

    # WHEN
    trace = helpers.get_traces(
        start_date=start_date,
        end_date=end_date,
        filter_expression=trace_filter_exporession,
        xray_client=boto3.client("xray"),
    )

    # THEN
    info = helpers.find_trace_additional_info(trace=trace)
    print(info)
    handler_trace_segment = [trace_segment for trace_segment in info if trace_segment.name == "## lambda_handler"][0]
    collect_payment_trace_segment = [
        trace_segment for trace_segment in info if trace_segment.name == "## collect_payment"
    ][0]

    annotation_key = config["environment_variables"]["ANNOTATION_KEY"]
    expected_value = config["environment_variables"]["ANNOTATION_VALUE"]
    expected_async_value = config["environment_variables"]["ANNOTATION_ASYNC_VALUE"]

    assert handler_trace_segment.annotations["Service"] == "e2e-tests-app"
    assert handler_trace_segment.metadata["e2e-tests-app"][annotation_key] == expected_value
    assert collect_payment_trace_segment.metadata["e2e-tests-app"][annotation_key] == expected_async_value
```

## After

```python
def test_lambda_handler_trace_is_visible(basic_handler_fn_arn: str, basic_handler_fn: str):
    # GIVEN
    handler_name = basic_handler.lambda_handler.__name__
    handler_subsegment = f"## {handler_name}"
    handler_metadata_key = f"{handler_name} response"
    trace_query = data_builder.build_trace_default_query(function_name=basic_handler_fn)

    # WHEN
    _, execution_time = helpers.trigger_lambda(lambda_arn=basic_handler_fn_arn)
    helpers.trigger_lambda(lambda_arn=basic_handler_fn_arn)

    # THEN
    trace = data_fetcher.get_traces(start_date=execution_time, filter_expression=trace_query, minimum_traces=2)

    assert len(trace.get_annotation(key="ColdStart", value=True)) == 1
    assert len(trace.get_metadata(key=handler_metadata_key, namespace=TracerStack.SERVICE_NAME)) == 2
    assert len(trace.get_subsegment(name=handler_subsegment)) == 2
```

leandrodamascena:
Huge improvement in readability and code quality @heitorlessa. Too easy to understand now :guitar:

heitorlessa:
@mploski @rubenfonseca could either of you review before I start Logger and complete the refactoring?

Thanks a lot!

tests/e2e/utils/data_fetcher/logs.py
```diff
@@ -0,0 +1,39 @@
...
+    cold_start: Optional[bool]
+    function_name: Optional[str]
+    function_memory_size: Optional[str]
+    function_arn: Optional[str]
+    function_request_id: Optional[str]
+    xray_trace_id: Optional[str]
+    extra_info: Optional[str]
+
+
+@lru_cache(maxsize=10, typed=False)
```
rubenfonseca, line 25: Should we remove this now?

tests/e2e/utils/data_fetcher/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+        End window to fetch metrics, by default start_date + 2 minutes window
+    period : int, optional
+        Time period to fetch metrics for, by default 60
+    stat : str, optional
+        Aggregation function to use when fetching metrics, by default "Sum"
+
+    Returns
+    -------
+    MetricDataResultTypeDef
+        _description_
```
rubenfonseca: Sorry in advance if I missed something, but is something missing here?

rubenfonseca (changes requested):
image

Just two tiny comments :)

tests/e2e/conftest.py
```diff
@@ -14,7 +16,7 @@
 from typing import Dict, Generator, Optional
 
 import pytest
-from e2e.utils import helpers, infrastructure
+from e2e.utils import infrastructure
```
mploski: should we merge this  line with line 9?

tests/e2e/utils/data_fetcher/__init__.py
```diff
@@ -0,0 +1,6 @@
+from tests.e2e.utils.data_fetcher.common import get_lambda_response
+from tests.e2e.utils.data_fetcher.logs import get_logs
+from tests.e2e.utils.data_fetcher.metrics import get_metrics
+from tests.e2e.utils.data_fetcher.traces import get_traces
+
+__all__ = ["get_traces", "get_metrics", "get_logs", "get_lambda_response"]
```
mploski: since you do an explicit method import in lines 1-4 do we need __all__ list?

tests/e2e/tracer/conftest.py
```diff
@@ -0,0 +1,21 @@
...
+
+from tests.e2e.tracer.infrastructure import TracerStack
+from tests.e2e.utils.infrastructure import deploy_once
+
+
+@pytest.fixture(autouse=True, scope="module")
+def infrastructure(request: pytest.FixtureRequest, tmp_path_factory: pytest.TempPathFactory, worker_id: str):
+    """Setup and teardown logic for E2E test infrastructure
+
+    Parameters
```
mploski, line 11: Should we update parameters docstring to reflect input?

tests/e2e/tracer/handlers/sync_async_capture.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+@tracer.capture_method
+def get_todos():
+    return [{"id": f"{uuid4()}", "completed": False} for _ in range(5)]
+
+
+@tracer.capture_method
+async def async_get_users():
+    await asyncio.sleep(1)
```
mploski: why do we need sleep here?

heitorlessa: I'll add a comment. It's to simulate a coroutine to block and be successfully traced. I had false positives in the past where the async code ran immediately when in fact it only traced a coroutine to be primed, not actually executed.

tests/e2e/tracer/infrastructure.py
```diff
@@ -0,0 +1,17 @@
...
+
+class TracerStack(BaseInfrastructureV2):
+    # Maintenance: Tracer doesn't support dynamic service injection (tracer.py L310)
+    # we could move after handler response or adopt env vars usage in e2e tests
+    SERVICE_NAME: str = build_service_name()
+
+    def __init__(self, handlers_dir: Path, feature_name: str = "tracer") -> None:
+        super().__init__(feature_name, handlers_dir)
+
+    def create_resources(self):
```
mploski: Should we add  -> None to notify reader that this function doesnt return anything?

tests/e2e/utils/data_fetcher/__init__.py
heitorlessa: I know I don't like either. Since you too are bothered, lemme play with flake8 to ignore unused imports in `__init__.py` files.

tests/e2e/utils/asset.py
```diff
@@ -2,18 +2,46 @@
...
+
+class AssetTemplateConfig(BaseModel):
+    source: AssetTemplateConfigSource
+    destinations: AssetTemplateConfigDestinations
+
+
+class TemplateAssembly(BaseModel):
+    version: str
+    files: Dict[str, AssetTemplateConfig]
+    docker_images: Dict = Field(Dict, alias="dockerImages")
```
mploski: I think we can remove this attribute as we don't need it in our implementation

tests/e2e/conftest.py
heitorlessa: YES! auto-refactor from PyCharm when moving things around.

tests/e2e/utils/data_fetcher/metrics.py
heitorlessa: Great catch!! Leftover from the last PR

tests/e2e/utils/data_fetcher/logs.py
heitorlessa: It will be in the next PR specifically for Logger

tests/e2e/utils/asset.py
```diff
@@ -2,18 +2,46 @@
...
 from mypy_boto3_s3 import S3Client
+from pydantic import BaseModel, Field
 
 from aws_lambda_powertools import Logger
-from tests.e2e.utils.models import AssetTemplateConfig, TemplateAssembly
 
 logger = Logger(service="e2e-utils")
 
 
+class AssetTemplateConfigSource(BaseModel):
```
mploski: I would rename AssetTemplateConfig to AssetManifest following official naming convention

tests/e2e/utils/data_builder/__init__.py
```diff
@@ -0,0 +1,24 @@
...
+    build_add_metric_input,
+    build_metric_name,
+    build_metric_query_data,
+    build_multiple_add_metric_input,
+    build_put_annotations_input,
+    build_put_metadata_input,
+)
+from tests.e2e.utils.data_builder.traces import build_trace_default_query
+
+__all__ = [
```
mploski: Should we keep all list, considering that you explicitly import methods from modules into the namespace?

tests/e2e/utils/data_builder/common.py
```diff
@@ -0,0 +1,9 @@
+import secrets
```
mploski, line 1: Nice! I usually use random for building random value

tests/e2e/utils/data_builder/metrics.py
```diff
@@ -0,0 +1,130 @@
...
+from aws_lambda_powertools.metrics import MetricUnit
+from tests.e2e.utils.data_builder.common import build_random_value
+
+
+def build_metric_query_data(
+    namespace: str,
+    metric_name: str,
+    period: int = 60,
+    stat: str = "Sum",
+    dimensions: Optional[List[DimensionTypeDef]] = None,
```
mploski, line 14: Instead of None can we set default value to empty list and remove line 16?

tests/e2e/utils/data_builder/metrics.py
```diff
@@ -0,0 +1,130 @@
...
+) -> List[MetricDataQueryTypeDef]:
+    dimensions = dimensions or []
```
mploski, line 36: Should we add a docstring here the same way you do in other methods? I guess we want to add it by default to every method?

tests/e2e/utils/data_builder/metrics.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Returns
+    -------
+    List[DimensionTypeDef]
+        Metric dimension input
+    """
+    return [{"Name": name, "Value": value} for name, value in dimensions.items()]
+
+
+def build_put_annotations_input(**annotations: str) -> List[Dict]:
```
mploski: are annotations str type?

tests/e2e/utils/data_builder/metrics.py
```diff
@@ -0,0 +1,130 @@
...
+
+    Returns
+    -------
+    List[Dict]
+        List of put annotations input
+    """
+    return [{"key": key, "value": value} for key, value in annotations.items()]
+
+
+def build_put_metadata_input(namespace: Optional[str] = None, **metadata: Any) -> List[Dict]:
```
mploski: shouldn't metadata be a type of Mapping here?

tests/e2e/utils/data_fetcher/logs.py
```diff
@@ -0,0 +1,39 @@
...
+@retry(ValueError, delay=1, jitter=1, tries=20)
```
mploski, line 26: I think we may adjust this number similar way you did here: https://github.com/awslabs/aws-lambda-powertools-python/pull/1457/files#diff-ae1df3bf5a60a4705238094f7903173aa2111914c9571300fb2c97d80a6bd150R12

tests/e2e/utils/data_fetcher/metrics.py
```diff
@@ -0,0 +1,71 @@
...
+    metric_name: str,
+    dimensions: Optional[List[DimensionTypeDef]] = None,
+    cw_client: Optional[CloudWatchClient] = None,
+    end_date: Optional[datetime] = None,
+    period: int = 60,
+    stat: str = "Sum",
+) -> MetricDataResultTypeDef:
+    """Fetch CloudWatch Metrics
+
+    It takes into account eventual consistency with up to 10 retries and 1s jitter.
```
mploski: 1.5s jitter :-)

tests/e2e/utils/data_fetcher/logs.py
```diff
@@ -0,0 +1,39 @@
...
+def get_logs(lambda_function_name: str, log_client: CloudWatchClient, start_time: int, **kwargs: dict) -> List[Log]:
+    response = log_client.filter_log_events(logGroupName=f"/aws/lambda/{lambda_function_name}", startTime=start_time)
+    if not response["events"]:
+        raise ValueError("Empty response from Cloudwatch Logs. Repeating...")
+    filtered_logs = []
+    for event in response["events"]:
+        try:
+            message = Log(**json.loads(event["message"]))
```
mploski, line 34: I think we need to decide whether to use Poetry models in data_fetchers/data_builder for the sake of consistency. Here I used it, but I can see that in other places you use typdef taken from mypy_boto3 directly and return simple objects

tests/e2e/utils/data_fetcher/traces.py
```diff
@@ -0,0 +1,266 @@
...
+    def get_annotation(self, key: str, value: Optional[any] = None) -> List:
+        return [
+            annotation
+            for annotation in self.annotations
+            if (value is not None and annotation.get(key) == value) or (value is None and key in annotation)
+        ]
+
+    def get_metadata(self, key: str, namespace: str = "") -> List[Dict[str, Any]]:
+        seen = []
+        for meta in self.metadata:
```
mploski, line 116: maybe we can use list comprehension here?

tests/e2e/utils/data_fetcher/traces.py
```diff
@@ -0,0 +1,266 @@
...
+        List[TraceSubsegment]
+            List of subsegments
+        """
+        seen = []
+        for document in self.trace_documents.values():
+            if document.subsegments:
+                for subsegment in document.subsegments:
+                    if subsegment.name not in self.exclude_segment_name:
+                        seen.append(subsegment)
+                    self._find_nested_subsegments(subsegment, seen)
```
mploski: If I read this correctly, this one will find only directly nested subsegments. Should we also account for a recursive nested subsegments ( not only in first branch, but also in second/third)?

tests/e2e/utils/data_fetcher/traces.py
```diff
@@ -0,0 +1,266 @@
...
+
+        self.trace_ids = self._get_trace_ids(pages)
+        self.trace_documents = self._get_trace_documents()
+        self.subsegments = self._get_subsegments()
+
+    def get_annotation(self, key: str, value: Optional[any] = None) -> List:
+        return [
+            annotation
+            for annotation in self.annotations
+            if (value is not None and annotation.get(key) == value) or (value is None and key in annotation)
```
mploski, line 111: I usually write variable checking against None as  ` if value` and `if not value`. Maybe we could agree on some convention to avoid multiple styles in the same codebase?

tests/e2e/utils/data_fetcher/traces.py
```diff
@@ -0,0 +1,266 @@
...
+        origin : Optional[List[str]], optional
+            Trace origin name to filter traces, by default ["AWS::Lambda::Function"]
+        minimum_traces : int
+            Minimum number of traces to be retrieved before exhausting retry attempts
+        """
+        self.filter_expression = filter_expression
+        self.start_date = start_date
+        self.end_date = end_date or self.start_date + timedelta(minutes=5)
+        self.xray_client: XRayClient = xray_client or boto3.client("xray")
+        self.trace_ids: List[str] = []
```
mploski, line 85: Is it used here? I can see that we overwrite trace_ids, trace_documents and subsegments in lines 104-106

tests/e2e/utils/data_fetcher/traces.py
```diff
@@ -0,0 +1,266 @@
...
+        paginator = self.xray_client.get_paginator("get_trace_summaries")
+        pages = paginator.paginate(
+            StartTime=self.start_date,
+            EndTime=self.end_date,
+            TimeRangeType="Event",
+            Sampling=False,
+            FilterExpression=self.filter_expression,
+        )
+
+        self.trace_ids = self._get_trace_ids(pages)
```
mploski: I wonder why we store the result as class instance variable here? We use it only in one place as an input to get_trace_documents, so maybe passing this as a explicit parameter i.e `self._get_trace_documents(trace_ids= self._get_trace_ids(pages))`. What is your point of view on this?

tests/e2e/utils/data_fetcher/logs.py
heitorlessa: yeah, I'll do that in the next PR.

tests/e2e/utils/data_fetcher/logs.py
heitorlessa: My rational was `mypy_boto3` is for calls to AWS input/output. But some services like X-Ray or CDK don't have models for some requests we use, so we create a BaseModel for them. In this case (Log) it's easier - it's a BaseModel - it's our data.

tests/e2e/tracer/conftest.py
heitorlessa: YES! done now.

tests/e2e/tracer/infrastructure.py
heitorlessa: Done. `None` is implicit but it doesn't hurt being explicit that they don't have to return anything. Good catch!

tests/e2e/utils/data_builder/common.py
heitorlessa: Yeah, secrets solves the rng problem and chars that might not be allowed in CFN resources (token_urlsafe)

tests/e2e/utils/data_builder/metrics.py
heitorlessa: Not at function signature, because it'd make it a mutable default. https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

tests/e2e/utils/data_builder/metrics.py
heitorlessa: Good call! I'm adding to every "public" method known to us.

tests/e2e/utils/data_fetcher/__init__.py
heitorlessa: Solved it with a new feature in flake8 - per-file-ignore ;) TIL

tests/e2e/utils/asset.py
heitorlessa: As you wish ;)

tests/e2e/utils/asset.py
heitorlessa: Great catch!

tests/e2e/utils/data_builder/__init__.py
heitorlessa: Solved with the new flake8 per-file-ignore as these were unused imports hence the `__all__`

tests/e2e/utils/data_builder/metrics.py
heitorlessa: I was also surprised with the annotation. It's meant to be the value of the arg within a kwarg. 

`build_put_annotations_input(service="blah")`  It checks the value `="blah"`

tests/e2e/utils/data_builder/metrics.py
heitorlessa: Not because it's the value of the mapping not the key. It could be any JSON valid value. But this comment was great because this should be in `traces.py` not in `metrics.py`... bad refactor... fixing it.

tests/e2e/utils/data_fetcher/metrics.py
heitorlessa: Yes!!

tests/e2e/utils/data_fetcher/traces.py
heitorlessa: I tried, it became unreadable :/

tests/e2e/utils/data_fetcher/traces.py
heitorlessa: That's because `if value` and `if not value` will assert whether `value` is truthy or falsy. In this case we want to know if `value` is ever set.

tests/e2e/utils/data_fetcher/traces.py
heitorlessa: I was using in `test_tracer.py` too to quickly assert number of traces. Now that we ensure a minimum number of traces are found **and** that we assert on subsegments, we might not need it.

Removing it now. If we need later we can simply re-add it.

tests/e2e/utils/data_fetcher/traces.py
heitorlessa: true! Need to think of a cleaner way

heitorlessa: finally found it! Here's the solution using generator. [`15b3285` (#1457)](https://github.com/awslabs/aws-lambda-powertools-python/pull/1457/commits/15b3285318204e28e7f73c5992df3cd3246b524c)

Updating tests to make all of them use nested subsegments to cover them too - making way more robust too.

**PS**: That was challenging!!

image

image

heitorlessa:
All feedback addressed @mploski - when you're ready

heitorlessa:
I'm gonna simplify `get_metrics` return to return the Values so tests are easier to write too ...... but in another PR.

tests/e2e/utils/data_builder/metrics.py
mploski: oh, yes makes total sense! Thx

tests/e2e/utils/data_fetcher/traces.py
mploski: ah ok, we want to check if user set the variable even with empty string. Thx!

mploski (approved):

rubenfonseca (approved):

heitorlessa:
Thanks a lot for the review and suggested changes @mploski @rubenfonseca ! Merging.

Issue #1459, opened 2022-08-18
RFC: Lambda Powertools for Python v2


rubenfonseca:
### Which AWS Lambda Powertools utility does this relate to?

Powertools itself

### Summary

The [Python 3.6 Lambda Runtime was deprecated](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtime-support-policy) as of Aug 17th 2022. This follows Python 3.6 End-Of-Life (EOL) reached on December 23, 2021. This means it's not possible to create or update Lambda functions using that runtime.

We should take this opportunity to release a new major version of Powertools, and decide what breaking changes to include.

### Use case

Customers need to upgrade their Lambda Functions to one of the supported Python runtimes (at the moment of writing, Python 3.7-3.9).

We could use this major upgrade to introduce some breaking changes on Lambda Powertools that will simplify the code and/or improve support for existing features.

### Proposal

We strive to make minimal breaking changes due to timing constraints with Python 3.6 deprecation. In ideal scenarios, we would provide ample time, nightly builds, a Beta for a major version, including a linter to help detect and upgrade from breaking changes.

This means, V2 will focus primarily on the following themes:

* **Drop Python 3.6**. Follows alignment with Lambda runtime deprecation policy. It also allows us to upgrade development dependencies and documentation niceties we couldn't before due to prolonged Python 3.6 EOL life in Lambda.
* **Make all dependencies optional to optimize package size**. AWS SDK makes up for over 90% of our package size. We also [heard from customers](https://github.com/awslabs/aws-lambda-powertools-python/issues/1433) that they want to use additional observability providers, thus wanting X-Ray SDK to be optional. 
    - Lambda Layer will include all optional dependencies excluding those already available at Lambda runtime to further optimize cold start. This also means dropping [SAR Extras](https://serverlessrepo.aws.amazon.com/applications/eu-west-1/057560766410/aws-lambda-powertools-python-layer-extras) as it becomes redundant.
*  **Remove deprecated features**. Specifically `sqs_batch_processor` and `PartialSQSProcessor` in favour of `BatchProcessor` launched 11 months ago that improved accuracy, security, and speed due to the new Lambda integration.
[...]

PR #1463, opened 2022-08-19
chore(batch): deprecate sqs_batch_processor

rubenfonseca:
**Issue number:** #1459

### Changes

This PR adds a deprecation warning to the `sqs_batch_processor` decorator and `PartialSQSProcessor` class.

It also adds a banner on the documentation urging customers to upgrade before the next major version.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/batch/sqs.py
+++ b/aws_lambda_powertools/utilities/batch/sqs.py
@@ -80,7 +80,7 @@ class PartialSQSProcessor(BasePartialProcessor):
 
         warnings.warn(
             "The sqs_batch_processor decorator and PartialSQSProcessor class are now deprecated, "
-            "and will be removed in the next major version of Powertools. "
+            "and will be removed in the next major version. "
             "Please follow the upgrade guide at "
             "https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/batch/#legacy "
             "to use the native batch_processor decorator or BatchProcessor class."
--- a/aws_lambda_powertools/utilities/batch/sqs.py
+++ b/aws_lambda_powertools/utilities/batch/sqs.py
@@ -6,6 +6,7 @@ Batch SQS utilities
 import logging
 import math
 import sys
+import warnings
 from concurrent.futures import ThreadPoolExecutor, as_completed
 from typing import Any, Callable, Dict, List, Optional, Tuple, cast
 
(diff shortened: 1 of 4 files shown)


rubenfonseca:
@heitorlessa I do not know if the way I've implemented the banner on the docs is the correct way to do it.

aws_lambda_powertools/utilities/batch/sqs.py
```diff
@@ -77,6 +78,14 @@ def __init__(
         self.suppress_exception = suppress_exception
         self.max_message_batch = 10
 
+        warnings.warn(
+            "The sqs_batch_processor decorator and PartialSQSProcessor class are now deprecated, "
+            "and will be removed in the next major version of Powertools. "
```
heitorlessa:
```suggestion
            "and will be removed in the next major version. "
```

docs/overrides/main.html
```diff
@@ -1,5 +1,10 @@
 {% extends "base.html" %}
```
heitorlessa, line 2: This would add an announcement for the entire documentation, we'd want under `batch.md` instead with admonition:

```
???+ warning
    our message...
```

This type of announcement **will** be super useful to let people know we're going to launch v2. So perhaps rephrase to inform them that? I think it'll also need to make it auto-hide so those who already seen don't see it again: https://squidfunk.github.io/mkdocs-material/setup/setting-up-the-header/?h=announcement#automatic-hiding

heitorlessa (changes requested):
Awesome! Two notes on repurposing announcement bar to inform customers of v2, and admonition for `batch.md`

rubenfonseca:
Moved the warning to the batch doc page, and added a banner to documentation saying "v2 is coming". As always, feedback on wording is very important to me.

heitorlessa (approved):

docs/overrides/main.html
```diff
@@ -1,5 +1,10 @@
...
 
+{% block announce %}
+  ⚠️ Powertools for Python v2 is coming soon!
```
heitorlessa:
```suggestion
  👋 Powertools for Python v2 is coming soon!
```

heitorlessa: @rubenfonseca that's the only thing I'd change. Wording is great!

👋 is a more friendly "attention grab" than ⚠️ that could be misunderstood as a problem

Issue #1484, opened 2022-08-30
Feature request: Kafka Event Data Class


lyoung-confluent:
### Use case

When using a Kafka event source either from [self-managed Apache Kafka](https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html) or [Amazon MSK](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html) there should be a relevant data class for easy iteration/decoding.

Such functionality [is implemented already in aws-lambda-go](https://github.com/aws/aws-lambda-go/blob/main/events/kafka.go) and [in aws-lambda-java-libs](https://github.com/aws/aws-lambda-java-libs/blob/da593256513d3cec2849bb3955ad9ec3ebf38545/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java).

### Solution/User Experience

Similar to other streaming data sources a `records` property should be iterable returning a `KafkaEventRecord` that transparently decodes the values from JSON:
```py
from aws_lambda_powertools.utilities.data_classes import event_source, KafkaEvent

@event_source(data_class=KafkaEvent)
def handler(event: KafkaEvent, context: LambdaContext):
	for record in event.records:
		print(record.decoded_key, record.json_value)
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [X] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

rubenfonseca:
Thank you so much for the feature request @lyoung-confluent! It makes a lot of sense to cover Kafka events too. We're looking forward to review your PR tomorrow!

ran-isenberg:
@rubenfonseca  i think we should also create a similar parser event for this one. what do you think?

ran-isenberg:
see also  https://github.com/awslabs/aws-lambda-powertools-python/pull/1499

PR #1485, opened 2022-08-30
feat(data-classes): add KafkaEvent and KafkaEventRecord

lyoung-confluent:
**Issue number:** #1484

## Summary
This change adds a data class for Kafka event sources either from [self-managed Apache Kafka](https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html) or [Amazon MSK](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html).

### Changes
A new data class `KafkaEvent` is introduced which includes a `records` property/generator. The returned `KafkaEventRecord` class has a variety of useful properties for decoding the record key, value and headers. 

### User experience
```python
from aws_lambda_powertools.utilities.data_classes import event_source, KafkaEvent

@event_source(data_class=KafkaEvent)
def handler(event: KafkaEvent, context: LambdaContext):
	for record in event.records:
		print(record.decoded_key, record.json_value)
```

## Checklist

* [x] Changes are documented

<details>

Not a breaking change

**RFC issue number**: ?

Checklist:

</details>

## Acknowledgment

-----
[View rendered docs/utilities/data_classes.md](https://github.com/lyoung-confluent/aws-lambda-powertools-python/blob/patch-2/docs/utilities/data_classes.md)


leandrodamascena:
Hi @lyoung-confluent! Indeed this is an awesome PR! Thanks a lot for taking the time to help improve the project! 🥇 

I'm going to start doing a review on this PR today and let you know if we need to make some small changes.

aws_lambda_powertools/utilities/data_classes/kafka_event.py
```diff
@@ -0,0 +1,122 @@
...
+        """The AWS service from which the Kafka event record originated."""
+        return self["eventSource"]
+
+    @property
+    def event_source_arn(self) -> str:
+        """The AWS service ARN from which the Kafka event record originated."""
+        return self["eventSourceArn"]
+
+    @property
+    def bootstrap_servers(self) -> Optional[List[str]]:
```
leandrodamascena: Hi @lyoung-confluent! I don't think we need to add this "if" here. Even the AWS documentation [here](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html) doesn't mention the bootstrapServers field, I created an MSK cluster and added it as an event source of a Lambda and I was able to confirm that the bootstrapServers field is also present there.
image

btw, I've notified the documentation team to include it.

aws_lambda_powertools/utilities/data_classes/kafka_event.py
```diff
@@ -0,0 +1,122 @@
...
+    - https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html
+    """
+
+    @property
+    def event_source(self) -> str:
+        """The AWS service from which the Kafka event record originated."""
+        return self["eventSource"]
+
+    @property
+    def event_source_arn(self) -> str:
```
leandrodamascena: In case of selfManaged Kafka events, the eventSourceArn doesn't exists, so, I think that make sense changing this to `return self["eventSourceArn"] or None`

tests/functional/test_data_classes.py
```diff
@@ -1138,6 +1139,30 @@ def test_base_proxy_event_json_body_with_base64_encoded_data():
     assert event.json_body == data
 
 
+def test_kafka_event():
+    event = KafkaEvent(load_event("kafkaEvent.json"))
+    assert event.event_source == "aws:SelfManagedKafka"
```
leandrodamascena: Could you please change the event source to "aws:kafka" and add the eventSourceArn field?

leandrodamascena (changes requested):
Hello @lyoung-confluent! I suggested some changes in this PR.

aws_lambda_powertools/utilities/data_classes/kafka_event.py
lyoung-confluent: Great, I've updated it to assume the field is always present and just return `List[str]` instead. 

Do you think it makes sense though for this property to return a list or should it return the raw string and we offer a `decoded_bootstrap_servers` that does the split similar to other `decoded_` properties?

aws_lambda_powertools/utilities/data_classes/kafka_event.py
lyoung-confluent: I replaced it with `self.get("eventSourceArn")` so it doesn't throw a `KeyError` if it's not present (instead just returns `None`)

aws_lambda_powertools/utilities/data_classes/kafka_event.py
leandrodamascena: OK I see! Great idea, so in this case we can have the raw string and another property with a decoded value as List of string.

Please, do this.

leandrodamascena:
Hey @lyoung-confluent. I split the tests into 2 to make it clearer what is being tested and added 1 more field to test the bootstrapserver as a raw string. I think it's good for me to merge. Do you have any more considerations?

:rocket: :airplane:

lyoung-confluent:
Looks good to me, thanks for the quick feedback/review @leandrodamascena!

Issue #1454, opened 2022-08-15
Feature request: Span operation names to include class name


petarlishov:
### Use case

Currently if you add a span to a function with the provided annotation `tracer.capture_method`, the created span only uses the function's name. So if it is a function within a class, say for example some implementation of an abstract class, it is hard to realise which particular method was called as there are multiple classes with the same method.

### Solution/User Experience

If we include the class in the operation name of the span, I think that would be helpful. I am not really sure if I won't start  doscussion on whether the full module name needs included as well, but I think in terms of usefulness, asking for the class name is a sensible middle ground that would cover most people's use cases (hopefully others don't often have the same class name within different modules of the same project)

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

rubenfonseca:
Hi @petarlishov thanks for opening this. We're actively trying to prioritize this issue for our v2 release in the next couple of weeks. We will also apply a [similar fix to the idempotency utility](https://github.com/awslabs/aws-lambda-powertools-python/issues/1330). Stay tuned :)

leandrodamascena:
Hi @petarlishov, how are you? I'm working on it and fixing some issues in the tests before opening a PR. This feature will be included in [v2](https://github.com/awslabs/aws-lambda-powertools-python/issues/1459) and it will take a few weeks for us to release this new version.

I was wondering if this feature is blocking something you need to work with, if so we can find a solution (as a feature flag) where you can set a custom name for the `subsegment` instead of getting this [automatically](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/tracing/tracer.py#L511).

Please let us know if you are comfortable waiting for v2 to be released or if an immediate solution would help your work further.

Thank you

heitorlessa:
Closing as we're wrap to launch V2

PR #1486, opened 2022-08-31
feat(tracer): support methods with the same name (ABCs) by including fully qualified name in v2

leandrodamascena:
**Issue number:** #1454 

### Changes

This change modifies the name of the range generated by the capture_method tracker. In some cases it is very difficult to identify which functions are added in XRay just using the function name, the reason for this is because a python function/file/system can use abstract methods/classes or even functions with the same name in different classes.

image

### User experience

Before to this change, the span name only used the function name:
image

After this change, the span name includes the module and Python qualified name ([PEP 3155](https://peps.python.org/pep-3155/))

image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/tracing/tracer.py
```diff
@@ -508,7 +508,7 @@ async def async_tasks():
                 functools.partial(self.capture_method, capture_response=capture_response, capture_error=capture_error),
             )
 
-        method_name = f"{method.__name__}"
+        method_name = f"{method.__module__}.{method.__qualname__}"
```
heitorlessa, line 513: Add a comment above this line so we remember the expected format.

tests/unit/test_tracing.py
```diff
@@ -125,9 +125,11 @@ def greeting(name, message):
     # and add its response as trace metadata
     # and use service name as a metadata namespace
     assert in_subsegment_mock.in_subsegment.call_count == 1
-    assert in_subsegment_mock.in_subsegment.call_args == mocker.call(name="## greeting")
+    assert in_subsegment_mock.in_subsegment.call_args == mocker.call(
+        name="## unit.test_tracing.test_tracer_method.<locals>.greeting"
+    )
     assert in_subsegment_mock.put_metadata.call_args == mocker.call(
-        key="greeting response", value=dummy_response, namespace="booking"
+        key="unit.test_tracing.test_tracer_method.<locals>.greeting response", value=dummy_response, namespace="booking"
```
heitorlessa: make it as a constant and use f-string to make it easier to read/maintain later.

tests/e2e/tracer/handlers/same_function_name.py
```diff
@@ -0,0 +1,26 @@
...
+    @tracer.capture_method
+    def get_all(self):
+        return [{"id": f"{uuid4()}", "completed": False} for _ in range(5)]
+
+
+class Todos:
+    @tracer.capture_method
+    def get_all(self):
+        return [{"id": f"{uuid4()}", "completed": False} for _ in range(5)]
+
```
heitorlessa: Create an ABC so it's closer to what customers reported to us when they hit this issue.

tests/e2e/tracer/test_tracer.py
```diff
@@ -46,14 +56,39 @@ def test_lambda_handler_trace_is_visible(basic_handler_fn_arn: str, basic_handle
...
+
+    method_name_2 = "same_function_name.Comments.get_all"
+    method_subsegment_2 = f"## {method_name_2}"
+    method_metadata_key_2 = f"{method_name_2} response"
+
+    trace_query = data_builder.build_trace_default_query(function_name=same_function_name_fn)
+
+    # WHEN
+    _, execution_time = data_fetcher.get_lambda_response(lambda_arn=same_function_name_arn)
+    data_fetcher.get_lambda_response(lambda_arn=same_function_name_arn)
```
heitorlessa: no need as we're already asserting on Cold Start in another test.

tests/e2e/tracer/test_tracer.py
```diff
@@ -46,14 +56,39 @@ def test_lambda_handler_trace_is_visible(basic_handler_fn_arn: str, basic_handle
...
 
+def test_lambda_handler_trace_multiple_functions_same_name(same_function_name_arn: str, same_function_name_fn: str):
+    # GIVEN
+    method_name_1 = "same_function_name.Todos.get_all"
+    method_subsegment_1 = f"## {method_name_1}"
+    method_metadata_key_1 = f"{method_name_1} response"
+
+    method_name_2 = "same_function_name.Comments.get_all"
+    method_subsegment_2 = f"## {method_name_2}"
+    method_metadata_key_2 = f"{method_name_2} response"
```
heitorlessa: make the names explicit like `todo` and `comments`

heitorlessa (changes requested):
looks great!! Made a few comments to enhance maintenance.

aws_lambda_powertools/tracing/tracer.py
leandrodamascena: Hi @heitorlessa! Hmm probably not a good idea. Flake mark it as possible commented code and adding an exception can make the code confusing.

```
linting::flake8..........................................................Failed
- hook id: flake8
- exit code: 1

aws_lambda_powertools/tracing/tracer.py:512:1: E800 Found commented out code
```

Instead, I updated the docstring:
image

Pls let me know what do you think about.

tests/e2e/tracer/handlers/same_function_name.py
leandrodamascena: Looks great now :rocket: 
image

tests/e2e/tracer/test_tracer.py
leandrodamascena: Removed

tests/e2e/tracer/test_tracer.py
leandrodamascena: renamed

tests/unit/test_tracing.py
leandrodamascena: Now it's a constant

image

aws_lambda_powertools/tracing/tracer.py
leandrodamascena: Hello @heitorlessa! I added the comment above the suggested line.

heitorlessa:
Looking...

heitorlessa (approved):

Issue #1192, opened 2022-04-29
Feature request: Support cookies in Lambda function response for format 2.0


eflopez:
### Use case

I am using the `Router` class to setup some resolvers for use with `APIGatewayHttpResolver`, and I would like to be able to set cookies in my customised `Response` object, but this is not currently possible.

According to the [APIGateway response documentation](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html), the format for this is 
```json
{
    "cookies" : ["cookie1", "cookie2"],
    "isBase64Encoded": true|false,
    "statusCode": httpStatusCode,
    "headers": { "headername": "headervalue", ... },
    "body": "Hello from Lambda!"
}
```

### Solution/User Experience

Would be great to be able to just be able to do something like 
```python
    return Response(
        status_code=418,
        cookies=["cookie1", "cookie2"],
        content_type="application/json",
        body=payload,
        headers=custom_headers,
    )
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

michaelbrewer:
Good idea! There are more cumbersome way, but format 2.0 does make it easier. Once this has been "approved" / "triaged", it should be an easy addition to the API GW handler.

michaelbrewer:
@eflopez For discussion purpose i created a PR #1196

This will also allow for a more convenient method to set cookies for ALB, Http and Rest apis.

heitorlessa:
Removing `help wanted` as @rubenfonseca will start working on this next week.

heitorlessa:
Closing as we're wrap to launch V2

PR #1487, opened 2022-08-31
feat(event_handler): add cookies as 1st class citizen in v2

rubenfonseca:
**Issue number:** #1192

## Summary

This PR adds first class support for Cookies. After the last PR where we added multiple header values and cookies, we noticed we could improve the experience of Cookies by having a strongly typed class based approach.

carbon (15)

### Changes

I've added a class for `Cookie` modeling the most common options that are used today according to the [RFC](https://www.rfc-editor.org/rfc/rfc6265).

It also updated and extends the E2E tests to be more comprehensive and generic.

### User experience

Before, the user would have to encode the cookie himself, producing a string with the desired output:

```python
return Response(..., cookies=["session_id=12345; Secure; HttpOnly"])
```

This would be error prone, as getting the string right can be the source of bugs.

With the new class based Cookie, this becomes simpler:

```python
return Response(..., cookies=[Cookie(name="session_id", value="12345", secure=True, http_only=True])
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


rubenfonseca:
@heitorlessa @leandrodamascena would appreciate early feedback on this one before I continue to apply changes

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -206,7 +207,7 @@ def _add_cache_control(self, cache_control: str):
 
     def _compress(self):
         """Compress the response body, but only if `Accept-Encoding` headers includes gzip."""
-        self.response.headers["Content-Encoding"] = "gzip"
+        self.response.headers["Content-Encoding"].append("gzip")
```
rubenfonseca: Please ignore this line

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,75 @@
...
+    return timestamp.strftime("%a, %d %b %Y %H:%M:%S GMT")
+
+
+class Cookie:
+    def __init__(
+        self,
+        name: str,
+        value: str,
+        path: Optional[str] = None,
+        domain: Optional[str] = None,
```
heitorlessa:
```suggestion
        path: str = "",
        domain: str = "",
```

Setting as an empty string will simplify your conditional branch (`if path and len(path) > 0`) to `if path`.

`path: Optional[str] = None` is useful in scenarios when an empty string is valid too - `if path is not None` would accept a `""`.

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,75 @@
...
+        expires: Optional[datetime] = None,
+        max_age: Optional[int] = None,
+        secure: Optional[bool] = None,
```
heitorlessa: what problems do we cause by having it `True` by default?

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,75 @@
...
+        http_only: Optional[bool] = None,
```
heitorlessa: what problems do we cause by having it `True` by default?

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,75 @@
...
+            payload.write(f"; Domain={self.domain}")
+
+        if self.expires:
+            payload.write(f"; Expires={_getdate(self.expires)}")
+
+        if self.max_age:
+            if self.max_age > 0:
+                payload.write(f"; MaxAge={self.max_age}")
+            if self.max_age < 0:
+                payload.write("; MaxAge=0")
```
heitorlessa:
```suggestion
            if self.max_age >= 0:
                payload.write(f"; MaxAge={self.max_age}")
            else:
                payload.write("; MaxAge=0")
```

it could be me after fighting with CDK non-sense, but reading the initial logic meant that `Cookie(MaxAge=0)` wasn't possible.

aws_lambda_powertools/shared/headers_serializer.py
```diff
@@ -44,11 +46,11 @@ def serialize(self, headers: Dict[str, Union[str, List[str]]], cookies: List[str
             else:
                 combined_headers[key] = ", ".join(values)
 
-        return {"headers": combined_headers, "cookies": cookies}
+        return {"headers": combined_headers, "cookies": list(map(str, cookies))}
```
heitorlessa, line 49: neat FP!

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,75 @@
...
+        self.domain = domain
+        self.expires = expires
+        self.max_age = max_age
+        self.secure = secure
+        self.http_only = http_only
+        self.same_site = same_site
+        self.custom_attributes = custom_attributes
+
+    def __str__(self) -> str:
+        payload = StringIO()
```
heitorlessa, line 86: great to see the efficient use of memory ;)

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,75 @@
...
+        self.path = path
+        self.domain = domain
+        self.expires = expires
+        self.max_age = max_age
+        self.secure = secure
+        self.http_only = http_only
+        self.same_site = same_site
+        self.custom_attributes = custom_attributes
+
+    def __str__(self) -> str:
```
heitorlessa, line 85: For the length of string this is fine, but careful with `str` as it's one the slowest part in Python as it's actually a Class, and every call to `str()` creates an object + permutations.

I ran a quick benchmark and this part of the code is ~6ns so it's more than fine.

tests/e2e/event_handler/test_header_serializer.py
```diff
@@ -122,20 +125,32 @@ def test_api_gateway_http_headers_serializer(apigw_http_endpoint):
...
+            json={"body": body, "status_code": status_code, "headers": headers, "cookies": list(map(str, cookies))},
+        )
+    )
 
     # THEN
-    assert response.status_code == 200
-    assert response.content == b"Hello world"
-    assert response.headers["content-type"] == "text/plain"
+    assert response.status_code == status_code
+    assert response.content.decode("ascii") == body
```
heitorlessa, line 209: could you add a note here on the assertion?  it has quite a bit of logic we will forget when returning at some point.

heitorlessa (changes requested):
super clean, I loved it. Some suggestions and two questions on why not enabling `httpOnly` and `Secure` attrs by default (there's probably good reasons that I prolly don't know)

aws_lambda_powertools/shared/cookies.py
rubenfonseca: > When a cookie has the Secure attribute, the user agent will include the cookie in an HTTP request only if the request is transmitted over a secure channel (typically HTTPS).

So I'm inclined to have it ON by default. The problem would manifest if hitting the Lambda from a browser via a HTTP endpoint: the browser will ignore the cookie.

aws_lambda_powertools/shared/cookies.py
rubenfonseca: > The HttpOnly attribute restricts the cookie from being accessed by, for instance, [JavaScript](https://en.wikipedia.org/wiki/JavaScript),

I would be more worried to turn this on by default... I'll leave it as optional for now

aws_lambda_powertools/shared/cookies.py
```diff
@@ -0,0 +1,117 @@
...
+    """
+
+    DEFAULT_MODE = ""
+    LAX_MODE = "Lax"
+    STRICT_MODE = "Strict"
+    NONE_MODE = "None"
+
+
+def _format_date(timestamp: datetime) -> str:
+    return timestamp.strftime("%a, %d %b %Y %H:%M:%S GMT")
```
heitorlessa:
```suggestion
def _format_date(timestamp: datetime) -> str:
    # Specification example: Wed, 21 Oct 2015 07:28:00 GMT
    return timestamp.strftime("%a, %d %b %Y %H:%M:%S GMT")
```

examples/event_handler_rest/src/fine_grained_responses.py
```diff
@@ -26,7 +27,7 @@ def get_todos():
         content_type=content_types.APPLICATION_JSON,
         body=todos.json()[:10],
         headers=custom_headers,
-        cookies=["<cookie-name>=<cookie-value>; Secure; Expires=<date>"],
+        cookies=[Cookie(name="session_id", value="12345", secure=True)],
```
heitorlessa:
```suggestion
        cookies=[Cookie(name="session_id", value="12345")],
```

tests/e2e/event_handler/test_header_serializer.py
```diff
@@ -36,106 +39,179 @@ def lambda_function_url_endpoint(infrastructure: dict) -> str:
...
-    assert sorted(foo_headers) == ["bar", "zbr"]
+    for key, value in headers.items():
+        assert key in response.headers
+        value = value if isinstance(value, str) else ", ".join(sorted(value))
 
-    # Only the last cookie should be set
-    assert "MonsterCookie" in response.cookies.keys()
-    assert "CookieMonster" in response.cookies.keys()
+        # ALB sorts the header values randomly, so we have to re-order them for comparison here
+        returned_value = ", ".join(sorted(response.headers[key].split(", ")))
```
heitorlessa, line 107: Note for us: maybe we should have an utility function for this as we'll forget at some point (syntax et al)

heitorlessa (approved):

Issue #1498, opened 2022-09-04
Feature request: Kafka Parser support


ran-isenberg:
### Use case

see https://github.com/awslabs/aws-lambda-powertools-python/issues/1484

When using a Kafka event source either from [self-managed Apache Kafka](https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html) or [Amazon MSK](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html) there should be a relevant data class for easy iteration/decoding.

Such functionality [is implemented already in aws-lambda-go](https://github.com/aws/aws-lambda-go/blob/main/events/kafka.go) and [in aws-lambda-java-libs](https://github.com/aws/aws-lambda-java-libs/blob/da593256513d3cec2849bb3955ad9ec3ebf38545/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/KafkaEvent.java).

### Solution/User Experience

```
@event_parser(model=MyALambdaKafkaBusiness, envelope=envelopes.KafkaEnvelope)
def handle_lambda_kafka_with_envelope(event: List[MyALambdaKafkaBusiness], _: LambdaContext):
    assert event[0].key == "value"
    assert len(event) == 1


@event_parser(model=KafkaEventModel)
def handle_kafka_event(event: KafkaEventModel, _: LambdaContext):
    return event
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [X] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

PR #1499, opened 2022-09-04
feat(parser): add KafkaMskEventModel and KafkaSelfManagedEventModel

ran-isenberg:
**Issue number: #1498**

## Summary
parser support for  kafka swelf managed events to lambda.
TBH still debating if the envelope is really useful, i think in most cases people should extend kafkamodel class and use it instead.

### Changes
added envelope, models, tests docs.

### User experience
see #1498 

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered CHANGELOG.md](https://github.com/ran-isenberg/aws-lambda-powertools-python/blob/kafka2/CHANGELOG.md)
[View rendered docs/core/event_handler/api_gateway.md](https://github.com/ran-isenberg/aws-lambda-powertools-python/blob/kafka2/docs/core/event_handler/api_gateway.md)
[View rendered docs/utilities/parser.md](https://github.com/ran-isenberg/aws-lambda-powertools-python/blob/kafka2/docs/utilities/parser.md)


ran-isenberg:
i think there's something with spaces/ \n in  2 files for some reason

leandrodamascena:
Hi @ran-isenberg. Thank you very much for submitting this PR to improve the parser features. We were working on #1485 to make sure we had consistency in the fields and nothing was missing.

At first glance, I think we need to change things up in this PR.

1 - I can't see the eventSourceArn field. This is mandatory when the event source comes from [MSK](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html).
2 - Are you considering events coming only from [Self-Managed Kafka](https://github.com/awslabs/aws-lambda-powertools-python/pull/1499/files#diff-931129498022cf7ce019c09293e2cdf20b0e8983786da18e5b1a316b80fbe4a0R64)? I think we should consider both: MSK and Self-Managed Kafka.

Feel free to ping me on Discord if you want.

Leandro

ran-isenberg:
> Hi @ran-isenberg. Thank you very much for submitting this PR to improve the parser features. We were working on #1485 to make sure we had consistency in the fields and nothing was missing.
> 
> At first glance, I think we need to change things up in this PR.
> 
> 1 - I can't see the eventSourceArn field. This is mandatory when the event source comes from [MSK](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html). 2 - Are you considering events coming only from [Self-Managed Kafka](https://github.com/awslabs/aws-lambda-powertools-python/pull/1499/files#diff-931129498022cf7ce019c09293e2cdf20b0e8983786da18e5b1a316b80fbe4a0R64)? I think we should consider both: MSK and Self-Managed Kafka.
> 
> Feel free to ping me on Discord if you want.
> 
> Leandro

@leandrodamascena 
I can add the second one, no worries.
However, you didnt answer my question about the envelope - do you think it's required in this case? i dont think it makes sense to get the value param without the extra data. maybe we should remove the envelope and stay with the model only like in other events?

ran-isenberg:
@leandrodamascena pushed the fixes

ran-isenberg:
so, i found a small bug in the test file, it says there: aws:aws:SelfManagedKafka (aws:aws..).

ran-isenberg:
@heitorlessa  @leandrodamascena  added some data class kafka docs fixes

leandrodamascena:
Hi @ran-isenberg! Answering your question: yes, we think it makes sense to keep the Kafka envelope to extract the logs from the record. We believe it makes the code cleaner and we follow the same one we used in Kinesis, for example.

Could you please rebase the repository locally and just send the code related to this PR? For some reason this PR came with previous changes to github actions and this makes it a little difficult to read the PR and understand everything that is being changed.

I will make considerations that we could consider changing before merging this PR. But as I see, this is a nice feature to parser events from Kafka.

Thanks :rocket:

ran-isenberg:
> Hi @ran-isenberg! Answering your question: yes, we think it makes sense to keep the Kafka envelope to extract the logs from the record. We believe it makes the code cleaner and we follow the same one we used in Kinesis, for example.
> 
> Could you please rebase the repository locally and just send the code related to this PR? For some reason this PR came with previous changes to github actions and this makes it a little difficult to read the PR and understand everything that is being changed.
> 
> I will make considerations that we could consider changing before merging this PR. But as I see, this is a nice feature to parser events from Kafka.
> 
> Thanks 🚀

take a look now ;)

ran-isenberg:
you might notice that there's a clear distinction between the two kafka events, unlike data classes, which is what i'd expect from a parser.

leandrodamascena:
> you might notice that there's a clear distinction between the two kafka events, unlike data classes, which is what i'd expect from a parser.

Yes, I've seen that and it makes it clearer events from selfManaged and MSK. I'll open a issue to split the data classes into 2 events, just like the parser. :medal_sports:

aws_lambda_powertools/utilities/parser/models/kafka.py
```diff
@@ -0,0 +1,84 @@
...
+    @validator("headers", pre=True, allow_reuse=True)
+    def decode_headers_list(cls, value):
+        for header in value:
+            for key, values in header.items():
+                header[key] = bytes(values)
+        return value
+
+
+class KafkaBaseEventModel(BaseModel):
+    bootstrapServers: Optional[List[str]]
```
leandrodamascena: This field is not optional. So, we must remove the `Optional`. Even the AWS documentation [here](https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html) doesn't mention the bootstrapServers field, I created an MSK cluster and added it as an event source of a Lambda and I was able to confirm that the bootstrapServers field is also present there.

image

I've notified the documentation team to include it.

tests/functional/parser/test_kafka.py
```diff
@@ -0,0 +1,84 @@
+from typing import List
+
+from aws_lambda_powertools.utilities.parser import envelopes, event_parser
+from aws_lambda_powertools.utilities.parser.models import KafkaEventModel, KafkaRecordModel, MskEventModel
+from aws_lambda_powertools.utilities.typing import LambdaContext
+from tests.functional.parser.schemas import MyALambdaKafkaBusiness
+from tests.functional.utils import load_event
+
+
+@event_parser(model=MyALambdaKafkaBusiness, envelope=envelopes.KafkaEnvelope)
```
leandrodamascena:
```suggestion
@event_parser(model=MyLambdaKafkaBusiness, envelope=envelopes.KafkaEnvelope)
```

tests/functional/parser/test_kafka.py
```diff
@@ -0,0 +1,84 @@
...
+    assert len(record.headers) == 1
+    assert record.headers[0]["headerKey"] == b"headerValue"
+
+
+@event_parser(model=MskEventModel)
+def handle_msk_event(event: MskEventModel, _: LambdaContext):
+    return event
+
+
+def test_msk_event():
```
leandrodamascena:
```suggestion
def test_kafka_msk_event():
```

aws_lambda_powertools/utilities/parser/models/kafka.py
```diff
@@ -0,0 +1,84 @@
...
+        return base64.b64decode(value)
+    except (BinAsciiError, TypeError):
+        raise ValueError("base64 decode failed")
+
+
+def _bytes_to_string(value: bytes) -> str:
+    try:
+        return value.decode("utf-8")
+    except (BinAsciiError, TypeError):
+        raise ValueError("base64 UTF-8 decode failed")
```
leandrodamascena: As we are adding new features and always improving our code, what do you think about moving these 2 functions to the shared functions [file](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/shared/functions.py)?

tests/functional/parser/test_kafka.py
```diff
@@ -0,0 +1,84 @@
+from typing import List
+
+from aws_lambda_powertools.utilities.parser import envelopes, event_parser
+from aws_lambda_powertools.utilities.parser.models import KafkaEventModel, KafkaRecordModel, MskEventModel
+from aws_lambda_powertools.utilities.typing import LambdaContext
+from tests.functional.parser.schemas import MyALambdaKafkaBusiness
```
leandrodamascena:
```suggestion
from tests.functional.parser.schemas import MyLambdaKafkaBusiness
```

leandrodamascena (changes requested):

aws_lambda_powertools/utilities/parser/models/kafka.py
ran-isenberg: sure, np

aws_lambda_powertools/utilities/parser/models/kafka.py
ran-isenberg: np, will fix it

ran-isenberg:
@leandrodamascena  all fixed

leandrodamascena:
> @leandrodamascena all fixed

Thanks for addressing all changes very quickly! I made some minor changes to change the `models` name and make it clear to users.

Hi @heitorlessa, could you please take a look here? I would like to have a double check on this.

tests/functional/parser/test_kafka.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@event_parser(model=KafkaSelfManagedEventModel)
+def handle_kafka_event(event: KafkaSelfManagedEventModel, _: LambdaContext):
+    return event
+
+
+def test_kafka_event_with_envelope():
+    event = load_event("kafkaEventMsk.json")
+    handle_lambda_kafka_with_envelope(event, LambdaContext())
```
heitorlessa: We're missing a test for an envelope with `kafkaEventSelfManaged.json`. The envelope is using `KafkaMskEventModel` to parse the obj, and I'm not sure if this will work when you use the SelfManaged envelope piece.

leandrodamascena: Thanks for pointing out this @heitorlessa! The envelope doesn't actually work with a KafkaSelfManagedEvent.

image

ran-isenberg: good catch!
However, i want to bring back the issue of the envelope, i think it's not reauired at all here. you'd want to use all the data, topic name etxc, not just the value payload. 
WDYT?

leandrodamascena: Hi @ran-isenberg, I understand your concern about this, but since it's a library and widely used, we can't predict every use case.
Keeping this doesn't cause a breaking change or even make things slower than expected, so it's okay to keep that. I made a small change to support both events.

leandrodamascena:
Hi @ran-isenberg! If you have no other considerations, it's okay to merge this PR.

Thanks again for helping to improve the project. :rocket:

leandrodamascena (approved):

ran-isenberg:
sure, thx for the changes.
Let's merge :)

Issue #1500, opened 2022-09-05
Maintenance: Migrate to CDK CLI and off Docker for E2E tests


heitorlessa:
### Summary

This tracks the work necessary to use CDK CLI for synthesizing and deploying E2E stacks in parallel. It also tracks our initial move away from Docker to build Lambda Layer, as it slows down E2E authoring significantly (~1.3m) for every individual run.

### Why is this needed?

For v2, @rubenfonseca needed to use the default VPC for Event Handler E2E test. We were caught by surprise that [CDK Context Lookup methods](https://docs.aws.amazon.com/cdk/v2/guide/context.html#context_methods) only work via CDK CLI, which invalidates our initial implementation.

### Which area does this relate to?

Tests

### Solution

PR incoming. We've successfully moved to use CDK CLI while keeping our authoring and parallel deployment experience. 

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

PR #1501, opened 2022-09-05
chore(ci): migrate E2E tests to CDK CLI and off Docker

heitorlessa:
**Issue number:**  #1500

## Summary

This PR moves our E2E deployment details to CDK CLI and off Docker for Lambda Layer build. It simplifies the E2E authoring experience as a bonus, and removes our previous drop-in replacement for Asset bundling in favour of CDK CLI synth.

We can now use `from_lookup` methods. I've also made a minor modification in Tracer to move the annotation for `Service` to allow updating service name at runtime to ease E2E tests.

### Changes

- [x] Generate CDK App files at runtime to support parallel deployments later
- [x] Abstract Lambda Layer build to support before/after and local build
- [x] Migrate away from Docker Lambda Layer build w/ Pip manylinux wheels
- [x] Isolate CDK Apps and Stack Outputs to support parallel deployments and step-through debugging
- [x] Infer `feature_name` and `handlers_dir` at init to reduce authoring boilerplate. It will also be the basis to support reuse in other scenarios (`integ test`)
- [x] Remove previous Asset implementation
- [x] Remove side effects from Stack classes as CDK CLI can call it multiple times upon error
- [x] Synthesize CDK Apps at runtime and deploy from isolated cdk.out Apps
- [x] Bring CDK CLI as a dependency, and update E2E workflows to install Node
- [x] Document that VSCode Debugger will no longer work with E2E due to lack of support for background processes (PyCharm supports!)
- [ ] Update E2E RFC with notes on our change to CDK CLI
- [x] Revisit and update E2E Maintainer docs on structure and workflow
- [x] Ping CDK team on how to contribute docs for parallel deployments

### User experience

See comments due to verbosity.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
### Authoring experience change

We no longer need `feature_name` and `handlers_dir`  as we infer them now at constructor level. `layer_arn` is also no longer needed as we abstract Lambda Layer build and implement caching.

**BEFORE**

```python
from pathlib import Path

from tests.e2e.utils.data_builder import build_service_name
from tests.e2e.utils.infrastructure import BaseInfrastructure


class TracerStack(BaseInfrastructure):
    # Maintenance: Tracer doesn't support dynamic service injection (tracer.py L310)
    # we could move after handler response or adopt env vars usage in e2e tests
    SERVICE_NAME: str = build_service_name()
    FEATURE_NAME = "tracer"

    def __init__(self, handlers_dir: Path, feature_name: str = FEATURE_NAME, layer_arn: str = "") -> None:
        super().__init__(feature_name, handlers_dir, layer_arn)

    def create_resources(self) -> None:
        env_vars = {"POWERTOOLS_SERVICE_NAME": self.SERVICE_NAME}
        self.create_lambda_functions(function_props={"environment": env_vars})
```


**AFTER**

```python
from tests.e2e.utils.infrastructure import BaseInfrastructure


class TracerStack(BaseInfrastructure):
    def create_resources(self) -> None:
        self.create_lambda_functions()
```

heitorlessa:
### Lambda Layer improvements

We no longer deploy a stack for a Lambda Layer, no more Docker, and instead use a local abstraction for Pip install including extensibility for future changes (e.g., boto3 removal).

Our new Layer abstraction only builds if the source code has changed, improving the author experience for E2E speed. Time to hit first breakpoint is now at 80 seconds compared to 160s w/ Docker and no caching.

```python
        layer_build = LocalLambdaPowertoolsLayer().build()
        layer = LayerVersion(
            self.stack,
            "aws-lambda-powertools-e2e-test",
            layer_version_name="aws-lambda-powertools-e2e-test",
            compatible_runtimes=[
                Runtime.PYTHON_3_7,
                Runtime.PYTHON_3_8,
                Runtime.PYTHON_3_9,
            ],
            code=Code.from_asset(path=layer_build),
        )
```

heitorlessa:
### Lessons learned

> Will update more when bandwidth allows.

We've tried the following approaches before deciding we had to move to CDK CLI:

* Create a new VPC but we quickly reached default limits given the amount of stacks we need across all Python versions
* Look up default VPC using AWS SDK and use [`Vpc.from_attributes`](https://docs.aws.amazon.com/cdk/api/v1/docs/@aws-cdk_aws-ec2.Vpc.html#static-fromwbrvpcwbrattributesscope-id-attrs) to create an `IVpc` object that is expected from L2 constructs - This generated all sorts of errors w/ Subnet. It confirmed what the CDK documentation says: "sometimes accidentally works"
* Use L1 construct only but the authoring value diminished (as expected) from using CDK when trying to setup ALB and API Gateway

When migrating to CDK CLI, we learned a new side effects we had to work around in order to support our parallel deployment requirement:

1. CDK CLI calls your Stack class multiple times when using context lookup methods. This caused problems with random stack values, where assertions failed intermittently.
2. CDK CLI is initially designed to have a single CDK App with one or more Stacks, not distinct CDK Apps to synthetize and deploy. This caused problems with `manifest.json` intermittently being overwritten and race conditions where the file would be missing in between deployments.

To amortize the performance cost of bringing CDK CLI (Node init, CLI init, process spawning), we were incentivized to improve Lambda Layer and build it locally instead of Docker. However, we've also hit issues since `ILocalBundling` is only [documented for TypeScript](https://github.com/aws/aws-cdk/issues/17928) and we failed in the several attempts to make it work.

---

Some of the intermittent errors when trying to make parallel deployments with CDK CLI work. 

```javascript
 ❌  testV39-metrics-cae198e3-e4d1-413c-9d8d-a2a408b731ab failed: InvalidChangeSetStatus: Cannot delete ChangeSet in status CREATE_IN_PROGRESS
    at Request.extractError (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/protocol/query.js:50:29)
    at Request.callListeners (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/request.js:688:12)
    at Request.callListeners (/Users/lessa/.npm/_npx/e72b144743208263/node_modules/aws-sdk/lib/sequential_executor.js:116:18) {
  code: 'InvalidChangeSetStatus',
  time: 2022-09-03T18:38:32.058Z,
```

```python
jsii.errors.JSIIError: EEXIST: file already exists, mkdir '/Users/lessa/DEV/aws-lambda-powertools-python/cdk.out/asset.da705b5776c4e58b759b0a7027cebe7cae58195baada0a1281389eeeaa0eedb5'
```


```javascript
 ❌ Deployment failed: Error: Stack Deployments Failed: AlreadyExistsException: ChangeSet cdk-deploy-change-set cannot be created due to a mismatch with existing attribute Description
    at deployStacks (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-cdk/lib/deploy.ts:61:11)
    at CdkToolkit.deploy (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-cdk/lib/cdk-toolkit.ts:312:7)
    at initCommandLine (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-cdk/lib/cli.ts:349:12)

Stack Deployments Failed: AlreadyExistsException: ChangeSet cdk-deploy-change-set cannot be created due to a mismatch with existing attribute Description
```


```javascript
 ❌  testV39-event-handlers-dc05b551-0baa-43d9-944d-1b5f703d0016 failed: InvalidChangeSetStatus: Cannot delete ChangeSet in status CREATE_IN_PROGRESS
    at Request.extractError (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/protocol/query.js:50:29)
    at Request.callListeners (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/request.js:686:14)
    at Request.transition (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/request.js:688:12)
    at Request.callListeners (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-sdk/lib/sequential_executor.js:116:18) {
  code: 'InvalidChangeSetStatus',
  time: 2022-09-03T19:52:55.176Z,
  requestId: '6c56e040-5ee4-4f1c-b821-fa844a6ca90f',
  statusCode: 400,
  retryable: false,
  retryDelay: 386.4977584554941
}

 ❌ Deployment failed: Error: Stack Deployments Failed: InvalidChangeSetStatus: Cannot delete ChangeSet in status CREATE_IN_PROGRESS
    at deployStacks (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-cdk/lib/deploy.ts:61:11)
    at CdkToolkit.deploy (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-cdk/lib/cdk-toolkit.ts:312:7)
    at initCommandLine (/Users/lessa/.nvm/versions/node/v14.17.2/lib/node_modules/aws-cdk/lib/cli.ts:349:12)
```

alexpulver:
Adding an example of a top-level AWS CDK application that defines E2E tests cloud infrastructure + how to synthesize, deploy and destroy it.

```python
import aws_cdk as cdk

import constants
from deployment import Tracer, Logger

app = cdk.App()

if app.node.try_get_context("tracer") == "true":
    Tracer(app, "Tracer", env=constants.DEV_ENV)
if app.node.try_get_context("logger") == "true":
    Logger(app, "Logger", env=constants.DEV_ENV)

app.synth()

# Synth
#
# stacks=<get stack names to synth based on selected test groups>
# synth_command="cdk synth"
# foreach stack in stacks:
#   synth_command += " --context ${stack}=true"
# $(synth_command)

# Deploy
#
# number_of_stacks=<get number of stacks from previous step>
# cdk deploy --app cdk.out --all --require-approval never --concurrency ${number_of_stacks} --outputs-file outputs.json

# Destroy
#
# cdk destroy --app cdk.out --all --force
```

heitorlessa:
Extended the E2E framework documentation to include the new CDK CLI parallelization, and how a maintainer/contributor would add a new E2E test altogether (infra+tests)... and the internals behind it.

Adding as a new comment as it might be hard to follow in the GH PR file diff UI

heitorlessa:
## E2E framework

### Structure

Our E2E framework relies on [Pytest fixtures](https://docs.pytest.org/en/6.2.x/fixture.html) to coordinate infrastructure and test parallelization - see [Test Parallelization](#test-runner-parallelization) and [CDK CLI Parallelization](#cdk-cli-parallelization).

**tests/e2e structure**

```shell
.
├── __init__.py
├── conftest.py # builds Lambda Layer once
├── logger
│   ├── __init__.py
│   ├── conftest.py  # deploys LoggerStack
│   ├── handlers
│   │   └── basic_handler.py
│   ├── infrastructure.py # LoggerStack definition
│   └── test_logger.py
├── metrics
│   ├── __init__.py
│   ├── conftest.py  # deploys MetricsStack
│   ├── handlers
│   │   ├── basic_handler.py
│   │   └── cold_start.py
│   ├── infrastructure.py # MetricsStack definition
│   └── test_metrics.py
├── tracer
│   ├── __init__.py
│   ├── conftest.py  # deploys TracerStack
│   ├── handlers
│   │   ├── async_capture.py
│   │   └── basic_handler.py
│   ├── infrastructure.py  # TracerStack definition
│   └── test_tracer.py
└── utils
    ├── __init__.py
    ├── data_builder  # build_service_name(), build_add_dimensions_input, etc.
    ├── data_fetcher  # get_traces(), get_logs(), get_lambda_response(), etc.
    ├── infrastructure.py # base infrastructure like deploy logic, etc.
```

Where:

- **`<feature>/infrastructure.py`**. Uses CDK to define the infrastructure a given feature needs.
- **`<feature>/handlers/`**. Lambda function handlers to build, deploy, and exposed as stack output in PascalCase (e.g., `BasicHandler`).
- **`utils/`**. Test utilities to build data and fetch AWS data to ease assertion
- **`conftest.py`**. Deploys and deletes a given feature infrastructure. Hierarchy matters:
    - **Top-level (`e2e/conftest`)**. Builds Lambda Layer only once and blocks I/O across all CPU workers.
    - **Feature-level (`e2e/<feature>/conftest`)**. Deploys stacks in parallel and make them independent of each other.

### Mechanics

Under [`BaseInfrastructure`](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/tests/e2e/utils/infrastructure.py), we hide the complexity of deployment and delete coordination under `deploy`, `delete`, and `create_lambda_functions` methods.

This allows us to benefit from test and deployment parallelization, use IDE step-through debugging for a single test, run one, subset, or all tests and only deploy their related infrastructure, without any custom configuration.

> Class diagram to understand abstraction built when defining a new stack (`LoggerStack`)

```mermaid
classDiagram
    class InfrastructureProvider {
        <<interface>>
        +deploy() Dict
        +delete()
        +create_resources()
        +create_lambda_functions() Dict~Functions~
    }

    class BaseInfrastructure {
        +deploy() Dict
        +delete()
        +create_lambda_functions() Dict~Functions~
        +add_cfn_output()
    }

    class TracerStack {
        +create_resources()
    }

    class LoggerStack {
        +create_resources()
    }

    class MetricsStack {
        +create_resources()
    }

    class EventHandlerStack {
        +create_resources()
    }

    InfrastructureProvider <|-- BaseInfrastructure : implement
    BaseInfrastructure <|-- TracerStack : inherit
    BaseInfrastructure <|-- LoggerStack : inherit
    BaseInfrastructure <|-- MetricsStack : inherit
    BaseInfrastructure <|-- EventHandlerStack : inherit
```

### Authoring a new feature E2E test

Imagine you're going to create E2E for Event Handler feature for the first time. Keep the following mental model when reading:

```mermaid
graph LR
    A["1. Define infrastructure"]-->B["2. Deploy/Delete infrastructure"]-->C["3.Access Stack outputs" ]
```

#### 1. Define infrastructure

We use CDK as our Infrastructure as Code tool of choice. Before you start using CDK, you'd take the following steps:

1. Create `tests/e2e/event_handler/infrastructure.py` file
2. Create a new class `EventHandlerStack` and inherit from `BaseInfrastructure`
3. Override `create_resources` method and define your infrastructure using CDK
4. (Optional) Create a Lambda function under `handlers/alb_handler.py`

> Excerpt `tests/e2e/event_handler/infrastructure.py`

```python
class EventHandlerStack(BaseInfrastructure):
    def create_resources(self):
        functions = self.create_lambda_functions()

        self._create_alb(function=functions["AlbHandler"])
        ...

    def _create_alb(self, function: Function):
        vpc = ec2.Vpc.from_lookup(
            self.stack,
            "VPC",
            is_default=True,
            region=self.region,
        )

        alb = elbv2.ApplicationLoadBalancer(self.stack, "ALB", vpc=vpc, internet_facing=True)
        CfnOutput(self.stack, "ALBDnsName", value=alb.load_balancer_dns_name)
        ...
```

> Excerpt `tests/e2e/event_handler/handlers/alb_handler.py`

```python
from aws_lambda_powertools.event_handler import ALBResolver, Response, content_types

app = ALBResolver()


@app.get("/todos")
def hello():
    return Response(
        status_code=200,
        content_type=content_types.TEXT_PLAIN,
        body="Hello world",
        cookies=["CookieMonster", "MonsterCookie"],
        headers={"Foo": ["bar", "zbr"]},
    )


```

#### 2. Deploy/Delete infrastructure when tests run

We need to create a Pytest fixture for our new feature under `tests/e2e/event_handler/conftest.py`.

This will instruct Pytest to deploy our infrastructure when our tests start, and delete it when they complete whether tests are successful or not. Note that this file will not need any modification in the future.

> Excerpt `conftest.py` for Event Handler

```python
import pytest

from tests.e2e.event_handler.infrastructure import EventHandlerStack


@pytest.fixture(autouse=True, scope="module")
def infrastructure():
    """Setup and teardown logic for E2E test infrastructure

    Yields
    ------
    Dict[str, str]
        CloudFormation Outputs from deployed infrastructure
    """
    stack = EventHandlerStack()
    try:
        yield stack.deploy()
    finally:
        stack.delete()

```

#### 3. Access stack outputs for E2E tests

Within our tests, we should now have access to the `infrastructure` fixture we defined earlier in `tests/e2e/event_handler/conftest.py`.

We can access any Stack Output using pytest dependency injection.

> Excerpt `tests/e2e/event_handler/test_header_serializer.py`

```python
@pytest.fixture
def alb_basic_listener_endpoint(infrastructure: dict) -> str:
    dns_name = infrastructure.get("ALBDnsName")
    port = infrastructure.get("ALBBasicListenerPort", "")
    return f"http://{dns_name}:{port}"


def test_alb_headers_serializer(alb_basic_listener_endpoint):
    # GIVEN
    url = f"{alb_basic_listener_endpoint}/todos"
    ...
```

### Internals

#### Test runner parallelization

Besides speed, we parallelize our end-to-end tests to ease asserting async side-effects may take a while per test too, _e.g., traces to become available_.

The following diagram demonstrates the process we take every time you use `make e2e` locally or at CI:

```mermaid
graph TD
    A[make e2e test] -->Spawn{"Split and group tests <br>by feature and CPU"}

    Spawn -->|Worker0| Worker0_Start["Load tests"]
    Spawn -->|Worker1| Worker1_Start["Load tests"]
    Spawn -->|WorkerN| WorkerN_Start["Load tests"]

    Worker0_Start -->|Wait| LambdaLayer["Lambda Layer build"]
    Worker1_Start -->|Wait| LambdaLayer["Lambda Layer build"]
    WorkerN_Start -->|Wait| LambdaLayer["Lambda Layer build"]

    LambdaLayer -->|Worker0| Worker0_Deploy["Launch feature stack"]
    LambdaLayer -->|Worker1| Worker1_Deploy["Launch feature stack"]
    LambdaLayer -->|WorkerN| WorkerN_Deploy["Launch feature stack"]

    Worker0_Deploy -->|Worker0| Worker0_Tests["Run tests"]
    Worker1_Deploy -->|Worker1| Worker1_Tests["Run tests"]
    WorkerN_Deploy -->|WorkerN| WorkerN_Tests["Run tests"]

    Worker0_Tests --> ResultCollection
    Worker1_Tests --> ResultCollection
    WorkerN_Tests --> ResultCollection

    ResultCollection{"Wait for workers<br/>Collect test results"}
    ResultCollection --> TestEnd["Report results"]
    ResultCollection --> DeployEnd["Delete Stacks"]
```

#### CDK CLI parallelization

For CDK CLI to work with [independent CDK Apps](https://docs.aws.amazon.com/cdk/v2/guide/apps.html), we specify an output directory when synthesizing our stack and deploy from said output directory.

```mermaid
flowchart TD
    subgraph "Deploying distinct CDK Apps"
        EventHandlerInfra["Event Handler CDK App"] --> EventHandlerSynth
        TracerInfra["Tracer CDK App"] --> TracerSynth
       EventHandlerSynth["cdk synth --out cdk.out/event_handler"] --> EventHandlerDeploy["cdk deploy --app cdk.out/event_handler"]

       TracerSynth["cdk synth --out cdk.out/tracer"] --> TracerDeploy["cdk deploy --app cdk.out/tracer"]
    end
```

We create the typical CDK `app.py` at runtime when tests run, since we know which feature and Python version we're dealing with (locally or at CI).

> Excerpt `cdk_app_V39.py` for Event Handler created at deploy phase

```python
from tests.e2e.event_handler.infrastructure import EventHandlerStack
stack = EventHandlerStack()
stack.create_resources()
stack.app.synth()
```

When we run E2E tests for a single feature or all of them, our `cdk.out` looks like this:

```shell
total 8
drwxr-xr-x  18 lessa  staff   576B Sep  6 15:38 event-handler
drwxr-xr-x   3 lessa  staff    96B Sep  6 15:08 layer_build
-rw-r--r--   1 lessa  staff    32B Sep  6 15:08 layer_build.diff
drwxr-xr-x  18 lessa  staff   576B Sep  6 15:38 logger
drwxr-xr-x  18 lessa  staff   576B Sep  6 15:38 metrics
drwxr-xr-x  22 lessa  staff   704B Sep  9 10:52 tracer
```

```mermaid
classDiagram
    class CdkOutDirectory {
        feature_name/
        layer_build/
        layer_build.diff
    }

    class EventHandler {
        manifest.json
        stack_outputs.json
        cdk_app_V39.py
        asset.uuid/
        ...
    }

    class StackOutputsJson {
        BasicHandlerArn: str
        ALBDnsName: str
        ...
    }

    CdkOutDirectory <|-- EventHandler : feature_name/
    StackOutputsJson <|-- EventHandler
```

Where:

- **`<feature>`**. Contains CDK Assets, CDK `manifest.json`, our `cdk_app_<PyVersion>.py` and `stack_outputs.json`
- **`layer_build`**. Contains our Lambda Layer source code built once, used by all stacks independently
- **`layer_build.diff`**. Contains a hash on whether our source code has changed to speed up further deployments and E2E tests

Together, all of this allows us to use Pytest like we would for any project, use CDK CLI and its [context methods](https://docs.aws.amazon.com/cdk/v2/guide/context.html#context_methods) (`from_lookup`), and use step-through debugging for a single E2E test without any extra configuration.

> NOTE: VSCode doesn't support debugging processes spawning sub-processes (like CDK CLI does w/ shell and CDK App). Maybe [this works](https://stackoverflow.com/a/65339352). PyCharm works just fine.

heitorlessa:
Ready for review while I'm away next week - @rubenfonseca @leandrodamascena. This should delight Ruben with CDK `from_lookup` working as it should, plus a much faster experience when re-running E2E tests locally when source code hasn't changed (PLUS a lot less boilerplate!)

rubenfonseca:
This is just a pre-review comment to show my appreciation for this awesome work. Congratulations guys, it looks amazing.

cdk.context.json
```diff
@@ -0,0 +1,41 @@
...
+        ]
+      }
+    ]
+  },
+  "availability-zones:account=231436140809:region=eu-west-1": [
+    "eu-west-1a",
+    "eu-west-1b",
+    "eu-west-1c"
+  ]
+}
```
rubenfonseca: I believe this file should not be included in the repository:

https://github.com/aws-samples/aws-cdk-examples/blob/master/.gitignore#L378

tests/e2e/utils/infrastructure.py
```diff
@@ -138,25 +144,86 @@ def create_lambda_functions(self, function_props: Optional[Dict] = None) -> Dict
...
-        template, asset_manifest_file = self._synthesize()
-        assets = Assets(asset_manifest=asset_manifest_file, account_id=self.account_id, region=self.region)
-        assets.upload()
-        self.stack_outputs = self._deploy_stack(self.stack_name, template)
-        return self.stack_outputs
+        stack_file = self._create_temp_cdk_app()
+        synth_command = f"npx cdk synth --app 'python {stack_file}' -o {self._cdk_out_dir}"
+        deploy_command = (
+            f"npx cdk deploy --app '{self._cdk_out_dir}' -O {self._stack_outputs_file} --require-approval=never"
+        )
```
rubenfonseca, line 160: I'm under the impression that since we're using `npx` here, we might not have to use `package.json`. While I like that idea, this makes me wonder how we can control new CDK releases not affecting our E2E silently.

tests/e2e/utils/lambda_layer/powertools_layer.py
```diff
@@ -0,0 +1,48 @@
...
+logger = logging.getLogger(__name__)
+
+
+class LocalLambdaPowertoolsLayer(BaseLocalLambdaLayer):
+    IGNORE_EXTENSIONS = ["pyc"]
+
+    def __init__(self, output_dir: Path = CDK_OUT_PATH):
+        super().__init__(output_dir)
+        self.package = f"{SOURCE_CODE_ROOT_PATH}[pydantic]"
+        self.build_args = "--platform manylinux1_x86_64 --only-binary=:all: --upgrade"
```
rubenfonseca, line 20: I've tested this on my M1 (aarch64) mac and the E2E test ran. I guess here we're installing a dependency for the Lambda Function, so it would be fine.

In the future it would be great to have a parameter to create arm64 functions too.

rubenfonseca (changes requested):
THANK YOU SO MUCH FOR THIS, what a great evolution from the previous code! Just left some comments and questions, but it looks very mergeable to me :)

tests/e2e/utils/infrastructure.py
heitorlessa: It's the contrary, we're using `npx` to use a local CDK CLI installation to make it reliable. The way `npx` works is it'll try to find CDK CLI in `./node_modules/.bin` first, then lookup for a global CDK CLI, and if not found it'll install the latest version.

https://www.npmjs.com/package/npx

cdk.context.json
heitorlessa: sounds like that sample is not following its own CDK recommendation: https://docs.aws.amazon.com/cdk/v2/guide/context.html#context_construct

I triple checked with @alexpulver before committing it

tests/e2e/utils/lambda_layer/powertools_layer.py
heitorlessa: Yes, I wasn't sure on the best UX on that since it's abstracted. My initial thought is to accept an optional parameter `platform` where you could pass an enum (`x86_64`, `arm64`), and if it's `None` it'll auto-detect the platform.

What do you think?

rubenfonseca:
Unfortunately, because of our changes in `develop`, we will have to reconstruct this again :( Let me know if you need a hand.

rubenfonseca (approved):

rubenfonseca:
@heitorlessa targeting v2, and fix the e2e tracer specs.

Only problem remaining is the `cdk.context.json`. Already had to checkout the file multiple times before committing, because I didn't want to add my AWS account data on top of the file. The file will only get bigger as more maintainers run the tests. I think we should try to remove it for now and see how it goes.

heitorlessa:
Agree, let's remove. I think the "best practice" from CDK only makes sense
for projects owned by a given team that use the same account, not in our
case as a OSS lib.

tests/e2e/utils/lambda_layer/powertools_layer.py
heitorlessa: I take it back, this auto-detect will cause trouble because we also need to take into account Functions are created using ARM architecture. Let's leave for another PR when we enable ARM Lambda Layers.

Issue #1503, opened 2022-09-06
Bug: DictWrapper should fully implement Mapping interface


Tankanow:
### Expected Behaviour

`DictWrapper` and all of its subclasses should fully implement the Mapping interface so that it can be used interchangeably with code that expects the `event` to be a dictionary. This helps Powertools Middleware play nicely with existing middleware.

### Current Behaviour

`DictWrapper` does not implement the abstract `__iter__` and `__len__` methods from the `Mapping` interface. Currently, client code cannot iterate over any subclass of `DictWrapper`, nor can it use the dictionary built-ins `keys`, `values`, or `items`.

### Code snippet

```python
from aws_lambda_powertools.utilities.data_classes.common import DictWrapper

event = DictWrapper({'foo': 'bar'})

[k for k in event]  # raises KeyError: 0
# all of the following raise AttributeError: 'DictWrapper' object has no attribute '...'
event.keys()
event.items()
event.values()
```

### Possible Solution

`DictWrapper` should subclass either `Mapping` or `Dict`, for example:

```python
class DictWrapper(Mapping):
    """Provides a single read only access to a wrapper dict"""

    def __init__(self, data: Dict[str, Any]):
        self._data = data

    def __getitem__(self, key: str) -> Any:
        return self._data[key]

    def __eq__(self, other: Any) -> bool:
        if not isinstance(other, DictWrapper):
            return False

        return self._data == other._data

    def get(self, key: str) -> Optional[Any]:
        return self._data.get(key)

    @property
    def raw_event(self) -> Dict[str, Any]:
        """The original raw event dict"""
        return self._data

    # missing methods
    def __iter__(self) -> Iterator:
        return iter(self._data)

    def __len__(self) -> int:
        return len(self._data)
```

### Steps to Reproduce

# see comment above

### AWS Lambda Powertools for Python version

latest

### AWS Lambda function runtime

3.9

### Packaging format used

PyPi

### Debugging logs

```python
No logs here. I'm happy to submit the PR to fix this. This should be a non-breaking change.
```

rubenfonseca:
Hi @Tankanow thank you for the very clear explanation! Would you mind opening a PR with the changes, together with a couple of tests? We will be thrilled to review it and include it in our next release.

Tankanow:
@rubenfonseca, happy to do so. I'll get to it this week. 👍🏽

PR #1516, opened 2022-09-13
fix(event_sources): implement Mapping protocol on DictWrapper for better interop with existing middlewares

Tankanow:
**Issue #1503**

## Summary

DictWrapper and all of its subclasses should fully implement the Mapping interface so that it can be used interchangeably with code that expects the event to be a dictionary. This helps Powertools Middleware play nicely with existing middleware.

### Changes

Add `Mapping` as parent class of `DictWrapper`, implement missing `__len__` and `__iter__` abstract methods.

### User experience

Before this change, a user would encounter exceptions when trying to use many standard functions on dictionaries coerced into `DictWrapper`. Though a user could access the raw dictionary with the `raw_event` property, this was often clumsy, causing existing middleware to break (or to be order dependent) or causing the user to write lots of type-checking code.

## Checklist

* [ ] Changes are documented

<details>
<summary>Is this a breaking change? <b>No</b></summary>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered CHANGELOG.md](https://github.com/Tankanow/aws-lambda-powertools-python/blob/ISSUE-1503/CHANGELOG.md)

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -1,6 +1,7 @@
 import base64
 import json
-from typing import Any, Dict, Iterator, Mapping, Optional
+from collections.abc import Mapping
+from typing import Any, Dict, Iterator, Optional
 
 
 class DictWrapper(Mapping):
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -1,9 +1,9 @@
 import base64
 import json
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Iterator, Mapping, Optional
 
 
-class DictWrapper:
+class DictWrapper(Mapping):
     """Provides a single read only access to a wrapper dict"""
 
     def __init__(self, data: Dict[str, Any]):
@@ -18,6 +18,12 @@ class DictWrapper:
             return False
 
         return self._data == other._data
+
+    def __iter__(self) -> Iterator:
+        return iter(self._data)
+
+    def __len__(self) -> int:
+        return len(self._data)
 
     def get(self, key: str, default: Optional[Any] = None) -> Optional[Any]:
         return self._data.get(key, default)
--- a/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
+++ b/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
@@ -182,8 +182,10 @@ class StreamRecord(DictWrapper):
         item = self.get("ApproximateCreationDateTime")
         return None if item is None else int(item)
 
+    # This override breaks the Mapping protocol of DictWrapper, it's left here for backwards compatibility with
+    # a 'type: ignore' comment. This is currently the only subclass of DictWrapper that breaks this protocol.
     @property
-    def keys(self) -> Optional[Dict[str, AttributeValue]]:
+    def keys(self) -> Optional[Dict[str, AttributeValue]]:  # type: ignore
         """The primary key attribute(s) for the DynamoDB item that was modified."""
         return _attribute_value_dict(self._data, "Keys")
 
--- a/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
+++ b/aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
@@ -182,10 +182,10 @@ class StreamRecord(DictWrapper):
         item = self.get("ApproximateCreationDateTime")
         return None if item is None else int(item)
 
-    # This override breaks the Mapping protocol of DictWrapper, it's left here for backwards compatibility with
-    # a 'type: ignore' comment. This is currently the only subclass of DictWrapper that breaks this protocol.
+    # NOTE: This override breaks the Mapping protocol of DictWrapper, it's left here for backwards compatibility with
+    # a 'type: ignore' comment. See #1516 for discussion
     @property
-    def keys(self) -> Optional[Dict[str, AttributeValue]]:  # type: ignore
+    def keys(self) -> Optional[Dict[str, AttributeValue]]:  # type: ignore[override]
         """The primary key attribute(s) for the DynamoDB item that was modified."""
         return _attribute_value_dict(self._data, "Keys")
 
--- a/tests/functional/test_data_classes.py
+++ b/tests/functional/test_data_classes.py
@@ -74,6 +74,7 @@ from aws_lambda_powertools.utilities.data_classes.dynamo_db_stream_event import 
     AttributeValueType,
     DynamoDBRecordEventName,
     DynamoDBStreamEvent,
+    StreamRecord,
     StreamViewType,
 )
 from aws_lambda_powertools.utilities.data_classes.event_source import event_source
@@ -99,6 +100,19 @@ def test_dict_wrapper_equals():
     assert data1 is not DataClassSample(data1)
 
     assert DataClassSample(data1).raw_event is data1
+
+
+def test_dict_wrapper_imlements_mapping():
+    class DataClassSample(DictWrapper):
+        pass
+
+    data = {"message": "foo1"}
+    dcs = DataClassSample(data)
+    assert len(dcs) == len(data)
+    assert list(dcs) == list(data)
+    assert dcs.keys() == data.keys()
+    assert list(dcs.values()) == list(data.values())
+    assert dcs.items() == data.items()
 
 
 def test_cloud_watch_dashboard_event():
@@ -615,6 +629,23 @@ def test_dynamo_attribute_value_type_error():
         print(attribute_value.get_value)
     with pytest.raises(ValueError):
         print(attribute_value.get_type)
+
+
+def test_stream_record_keys_with_valid_keys():
+    attribute_value = {"Foo": "Bar"}
+    sr = StreamRecord({"Keys": {"Key1": attribute_value}})
+    assert sr.keys == {"Key1": AttributeValue(attribute_value)}
+
+
+def test_stream_record_keys_with_no_keys():
+    sr = StreamRecord({})
+    assert sr.keys is None
+
+
+def test_stream_record_keys_overrides_dict_wrapper_keys():
+    data = {"Keys": {"key1": {"attr1": "value1"}}}
+    sr = StreamRecord(data)
+    assert sr.keys != data.keys()
 
 
 def test_event_bridge_event():
--- a/tests/functional/test_data_classes.py
+++ b/tests/functional/test_data_classes.py
@@ -102,17 +102,17 @@ def test_dict_wrapper_equals():
     assert DataClassSample(data1).raw_event is data1
 
 
-def test_dict_wrapper_imlements_mapping():
+def test_dict_wrapper_implements_mapping():
     class DataClassSample(DictWrapper):
         pass
 
     data = {"message": "foo1"}
-    dcs = DataClassSample(data)
-    assert len(dcs) == len(data)
-    assert list(dcs) == list(data)
-    assert dcs.keys() == data.keys()
-    assert list(dcs.values()) == list(data.values())
-    assert dcs.items() == data.items()
+    event_source = DataClassSample(data)
+    assert len(event_source) == len(data)
+    assert list(event_source) == list(data)
+    assert event_source.keys() == data.keys()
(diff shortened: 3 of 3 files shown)


Tankanow:
Has something gone wild with the git history? I definitely did not change 52 files. =) Gimme a bit to fix.

rubenfonseca:
@Tankanow yes we had a small problem with a pipeline process. I think the best way is to reset your local repo from `develop` and cherry-pick your new commit on top. Thanks for the effort!

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -1,9 +1,9 @@
 import base64
 import json
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Iterator, Mapping, Optional
 
 
-class DictWrapper:
+class DictWrapper(Mapping):
```
Tankanow, line 7: NOTE: I thought for a long time whether or not to explicitly add the `Mapping` protocol here or to simply implement the missing dunder methods and allow duck-typing to carry the day. 

In the end, I decided it was better to explicitly add the `Mapping` protocol because, all things being equal, I think it is bad for subclasses of this class to break that protocol. It is good for `mypy` to warn when an implementer chooses to break the protocol. It should require an explicit `type: ignore` comment when choosing to break the protocol.

There is one existing class, `StreamRecord`, that breaks the protocol. There are no subclasses of `StreamRecord` and it is used only within the context of a `DynamoDBRecord` object. Because there is only one instance of this type breakage, and it is so localized, I marked `StreamRecord.keys` as `type: ignore` and added a comment and some tests to indicate that this class explicitly breaks the `Mapping` protocol.

Tankanow:
@rubenfonseca, this is ready for review. I added a review comment to describe one design choice. I'm excited for your feedback.

rubenfonseca:
Thank you so much for following up! After discussing this internally, we would like to move the discussion to our [v2 release](https://github.com/awslabs/aws-lambda-powertools-python/issues/1459).

We are concerned with the changes that this PR might introduce to consumers of `DictWrapper` and its many subclasses. We want to take time to better understand and document the eventual changes of this PR. The release of v2 is being tracked to the end of September.

Action: move PR base to v2

heitorlessa:
Adding to @rubenfonseca comments on our concern about the MappingInterface.

The way Event Source Data Classes was done it abuses the word "Dict", when in fact is a class. 

If we implement an __iter__ for instance, it wouldn't be true for most Event Source Data Classes as the data varies in format - e.g., DynamoDB keeps a list of records in data.records, while API GW keeps in data.body.

A long-term strategy is to get rid of the duplication of Event Source Data Classes and Parser (Pydantic). We're waiting for Pydantic v2 to be launched to analyse its package size -- doing so would improve speed (~17x faster), validation, and open the door to more easily implement strict typing for all let alone a single source of truth for Event Source Models.

We'd still love to hear from you if you feel strongly inclined in having the protocol implemented either way in v2

Tankanow:
Thanks for the response @heitorlessa. 

I respectfully disagree on this point:

> If we implement an **iter** for instance, it wouldn't be true for most Event Source Data Classes as the data varies in format - e.g., DynamoDB keeps a list of records in data.records, while API GW keeps in data.body.

The `event` is _always_ a dictionary so you can always iterate over its keys. I think you can rightly call the class `DictWrapper` to turn raw data into objects. In my humble opinion, classes wrapping the Lambda `event` should implement the Mapping interface (decorator pattern if you like) so that code can benefit from the Python standard library and data model without resorting to typing checking and access via `raw_event`.

To this point, in practice - that is in the entire code base - I could only find [1 subclass that breaks the interface](https://github.com/awslabs/aws-lambda-powertools-python/pull/1516/files#r969683429). It seems that there are no consumers that expect the behavior you are worried about, i.e. that **iter** should behave differently for different event subclasses.

Lastly, it is true that this is a breaking change for anyone using Powertools who has currently implemented a *negative* check on the `event` type in order to coerce things _back_ to a dict via `raw_event` if needed, as in this example:

```python
from aws_lambda_powertools.utilities.data_classes import APIGatewayProxyEvent
from typing import Mapping, Union

def handler(event: Union[Mapping, DictWrapper], context):
    event_dict = event.raw_event if not isinstance(event, Mapping) else event
    return event_dict
        
event_dict = {'foo': 'bar'}
event_obj = APIGatewayProxyEvent(event_dict)
assert handler(event_dict, None) == handler(event_obj, None)
```

I feel like it is a corner case. As I feel like most consumers would use the positive check `if isinstance(event, DictWrapper)` to ensure the event has the `.raw_event` property.

heitorlessa:
Ah, I missed an important detail on iter - you want to access the raw_event property (actual dict), and not the data property where a payload might be (records, body etc).

If I got that right I take it back what I alluded earlier with different subclasses, and I fully agree we should do that.

I don't have data points on whether people are checking the superclass instance (DictWrapper) as opposed to the Event Source Class itself like we do even internally -- for the sake of Hyrum's law as customers might have taken a dependency on this, we should do it safely for V2 and document in the upgrade guide.

Let me know if I still haven't processed this correctly

Tankanow:
@heitorlessa,

> If I got that right I take it back what I alluded earlier with different subclasses, and I fully agree we should do that.

Yes!

> - for the sake of Hyrum's law 

I totally agree. I regret missing this when I opened this PR against V1. I'm glad you and @rubenfonseca caught it. I love your commitment to non-breaking changes as even breaking changes in a major release can be really frustrating to users. I would be happy to help V2 by testing against our code bases. Please let me know how I can help.

heitorlessa:
Perfect, we have a plan then.

Could you create a PR implementing the mapping interface against the v2 branch? 

We can take specific questions, if any, about implementation and the messaging on upgrade guide.

@rubenfonseca @leandrodamascena and I can support you 

Thank you!!!

Tankanow:
@heitorlessa and @rubenfonseca,

Regarding this request:

> Could you create a PR implementing the mapping interface against the v2 branch?

It looks like someone moved this PR destination to `awslabs:v2`. Is there any action item for me on this ticket?

heitorlessa:
@Tankanow we'll be looking into the PR this week ;) I mistakenly added a comment in the PR asking to create a PR :D, I'll be flying back home tomorrow, so EOW in the worst case.

heitorlessa:
Looking into this now...

aws_lambda_powertools/utilities/data_classes/common.py
heitorlessa: In older days I'd ask to remove it as we're breaking it in `StreamRecord` - I'd take this as an instance of `Practicality Beats Purity`.

It's highly unlikely a customer will use `.keys` to get the keys of a `StreamRecord` dict given the intent. However, if we do receive any customer complaint we'll consider removing the subclassing.

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -1,9 +1,9 @@
 import base64
 import json
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Iterator, Mapping, Optional
```
heitorlessa:
```suggestion
from collections.abc import Mapping
from typing import Any, Dict, Iterator, Optional
```

Because your intent is to apply the protocol.

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -182,8 +182,10 @@ def approximate_creation_date_time(self) -> Optional[int]:
         item = self.get("ApproximateCreationDateTime")
         return None if item is None else int(item)
 
+    # This override breaks the Mapping protocol of DictWrapper, it's left here for backwards compatibility with
+    # a 'type: ignore' comment. This is currently the only subclass of DictWrapper that breaks this protocol.
     @property
-    def keys(self) -> Optional[Dict[str, AttributeValue]]:
+    def keys(self) -> Optional[Dict[str, AttributeValue]]:  # type: ignore
```
heitorlessa:
```suggestion
    def keys(self) -> Optional[Dict[str, AttributeValue]]:  # type: ignore[override]
```

It's best to ignore a specific mypy error.

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -182,8 +182,10 @@ def approximate_creation_date_time(self) -> Optional[int]:
         item = self.get("ApproximateCreationDateTime")
         return None if item is None else int(item)
 
+    # This override breaks the Mapping protocol of DictWrapper, it's left here for backwards compatibility with
```
heitorlessa:
```suggestion
    # NOTE: This override breaks the Mapping protocol of DictWrapper, it's left here for backwards compatibility with
```

Capital `NOTE` helps highlight in IDE when looking for notes worthy of attention

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -182,8 +182,10 @@ def approximate_creation_date_time(self) -> Optional[int]:
...
+    # a 'type: ignore' comment. This is currently the only subclass of DictWrapper that breaks this protocol.
```
heitorlessa:
```suggestion
    # a 'type: ignore' comment. See #1516 for discussion
```

This is true atm of writing but may not be. Minor update to point to PR instead if one needs more info.

tests/functional/test_data_classes.py
```diff
@@ -101,6 +102,19 @@ def message(self) -> str:
     assert DataClassSample(data1).raw_event is data1
 
 
+def test_dict_wrapper_imlements_mapping():
```
heitorlessa:
```suggestion
def test_dict_wrapper_implements_mapping():
```

tests/functional/test_data_classes.py
```diff
@@ -101,6 +102,19 @@ def message(self) -> str:
...
+    class DataClassSample(DictWrapper):
+        pass
+
+    data = {"message": "foo1"}
+    dcs = DataClassSample(data)
+    assert len(dcs) == len(data)
+    assert list(dcs) == list(data)
+    assert dcs.keys() == data.keys()
+    assert list(dcs.values()) == list(data.values())
+    assert dcs.items() == data.items()
```
heitorlessa:
```suggestion
    event_source = DataClassSample(data)
    assert len(event_source) == len(data)
    assert list(event_source) == list(data)
    assert event_source.keys() == data.keys()
    assert list(event_source.values()) == list(data.values())
    assert event_source.items() == data.items()
```

np: better describes intent. Over time, we tend to forget those short letter variable names. Any contributor with OCD or ADHD might unintentionally obsess over its meaning too.

tests/functional/test_data_classes.py
```diff
@@ -617,6 +631,23 @@ def test_dynamo_attribute_value_type_error():
...
+
+def test_stream_record_keys_with_no_keys():
+    sr = StreamRecord({})
+    assert sr.keys is None
+
+
+def test_stream_record_keys_overrides_dict_wrapper_keys():
+    data = {"Keys": {"key1": {"attr1": "value1"}}}
+    sr = StreamRecord(data)
+    assert sr.keys != data.keys()
```
heitorlessa:
```suggestion
def test_stream_record_keys_with_valid_keys():
    attribute_value = {"Foo": "Bar"}
    record = StreamRecord({"Keys": {"Key1": attribute_value}})
    assert record.keys == {"Key1": AttributeValue(attribute_value)}


def test_stream_record_keys_with_no_keys():
    record = StreamRecord({})
    assert record.keys is None


def test_stream_record_keys_overrides_dict_wrapper_keys():
    data = {"Keys": {"key1": {"attr1": "value1"}}}
    record = StreamRecord(data)
    assert record.keys != data.keys()
```

np: better describes intent. Over time, we tend to forget those short letter variable names. Any contributor with OCD or ADHD might unintentionally obsess over its meaning too.

heitorlessa (changes requested):
Looks great @Tankanow - a minor note on subclassing (practicality vs purity) with our intent to keep but will revert if it affects any use case we couldn't account for, and code suggestions to improve maintenance.

We can merge once these are in ;)

Thank you!!!!

rubenfonseca:
Just re-checking this PR, I agree that we should merge this into `develop` instead :)

heitorlessa (approved):

heitorlessa:
Thanks for addressing the feedback @Tankanow - merging now. THANK YOU <3

Issue #1330, opened 2022-07-21
Computation of idempotency hash key using unqualified name 


dispyfree:
### Expected Behaviour

The idempotency layer should be using the fully qualified name of the input function to compute the hash key.
Using unqualified names may result in two distinct functions and/or methods contending for the same set of idempotency items. 

### Current Behaviour

In [IdempotencyHandler](https://github.com/awslabs/aws-lambda-powertools-python/blob/a320dddf495b0f83f9bf074075aa0f4205859b69/aws_lambda_powertools/utilities/idempotency/base.py#L42), line 77:

`persistence_store.configure(config, self.function.__name__)`

### Possible Solution

`persistence_store.configure(config, self.function.__qualname__)`

Note that it should be checked whether `__qualname__` is subject to Python module import peculiarities. In particular, whether `import module` and `from module import funcName` yield the same `__qualname__`. 

### Steps to Reproduce

```python
class A:
    @idempotent_function
    def test(self, myParam):
        return myParam
    

class B(A):
    
    @idempotent_function
    def test(self, myParam):
        return myParam
```

### AWS Lambda Powertools for Python version

latest

### AWS Lambda function runtime

3.9

### Packaging format used

PyPi

heitorlessa:
Thanks for flagging this @dispyfree - we indeed use qualified name [elsewhere in the code](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/middleware_factory/factory.py#L122) but not in the idempotency. According to the [PEP 3155](https://peps.python.org/pep-3155/), it excludes the module name so the result would be the same regardless of how it was imported.

Would you like to send a fix as a first contributor? Alternatively, we can also fix it ourselves and give you credit as always.

cc @rubenfonseca 

Thank you!

heitorlessa:
Assigning to me so I can push a fix and make a patch release tomorrow

dispyfree:
That's an interesting bit @heitorlessa.
So using `__qualname__` will fix it with respect to nesting inside of
- functions
- classes 

By excluding the module, however, we can - in theory - still have two _distinct_ functions contest for the same idempotency items. They just bear the same name, have the same parameters and sit in two modules.
Would it make sense to explicitly include `__module__` as well? By definition, this is the module that the function was originally defined in, so it is not subject to module aliasing, monkey patching and other dark magic. 

[...]

PR #1535, opened 2022-09-21
feat(idempotency): support methods with the same name (ABCs) by including fully qualified name in v2

leandrodamascena:
**Issue number:** #1330 

### Changes

This change modifies the name of the key generated by the idempotency methods. In case you have [2 functions](https://github.com/awslabs/aws-lambda-powertools-python/issues/1330#issuecomment-1192054984) with the same name, the hash is calculated only for the first one and the other one cannot use the idempotency function.

### User experience

Before to this change, the idempotency key was just using the function name:
image

After this change, the idempotency key includes the module and Python qualified name ([PEP 3155](https://peps.python.org/pep-3155/))
image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered docs/utilities/idempotency.md](https://github.com/leandrodamascena/aws-lambda-powertools-python/blob/feat/idempotency-qual-name/docs/utilities/idempotency.md)


tests/e2e/idempotency/test_idempotency_dynamodb.py
```diff
@@ -0,0 +1,30 @@
...
+
+@pytest.fixture
+def basic_handler_fn_arn(infrastructure: dict) -> str:
+    return infrastructure.get("BasicHandlerArn", "")
+
+
+def test_basic_idempotency_record(basic_handler_fn_arn: str, basic_handler_fn: str):
+    # GIVEN
+    function_name = "basic_handler.lambda_handler"
+    table_name = "IdempotencyTable"
```
rubenfonseca: Can we get this from the CloudFormation output?

```suggestion

@pytest.fixture
def idempotency_table_name(infrastructure: dict) -> str:
    return infrastructure.get("DynamoDBTable", "")

def test_basic_idempotency_record(basic_handler_fn_arn: str, basic_handler_fn: str, idempotency_table_name: str):
    # GIVEN
    function_name = "basic_handler.lambda_handler"
    table_name = idempotency_table_name
```

tests/e2e/utils/data_fetcher/idempotency.py
```diff
@@ -0,0 +1,57 @@
+import boto3
+from retry import retry
+
+
+class DynamoDB:
```
rubenfonseca: Can we call it DynamoDBFetcher?

```suggestion
class DynamoDBFetcher:
```

tests/e2e/utils/data_fetcher/idempotency.py
```diff
@@ -0,0 +1,57 @@
...
+        Name of Lambda function to fetch dynamodb record
+    table_name : str
+            Name of DynamoDB table
+
+    Returns
+    -------
+    DynamoDB
+        DynamoDB instance with dynamodb record
+    """
+    return DynamoDB(function_name=function_name, table_name=table_name)
```
rubenfonseca: I was comparing this with the other data fetchers and notice two things:

- the other utilities actually fetch the data on the constructor. The rest of the methods (like `get_records` in your case) just return data that is already in memory. Do you think we can do the same here?
- another option, since this is really simple, maybe we don't need the intermediate `DynamoDBFetcher` for now, and just go straight to the fetching of the item (like we do on `data_fetcher.metrics`)

What do you think? Can we simplify it?

rubenfonseca (changes requested):
It seems the Lambda function is lacking permissions to RW to DDB.

Great job!  I left some comments, but I would like to share some ideas about improving the e2e tests

tests/e2e/idempotency/test_idempotency_dynamodb.py
leandrodamascena: Done

tests/e2e/utils/data_fetcher/idempotency.py
leandrodamascena: I removed this class. As you said: this is very simple to use an intermediate class.

tests/e2e/utils/data_fetcher/idempotency.py
leandrodamascena: Thanks for this feedback @rubenfonseca! Thinking simple is very good. I used the second option, this is really simple.

leandrodamascena:
Actually I was missing the RW Permissions for DDB, but I'll investigate why I'm running fine here even without this permission.

tests/e2e/idempotency/test_idempotency_dynamodb.py
```diff
@@ -0,0 +1,97 @@
...
+    arguments = {"lambda_arn": parallel_execution_handler_fn_arn}
+
+    # WHEN
+    # executing Lambdas in parallel
+    execution_result_list = execute_lambdas_in_parallel(
+        [data_fetcher.get_lambda_response, data_fetcher.get_lambda_response], arguments
+    )
+
+    timeout_execution_response = execution_result_list[0][0]["Payload"].read().decode("utf-8")
+    error_idempotency_execution_response = execution_result_list[1][0]["Payload"].read().decode("utf-8")
```
rubenfonseca, line 92: I'm a little bit worried about the concurrency here. If things are really executed in parallel, what guarantees us that the first function is the one that is going to fail? Has it ever happened during your tests that they randomly fail?

rubenfonseca (approved):
This is SOOOOOOOOOO awesome Leandro, fantastic job! Just left a slight note for us to think about, but maybe it's not a problem.

Should we also update the `docs/upgrade.md` explaining the breaking change?

leandrodamascena:
> This is SOOOOOOOOOO awesome Leandro, fantastic job! Just left a slight note for us to think about, but maybe it's not a problem.
> 
> Should we also update the `docs/upgrade.md` explaining the breaking change?

Heyyyy @rubenfonseca, thank you so much for this feedback, I appreciate a lot! I like the way you encourage the work of others! :v: :rocket: 

I agree 100% that we must update the documentation and I did. Please review this and add what you think is important.

tests/e2e/idempotency/test_idempotency_dynamodb.py
leandrodamascena: Ruben, I was also concerned and did some additional research on this. So I changed `submit` to `map` because using map we can keep the order of inputs and returns even if it runs in parallel. 

I think it is now resilient now.

Thank you for pointing this.

References:
https://pymotw.com/3/concurrent.futures/
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map
https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

rubenfonseca (approved):
image

Issue #1527, opened 2022-09-20
Feature request: Simplified logger formatter when in AWS SAM Local


pmarko1711:
### Use case

I use `aws_lambda_powertools.Logger` which works very nicely when looking at logs in Cloudwatch. 
Nevertheless, when testing/debugging lambda functions using AWS SAM local invoke, I find the json representation difficult to read and cluttering the console too much. 

I know I can swap the logger, or use a custom formatter. But would it not be meaningful to include a simple formatter (instead of the `LambdaPowertoolsFormatter` that prints out jsons) and use it automatically when not really in Lambda (but e.g. in AWS SAM local invoke) ?

Related issue: #409

Note:
- Strictly speaking this request might breach your tenet 1 ("AWS Lambda only"), but it depends where you place AWS SAM local invoke or development & testing of Lambda functions outside Lambda

### Solution/User Experience

- Add a "simple" formatter that would print a single line, without serializing to json
   - optionally with a custom format
- With that, use the existing json formatter only when really in Lambda
   - or alternatively, use the simple one when in AWS SAM Local (and similar, if it can be detected).
   - or make it easy to swap, by e.g. an envvar

### Alternative solutions

```markdown
There exists a solution already and that is to use a custom formatter. 

I opened the issue to consider building this in thinking that it might be beneficial for more users developing and testing Lambda functions.
```

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

heitorlessa:
Hey @pmarko1711, thanks a lot for taking the time to help us improve! Would you have the output handy to make it easier to reason the next steps?

I need to spin up a project to test. I can't remember if it's an one line JSON (cost optimization), or indented -- would it be more helpful to have indented JSON when running locally if that's not the case yet?

My initial concern: transparently swapping for a no JSON output could lead to incorrect assumptions how the final output would be when deployed.

Look forward to hearing from you

heitorlessa:
### Example

Demonstrating how it could look like without any code change.

#### Before

```python
sam local invoke     
Invoking app.lambda_handler (python3.9)
[...]

PR #1548, opened 2022-09-27
feat(logger): pretty-print JSON when POWERTOOLS_DEV is set

pmarko1711:
**Issue number: #1527**

### Changes

* Logger's default serialize function `json.dumps` will use indentation by four spaces when in AWS SAM Local. Detection happens using [`AWS_SAM_LOCAL` env var](https://github.com/aws/aws-sam-cli/commit/5e56d94c22d71b1ae2a18057ba9182372cade11b) (if `=="true"`).

### User experience

There should be no impact unless in AWS SAM Local. 

In AWS SAM Local, logs are output to console. To make the serialized json messages more readable and consequently debugging with AWS SAM Local easier, logger's default formatter will in this case automatically use indentation by four spaces. That means that messages normally serialized as
```
{"level": "INFO", "location": "lambda_handler:39", "message": "Some log message...", "timestamp": "2022-09-26 09:03:38,349+0000", "service": "issue-1527"}
```
will be in AWS SAM Local by default indented by four spaces:
```
{
    "level": "INFO",
    "location": "lambda_handler:39",
    "message": "Some log message...",
    "timestamp": "2022-09-26 09:02:05,262+0000",
    "service": "issue-1527"
}
This is to make debugging easier
```

## Checklist

    * (only using pytest)
* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered docs/core/logger.md](https://github.com/pmarko1711/aws-lambda-powertools-python/blob/develop/docs/core/logger.md)

### DIFF
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -9,6 +9,7 @@ from functools import partial
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Union
 
 from ..shared import constants
+from ..shared.functions import strtobool
 
 RESERVED_LOG_ATTRS = (
     "name",
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -111,9 +111,15 @@ class LambdaPowertoolsFormatter(BasePowertoolsFormatter):
             Key-value to be included in log messages
 
         """
+
         self.json_deserializer = json_deserializer or json.loads
         self.json_default = json_default or str
-        self.json_serializer = json_serializer or partial(json.dumps, default=self.json_default, separators=(",", ":"))
+        self.json_indent = (
+            4 if os.getenv("AWS_SAM_LOCAL", "").lower() == "true" else None
+        )  # indented json serialization when in AWS SAM Local
+        self.json_serializer = json_serializer or partial(
+            json.dumps, default=self.json_default, separators=(",", ":"), indent=self.json_indent
+        )
 
         self.datefmt = datefmt
         self.use_datetime_directive = use_datetime_directive
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -115,7 +115,7 @@ class LambdaPowertoolsFormatter(BasePowertoolsFormatter):
         self.json_deserializer = json_deserializer or json.loads
         self.json_default = json_default or str
         self.json_indent = (
-            constants.PRETTY_INDENT if os.getenv("POWERTOOLS_DEV", "").lower() == "true" else constants.COMPACT_INDENT
+            constants.PRETTY_INDENT if strtobool(os.getenv("POWERTOOLS_DEV", "0")) else constants.COMPACT_INDENT
         )  # indented json serialization when in AWS SAM Local
         self.json_serializer = json_serializer or partial(
             json.dumps, default=self.json_default, separators=(",", ":"), indent=self.json_indent
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -115,7 +115,7 @@ class LambdaPowertoolsFormatter(BasePowertoolsFormatter):
         self.json_deserializer = json_deserializer or json.loads
         self.json_default = json_default or str
         self.json_indent = (
-            4 if os.getenv("AWS_SAM_LOCAL", "").lower() == "true" else None
+            constants.PRETTY_INDENT if os.getenv("POWERTOOLS_DEV", "").lower() == "true" else constants.COMPACT_INDENT
         )  # indented json serialization when in AWS SAM Local
         self.json_serializer = json_serializer or partial(
             json.dumps, default=self.json_default, separators=(",", ":"), indent=self.json_indent
--- a/aws_lambda_powertools/shared/constants.py
+++ b/aws_lambda_powertools/shared/constants.py
@@ -32,3 +32,6 @@ LOGGER_LAMBDA_CONTEXT_KEYS = [
     "cold_start",
     "xray_trace_id",
 ]
+
+PRETTY_INDENT: int = 4
+COMPACT_INDENT = None
--- a/aws_lambda_powertools/shared/constants.py
+++ b/aws_lambda_powertools/shared/constants.py
@@ -33,5 +33,6 @@ LOGGER_LAMBDA_CONTEXT_KEYS = [
     "xray_trace_id",
 ]
 
+# JSON indentation level
 PRETTY_INDENT: int = 4
 COMPACT_INDENT = None
--- a/docs/core/logger.md
+++ b/docs/core/logger.md
@@ -369,6 +369,9 @@ If you prefer configuring it separately, or you'd want to bring this JSON Format
 | **`log_record_order`**       | set order of log keys when logging                                                                                       | `["level", "location", "message", "timestamp"]`               |
 | **`kwargs`**                 | key-value to be included in log messages                                                                                 | `None`                                                        |
 
+???+ info
+    When `POWERTOOLS_DEV` env var is present and set to `"true"`, Logger's default serializer (`json.dumps`) will pretty-print log messages for easier readability.
+
 ```python hl_lines="2 7-8" title="Pre-configuring Lambda Powertools Formatter"
 --8<-- "examples/logger/src/powertools_formatter_setup.py"
 ```
@@ -554,9 +557,6 @@ As parameters don't always translate well between them, you can pass any callabl
 ```python hl_lines="1 3 7-8 13" title="Using Rust orjson library as serializer"
 --8<-- "examples/logger/src/bring_your_own_json_serializer.py"
 ```
-
-???+ info
-    When your code runs in [AWS SAM local invoke](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-cli-command-reference-sam-local-invoke.html) (marked by the `AWS_SAM_LOCAL` env var), Logger's default `json.dumps` will apply indentation by four spaces.
 
 ## Testing your code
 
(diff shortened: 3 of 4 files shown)


aws_lambda_powertools/logging/formatter.py
```diff
@@ -111,9 +111,15 @@ def __init__(
             Key-value to be included in log messages
 
         """
+
         self.json_deserializer = json_deserializer or json.loads
         self.json_default = json_default or str
-        self.json_serializer = json_serializer or partial(json.dumps, default=self.json_default, separators=(",", ":"))
+        self.json_indent = (
+            4 if os.getenv("AWS_SAM_LOCAL", "").lower() == "true" else None
```
heitorlessa: Suggestion: Use a constant to ease maintaining this later, e.g.

```suggestion
            PRETTY_INDENT if os.getenv("AWS_SAM_LOCAL", "").lower() == "true" else COMPACT_INDENT
```

heitorlessa (changes requested):
Thank you so much for kicking the tires here. Made one suggestion, and I have one important note: let's use `POWERTOOLS_DEV` instead of `AWS_SAM_LOCAL`, I'll create a RFC today.

During maintainers sync, the TypeScript team also had [customers for the same request ](https://github.com/awslabs/aws-lambda-powertools-typescript/issues/751) except that they want to use `node <app.ts>`, or Amplify-CLI, which makes this scenario muddy.

We thought of a new environment variable `POWERTOOLS_DEV` to detect intent on overriding decisions like this and others in the future (e.g., Event Handler has a debug mode which we can deprecate in favour of this in v2). This helps provide a consistent experience across Powertools languages, document behaviours this would enable upfront in our docs, and avoid a proliferation of multiple env vars.

This also futureproof a potential move to AWS Powertools where we could support[1] both Lambda, Fargate, etc.

If your're busy, I'm happy to contribute these changes directly to your fork.

[1] Despite our tenets and name, we've been told by customers who use Powertools in Glue jobs, Fargate, and even IoT; classic [Hyrum's law](https://www.hyrumslaw.com/).

pmarko1711:
OK, I will change this later today. Is there a good place to put the constants in, pls?

heitorlessa:
> OK, I will change this later today. Is there a good place to put the constants in, pls?

Yes! `shared/constants.py` is where we keep them.

heitorlessa:
We're making progress in the RFC but this might be of interest if others agree this is a clearer intent:

https://github.com/awslabs/aws-lambda-powertools/issues/86#issuecomment-1261126294

pmarko1711:
OK, I changed everything to `POWERTOOLS_DEV` and also renamed the test functions and this PR, @heitorlessa .

Since this env var will have multiple effects, shouldn't it be explained in a more prominent place in the documentation? For now I only put an info note in the page about logger. 

Maybe it's not for this PR though.

heitorlessa:
thanks a lot @pmarko1711 and yes, I'll create a separate PR [documenting the effects](https://github.com/awslabs/aws-lambda-powertools/issues/86#issuecomment-1262485237).

I'll look into the PR as soon as I complete another task today, hopefully by EOD the latest. With your permission, I might push changes directly to your fork to speed up merging this change (in case I have delays with my primary task) - I'll share feedback nevertheless.

pmarko1711:
ok, yes, no problem @heitorlessa . If you need me to set something (permissions wise) on my fork, let me know.

heitorlessa:
Looking at it now ;) Appreciate the patience very much

aws_lambda_powertools/logging/formatter.py
```diff
@@ -111,9 +111,15 @@ def __init__(
             Key-value to be included in log messages
 
         """
+
         self.json_deserializer = json_deserializer or json.loads
         self.json_default = json_default or str
-        self.json_serializer = json_serializer or partial(json.dumps, default=self.json_default, separators=(",", ":"))
+        self.json_indent = (
+            constants.PRETTY_INDENT if os.getenv("POWERTOOLS_DEV", "").lower() == "true" else constants.COMPACT_INDENT
```
heitorlessa: Tip: use [`strtobool`](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/shared/functions.py#L9) as that accounts for variations like `"1"`, `"0"`, etc..

```suggestion
            constants.PRETTY_INDENT if strtobool(os.getenv("POWERTOOLS_DEV", "0")) else constants.COMPACT_INDENT
```

You'll need to import it like

```
from aws_lambda_powertools.shared.functions import strtobool
```

aws_lambda_powertools/shared/constants.py
```diff
@@ -32,3 +32,6 @@
     "cold_start",
     "xray_trace_id",
 ]
+
+PRETTY_INDENT: int = 4
+COMPACT_INDENT = None
```
heitorlessa:
```suggestion

# JSON indentation level
PRETTY_INDENT: int = 4
COMPACT_INDENT = None
```

Added a quick comment so we don't forget in the future.

docs/core/logger.md
```diff
@@ -369,6 +369,9 @@ If you prefer configuring it separately, or you'd want to bring this JSON Format
 | **`log_record_order`**       | set order of log keys when logging                                                                                       | `["level", "location", "message", "timestamp"]`               |
 | **`kwargs`**                 | key-value to be included in log messages                                                                                 | `None`                                                        |
 
+???+ info
+    When `POWERTOOLS_DEV` env var is present and set to `"true"`, Logger's default serializer (`json.dumps`) will pretty-print log messages for easier readability.
```
heitorlessa, line 373: Great wording!

tests/functional/test_logger_powertools_formatter.py
```diff
@@ -288,3 +290,22 @@ def test_log_formatting(stdout, service_name):
 
     # THEN the formatting should be applied (NB. this is valid json, but hasn't be parsed)
     assert log_dict["message"] == '["foo bar 123 [1, None]", null]'
+
+
+def test_log_json_indent_compact_indent(stdout, service_name, monkeypatch):
+    # GIVEN a logger with default settings and WHEN POWERTOOLS_DEV is not set
+    if "POWERTOOLS_DEV" in os.environ:
+        monkeypatch.delenv(name="POWERTOOLS_DEV")
```
heitorlessa:
```suggestion
    monkeypatch.delenv(name="POWERTOOLS_DEV", raising=False)
```

[`delenv`](https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest.MonkeyPatch.delenv) supports a flag to not raise an exception if the env var doesn't exist.

tests/functional/test_logger_powertools_formatter.py
```diff
@@ -288,3 +290,22 @@ def test_log_formatting(stdout, service_name):
...
+    logger = Logger(service=service_name, stream=stdout)
+    logger.info("Test message")
+    # THEN the json should not be indented using constant.PRETTY_INDENT blank spaces
+    assert " " * constants.PRETTY_INDENT not in stdout.getvalue()
```
heitorlessa:
```suggestion
    new_lines = stdout.getvalue().count("\n")
    assert new_lines == 1
```

Tip: non-indented log lines should have a single `\n`. While this assertion works for this test, it isn't always correct in the `test_log_json_pretty_indent` test because `assert " " * 0 in stdout.getvalue()` is also True.

tests/functional/test_logger_powertools_formatter.py
```diff
@@ -288,3 +290,22 @@ def test_log_formatting(stdout, service_name):
...
+
+
+def test_log_json_pretty_indent(stdout, service_name, monkeypatch):
+    # GIVEN a logger with default settings and WHEN POWERTOOLS_DEV=="true"
+    monkeypatch.setenv(name="POWERTOOLS_DEV", value="true")
+    logger = Logger(service=service_name, stream=stdout)
+    logger.info("Test message")
+    # THEN the json should contain indentation (of constant.PRETTY_INDENT blank spaces)
+    assert " " * constants.PRETTY_INDENT in stdout.getvalue()
```
heitorlessa:
```suggestion
    new_lines = stdout.getvalue().count("\n")
    assert new_lines > 1
```

See my earlier comment. You could also do `assert new_lines == 7` if you want to be 100% precise, but it's brittle if we add additional keys.

heitorlessa (changes requested):
Nice work!! I love the wording - straight to the point.

Made a few suggestions to improve test accuracy, and use our `strtobool` to support cases where customers may use boolean strings like `0` or `1`

docs/core/logger.md
pmarko1711: I am not a native English speaker, so many thanks

tests/functional/test_logger_powertools_formatter.py
pmarko1711: Hmm, at first I thought the same but then this test failed. For some reason,there was one new line (but I think my message at the time was two words, "Test message" and for some reason it got split in the middle.)

pmarko1711: also, should we not rather use `os.linesep`  instead of `\n`?

pmarko1711: I've commited your suggestion, but using `os.linesep`

tests/functional/test_logger_powertools_formatter.py
pmarko1711: I've commited your suggestion, but using os.linesep

pmarko1711:
I've committed all your suggestions, @heitorlessa , with a small change - I counted `os.linesep` instead of `\n` in the functional tests

heitorlessa:
Looking into now ;) Thank you so much for these changes!

heitorlessa (approved):

heitorlessa:
Looks great @pmarko1711 - thank you so much for working throughout the review process. Merging now.

**Next steps**: I'll **(1)** update the RFC with the latest info on `POWERTOOLS_DEV` and `POWERTOOLS_DEBUG` (separate feat altogether), **(2)** create a PR to update the documentation on these flags, and **(3)** release this feature by the end of the week.

pmarko1711:
thanks for all the support, @heitorlessa

Issue #1555, opened 2022-09-29
Feature request: Kinesis Data Firehose event envelope and parser model


ran-isenberg:
### Use case

Similar to https://github.com/awslabs/aws-lambda-powertools-python/issues/1539 but for the parser utility
Kinesis Firehose supports data transformation via Lambda. The event is a standard format, documented [here](https://docs.aws.amazon.com/lambda/latest/dg/services-kinesisfirehose.html).

Ref: [Kinesis Data Transformation](https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html)
Example event from [aws-lambda-go sdk](https://github.com/aws/aws-lambda-go/blob/v1.34.1/events/firehose.go#L14)

### Solution/User Experience

Add kinesis firehose envelope and model.

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

ran-isenberg:
see https://github.com/awslabs/aws-lambda-powertools-python/pull/1556

PR #1556, opened 2022-09-29
feat(parser): add KinesisFirehoseModel

ran-isenberg:
**Issue number:** #1555 

## Summary
Add Kinesis Data Firehose event envelope and parser model

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered docs/utilities/parser.md](https://github.com/ran-isenberg/aws-lambda-powertools-python/blob/firehose/docs/utilities/parser.md)


aws_lambda_powertools/shared/functions.py
```diff
@@ -67,7 +67,7 @@ def resolve_env_var_choice(
 
 def base64_decode(value: str) -> bytes:
     try:
-        logger.debug("Decoding base64 Kafka record item before parsing")
+        logger.debug("Decoding base64 record item before parsing")
```
ran-isenberg, line 74: removed the kafka from the shared function

aws_lambda_powertools/utilities/parser/envelopes/kinesis.py
```diff
@@ -16,7 +16,7 @@ class KinesisDataStreamEnvelope(BaseEnvelope):
     Regardless of its type it'll be parsed into a BaseModel object.
 
     Note: Records will be parsed the same way so if model is str,
-    all items in the list will be parsed as str and npt as JSON (and vice versa)
+    all items in the list will be parsed as str and not as JSON (and vice versa)
```
ran-isenberg, line 19: typo :)

aws_lambda_powertools/utilities/parser/models/kinesis.py
```diff
@@ -19,11 +15,7 @@ class KinesisDataStreamRecordPayload(BaseModel):
 
     @validator("data", pre=True, allow_reuse=True)
     def data_base64_decode(cls, value):
-        try:
-            logger.debug("Decoding base64 Kinesis data record before parsing")
-            return base64.b64decode(value)
-        except (BinAsciiError, TypeError):
-            raise ValueError("base64 decode failed")
+        return base64_decode(value)
```
ran-isenberg, line 18: old kinesis improvement

heitorlessa:
@leandrodamascena we might want to do an E2E for this. Due to parser's nature, it has the highest risk of causing runtime disruption.

leandrodamascena:
Hii @ran-isenberg thanks for one more PR :smiley:.. I'm starting to work on it now.

@heitorlessa in fact adding e2e tests is a big improvement in the parser area. I just need to check how much data we need to generate in `Kinesis Firehose with PUT source` and `Kinesis Firehose with Kinesis Stream source` to trigger an event in Lambda. I think we should open a new PR to include e2e tests in `v2 branch` while this is not the official branch.

ran-isenberg:
+1 for E2E tests.
but on the flip side, it's going to make adding new services way harder for newcomers.

leandrodamascena:
> +1 for E2E tests. but on the flip side, it's going to make adding new services way harder for newcomers.

Yeah Ran, it makes sense and I understand your concern.. But for now we can write e2e tests when someone adds a new service, feature or even extends an existing feature and make life easier for new contributors.

ran-isenberg:
> > +1 for E2E tests. but on the flip side, it's going to make adding new services way harder for newcomers.
> 
> Yeah Ran, it makes sense and I understand your concern.. But for now we can write e2e tests when someone adds a new service, feature or even extends an existing feature and make life easier for new contributors.

You use CDK right?

leandrodamascena:
> You use CDK right?

Yeah, we use CDK for e2 testing. But I have some considerations to make about e2e testing before we think about any implementation on Parser.

- e2e tests will only work in v2, so in this PR we should not write end2end tests for the parser.
- We have a release date for v2 later this month and we are working hard for it.
- We believe that only the e2e tests are not enough in all cases. For e.g, if the parser fails for any change in the payload, we run a risk of breaking the function, because it is a runtime error. We need to write a RFC to think of a solution where we can update all lambda eventSources payload every X days/hours, because the payload can change on the service/lambda side and if we don't have the updated project it can be a problem. Maybe a separate project to deal with this, I'm not sure yet, but we'll start that discussion soon.

leandrodamascena (changes requested):
We need to update the code to account for differences between payloads.

Payload using Kinesis Firehose with a **Kinesis Stream as source**:

```
{
   "invocationId":"77c0024f-9869-4af0-951a-d83c102e8a63",
   "sourceKinesisStreamArn":"arn:aws:kinesis:us-east-1:200984112386:stream/kinesis-evento",
   "deliveryStreamArn":"arn:aws:firehose:us-east-1:200984112386:deliverystream/KDS-S3-10POr",
   "region":"us-east-1",
   "records":[
      {
         "recordId":"49633602406469139358835205474851822974501572780333465650000000",
         "approximateArrivalTimestamp":1664028793294,
         "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg==",
         "kinesisRecordMetadata":{
            "sequenceNumber":"49633602406469139358835205474851822974501572780333465650",
            "subsequenceNumber":0,
            "partitionKey":"0",
            "shardId":"shardId-000000000003",
            "approximateArrivalTimestamp":1664028793294
         }
      },
      {
         "recordId":"49633602406469139358835205474853031900321189264934043698000000",
         "approximateArrivalTimestamp":1664028820148,
         "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg==",
         "kinesisRecordMetadata":{
            "sequenceNumber":"49633602406469139358835205474853031900321189264934043698",
            "subsequenceNumber":0,
            "partitionKey":"0",
            "shardId":"shardId-000000000003",
            "approximateArrivalTimestamp":1664028820148
         }
      }
   ]
}
```

Payload using Kinesis Firehose with **Direct PUT**:
```
{
   "invocationId":"25326478-5c08-45ad-a34d-7342edbc942d",
   "deliveryStreamArn":"arn:aws:firehose:us-east-1:200984112386:deliverystream/PUT-S3-PCYyW",
   "region":"us-east-1",
   "records":[
      {
         "recordId":"49633602747871247603140515133584305740288993709252411394000000",
         "approximateArrivalTimestamp":1664029185290,
         "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg=="
      },
      {
         "recordId":"49633602747871247603140515133585514666108608544585547778000000",
         "approximateArrivalTimestamp":1664029186945,
         "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg=="
      }
   ]
}

```

We can go in the same direction as we went here #1499 and create separate classes for each type of payload.

ran-isenberg:
> > You use CDK right?
> 
> Yeah, we use CDK for e2 testing. But I have some considerations to make about e2e testing before we think about any implementation on Parser.
> 
> * e2e tests will only work in v2, so in this PR we should not write end2end tests for the parser.
> * We have a release date for v2 later this month and we are working hard for it.
> * We believe that only the e2e tests are not enough in all cases. For e.g, if the parser fails for any change in the payload, we run a risk of breaking the function, because it is a runtime error. We need to write a RFC to think of a solution where we can update all lambda eventSources payload every X days/hours, because the payload can change on the service/lambda side and if we don't have the updated project it can be a problem. Maybe a separate project to deal with this, I'm not sure yet, but we'll start that discussion soon.

The way I'd handle schemas is to put the models in a new repo and maintain proper versions BUT it should be managed and updated by the AWS services, not the community. They should be aware of how we use and model the events since sometime the docs are not good enough. 
If there are any unit/e2e tests, I'd expect the services to run them each time they make a change, some sort of  a  contract test. The E2E tests will run here per PR but you probably wont catch.

ran-isenberg:
> We need to update the code to account for differences between payloads.
> 
> Payload using Kinesis Firehose with a **Kinesis Stream as source**:
> 
> ```
> {
>    "invocationId":"77c0024f-9869-4af0-951a-d83c102e8a63",
>    "sourceKinesisStreamArn":"arn:aws:kinesis:us-east-1:200984112386:stream/kinesis-evento",
>    "deliveryStreamArn":"arn:aws:firehose:us-east-1:200984112386:deliverystream/KDS-S3-10POr",
>    "region":"us-east-1",
>    "records":[
>       {
>          "recordId":"49633602406469139358835205474851822974501572780333465650000000",
>          "approximateArrivalTimestamp":1664028793294,
>          "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg==",
>          "kinesisRecordMetadata":{
>             "sequenceNumber":"49633602406469139358835205474851822974501572780333465650",
>             "subsequenceNumber":0,
>             "partitionKey":"0",
>             "shardId":"shardId-000000000003",
>             "approximateArrivalTimestamp":1664028793294
>          }
>       },
>       {
>          "recordId":"49633602406469139358835205474853031900321189264934043698000000",
>          "approximateArrivalTimestamp":1664028820148,
>          "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg==",
>          "kinesisRecordMetadata":{
>             "sequenceNumber":"49633602406469139358835205474853031900321189264934043698",
>             "subsequenceNumber":0,
>             "partitionKey":"0",
>             "shardId":"shardId-000000000003",
>             "approximateArrivalTimestamp":1664028820148
>          }
>       }
>    ]
> }
> ```
> 
> Payload using Kinesis Firehose with **Direct PUT**:
> 
> ```
> {
>    "invocationId":"25326478-5c08-45ad-a34d-7342edbc942d",
>    "deliveryStreamArn":"arn:aws:firehose:us-east-1:200984112386:deliverystream/PUT-S3-PCYyW",
>    "region":"us-east-1",
>    "records":[
>       {
>          "recordId":"49633602747871247603140515133584305740288993709252411394000000",
>          "approximateArrivalTimestamp":1664029185290,
>          "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg=="
>       },
>       {
>          "recordId":"49633602747871247603140515133585514666108608544585547778000000",
>          "approximateArrivalTimestamp":1664029186945,
>          "data":"eyJ1c2VyX2lkIjoidXNlcjEiLCAic2NvcmUiOiAxMDB9Cg=="
>       }
>    ]
> }
> ```
> 
> We can go in the same direction as we went here #1499 and create separate classes for each type of payload.

I've modeled the metadata as optional so it works for both cases. The only reason to split them is if it's officially documented as such. do you think that's the case? I didnt find a mention that metadata params are applicable only for put events. If we add by ourselves, it's more likely to be broken in the future.
WDYT?

leandrodamascena:
> The way I'd handle schemas is to put the models in a new repo and maintain proper versions BUT it should be managed and updated by the AWS services, not the community. They should be aware of how we use and model the events since sometime the docs are not good enough. If there are any unit/e2e tests, I'd expect the services to run them each time they make a change, some sort of a contract test. The E2E tests will run here per PR but you probably wont catch.

I'll open an issue and add this discussion in there, so we can keep focus here in this PR. ok?

ran-isenberg:
> > The way I'd handle schemas is to put the models in a new repo and maintain proper versions BUT it should be managed and updated by the AWS services, not the community. They should be aware of how we use and model the events since sometime the docs are not good enough. If there are any unit/e2e tests, I'd expect the services to run them each time they make a change, some sort of a contract test. The E2E tests will run here per PR but you probably wont catch.
> 
> I'll open an issue and add this discussion in there, so we can keep focus here in this PR. ok?

No problem.

aws_lambda_powertools/utilities/parser/models/kinesis_firehose.py
```diff
@@ -9,15 +8,15 @@
 class KinesisFirehoseRecordMetadata(BaseModel):
     shardId: str
     partitionKey: str
-    approximateArrivalTimestamp: str
+    approximateArrivalTimestamp: int
```
leandrodamascena: Even in [documentation](https://docs.aws.amazon.com/lambda/latest/dg/services-kinesisfirehose.html) the field is a string, I created again a kinesis -> kinesis firehose -> lambda and in the service payload this is not a string, this a timestamp field.

image

I'll open a ticket to the documentation team to fix that.

aws_lambda_powertools/utilities/parser/models/kinesis_firehose.py
```diff
@@ -9,15 +8,15 @@
...
     sequenceNumber: str
     subsequenceNumber: str
 
 
 class KinesisFirehoseRecord(BaseModel):
     data: Union[bytes, Type[BaseModel]]  # base64 encoded str is parsed into bytes
     recordId: str
-    approximateArrivalTimestamp: datetime
+    approximateArrivalTimestamp: int
```
leandrodamascena: The reason I changed it to int was: imagine you need to save this information to a database or even needs a timestamp based control, so converting that to datetime might make things difficult.

What do you think? Make sense?

tests/functional/parser/test_kinesis_firehose.py
```diff
@@ -14,61 +14,83 @@
...
     assert metadata_02.shardId == "shardId-000000000001"
-    assert metadata_02.approximateArrivalTimestamp == "2012-04-23T19:25:43.511Z"
+    assert metadata_02.approximateArrivalTimestamp == 1664028793294
     assert metadata_02.sequenceNumber == "49546986683135544286507457936321625675700192471156785155"
 
 
-def test_kinesis_trigger_event():
-    event_dict = load_event("kinesisFirehoseEvent.json")
+@event_parser(model=KinesisFirehoseModel)
+def handle_firehose_no_envelope_put(event: KinesisFirehoseModel, _: LambdaContext):
```
leandrodamascena, line 61: New test using PUT payload.

leandrodamascena (changes requested):
@ran-isenberg could please check this changes?

leandrodamascena:
> I've modeled the metadata as optional so it works for both cases. The only reason to split them is if it's officially documented as such. do you think that's the case? I didnt find a mention that metadata params are applicable only for put events. If we add by ourselves, it's more likely to be broken in the future. WDYT?

I totally agree with you! :100:

aws_lambda_powertools/utilities/parser/models/kinesis_firehose.py
ran-isenberg: i think datetime has better validation. I used datetime in all the other models too.
I saw that you changed the example input from a string to an int? is that really the usecase?
If so, then perhaps int is better but you can a datetime validator just to make sure we'r good. I mean, it *is* the parser, so it should offer extra validations.

ran-isenberg: i changed it to positive int to make it extra safer

ran-isenberg:
@leandrodamascena added my small fix and fixed develop conflict

leandrodamascena (approved):

aws_lambda_powertools/utilities/parser/models/kinesis_firehose.py
leandrodamascena: Good catch, man! :1st_place_medal:

leandrodamascena:
> @leandrodamascena added my small fix and fixed develop conflict

Heyy good morning and happy monday @ran-isenberg!
If you don't have any other considerations, I think we are good to merge it!
Thanks for addressing all feedback so quickly. :fast_forward:

ran-isenberg:
LGTM

Issue #1552, opened 2022-09-29
Bug: APIGatewayRestResolver fails to route when "/" appears at end of route name


walmsles:
### Expected Behaviour

Given I am using APIGatewayRestResolver and I define a route as follows:
```python
from aws_lambda_powertools.event_handler import APIGatewayRestResolver

app = APIGatewayRestResolver()

@app.get("/hello")
def hello_world():
    return {"message": "Hello World"}
```

I expect the following 2 GET request URLs:

1. https://myapigwid.execute-api.ap-southeast-2.amazonaws.com/Prod/hello/
1. https://myapigwid.execute-api.ap-southeast-2.amazonaws.com/Prod/hello

to return the same response:

```JSON
{"message":"Hello World"}
```

### Current Behaviour

Given I am using APIGatewayRestResolver and I define a route as follows:
```python
@app.get("/hello")
def hello_world():
    return {"message": "Hello World"}
```

When I use the following URL:

1. https://myapigwid.execute-api.ap-southeast-2.amazonaws.com/Prod/hello/

AWS Lambda Powertools returns a 404 not found response:

```JSON
{"statusCode":404,"message":"Not found"}
```

### Code snippet

```python
from aws_lambda_powertools.event_handler import APIGatewayRestResolver
from aws_lambda_powertools import Logger

app = APIGatewayRestResolver()
logger = Logger()

@app.get("/hello")
def hello_world():
    return {"message": "Hello World"}

@logger.inject_lambda_context(log_event=True)
def lambda_handler(event, context):
    logger.info("Executing app.resolve")
    return app.resolve(event, context)
```

### Possible Solution

Route resolution uses the `path` attribute of the AWS API Gateway Lambda event so that the 2 URLs (one with trailing "/") are no longer the same even though the API Gateway Service treats the routes correctly and executes the expected Lambda function.

I would suggest that during route resolution trailing "/" characters be removed and not considered in route resolution. 

Unsure if this would be considered a breaking change and whether any customers are using this behaviour - I would not think this behaviour would be in use y customers but am unsure 😬 

### Steps to Reproduce

Create and deploy the SAM CLI Hello World function then replace the app.py with my code snippet above which has replaced it with AWS Lambda powertools resolver once deployed try loading the API in browser with and without trailing "/" character

### AWS Lambda Powertools for Python version

latest

### AWS Lambda function runtime

3.9

### Packaging format used

PyPi

### Debugging logs

```python
[...]

PR #1577, opened 2022-10-09
feat(router): Ignore trailing slashes in route resolvers inherited from APIGatewayResolver

walmsles:
**Issue number:** #1552

## Summary

Ingore trailing slash for routers inheriting from APIGatewayResolver

### Changes

AWS API Gateway routes the following 2 paths in the same way "/my/path", "/my/path/", i.e. when routing "/my/path/" it routes exactly as if "/my/path" had been called.  The Lambda Powertools router currently treats these as 2 distinct paths so "/my/path/" returns HTTP statusCode of 404.

### User experience

No change to user experience.  Users will create routes as normal.  In the given sample below making API gateway call to path "/my/path/" will correctly ensure the "get_lambda" route function is correctly called even though it is defined as "/my/path"

```python

    @app.get("/my/path")
    def get_lambda() -> Response:
        assert isinstance(app.current_event, APIGatewayProxyEvent)
        assert app.lambda_context == {}
        assert app.current_event.request_context.domain_name == "id.execute-api.us-east-1.amazonaws.com"
        return Response(200, content_types.APPLICATION_JSON, json.dumps({"foo": "value"}))
```

The code below will work in the same way when called using path "/my/path", "/my/path/"

```python

    @app.get("/my/path/")
    def get_lambda() -> Response:
        assert isinstance(app.current_event, APIGatewayProxyEvent)
        assert app.lambda_context == {}
        assert app.current_event.request_context.domain_name == "id.execute-api.us-east-1.amazonaws.com"
        return Response(200, content_types.APPLICATION_JSON, json.dumps({"foo": "value"}))
```

## Checklist

* [x] Changes are documented

## Acknowledgment

-----
[View rendered docs/core/event_handler/api_gateway.md](https://github.com/walmsles/aws-lambda-powertools-python/blob/develop/docs/core/event_handler/api_gateway.md)

### DIFF
--- a/.github/ISSUE_TEMPLATE/share_your_work.yml
+++ b/.github/ISSUE_TEMPLATE/share_your_work.yml
@@ -1,0 +1,56 @@
+name: I Made This (showcase your work)
+description: Share what you did with Powertools 💞💞. Blog post, workshops, presentation, sample apps, etc.
+title: "[I Made This]: <TITLE>"
+labels: ["community-content"]
+body:
+  - type: markdown
+    attributes:
+      value: Thank you for helping spread the word out on Powertools, truly!
+  - type: input
+    id: content
+    attributes:
+      label: Link to your material
+      description: |
+        Please share the original link to your material.
+
+        *Note: Short links will be expanded when added to Powertools documentation*
+    validations:
+      required: true
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: Describe in one paragraph what's in it for them (readers)
+    validations:
+      required: true
+  - type: input
+    id: author
+    attributes:
+      label: Preferred contact
+      description: What's your preferred contact? We'll list it next to this content
+    validations:
+      required: true
+  - type: input
+    id: author-social
+    attributes:
+      label: (Optional) Social Network
+      description: If different from preferred contact, what's your preferred contact for social interactions?
+    validations:
+      required: false
+  - type: textarea
+    id: notes
+    attributes:
+      label: (Optional) Additional notes
+      description: |
+        Any notes you might want to share with us related to this material.
+
+        *Note: These notes are explicitly to Powertools maintainers. It will not be added to the community resources page.*
+    validations:
+      required: false
+  - type: checkboxes
+    id: acknowledgment
+    attributes:
+      label: Acknowledgment
+      options:
+        - label: I understand this content may be removed from Powertools documentation if it doesn't conform with the [Code of Conduct](https://aws.github.io/code-of-conduct)
+          required: true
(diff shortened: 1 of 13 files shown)


walmsles:
Note: I started working on E2E tests - these are unfinished and will take some more time to complete.  

@heitorlessa you mentioned ALB E2E tests already being defined - I was not able to find can you share a link to the test?

heitorlessa:
> Note: I started working on E2E tests - these are unfinished and will take some more time to complete.
> 
> @heitorlessa you mentioned ALB E2E tests already being defined - I was not able to find can you share a link to the test?

My bad I forgot to mention it's in the V2 branch: https://github.com/awslabs/aws-lambda-powertools-python/blob/v2/tests/e2e/event_handler/test_header_serializer.py#L39

This was for a breaking change feature @rubenfonseca was doing, but we didn't merge the tests (and optimizations) to `develop` because of V2 proximity -- Would you mind changing the base to `v2` as we're close to launch?

There's only one major item pending for V2 (Serverless Application Repository) and we should be good to go.

walmsles:
> My bad I forgot to mention it's in the V2 branch: https://github.com/awslabs/aws-lambda-powertools-python/blob/v2/tests/e2e/event_handler/test_header_serializer.py#L39
> 
> This was for a breaking change feature @rubenfonseca was doing, but we didn't merge the tests (and optimizations) to `develop` because of V2 proximity -- Would you mind changing the base to `v2` as we're close to launch?

Ahhh, that makes sense now - nice - it's all in place, so can add my tests pretty quickly - let me move it across.

walmsles:
Hope my ramblings below make sense, the TL;DR version goes like this: 

> I have decided to restrict the implementation of this feature to REST API Gateway (V1) since this is the default behaviour of this gateway implementation for a defined route i.e. both "/hello" and "/hello/" resolve to the same lambda integration.  I will write E2E tests covering all the API behaviours around this feature so if anything changes it will be picked up.

The E2E tests are amazing!  I have been playing with APIs in various guises today (Much appreciation to the person responsible for writing those!) and have come to some conclusions:

- The only use-case this feature applies to IS the REST API (V1) Implementation - which works as per my original description
- ALB and Lambda URL - all routing is deferred to the actual Lambda, so we should be faithful to [RFC 3986 on Path](https://www.rfc-editor.org/rfc/rfc3986#section-3.3). so as not to close off possible customer use cases.
- HTTP API (V2) - actually does not allow the creation of a route with a trailing "/" character and behaves by sending back a "404" response when an invalid path is requested (unlike V1 which is a default 403 - forbidden).  If you setup a default route then the HTTP API will behave in similar fashion to ALB and Lambda URL - so again RFC 3986 should apply.

The code in this PR is not up to date as yet since have switched my feature branch to be from V2.  Will hopefully push changes tomorrow.

walmsles:
@heitorlessa - This PR is a little mucked up (my fault) - will re-issue a new PR based on V2 and my feature branch, which is required to fix this up.

Have raised new PR - #1609

heitorlessa:
Closing in favour of #1609

Issue #1590, opened 2022-10-12
SnsSqsEnvelope/SnsNotificationModel - Missing fields for FIFO SNS -> FIFO SQS -> Lambda


plondino:
### Expected Behaviour

I have a lambda triggered by a FIFO SQS which is subscribed to a FIFO SNS (see https://docs.aws.amazon.com/sns/latest/dg/fifo-message-delivery.html for this format).

The payload I receive is:
```{
    "Records": [
        {
            "messageId": "4136f084-6249-4a9a-89d8-e824bfa8c237",
            "receiptHandle": "AQEBKimXhFVwYo/BVlwqfx2KHOosqbnMpK8sT3Wo/9VL49dnWMcDgHtllmcYbL02YFakmTCK+1vLCtMNRidlZyLGFv5Z0L8H0Zw/xvdIteyXZiAR7m49u5lFp4f0nfst70Qykasg6dHkoz9s1xmEh03HIRJlNcOWarVF8jneaCoCw3mz77ehoY5WWNaVmw0hbZRlVpu8MLJ2k2RiIPNgJlShASyJ5KD0YTfUQFNC5T7657qiBwlGyFVfZs6BxYKB5Tv8QZy3n7OSjBZTsI7Db35gMlpMmVWcHyBv7XBjf6vdARp42oopStqYbDjcQhEYyl1taPPmfk9pCY4lwxaYQOoGiQ==",
            "body": "{\n  \"Type\" : \"Notification\",\n  \"MessageId\" : \"20aefd0b-ab3a-5911-b009-a6ca92db2664\",\n  \"SequenceNumber\" : \"10000000000000028000\",\n  \"TopicArn\" : \"arn:aws:sns:us-east-1:1234567890:topic.fifo\",\n  \"Message\" : \"...\",\n  \"Timestamp\" : \"2022-10-12T21:54:41.629Z\",\n  \"UnsubscribeURL\" : \"https://sns.us-east-1.amazonaws.com/?Action=Unsubscribe&SubscriptionArn=arn:aws:sns:us-east-1:1234567890:topic.fifo\"\n}\n",
            "attributes": {
                "ApproximateReceiveCount": "1",
                "SentTimestamp": "1665611681790",
                "SequenceNumber": "18873140664247791872",
                "MessageGroupId": "123456",
                "SenderId": "AIDAYRRVD2ENU4DSO2WBX",
                "MessageDeduplicationId": "CID1234",
[...]

PR #1606, opened 2022-10-14
fix(parser): loose validation on SNS fields to support FIFO

heitorlessa:
**Issue number:** #1590

### Changes

When enabling FIFO in SNS topics, SNS removes the following three keys leading to a `ValidationError` in Parser. 

**Error**

```python
SigningCertUrl
  none is not an allowed value (type=type_error.none.not_allowed)
Signature
  field required (type=value_error.missing)
SignatureVersion
  field required (type=value_error.missing)
```

This PR makes two changes:

* Marks all of these fields as Optional. In the future, we should look into having explicit FIFO Models to make strict typing better
* Removes `test_handle_sns_sqs_trigger_event_json_body_missing_signing_cert_url` that verified the field being tampered.. which is no longer valid due to FIFO behaviour completely removing this and other fields above

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/parser/models/sns.py
+++ b/aws_lambda_powertools/utilities/parser/models/sns.py
@@ -22,10 +22,10 @@ class SnsNotificationModel(BaseModel):
     MessageAttributes: Optional[Dict[str, SnsMsgAttributeModel]]
     Message: Union[str, TypingType[BaseModel]]
     MessageId: str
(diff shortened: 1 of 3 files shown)


leandrodamascena (approved):
Approved!!

PR #1608, opened 2022-10-14
chore(deps): remove email-validator; use Str over EmailStr in SES model

### LINKED ISSUE #1607: Maintenance: Remove email-validator dependency for v2
### Summary

We should remove the `email-validator` dependency from the v2 Lambda Layer and SAR.

### Why is this needed?

As we try to reduce the size of our v2 layer, we noticed that the [email-validator](https://pypi.org/project/email-validator/) is only needed if you use the [pydantic's  `EmailStr` validator](https://pydantic-docs.helpmanual.io/usage/types/#pydantic-types). 

Currently, adding `email-validator` adds ~400KB to our compressed layer size.

Validation

### Solution

_No response_

### Acknowledgment

rubenfonseca:
**Issue number:** #1607

### Changes

Removes `email-validator` from the extra group dependencies.

### User experience

This will also reduce the total Lambda Layer and SAR app size.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/parser/models/ses.py
+++ b/aws_lambda_powertools/utilities/parser/models/ses.py
@@ -2,7 +2,6 @@ from datetime import datetime
 from typing import List, Optional
 
 from pydantic import BaseModel, Field
-from pydantic.networks import EmailStr
 from pydantic.types import PositiveInt
 
 from ..types import Literal
@@ -21,7 +20,7 @@ class SesReceiptAction(BaseModel):
 class SesReceipt(BaseModel):
     timestamp: datetime
     processingTimeMillis: PositiveInt
-    recipients: List[EmailStr]
+    recipients: List[str]
     spamVerdict: SesReceiptVerdict
     virusVerdict: SesReceiptVerdict
     spfVerdict: SesReceiptVerdict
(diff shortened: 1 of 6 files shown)


aws_lambda_powertools/utilities/parser/models/ses.py
```diff
@@ -21,7 +20,7 @@ class SesReceiptAction(BaseModel):
 class SesReceipt(BaseModel):
     timestamp: datetime
     processingTimeMillis: PositiveInt
-    recipients: List[EmailStr]
+    recipients: List[str]
```
ran-isenberg, line 23: Hey @rubenfonseca @heitorlessa  , that's a breaking change. users are expecting here proper email address validation.
I think a better solution would be:
1. During import try to import EmailStr, if there's an except redefine the type as str
2. add docs and specify that for Ses model, you should install the extra dependency

WDYT?

rubenfonseca: Hi @ran-isenberg , we hear you and appreciate your feedback. We’re including this change in V2 to not impact customers in any shape or form.

To add to my [comment earlier](https://github.com/awslabs/aws-lambda-powertools-python/issues/1607#issuecomment-1280864915), as helpful as e-mail validator may be for an untrusted payload, this is not the case for SES. It would, however, for something like “body” in API Gateway.

I’ll update the issue linked with this and related Pydantic overhead that are making us more cautious on double validation.

heitorlessa:
Reviewing it now. 

Before merging, we need to make sure @huonw @mgochoa are credited. Let's try using the recently discovered manual co-authors feature in GitHub. We should do it in the Release notes regardless

heitorlessa (approved):

Issue #1432, opened 2022-08-07
Feature request: support DynamoDB image values to dict with built-in Python types 


shanab:
### Use case

When using `DynamoDBRecord` class, the values for `old_image` and `new_image` are represented as `Dict[str, AttributeValue]`. This representation is sufficient when the data parsed is simple and does not contain nested maps/lists.
But when the data parsed contains nested maps/lists, then it becomes hard to get to the data that `AttributeValue` represents.
One would need to recursively parse the `AttributeValue` objects in the nested dicts to get to the underlying data.

### Solution/User Experience

Using a deserializer similar to the `TypeDeserializer` provided by boto3 ([link to code](https://github.com/boto/boto3/blob/d581014c3f4bc7810e1868807481d3fe69e9e3b6/boto3/dynamodb/types.py#L242)), one would be able to convert image to a Python dict where the values would be built-in Python types.

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

heitorlessa:
Thanks a lot for taking the time to raise it Ahmed. I can’t wait to see
this PR.

ran-isenberg:
@shanab 
did you try using the parser? you can specify models for the images and it all gets parsed nicely.
https://awslabs.github.io/aws-lambda-powertools-python/latest/utilities/parser/#built-in-envelopes
There's a DynamoDBStreamEnvelope class.

It seems like the data classes features is trying to implement parsing features IMO.

shanab:
@ran-isenberg the problem in our case was that the service was not using Pydantic, and we did not want to introduce it only to parse the DynamoDB record.
I agree that there is an overlap between data classes and Parser, but IMO data classes should still allow traversing the fields of a record.
Also looking at the [data classes docs](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/docs/utilities/data_classes.md#key-features), one of the key features mentioned is "Helper functions for decoding/deserializing nested fields"

heitorlessa:
hey @shanab Ahmed, checking in to see if you still have the bandwidth to contribute as you already had the code working.

Thank you!

[...]

PR #1619, opened 2022-10-19
feat(data-classes): replace AttributeValue in DynamoDBStreamEvent with deserialized Python values

shanab:
**Issue number:** #1432 

## Summary

Add `DynamoDBImageDeserializer` class to data classes. The class can be used to parse DynamoDB StreamRecord's `old_image` and `new_image` properties of type `Dict[str, AttrValue]` to Dict with Python types.

### Changes

- Replace `AttributeValue` with deserialized Python types.

### User experience

Users will be able to convert a StreamRecord's complex image type into a Python dictionary with Python types; without needing to recurse through the data, which was a limitation when using `AttributeValue`.

Please note that the following functions: `new_image`, `old_image` and `keys` will introduce breaking changes to users.

#### Old usage example:
```python
def lambda_handler(event: DynamoDBStreamEvent, context):
    for record in event.records:
        new_image: Dict[str, AttributeValue] = record.dynamodb.new_image
        event_type: AttributeValue = new_image["eventType"].get_value
        if event_type == "PENDING":
            data = get_image_data(new_image)
            send_to_sqs(data)

def get_image_data(image: Dict[str, AttributeValue]) -> Dict[str, Any]:
    # Note: the following line does not work with DynamoDB list and map types.
    # If such types are present, then one will have to recursively call `get_value`
    # to get to the underlying data.
    return {k: v.get_value for k, v in image.items()}

def send_to_sqs(data: Dict):
    body = json.dumps(data)
    # send to sqs
```

#### New usage example:
```python
def lambda_handler(event: DynamoDBStreamEvent, context):
    for record in event.records:
        new_image: Dict[str, Any] = record.dynamodb.new_image
        if new_image.get("eventType") == "PENDING":
            send_to_sqs(new_image)  # Here new_image is just a Python Dict type

def send_to_sqs(data: Dict[str, Any]):
    body = json.dumps(data)
    # send to sqs
```

## Checklist

* [x] Changes are documented

<details>

Yes
**RFC issue number**:

Checklist:

</details>

## Acknowledgment

## TODOLIST

* [ ] Remove all references of `AttributeValue`
* [ ] Document with examples how `new_image`, `old_image` and `keys` are now different from how they used to work previously.


aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -1,169 +1,78 @@
...
-    --------------
-    - https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_AttributeValue.html
-    - https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.NamingRulesDataTypes.html
+class TypeDeserializer:
+    """
+    This class deserializes DynamoDB types to Python types.
+    It based on boto3's DynamoDB TypeDeserializer found here:
+    https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/types.html
+    Except that it deserializes DynamoDB numbers into strings, and does not wrap binary
+    with a Binary class.
```
heitorlessa: any particular reason why we should not handle Binary (`B`, `BS`) too?

shanab: We do actually, but we do not wrap it in a `Binary` class like the one boto3 provides here:
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/dynamodb.html#boto3.dynamodb.types.Binary

Maybe the wording can be improved. Any suggestions?

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -176,28 +85,43 @@ class StreamViewType(Enum):
...
+        """
+        super().__init__(data)
+        self._deserializer = TypeDeserializer()
+
+    def _deserialize_dynamodb_dict(self, key: str) -> Optional[Dict[str, Any]]:
+        dynamodb_dict = self._data.get(key)
+        return (
+            None if dynamodb_dict is None else {k: self._deserializer.deserialize(v) for k, v in dynamodb_dict.items()}
+        )
+
```
heitorlessa:
```suggestion
        dynamodb_dict = self._data.get(key)
        if dynamodb_dict is None:
            return None
        
        return {k: self._deserializer.deserialize(v) for k, v in dynamodb_dict.items()}
```

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -1,169 +1,78 @@
...
+        :returns: The pythonic value of the DynamoDB type.
         """
-        return self.get("NS")
 
-    @property
-    def null_value(self) -> None:
-        """An attribute of type Null.
+        if not value:
+            raise TypeError("Value must be a nonempty dictionary whose key " "is a valid dynamodb type.")
```
heitorlessa:
```suggestion
```

pr: We'll always receive the DynamoDB type as a value, even if the value itself is null `{"S": ""}` so this will never hit. We can safely remove it.

Boto implementation takes into account a manually crafted DynamoDB data which we don't need to (this entire class serializes a real DynamoDB event)

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -1,169 +1,78 @@
...
+        dynamodb_type = list(value.keys())[0]
+        try:
+            deserializer = getattr(self, f"_deserialize_{dynamodb_type}".lower())
+        except AttributeError:
+            raise TypeError(f"Dynamodb type {dynamodb_type} is not supported")
+        return deserializer(value[dynamodb_type])
```
heitorlessa:
```suggestion
        deserializer = getattr(self, f"_deserialize_{dynamodb_type}".lower(), None)
        if deserializer is None:
            raise TypeError(f"Dynamodb type {dynamodb_type} is not supported")

        return deserializer(value[dynamodb_type])
```

heitorlessa: pr: Exception handling isn't zero cost until Python 3.11. We can use a sentinel value (`None`) and use a quick if that has not much cost in bytecode.

heitorlessa (changes requested):
peer review comments in

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -176,28 +85,43 @@ class StreamViewType(Enum):
...
+    def __init__(self, data: Dict[str, Any]):
+        """StreamRecord constructor
+
+        Parameters
+        ----------
+        data: Dict[str, Any]
+            Represents the dynamodb dict inside DynamoDBStreamEvent's records
+        """
+        super().__init__(data)
+        self._deserializer = TypeDeserializer()
```
heitorlessa:
```suggestion
class StreamRecord(DictWrapper):
    _deserializer = TypeDeserializer()  # init once instead of per record

    def __init__(self, data: Dict[str, Any]):
        """StreamRecord constructor

        Parameters
        ----------
        data: Dict[str, Any]
            Represents the dynamodb dict inside DynamoDBStreamEvent's records
        """
        super().__init__(data)
```

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
heitorlessa:
```suggestion
    Deserializes DynamoDB types to Python types.
    
    It based on boto3's [DynamoDB TypeDeserializer](https://boto3.amazonaws.com/v1/documentation/api/latest/_modules/boto3/dynamodb/types.html).
    
    The only notable difference is that for Binary (`B`, `BS`) values we return Python Bytes directly, since we don't support Python 2.
```

heitorlessa: had a discussion with @rubenfonseca and @leandrodamascena - both prefer staying in sync with AWS SDK Decimal handling.

**Reasoning**

* Customer handling `DynamoDB Stream -> Lambda -> SQS`  would need to handle a Decimal here with AWS SDK. With our `AttributeValue`, they would need to do more and call `get_value` before they're able to serialize it
* Customer handling `API Gateway -> Lambda` with Event Handler, or Idempotency, will automatically serialize `Decimal` to `str`.

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -1,169 +1,78 @@
...
-    @property
-    def ns_value(self) -> Optional[List[str]]:
-        """An attribute of type Number Set
-
-        Example:
-            >>> {"NS": ["42.2", "-19", "7.5", "3.14"]}
+    def deserialize(self, value):
+        """The method to deserialize the DynamoDB data types.
+
+        :param value: A DynamoDB value to be deserialized to a pythonic value.
```
heitorlessa, line 19:
```suggestion
         Parameters
         ----------
         value: Any
            DynamoDB value to be deserialized to a python type
```

We use a different docstring style than AWS SDK

aws_lambda_powertools/utilities/data_classes/dynamo_db_stream_event.py
```diff
@@ -1,169 +1,78 @@
...
+            {'N': str(value)}                       str(value)
+            {'S': string}                           string
+            {'B': bytes}                            bytes
+            {'NS': [str(value)]}                    set([str(value)])
+            {'SS': [string]}                        set([string])
+            {'BS': [bytes]}                         set([bytes])
+            {'L': list}                             list
+            {'M': dict}                             dict
+
+        :returns: The pythonic value of the DynamoDB type.
```
heitorlessa:
```suggestion

           Here are the various conversions:

           DynamoDB                                Python
           --------                                ------
           {'NULL': True}                          None
           {'BOOL': True/False}                    True/False
           {'N': str(value)}                       str(value)
           {'S': string}                           string
           {'B': bytes}                            bytes
           {'NS': [str(value)]}                    set([str(value)])
           {'SS': [string]}                        set([string])
           {'BS': [bytes]}                         set([bytes])
           {'L': list}                             list
           {'M': dict}                             dict

        Parameters
        ----------
        value: Any
           DynamoDB value to be deserialized to a python type

        Returns
        --------
        any
            Python native type converted from DynamoDB type
```

We  use a different docstring (Numpy) style compared to Boto (Sphinx). This adjusts it so it can render nicely in the API docs.

heitorlessa:
Pushed the upgrade guide for this with a note on `Decimal` too.

image

heitorlessa (approved):

heitorlessa:
Added static typing for methods signatures and return, added last comments to help remember Decimal logic in the far future, and added a note about this being related to Batch Processing too.

Waiting for CI and we can merge!!!!

image

PR #1639, opened 2022-10-23
fix: parser S3Model not compatible with ObjectRemoved events

ran-isenberg:
issue https://github.com/awslabs/aws-lambda-powertools-python/issues/1637

### DIFF
--- a/aws_lambda_powertools/utilities/parser/models/s3.py
+++ b/aws_lambda_powertools/utilities/parser/models/s3.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from typing import List, Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, root_validator
 from pydantic.fields import Field
 from pydantic.networks import IPvAnyNetwork
 from pydantic.types import NonNegativeFloat
@@ -43,8 +43,8 @@ class S3Bucket(BaseModel):
 
 class S3Object(BaseModel):
     key: str
-    size: NonNegativeFloat
-    eTag: str
+    size: Optional[NonNegativeFloat]
+    eTag: Optional[str]
     sequencer: str
     versionId: Optional[str]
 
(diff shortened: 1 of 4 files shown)


ran-isenberg:
@heitorlessa @barreeeiroo FYI

barreeeiroo:
Thanks @ran-isenberg! I'm back with my laptop, so I was about to test it now. Would it be fine if I cherrypick your commit into a single PR?

ran-isenberg:
Yeah sure np! I had to make some fixes to the validator.

barreeeiroo:
I see you added an `if s3_object`. Is it really needed? I added an additional test case (to make sure we throw the exception), and it seems to be passing without it.
The `root_validator` should run after parsing the object, so technically `s3_object` is never `None` (otherwise a `ValidationException` is thrown earlier.

ran-isenberg:
> I see you added an `if s3_object`. Is it really needed? I added an additional test case (to make sure we throw the exception), and it seems to be passing without it. The `root_validator` should run after parsing the object, so technically `s3_object` is never `None` (otherwise a `ValidationException` is thrown earlier.

no, it's not needed, i think you can remove the extra test case too.
I wrote an earlier version with pre=True but then removed it, forgot to remove this if.

barreeeiroo:
I think it should be done now then. Can you double-check https://github.com/awslabs/aws-lambda-powertools-python/pull/1638 to see if I missed anything from your changes?

heitorlessa:
@leandrodamascena let's merge @ran-isenberg into @barreeeiroo PR, and make sure @ran-isenberg is a co-author in that PR

heitorlessa:
nvm, it is already in (need coffee). I'll merge and add Ran as a co-author

ran-isenberg:
@heitorlessa  it's already merged there

PR #1662, opened 2022-10-27
feat(logger): add use_rfc3339 and auto-complete formatter opts in Logger

### LINKED ISSUE #946: Documentation on using custom date format for logging
**What were you initially searching for in the docs?**
I was searching for a way to use ISO8601 especially the RFC3339 variant while printing the event time stamps. Then i found this discussion link https://github.com/awslabs/aws-lambda-powertools-python/issues/905#issuecomment-1003420845

**Is this related to an existing part of the documentation? Please share a link**
https://awslabs.github.io/aws-lambda-powertools-python/latest/core/logger/#

**Describe how we could make it clearer**
I would be really helpful if we can provide an example of how to change the timeformat to RFC3339 in the documentation site here

**If you have a proposed update, please share it here**
Example on how to instantiate logger using RFC3339 format

heitorlessa:
**Issue number:** #946

## Summary

This makes it easier to use a compatible date format with RFC3339 and ISO8601: `2022-10-27T16:27:43.738+02:00`

### Changes

### User experience

You can now simply use `use_rfc3339` flag to enable it

```python
logger = Logger(service=service_name, stream=stdout, use_rfc3339=True)
```

This PR also raised a suboptimal UX we had on propagating `LambdaPowertoolsFormatter` options in `Logger` as `kwargs`. When implementing this feature, it made it difficult to remember the entire param. For this reason, this PR exposes all of them within `Logger` constructor, so that: `Logger(utc=, ...)` autocompletes.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
@leandrodamascena need your review as this must be included in today's release

aws_lambda_powertools/logging/formatter.py
```diff
@@ -72,6 +73,7 @@ def __init__(
         use_datetime_directive: bool = False,
         log_record_order: Optional[List[str]] = None,
         utc: bool = False,
+        use_rfc3339_iso8601: bool = False,
```
leandrodamascena: This is a personal opinion, but I think this variable name makes me a little confused. I know RFC3339 is a profile/subset of the ISO8601 format, but it seems a little confusing to me. Maybe just `use_rfc3339_format`?

But again, it's just a personal opinion as a user and makes no difference to the intended purpose.

leandrodamascena (commented):
Hi @heitorlessa! I tested it locally and it's working as expected!

I just made a comment, but if you don't agree with that I think this PR is read to merge.

aws_lambda_powertools/logging/formatter.py
heitorlessa: Makes sense, I'll shorten to `use_rfc3339`. What I meant here is that our new format is compliant with both RFC3339 and ISO8601.

I thought leaving `ISO8601` would give the impression we don't support. But if you are confused, I'm sure others will too - I'll keep `ISO8601` as a docstring but shorten the param.

heitorlessa: Done!

heitorlessa:
Addressed your feedback and shorten to `use_rfc3339`. I'll work on docs now to make this more visible.

heitorlessa:
Docs done - @leandrodamascena if you approve, feel free to merge.

image


image

leandrodamascena:
> Docs done - @leandrodamascena if you approve, feel free to merge.
> 

Amazing @heitorlessa!

jplock:
Question on the usage of this, if I set `use_rfc3339=True` in the main `Logger`, and then reference another logger in another file with `Logger(child=True)`, do I also have to set `use_rfc3339=True` in the child loggers?

heitorlessa:
Nope it should just work, as a child logger will use the parent Logger
formatter.

I’ve also updated docs on Child Loggers that unless you truly need an
explicit child Logger, you can use a Logger() if POWERTOOLS_SERVICE_NAME is
set (Logger(service=service) otherwise)

PR #1666, opened 2022-10-28
Feature request: Time based feature flags actions

### LINKED ISSUE #1554: Feature request: Time based feature flags actions
### Use case

I've got a few use cases for feature flags rule actions. It's possible to due that today but its really finicky.

1. I'd like to turn on/enable a feature flags rule only during a certain time of the day (between 10:00-23:00 24 hours clock)
2. i'd like to turn on/enable a feature flags rule during a certain date range (UTC time)
3. i'd like to turn on/enable a feature flags rule only during weekends, or just weekdays
4. i'd like to turn on/enable a feature flags rule during certain week days and certain hours

I'd like to be able to do combine these conditions with other conditions.

I think the implementation is simple, just need to define the UX.

### Solution/User Experience

Time based conditions are sorely missing from the current solution. 

### Alternative solutions

```markdown
use regular actions. Use rules that define time as unix time ints and define actions like KEY_GREATER_THAN_VALUE and KEY_LESS_THAN_VALUE.
Not really readable when using unix time ints but it can work.
Not agreat UX.
```

### Acknowledgment

ran-isenberg:
**#1554:**
Feature request: Time based feature flags actions

## Summary

Use cases:

* Enable maintenance mode during a weekend
* Disable support/chat feature after working hours
* Launch a new feature on a specific date and time

### Changes

1. Added new condition actions, key and values (see user experience examples)
2. Added tests
3. Added documentation
4. Added new schema validation

New actions: SCHEDULE_BETWEEN_TIME_RANGE, SCHEDULE_BETWEEN_DATETIME_RANGE, SCHEDULE_BETWEEN_DAYS_OF_WEEK

New keys: CURRENT_TIME, CURRENT_DATETIME, CURRENT_DAY_OF_WEEK

New values : a dict of START and END string values and a list of all weekdays. Also an optional TIMEZONE field supporting [IANA Time zones](https://www.iana.org/time-zones). When not specified, we assume UTC. 

Assumption: time is in 24 hours format, start time is always smaller than end time and it does not overlap a day.

### User experience

1. Time range

As a customer, I'd like to flip a static flag value when evaluated between 11:11 and 23:59.

```
{
    "my_feature": {
        "default": false,
        "rules": {
            "lambda time is between UTC 11:11-23:59": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_TIME_RANGE",
                        "key": "CURRENT_TIME",
                        "value": {
                            "START_TIME": "11:11",
                            "END_TIME": "23:59",
                            "TIMEZONE": "Europe/Copenahgen"
                        }
                    }
                ]
            }
        }
    }
}
```
2. Date range

As a customer, I'd like to flip a static flag when evaluated between full datetime ranges 2022-10-05T12:15:00Z to 2022-10-10T12:15:00Z.

```
{
    "my_feature": {
        "default": false,
        "rules": {
            "lambda time is between UTC october 5th 2022 12:14:32PM to october 10th 2022 12:15:00 PM": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_DATETIME_RANGE",
                        "key": "CURRENT_DATETIME",
                        "value": {
                            "START_TIME": "2022-10-05T12:15:00",
                            "END_TIME": "2022-10-10T12:15:00",
                            "TIMEZONE": "America/Los_Angeles"
                        }
                    }
                ]
            }
        }
    }
}
```
Selected days

As a customer, I'd like to flip a static flag when evaluated between Monday to Friday.

```
{
    "my_feature": {
        "default": false,
        "rules": {
            "match only monday through friday": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_DAYS_OF_WEEK",
                        "key": "CURRENT_DAY_OF_WEEK",
                        "value": {
                            "DAYS": [
                               "MONDAY",
                               "TUESDAY",
                               "WEDNESDAY",
                               "THURSDAY",
                               "FRIDAY"
                             ],
                             "TIMEZONE": "America/New_York" 
                        }
                    }
                ]
            }
        }
    }
}
```
Day and time range combined

As a customer, I'd like to flip a static flag when evaluated Monday-Friday between 11:00 and 23:00 (UTC).

```
{
    "my_feature": {
        "default": false,
        "rules": {
            "match when lambda time is between UTC 11:00-23:00 and day is either monday or thursday": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_TIME_RANGE",
                        "key": "CURRENT_TIME",
                        "value": {
                            "START_TIME": "11:00",
                            "END_TIME": "23:00"
                        }
                    },
                    {
                        "action": "SCHEDULE_BETWEEN_DAYS_OF_WEEK",
                        "key": "CURRENT_DAY_OF_WEEK",
                        "value": 
                            "DAYS": [
                                "MONDAY",
                                "THURSDAY"
                            ]
                    }
                ]
            }
        }
    }
}
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

## Tasks

- [x] Update documentation with realistic use cases
- [x] Rename `_UTC` suffix in keys, and implement the `timezone` field on the value (use `python-dateutil`)
- [x] Run a performance assessment
- [ ] Review logging information
- [x] Review docstrings and comments


ran-isenberg:
@leandrodamascena & @heitorlessa   FYI.
I implemented all the actions but there's a lot of work to be done.

1. Validation - i think we need to move to pydantic and validate ALL the configuration with it
2. Think about enums structure, maybe it's a bit confusing the way it is now, maybe split the time keys/values into even smaller enumbs.
3. mock time in tests - they fail becuase the conditions dont match
4. add tests where we mock them not to match
5. optimizations - i did a naive solution where i always calc the current time again and reparse the conditions into datetime. maybe we can use a cache for those conditions datetime parsing etc.

WDYT?
do you want to work on this together and split it?

i think we should start by merging our code and doing a quick CR

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -87,6 +90,10 @@ def _evaluate_conditions(
             cond_action = condition.get(schema.CONDITION_ACTION, "")
             cond_value = condition.get(schema.CONDITION_VALUE)
 
+            # rule based actions have no user context. the context is the condition key
+            if cond_action == schema.RuleAction.TIME_RANGE.value or schema.RuleAction.TIME_SELECTED_DAYS:
+                context_value = condition.get(schema.CONDITION_KEY)
+
```
leandrodamascena:
```suggestion
            # rule based actions have no user context. the context is the condition key
            if cond_action == schema.RuleAction.TIME_RANGE.value or schema.RuleAction.TIME_SELECTED_DAYS:
                context_value = condition.get(schema.CONDITION_KEY)
                
                if cond_action == schema.RuleAction.TIME_SELECTED_DAYS:
	                if not isinstance(cond_value,list):
	                    cond_value = [cond_value]
```

With user experience in mind, a value for this condition can be a simple string or a list, so we convert it to a list if it's a string. 
That makes sense?

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,55 @@
...
+def time_selected_days_compare(action: str, values: List[str]) -> bool:
+    if action == TimeKeys.CURRENT_DAY_UTC.value:
+        return _time_selected_days_current_days_compare(values)
+    # we assume it passed validation right? so no need to raise an error
+    return False
+
+
+def _time_selected_days_current_days_compare(values: List[str]) -> bool:
+    current_day_number: datetime = datetime.now(timezone.utc).isoweekday()
+    return DAY_MAPPING.get(current_day_number, "") in values
```
leandrodamascena:
```suggestion
    current_day: datetime = datetime.now(timezone.utc).strftime('%A').lower()
    return current_day in values
```

Maybe this to get the name of the week directly?

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,55 @@
...
+HOUR_MIN_SEPARATOR = ":"
+DAY_MAPPING = {
+    1: TimeValues.MONDAY,
+    2: TimeValues.TUESDAY,
+    3: TimeValues.WEDNESDAY,
+    4: TimeValues.THURSDAY,
+    5: TimeValues.FRIDAY,
+    6: TimeValues.SATURDAY,
+    7: TimeValues.SUNDAY,
+}
```
leandrodamascena:
```suggestion
```

leandrodamascena:
> @leandrodamascena & @heitorlessa FYI. I implemented all the actions but there's a lot of work to be done.
> 
Hi @ran-isenberg! Sorry for the delay in responding, but the days have been busy around here.
I tested this initial code and was able to validate a rule using this, but yes, there is a lot of work we need to do to do this. I made some suggestions to make things easier.

> 1. Validation - i think we need to move to pydantic and validate ALL the configuration with it
> 2. Think about enums structure, maybe it's a bit confusing the way it is now, maybe split the time keys/values into even smaller enumbs.
> 3. mock time in tests - they fail becuase the conditions dont match
> 4. add tests where we mock them not to match
> 5. optimizations - i did a naive solution where i always calc the current time again and reparse the conditions into datetime. maybe we can use a cache for those conditions datetime parsing etc.
> 
For these questions, @heitorlessa will help us work it out. I'll be on parental leave for the next few days :baby:.

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
ran-isenberg: Not sure I agree with this one. I see this action as "IN" so it should be a list.
If you open it to duality of values, it's more confusing.
It also makes inputs such as 'MONDAY, TUESDAY' break the code since you will turn it into a ['MONDAY, TUESDAY' ] which is not correct. I think having it as a list, and validating it as a list will keep it simple.

ran-isenberg:
> > @leandrodamascena & @heitorlessa FYI. I implemented all the actions but there's a lot of work to be done.
> 
> Hi @ran-isenberg! Sorry for the delay in responding, but the days have been busy around here. I tested this initial code and was able to validate a rule using this, but yes, there is a lot of work we need to do to do this. I made some suggestions to make things easier.
> 
> > 1. Validation - i think we need to move to pydantic and validate ALL the configuration with it
> > 2. Think about enums structure, maybe it's a bit confusing the way it is now, maybe split the time keys/values into even smaller enumbs.
> > 3. mock time in tests - they fail becuase the conditions dont match
> > 4. add tests where we mock them not to match
> > 5. optimizations - i did a naive solution where i always calc the current time again and reparse the conditions into datetime. maybe we can use a cache for those conditions datetime parsing etc.
> 
> For these questions, @heitorlessa will help us work it out. I'll be on parental leave for the next few days 👶.

I hope everything goes well @leandrodamascena !
I committed two of your suggestions and replied to the third.
I'll try to continue the work on it, it's a busy week.

I'll work on getting the time mocked properly and adding tests for fail use cases. That way when we refactor the code, it would be much easier. TDD for the win :)

ran-isenberg:
@heitorlessa  @leandrodamascena  added mocked time. added test for no match. fixed bugs.
it actually works now! 
docs/optimisations/schema validation are left. let me know what you think. I'm stopping here for now.

heitorlessa:
Looking into the UX this morning as promised - One thing to speed up PR reviews later is to add some UX example(s) in the PR body.

heitorlessa:
Adding this here so I can come back to it later today (and for @rubenfonseca after re:Invent)

Stream of consciousness for now:

* ~~`TIME_RANGE` action is also being applied for date range intent; `DATETIME_RANGE` might be more accurate~~
* ~~Might be a good idea to have a prefix for all scheduled-like tasks, e.g., `SCHEDULE_TIME_RANGE`, `SCHEDULE_DAYS`, `SCHEDULE_DATETIME_RANGE`~~
* ~~Review: Revisit `key` values to find alternative UX candidates.~~
* Tests: for a separate PR maybe, we need to revisit the reading and writing tests for this utility - we can simplify a lot with builder functions.
* Malformed dates: Datetime range missing a `Z` or any timezone info will cause a `ValueError` which results in a `False` value during evaluation - ask @ran-isenberg what's the intended behaviour here (warn customers?)

<details>
<summary>Discuss usage of `strptime` as it's the slowest way to convert ISO string to date</summary>

We're converting ISO8601 strings to datetime objects. There are two known ways to do it:

1. `datetime.strptime(value, "%Y-%m-%dT%H:%M:%S")`
2. `datetime.fromisoformat(value)`

The first option (currently implemented) is considered the slowest [for various reasons](https://ehmatthes.com/blog/faster_than_strptime/) (string interpolation, string to integer, etc.).

The second option can be up to 40x faster (depending on Python version and machine; 3.9 on mine is 2x). However it doesn't parse the `Z` as UTC timezone like ECMA, and instead requires `+00:00` format. We could easily replace `Z` with `+00:00` - even with a string replacement is much faster than `strptime`.

Why worry about this?

We don't know how many combinations a single rule could have, so parsing many dates over on every execution can cause a significant slowdown in the execution. In large number of dates (thousands) the slowness can be measured in seconds.

Did a super contrived single date parsing with snakeviz to visualize it.

> strptime output

image

> fromisoformat

image


</details>

**Quick changes pushed as I looked to understand the logic from a high level**

* Performance changes (2.15s to 1.90s on functional tests)
  * Replaced `dateutil` TZ with `datetime(..., tzinfo=timezone.utc)` - reduce 3P dependency as it's available in stdlib
    * **BEFORE**: `datetime.datetime(..., tz.gettz("UTC"))`
    * **AFTER**:  `datetime.datetime(..., tzinfo=datetime.timezone.utc)`
  * Coerced once
    * **BEFORE**: `context.get(str(condition.get(schema.CONDITION_KEY)))`
    * **AFTER**: `context.get(condition.get(schema.CONDITION_KEY, ""))`
  * Enum shouldn't subclass from str - it'll prevent Mypyc compilation. Looks like an old mistake (prior to this PR)
    * **BEFORE**: `class RuleAction(str, Enum):`
    * **AFTER**: `class RuleAction(Enum):`
* Fixed time/date range function comparison to use enum values over member. I highly suspect this was due to `test_time_based_multiple_conditions_utc_in_between_time_range_rule_match` test accidentally using an enum key over value.

---

### Current schema UX at time of writing

**Time range**

> As a customer, I'd like to flip a static flag value when evaluated between 11:11 and 23:59.

```json
{
    "my_feature": {
        "default": false,
        "rules": {
            "lambda time is between UTC 11:11-23:59": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_TIME_RANGE",
                        "key": "CURRENT_HOUR_UTC",
                        "value": {
                            "START_TIME": "11:11",
                            "END_TIME": "23:59"
                        }
                    }
                ]
            }
        }
    }
}
```

**Date range**

> As a customer, I'd like to flip a static flag when evaluated between 11:11 and 23:59.

```json
{
    "my_feature": {
        "default": false,
        "rules": {
            "lambda time is between UTC october 5th 2022 12:14:32PM to october 10th 2022 12:15:00 PM": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_DATETIME_RANGE",
                        "key": "CURRENT_TIME_UTC",
                        "value": {
                            "START_TIME": "2022-10-05T12:15:00Z",
                            "END_TIME": "2022-10-10T12:15:00Z"
                        }
                    }
                ]
            }
        }
    }
}
```

**Selected days**

> As a customer, I'd like to flip a static flag when evaluated between Monday to Friday.

```json
{
    "my_feature": {
        "default": false,
        "rules": {
            "match only monday through friday": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_DAYS",
                        "key": "CURRENT_DAY_UTC",
                        "value": [
                            "MONDAY",
                            "TUESDAY",
                            "WEDNESDAY",
                            "THURSDAY",
                            "FRIDAY"
                        ]
                    }
                ]
            }
        }
    }
}
```

**Day and time range combined**

> As a customer, I'd like to flip a static flag when evaluated Monday-Friday between 11:00 and 23:00.

```json
{
    "my_feature": {
        "default": false,
        "rules": {
            "match when lambda time is between UTC 11:00-23:00 and day is either monday or thursday": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "TIME_RANGE",
                        "key": "CURRENT_HOUR_UTC",
                        "value": {
                            "START_TIME": "11:00",
                            "END_TIME": "23:00"
                        }
                    },
                    {
                        "action": "TIME_SELECTED_DAYS",
                        "key": "CURRENT_DAY_UTC",
                        "value": [
                            "MONDAY",
                            "THURSDAY"
                        ]
                    }
                ]
            }
        }
    }
}
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -31,6 +31,26 @@ class RuleAction(str, Enum):
     KEY_NOT_IN_VALUE = "KEY_NOT_IN_VALUE"
     VALUE_IN_KEY = "VALUE_IN_KEY"
     VALUE_NOT_IN_KEY = "VALUE_NOT_IN_KEY"
+    TIME_RANGE = "TIME_RANGE"  # 24 hours clock UTC time
```
heitorlessa: UX suggestion: add a `DATETIME_RANGE`, since you are already doing at `test_time_based_utc_in_between_full_time_range_rule_match` but it wasn't obvious to me it was possible too.

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -31,6 +31,26 @@ class RuleAction(str, Enum):
...
+    TIME_SELECTED_DAYS = "TIME_SELECTED_DAYS"
+
+
+class TimeKeys(Enum):
+    CURRENT_HOUR_UTC = "CURRENT_HOUR_UTC"
+    CURRENT_DAY_UTC = "CURRENT_DAY_UTC"
+    CURRENT_TIME_UTC = "CURRENT_TIME_UTC"
```
heitorlessa: UX suggestion: `CURRENT_DATETIME_UTC`.

Looking at `test_time_based_utc_in_between_full_time_range_rule_match` it looks like you're expecting a datetime string not time alone.

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -31,6 +31,26 @@ class RuleAction(str, Enum):
     KEY_NOT_IN_VALUE = "KEY_NOT_IN_VALUE"
     VALUE_IN_KEY = "VALUE_IN_KEY"
     VALUE_NOT_IN_KEY = "VALUE_NOT_IN_KEY"
+    TIME_RANGE = "TIME_RANGE"  # 24 hours clock UTC time
+    TIME_SELECTED_DAYS = "TIME_SELECTED_DAYS"
```
heitorlessa: UX suggestion: `SCHEDULE_BETWEEN_DAYS` because you also have a start/end date

heitorlessa (changes requested):
@ran-isenberg UX wise, if I understood this right, this should support flags that should change state on **time**, **datetime**, and **days** - is that right?

If so, I'd keep `TimeKeys` as is, but change `RuleAction` to be more precise:

* `TIME_SELECTED_DAYS` -> `SCHEDULE_BETWEEN_DAYS`
* `TIME_RANGE` -> `SCHEDULE_BETWEEN_TIME_RANGE`
* `TIME_RANGE` for full dates -> `SCHEDULE_BETWEEN_DATETIME_RANGE`

The `SCHEDULE` prefix makes it easier to spot what the rule is about (time or value based). The addition of `DATETIME_RANGE` makes it explicit based on your logic - if I got it right - that you're comparing between the current_time, start_date, and end_date.

It'll become clearer any better variation (if any) when we write docs for it.

aws_lambda_powertools/utilities/feature_flags/schema.py
ran-isenberg: fixed

aws_lambda_powertools/utilities/feature_flags/schema.py
ran-isenberg: fixed

aws_lambda_powertools/utilities/feature_flags/schema.py
ran-isenberg: fixed

ran-isenberg:
@heitorlessa  Thank you for the feedback!
I've updated everything according to your suggestion and will start working on the validation.

ran-isenberg:
@heitorlessa 
"Malformed dates: Datetime range missing a Z or any timezone info will cause a ValueError which results in a False value during evaluation - ask @ran-isenberg what's the intended behaviour here (warn customers?)" 

I think we need to fail with a schema validation error upon load.

ran-isenberg:
@heitorlessa  @leandrodamascena  XXL now :)
finished validation so it's done now. all tests are green.
feel free to edit/change, no more tasks on my end for now.

ran-isenberg:
@leandrodamascena  will you have time this week to review and make suggestions for changes/optimizations?

rubenfonseca:
Hi @ran-isenberg I was the one meant to look into your PR this week, but unfortunately I got sick. I'm getting better now and should start looking into your work between today and tomorrow. Looking forward to review it!

rubenfonseca:
First time looking at this PR, with no context :) I'm very tired now, so I'll just add the main thing that worries me, and write more about it tomorrow:

I find it very arbitrary that we're creating rules for matching DAYS, and TIMES, and even entire DATERANGES. However, I left wondering, what if I also want to match certain HOURS, MINUTES or MONTHS? Will we be keep adding rule types to cover all those scenarios in the future?

**OR,** should we think of a better, generic way of representing (recurrent) time intervals and cover it in a just two calls (one-time time interval / recurrent time intervals)? Right now I'm [re-reading ISO8601](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals), as I feel this problem was solved before I was born :)

ran-isenberg:
> First time looking at this PR, with no context :) I'm very tired now, so I'll just add the main thing that worries me, and write more about it tomorrow:
> 
> I find it very arbitrary that we're creating rules for matching DAYS, and TIMES, and even entire DATERANGES. However, I left wondering, what if I also want to match certain HOURS, MINUTES or MONTHS? Will we be keep adding rule types to cover all those scenarios in the future?
> 
> **OR,** should we think of a better, generic way of representing (recurrent) time intervals and cover it in a just two calls (one-time time interval / recurrent time intervals)? Right now I'm [re-reading ISO8601](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals), as I feel this problem was solved before I was born :)

The daterange is standard ISO definition UTC time.
The days and hours:min range is a simplification for recurring events. So we are already covering hours or minutes, just not months.
In the end, it depends on user input. I find the datetime and days the most useful, at least in my company's usages but who knows. They seem pretty standard to me.

rubenfonseca:
> The daterange is standard ISO definition UTC time.

Humm I believe [ISO8601 dateranges](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals) are **not** supported in Python's stdlib. Only ISO8601 timestamps. Is that what you mean?

> The days and hours:min range is a simplification for recurring events. So we are already covering hours or minutes, just not months.

Ok, I was struggling to find a way to say "between minute 49 and 52 of every hour" with the current implementation. Are you sure we are covering that already? What about stuff like "every last day of every month"? Or "on the 25th of December, every year" Is it possible right now?

> In the end, it depends on user input. I find the datetime and days the most useful, at least in my company's usages but who knows. They seem pretty standard to me.

Sure! Let's just try to find the right balance between flexibility and complexity, while trying to cover more scenarios! I'm thinking a couple of things, will write back in the next hour or so

ran-isenberg:
> on the 25th of December, every year" Is it possible right now? - Nope, but it wasnt the use case I tried to solve.
This is a super generic date condition system. I started with what I believe where common use case for feature flags. I hardly van think of a use case where you'd use rule like that for feature flags, but i might be wrong.

BTW, the way we started with the actions in the feature flags was that I provided 4 basic actions and the community added more actions and feedback, and now we have over 10.
I think it's ok to start with these conditions as they cover a lot of ground for the common user IMO.

ran-isenberg:
> > The daterange is standard ISO definition UTC time.
> 
> Humm I believe [ISO8601 dateranges](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals) are **not** supported in Python's stdlib. Only ISO8601 timestamps. Is that what you mean?
> 
> > The days and hours:min range is a simplification for recurring events. So we are already covering hours or minutes, just not months.
> 
> Ok, I was struggling to find a way to say "between minute 49 and 52 of every hour" with the current implementation. Are you sure we are covering that already? What about stuff like "every last day of every month"? Or "on the 25th of December, every year" Is it possible right now?
> 
> > In the end, it depends on user input. I find the datetime and days the most useful, at least in my company's usages but who knows. They seem pretty standard to me.
> 
> Sure! Let's just try to find the right balance between flexibility and complexity, while trying to cover more scenarios! I'm thinking a couple of things, will write back in the next hour or so

You guys

ran-isenberg:
closed by mistake

rubenfonseca:
All right Ran, let's not make this harder than it needs to be. Before we review the code, I'll just ask for the following naming changes:

- `SCHEDULE_BETWEEN_DAYS` --> `SCHEDULE_BETWEEN_DAYS_OF_WEEK`

	A couple of times I've read this, and thought it was referring to calendar days of the month (1, 2, 15, 30, etc). Using `DAYS_OF_WEEK` is more explicit that we are talking about monday to sunday.

- `START_TIME` / `END_TIME` --> `START` / `END`

	The reason for this is that sometimes these keys are used to represent *times*, but other times they are used to represent full *dates*, which was very confusing to me. Changing it to "start" and "end" solves the problem in my head, because now I understand I'm setting the intervals for the context.

- `CURRENT_HOUR_UTC` --> `CURRENT_TIME_UTC`

	This context key represents the time, not just the hour. Since we have `CURRENT_DATETIME_UTC` and `CURRENT_DAY_UTC` already, I believe it makes sense to have this one called "time".

With these changes in place, I will start reviewing the code and ensuring all the feedback is included. Great job!

ran-isenberg:
> 

no problem. on it.

ran-isenberg:
done @rubenfonseca

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -59,6 +64,9 @@ def _match_by_action(self, action: str, condition_value: Any, context_value: Any
             schema.RuleAction.KEY_NOT_IN_VALUE.value: lambda a, b: a not in b,
             schema.RuleAction.VALUE_IN_KEY.value: lambda a, b: b in a,
             schema.RuleAction.VALUE_NOT_IN_KEY.value: lambda a, b: b not in a,
+            schema.RuleAction.SCHEDULE_BETWEEN_TIME_RANGE.value: lambda a, b: compare_utc_date_range(a, b),
```
rubenfonseca:
```suggestion
            schema.RuleAction.SCHEDULE_BETWEEN_TIME_RANGE.value: lambda a, b: compare_utc_time_range(a, b),
```

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -59,6 +64,9 @@ def _match_by_action(self, action: str, condition_value: Any, context_value: Any
...
+            schema.RuleAction.SCHEDULE_BETWEEN_DATETIME_RANGE.value: lambda a, b: compare_utc_datetime_range(a, b),
+            schema.RuleAction.SCHEDULE_BETWEEN_DAYS_OF_WEEK.value: lambda a, b: compare_between_utc_days(a, b),
```
rubenfonseca:
```suggestion
            schema.RuleAction.SCHEDULE_BETWEEN_DAYS_OF_WEEK.value: lambda a, b: compare_utc_days_of_week(a, b),
```

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -228,7 +244,7 @@ def evaluate(self, *, name: str, context: Optional[Dict[str, Any]] = None, defau
         # method `get_matching_features` returning Dict[feature_name, feature_value]
         boolean_feature = feature.get(
             schema.FEATURE_DEFAULT_VAL_TYPE_KEY, True
-        )  # backwards compatability ,assume feature flag
+        )  # backwards compatibility ,assume feature flag
```
rubenfonseca: while we are here :) 

```suggestion
        )  # backwards compatibility, assume feature flag
```

aws_lambda_powertools/utilities/feature_flags/feature_flags.py
```diff
@@ -287,7 +303,7 @@ def get_enabled_features(self, *, context: Optional[Dict[str, Any]] = None) -> L
             feature_default_value = feature.get(schema.FEATURE_DEFAULT_VAL_KEY)
             boolean_feature = feature.get(
                 schema.FEATURE_DEFAULT_VAL_TYPE_KEY, True
-            )  # backwards compatability ,assume feature flag
+            )  # backwards compatibility ,assume feature flag
```
rubenfonseca:
```suggestion
            )  # backwards compatibility, assume feature flag
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -31,6 +32,27 @@ class RuleAction(str, Enum):
...
     VALUE_IN_KEY = "VALUE_IN_KEY"
     VALUE_NOT_IN_KEY = "VALUE_NOT_IN_KEY"
+    SCHEDULE_BETWEEN_TIME_RANGE = "SCHEDULE_BETWEEN_TIME_RANGE"  # hour:min 24 hours clock UTC time
+    SCHEDULE_BETWEEN_DATETIME_RANGE = "SCHEDULE_BETWEEN_DATETIME_RANGE"  # full datetime format
+    SCHEDULE_BETWEEN_DAYS_OF_WEEK = "SCHEDULE_BETWEEN_DAYS_OF_WEEK"
+
+
+class TimeKeys(Enum):
+    CURRENT_TIME_UTC = "CURRENT_TIME_UTC"
+    CURRENT_DAY_UTC = "CURRENT_DAY_UTC"
```
rubenfonseca: One more that I forgot. Otherwise I would think these are numbers (1-31)

```suggestion
    CURRENT_DAY_OF_WEEK_UTC = "CURRENT_DAY_OF_WEEK_UTC"
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -264,9 +290,79 @@ def validate_condition_key(condition: Dict[str, Any], rule_name: str):
...
+                f"'condition with a 'SCHEDULE_BETWEEN_TIME_RANGE' action must have a 'CURRENT_TIME_UTC' condition key, rule={rule_name}"  # noqa: E501
+            )
+        if action == RuleAction.SCHEDULE_BETWEEN_DATETIME_RANGE.value and key != TimeKeys.CURRENT_DATETIME_UTC.value:
+            raise SchemaValidationError(
+                f"'condition with a 'SCHEDULE_BETWEEN_DATETIME_RANGE' action must have a 'CURRENT_DATETIME_UTC' condition key, rule={rule_name}"  # noqa: E501
+            )
+        if action == RuleAction.SCHEDULE_BETWEEN_DAYS_OF_WEEK.value and key != TimeKeys.CURRENT_DAY_UTC.value:
+            raise SchemaValidationError(
+                f"'condition with a 'SCHEDULE_BETWEEN_DAYS_OF_WEEK' action must have a 'CURRENT_DAY_UTC' condition key, rule={rule_name}"  # noqa: E501
+            )
```
rubenfonseca:
```suggestion
        if action == RuleAction.SCHEDULE_BETWEEN_DAYS_OF_WEEK.value and key != TimeKeys.CURRENT_DAY_OF_WEEK_UTC.value:
            raise SchemaValidationError(
                f"'condition with a 'SCHEDULE_BETWEEN_DAYS_OF_WEEK' action must have a 'CURRENT_DAY_OF_WEEK_UTC' condition key, rule={rule_name}"  # noqa: E501
            )
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -264,9 +290,79 @@ def validate_condition_key(condition: Dict[str, Any], rule_name: str):
...
+        if action == RuleAction.SCHEDULE_BETWEEN_TIME_RANGE.value:
+            ConditionsValidator._validate_schedule_between_time_and_datetime_ranges(
+                value, rule_name, action, ConditionsValidator.TIME_RANGE_FORMAT
+            )
+        elif action == RuleAction.SCHEDULE_BETWEEN_DATETIME_RANGE.value:
+            ConditionsValidator._validate_schedule_between_time_and_datetime_ranges(
+                value, rule_name, action, ConditionsValidator.DATETIME_RANGE_FORMAT
+            )
+        elif action == RuleAction.SCHEDULE_BETWEEN_DAYS_OF_WEEK.value:
+            ConditionsValidator._validate_schedule_between_days(value, rule_name)
```
rubenfonseca:
```suggestion
            ConditionsValidator._validate_schedule_between_days_of_week(value, rule_name)
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -264,9 +290,79 @@ def validate_condition_key(condition: Dict[str, Any], rule_name: str):
...
+    def _validate_time_value(time: str, rule_name: str, date_format: str):
+        try:
+            datetime.strptime(time, date_format)
+        except Exception:
+            raise SchemaValidationError(
+                f"'START' and 'END' must be a valid time format, time_format={date_format}, rule={rule_name}"
+            )
+
+    @staticmethod
+    def _validate_schedule_between_days(value: Any, rule_name: str):
```
rubenfonseca:
```suggestion
    def _validate_schedule_between_days_of_week(value: Any, rule_name: str):
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -264,9 +290,79 @@ def validate_condition_key(condition: Dict[str, Any], rule_name: str):
...
+        if not isinstance(value, list) or not value:
+            raise SchemaValidationError(
+                f"condition with a CURRENT_DAY_UTC action must have a non empty condition value type list, rule={rule_name}"  # noqa: E501
```
rubenfonseca:
```suggestion
                f"condition with a CURRENT_DAY_OF_WEEK_UTC action must have a non empty condition value type list, rule={rule_name}"  # noqa: E501
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -264,9 +290,79 @@ def validate_condition_key(condition: Dict[str, Any], rule_name: str):
...
+                TimeValues.SUNDAY.value,
+                TimeValues.MONDAY.value,
+                TimeValues.TUESDAY.value,
+                TimeValues.WEDNESDAY.value,
+                TimeValues.THURSDAY.value,
+                TimeValues.FRIDAY.value,
+                TimeValues.SATURDAY.value,
+            ]:
+                raise SchemaValidationError(
+                    f"condition value must represent a week day string defined in 'TimeValues' enum, rule={rule_name}"
```
rubenfonseca: week day for me is Monday - Frday

```suggestion
                    f"condition value must represent a day of the week string defined in 'TimeValues' enum, rule={rule_name}"
```

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,34 @@
...
+from .schema import TimeValues
+
+HOUR_MIN_SEPARATOR = ":"
+
+
+def _get_utc_time_now() -> datetime:
+    return datetime.now(timezone.utc)
+
+
+def compare_between_utc_days(action: str, values: List[str]) -> bool:
```
rubenfonseca:
```suggestion
def compare_utc_days_of_week(action: str, values: List[str]) -> bool:
```

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,34 @@
...
+    current_day = _get_utc_time_now().strftime("%A").upper()
+    return current_day in values
+
+
+def compare_utc_datetime_range(action: str, values: Dict) -> bool:
+    current_time_utc: datetime = _get_utc_time_now()
+    start_date = datetime.strptime(values.get(TimeValues.START.value, ""), "%Y-%m-%dT%H:%M:%S%z")
+    end_date = datetime.strptime(values.get(TimeValues.END.value, ""), "%Y-%m-%dT%H:%M:%S%z")
```
rubenfonseca: We already use this datetime format string in multiple places, including in the `ConditionsValidator` class. Can we move it to a more global constant and use it everywhere, instead of repeating it?

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,34 @@
...
+    return current_time_utc >= start_date and current_time_utc <= end_date
+
+
+def compare_utc_date_range(action: str, values: Dict) -> bool:
```
rubenfonseca:
```suggestion
def compare_utc_time_range(action: str, values: Dict) -> bool:
```

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,34 @@
...
+    current_time_utc: datetime = _get_utc_time_now()
+    start_hour, start_min = values.get(TimeValues.START.value, HOUR_MIN_SEPARATOR).split(HOUR_MIN_SEPARATOR)
+    end_hour, end_min = values.get(TimeValues.END.value, HOUR_MIN_SEPARATOR).split(HOUR_MIN_SEPARATOR)
```
rubenfonseca: I'm not sure I understand the logic here. If we don't have a START or END value, we're returning `":"`, which, when splitted, produces a tuple `("", "")`, which will crash on line 30.

Since the data is already validated, why don't we just return "" ?

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,34 @@
...
+    return (
+        current_time_utc.hour >= int(start_hour)
+        and current_time_utc.hour <= int(end_hour)
+        and current_time_utc.minute >= int(start_min)
+        and current_time_utc.minute <= int(end_min)
+    )
```
rubenfonseca: NOTE TO SELF: make sure that the documentation states that all intervals **include the boundaries**

rubenfonseca (changes requested):
First pass on the code, I'll review the tests in the afternoon.

Note: right now, the following use case fails, but I think it's a valid one. Do you think we can make it work too?

```json
{
    "my_feature": {
        "default": false,
        "rules": {
            "during silent hours at night": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "SCHEDULE_BETWEEN_TIME_RANGE",
                        "key": "CURRENT_TIME_UTC",
                        "value": {
                            "START_TIME": "23:00",
                            "END_TIME": "7:00"
                        }
                    }
                ]
            }
        }
    }
}
````

File: aws_lambda_powertools/utilities/feature_flags/time_conditions.py (hunk 18 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/time_conditions.py]: yeah i refactored it and didnt use it here. np

**Comment by ran-isenberg** [comment]
> First pass on the code, I'll review the tests in the afternoon.
> 
> Note: right now, the following use case fails, but I think it's a valid one. Do you think we can make it work too?
> 
> ```json
> {
>     "my_feature": {
>         "default": false,
>         "rules": {
>             "during silent hours at night": {
>                 "when_match": true,
>                 "conditions": [
>                     {
>                         "action": "SCHEDULE_BETWEEN_TIME_RANGE",
>                         "key": "CURRENT_TIME_UTC",
>                         "value": {
>                             "START_TIME": "23:00",
>                             "END_TIME": "7:00"
>                         }
>                     }
>                 ]
>             }
>         }
>     }
> }
> ```

Interesting! I think i had a hidden assumption that you dont go over to the next day in this case. It will fail on schema validation right on the start. This is a tricky one.
One could argue that you could use two rules to get the same capability -  23-24 and 0-7 but that's less ideal.

if we allow such use case, is 7:00-6:00 also accepted? It weakens the schema validation as i didnt allow more than 24 hours use cases.

File: aws_lambda_powertools/utilities/feature_flags/time_conditions.py (hunk 20 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/time_conditions.py]: it's a pylint issue, it has to be a string and not None. why ":"? , it might not matter anyways since it's going to crash anyways in line 30 like you mentioned. 
I dont mind changing, it doesnt occur anyways, it's for pylint/mypy sake.

**Comment by rubenfonseca** [comment]
> One could argue that you could use two rules to get the same capability - 23-24 and 0-7 but that's less ideal.

Agree, let's not complicate this, I totally forgot you could add more than one rule :)

File: aws_lambda_powertools/utilities/feature_flags/time_conditions.py (hunk 20 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/time_conditions.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/time_conditions.py (hunk 19 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/time_conditions.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/time_conditions.py (hunk 17 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/time_conditions.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 11 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 12 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 13 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 14 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 15 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 16 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: i'll align it to match most of the world. we are special, it's my day off here.
starting with monday.

File: aws_lambda_powertools/utilities/feature_flags/feature_flags.py (hunk 10 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/feature_flags.py]: fixed both

File: aws_lambda_powertools/utilities/feature_flags/feature_flags.py (hunk 9 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/feature_flags.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/feature_flags.py (hunk 8 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/feature_flags.py]: fixed

File: aws_lambda_powertools/utilities/feature_flags/feature_flags.py (hunk 7 above)
Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/feature_flags.py]: fixed

**Comment by ran-isenberg** [comment]
@rubenfonseca  pushed all the fixes for the first pass.

File: aws_lambda_powertools/utilities/feature_flags/schema.py (hunk 16 above)
Comment (rubenfonseca) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: The point here was about "week day" vs "day of the week", not which days are work days :P

Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: lol

Comment (ran-isenberg) [reply aws_lambda_powertools/utilities/feature_flags/schema.py]: fixed

**Comment by rubenfonseca** [comment]
Status update: I took the liberty of re-writing some of the tests so they are easy to read and to navigate.
Next week we'll look into ironing out any performance quick-wins, and writing the feature documentation.

**Comment by ran-isenberg** [comment]
> Status update: I took the liberty of re-writing some of the tests so they are easy to read and to navigate. Next week we'll look into ironing out any performance quick-wins, and writing the feature documentation.

np, i'll take a look later but it seems that the tests are broken

**Comment by ran-isenberg** [comment]
The tests look much better now @rubenfonseca

**Comment by ran-isenberg** [comment]
@rubenfonseca  do you know why the tests work locally but fail on github?

**Comment by rubenfonseca** [comment]
@ran-isenberg tests were failing locally for me too, fixed them

Pushed the changes Heitor requested around using `datetime.fromisoformat`. Can you please check if it makes sense for you? Next step for me is to write documentation for the new feature.

File: tests/functional/feature_flags/test_schema_validation.py [hunk 22]
```diff
@@ -678,23 +678,16 @@ def test_validate_time_condition_between_datetime_range_invalid_condition_value_
         ConditionsValidator.validate_condition_value(condition=condition, rule_name=rule_name)
 
 
-@pytest.mark.parametrize(
-    "cond_value",
-    [
-        {TimeValues.END.value: "10:10", TimeValues.START.value: "2022-10-10T12:15:00Z"},
-        {TimeValues.END.value: "2022-10-10", TimeValues.START.value: "2022-10-10T12:15:00Z"},
```
rubenfonseca: Note: `2022-10-10` is actually a correct ISO8601 format parsed by Python, corresponding to `2022-10-10T00:00:00Z`.

ran-isenberg:
> @ran-isenberg tests were failing locally for me too, fixed them
> 
> Pushed the changes Heitor requested around using `datetime.fromisoformat`. Can you please check if it makes sense for you? Next step for me is to write documentation for the new feature.

looks good to me!

rubenfonseca:
@heitorlessa re-requesting your review to take a look at the docs (and anything else that your eye catches)

ran-isenberg:
@heitorlessa @rubenfonseca  @leandrodamascena  is there a chance to merge this before the holidays?

heitorlessa:
That's the plan @ran-isenberg.

Ruben had an emergency and is unreachable, and Leandro is on paternity leave.

I'll work on my vacation to complete the remaining parts:

- Docs
- Refactor UTC focus for timezone-aware (UTC by default)
- Performance assessment
- Last maintainability check (tests, internal implementation, ease of understanding, and extensibility)

Got 6 more hours flight (13 so far), and will look into it again next week on my interim timezone for the next 5 weeks (GMT-3).

If I can't get it done in 10 days, this will be for next year - it took us more than expected to get this PR in a good shape to merge. I'm cutting 10 days from my originally planned vacation to make additional progress.

heitorlessa:
In 2023, Ruben, Leandro and I will create a more detailed guideline on what good contribution looks like before we merge -- that is in addition to the checks we have in the body (tests, docs, etc).

This should, at least in theory, lessen the effort on status updates from both parts, and be clearer when we will deprioritise PRs if we don't have the bandwidth to improve them to a good maintainable level - like we've been trying on this PR for the last 30 days

ran-isenberg:
> That's the plan @ran-isenberg.
> 
> Ruben had an emergency and is unreachable, and Leandro is on paternity leave.
> 
> I'll work on my vacation to complete the remaining parts:
> 
> * Docs
> * Refactor UTC focus for timezone-aware (UTC by default)
> * Performance assessment
> * Last maintainability check (tests, internal implementation, ease of understanding, and extensibility)
> 
> Got 6 more hours flight (13 so far), and will look into it again next week on my interim timezone for the next 5 weeks (GMT-3).
> 
> If I can't get it done in 10 days, this will be for next year - it took us more than expected to get this PR in a good shape to merge. I'm cutting 10 days from my originally planned vacation to make additional progress.

I admire your dedication my friend! but please, don't work on your vacation on this PR!
I thought we finished the work here and didn't realize there were more open issues on your side.

I think it comes down to transparency and coordinating expectations.
Please enjoy your vacation, it will wait.

heitorlessa:
Thanks a lot @ran-isenberg - I spoke to Ruben and he shared he felt ill before he was able to share the areas of updates. This only reinforces the need for what I shared earlier to help both sides - we need a mechanism on:

* **1.** Continuously check whether we provided enough details on what's required for merging any PR. This keeps us honest, and it prevents confusion on whoever takes any PR next assuming it was shared on private channels - if it's not public we **must** assume it wasn't discussed.
* **2.** Provide guidelines, [similar to RFCs](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/MAINTAINERS.md#triage-rfcs), on what maintainers should be looking at PRs, and which PRs will be fast track for merging if X Y Z is included.

I'm gonna create a separate comment for details on what Ruben couldn't share, so we start January with the right foot.

heitorlessa:
Expanding on my comment earlier with more details on what's pending to merge

## PR details

We need a summary of what's being changed in the feature flags code - a list of bullet points helps anyone reviewing OR revisiting this change in the future to understand its reasoning.

We need a few sample schemas in the UX section demonstrating what this enables. This makes reviewing much faster (at least a day), and we can quickly contribute editing the PR details if we together find better ways. We will need that for the docs.

## Docs

Could you share a realistic use case we could use to emphasize and word docs around it? This would speed up at least 3-5 days depending who's completing docs.

## Make it timezone-aware

One of the issues Ruben identified was that while UTC is great by default, it'll soon enough lead to production issues for customers in AMER/APAC, where choosing a time in UTC will be error prone.

**Idea**

Since the Schema supports `ANY` as `value`, we could include an optional key named `timezone` where the default is UTC. The value for `timezone` being an [IANA identifie](https://data.iana.org/time-zones/tzdb-2020f/europe)r, e.g. `Europe/Amsterdam`. [**Benefits**] This makes it easier to prevent mistakes like a selected day `MONDAY` already happened in UTC, and whoever is authoring OR reading can do so in a less error-prone way.

Lambda runtime includes [python-dateutil](https://dateutil.readthedocs.io/en/stable/), which means we could use its built-in TZ DB in case Lambda removed from the base image.

**Example**

```json
{
    "my_feature": {
        "default": false,
        "rules": {
            "match when lambda time is between 11:00-23:00 and day is either monday or thursday": {
                "when_match": true,
                "conditions": [
                    {
                        "action": "TIME_RANGE",
                        "key": "CURRENT_HOUR_UTC",
                        "value": {
                            "START_TIME": "11:00",
                            "END_TIME": "23:00",
                            "timezone": "Europe/Amsterdam"        # HERE
                        }
                    },
                    {
                        "action": "TIME_SELECTED_DAYS",
                        "key": "CURRENT_DAY_UTC",
                        "value": {
                            "days": ["MONDAY", "THURSDAY"],    # HERE, to accommodate dict syntax
                           "timezone": "Europe/Amsterdam"         # HERE
                        }
                    }
                ]
            }
        }
    }
}
```

## Performance assessment

Line profiler to assess whether we're not missing any excessive memory allocation or CPU cycles when dealing with dates and schema validation. I ran one in my quick review and discovered that manual date parsing with string formatting was up to 40x slower compared to `fromisoformat`.

## Maintainability checks

Ruben made some changes to tests but the `expected_value` shouldn't be a parameter, instead we should return it and assert within our tests (less future surprises).

We should review whether the amount of logging is helpful (debug/info) for customers looking to understand why X doesn't work in the future - Feature Flags is the only one we accept a Logger instance, it's great we should do to others, including accept a Metrics instance.

Last but not least, we should verify whether we have appropriate comments and docstrings so anyone unfamiliar with the code isn't overwhelmed with cognitive load (short-term and working memory). Some might be examples, others might be a high-level flow with the decisions being made during evaluation at the highest possible level.

Please let me know if this isn't clearer. This is what I expected to have been communicated earlier.

ran-isenberg:
> PR details

@heitorlessa @rubenfonseca 
Added PR details at the first comment of the PR as your asked, I hope it's clearer now. It describes the current state.

Docs - I think the use cases at the PR description make sense but I can try and add:
I've had feature flags for maintenance mode that we wanted  to enable only on the weekend.
Another use case it to disable the call support/support chat feature after working hours (after 17:00), you can also add the work days example in this case.
Another use case was that we wanted to launch a new feature stating a specific date and time (full datetime) and enable it for lets say a period of two months, which after that, the feature would become permanent and retired as it's already stable etc.

About the timezone - maybe add it in another ticket? or ask community feedback on this? I think it's going to complicate things and make the schema even harder to write. Perhaps if you make it numerical - UTC +/- X it can be simpler but then you need to take into account summer/winter time etc.



expected_value - i made the change you asked and will push it shortly.

Waiting for your feedback and happy holidays :)

rubenfonseca:
Hi @ran-isenberg I'm back now and looking into this PR. Based on the feedback, I've updated the PR description with a list of tasks do we want to get done before merging.

The biggest is the removal of the `_UTC` suffix, and related logic, which we believe is essential to get this feature merged, otherwise it will have few practical applications for the feature. Heitor left good points on the UX and the use of `python-dateutil`. This way we could things like `Europe/Copenhagen` and get all the DST calculations for free.

I will start taking tasks top to bottom, but let me know if you can take any of the tasks, and go for it!

ran-isenberg:
> Hi @ran-isenberg I'm back now and looking into this PR. Based on the feedback, I've updated the PR description with a list of tasks do we want to get done before merging.
> 
> The biggest is the removal of the `_UTC` suffix, and related logic, which we believe is essential to get this feature merged, otherwise it will have few practical applications for the feature. Heitor left good points on the UX and the use of `python-dateutil`. This way we could things like `Europe/Copenhagen` and get all the DST calculations for free.
> 
> I will start taking tasks top to bottom, but let me know if you can take any of the tasks, and go for it!

I'm afraid i wont have a lot of time in the 1-2 of weeks as I need to prepare my presentation for the AWS community day TLV and a couple of podcast appearances i attend this month.

However, seems like there's only one epic right? removing UTC and adding a field for time zone which means:
1. Defining the approved time zone values
2. Adding support in the schema
3. Doing schema validation + add tests for invalid values
4. Updating the tests with timezones

@rubenfonseca  can you start with task number 1 so we'r aligned? i might be able to squeeze some work but #1 is a must for any parallel work here.

rubenfonseca:
I pushed a change to remove one more `strptime` from the code, which again was the biggest cause for slowdowns.

Before the change, on a 100.000 iteration test, we were spending a lot of time on `strptime` inside `validate_time_value`:

image

image

After the change, we spend 10x less :) 

image

rubenfonseca:
> @rubenfonseca can you start with task number 1 so we'r aligned? i might be able to squeeze some work but #1 is a must for any parallel work here.

I believe Heitor [mentioned](https://github.com/awslabs/aws-lambda-powertools-python/pull/1666#issuecomment-1353140788) supporting IANA format which [python-dateutil](https://dateutil.readthedocs.io/en/stable/index.html) takes care out of the box. So we just try to parse the tz with `dateutil`, and fail the schema validation if we can't parse it. Or use UTC if nothing is provided.

rubenfonseca:
@ran-isenberg I've also updated the PR description with the target UX

ran-isenberg:
@rubenfonseca  doesnt the regex allows times such as 29:11?

rubenfonseca:
> @rubenfonseca doesnt the regex allows times such as 29:11?

Yes :) and no :) https://github.com/awslabs/aws-lambda-powertools-python/pull/1666/files#diff-5d91869726bc42ea856402ece418345fd499512aaecdcdb2fd8363437fa9ac90R339

ran-isenberg:
i think it's easier to use the following regex:
[2][0-4]:[0-5]\d|[0-1]\d:[0-5]\d

and also remove the capture group and the special if. less logic. WDYT?

rubenfonseca:
> i think it's easier to use the following regex:
> [2][0-4]:[0-5]\d|[0-1]\d:[0-5]\d

great idea! Just make sure you use `0-3` instead of `0-4` :P

ran-isenberg:
> > i think it's easier to use the following regex:
> > [2][0-4]:[0-5]\d|[0-1]\d:[0-5]\d
> 
> great idea! Just make sure you use `0-3` instead of `0-4` :P

yep :) even simpler:

2[0-3]:[0-5]\d|[0-1]\d:[0-5]\d

rubenfonseca:
Feedback is welcome, the only thing left on my todo list is to improve debug logging.

ran-isenberg:
> Feedback is welcome, the only thing left on my todo list is to improve debug logging.

@rubenfonseca Awesome! I'll take a look tomorrow/Saturday.

ran-isenberg:
@rubenfonseca The code looks great.
I added another docs example for datetime use case and also fixed the timebased.
I think you meant that a premium tier customer AND also during specific hours would get a discount - two conditions under the same rule.
LGTM ;)

rubenfonseca:
@ran-isenberg cool thanks! Let's aim to release this next week after @leandrodamascena review!

leandrodamascena:
Hi @rubenfonseca and @ran-isenberg! One more PR to review and learn :smile: 

I started the review yesterday and hope to finish it on Monday! For now, I've created some scenarios to simulate a real use of this feature and so far so good.

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,56 @@
...
+    current_time: datetime = _get_now_from_timezone(gettz(timezone_name))
+
+    start_hour, start_min = values.get(TimeValues.START.value, "").split(HOUR_MIN_SEPARATOR)
+    end_hour, end_min = values.get(TimeValues.END.value, "").split(HOUR_MIN_SEPARATOR)
+    return (
+        current_time.hour >= int(start_hour)
+        and current_time.hour <= int(end_hour)
+        and current_time.minute >= int(start_min)
+        and current_time.minute <= int(end_min)
+    )
```
leandrodamascena: This code can cause problems when using start and end times with exact times, for example, start: 10:00 am and end: 11:00 am. This is also valid for any start and/or end where the minutes are 00. See the code example I made below:

image

I fixed this code locally and it is working as expected. My only concern here is why do I need to remove datetime type for the current_time variable because `strftime` returns a string. If we want to keep typing, we need to convert the dates using `strptime`, but from what I see you have removed that function due to performance.

```suggestion
    timezone_name = values.get(TimeValues.TIMEZONE.value, "UTC")
    current_hour = _get_now_from_timezone(gettz(timezone_name)).strftime("%H:%M")

    start_hour = values.get(TimeValues.START.value, "")
    end_hour = values.get(TimeValues.END.value, "")

    return start_hour <= current_hour <= end_hour
```

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
```diff
@@ -0,0 +1,56 @@
+from datetime import datetime, tzinfo
+from typing import Dict, Optional
+
+from dateutil.tz import gettz
+
+from .schema import HOUR_MIN_SEPARATOR, TimeValues
```
leandrodamascena, line 6:
```suggestion
from .schema import TimeValues
```

aws_lambda_powertools/utilities/feature_flags/schema.py
```diff
@@ -14,9 +18,12 @@
 CONDITION_VALUE = "value"
 CONDITION_ACTION = "action"
 FEATURE_DEFAULT_VAL_TYPE_KEY = "boolean_type"
+TIME_RANGE_FORMAT = "%H:%M"  # hour:min 24 hours clock
+TIME_RANGE_RE_PATTERN = re.compile(r"2[0-3]:[0-5]\d|[0-1]\d:[0-5]\d")  # 24 hour clock
+HOUR_MIN_SEPARATOR = ":"
```
leandrodamascena, line 23:
```suggestion
```

tests/functional/feature_flags/test_time_based_actions.py
```diff
@@ -0,0 +1,476 @@
...
+                            TimeValues.START.value: "11:11",
+                            TimeValues.END.value: "23:59",
+                            TimeValues.TIMEZONE.value: timezone_name,
+                        },
+                    },
+                ],
+            }
+        },
+        mocked_time=(2022, 2, 15, 11, 11, 0, gettz(timezone_name)),  # rule match 11:11 am, Europe/Copenhagen
+    )
```
leandrodamascena, line 189: We should add a new test contemplating intervals between 11:00 and 13:00, for example.

leandrodamascena (changes requested):
Hi @rubenfonseca and @ran-isenberg! 
I made some suggestions of what could be changed to improve this PR, but it's a great job indeed!

aws_lambda_powertools/utilities/feature_flags/time_conditions.py
rubenfonseca: Thanks for this @leandrodamascena ! I digg a little further and doing a lexical comparison would also fail. However, I've addressed the problem by doing a little bit more work. I've updated the code and added some explanation. I also added a few new tests to test your scenario and two new ones. Please let me know how it looks now :)

leandrodamascena (approved):
@rubenfonseca thanks for the fix and feedback! I have no further considerations, everything is working fine!

Hoping to hear something from @heitorlessa before merge it.

image

heitorlessa:
hey! back today and got a few internal escalations to handle - ship it if @leandrodamascena and @rubenfonseca are happy

leandrodamascena:
@rubenfonseca and @ran-isenberg! Before merging, we should rebase this branch to avoid unnecessary changes like this https://github.com/awslabs/aws-lambda-powertools-python/pull/1666/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

ran-isenberg:
@heitorlessa @rubenfonseca @leandrodamascena Thank you guys so much for the time and effort you put into this major feature!

ran-isenberg:
when do you expect it to get merged and released?

leandrodamascena:
> when do you expect it to get merged and released?

Hey Ran! Yesterday and today the team was in an internal workshop and we were not able to merge this. Tomorrow we can merge that and start planning the next release.

leandrodamascena:
This PR was closed without merging due to the fact that there were some unnecessary changes in the files and the rebase would be complicated. In consensus with @ran-isenberg, the original author of this PR, it was agreed that a new PR would be opened.

Closed in favor of #1846

PR #1669, opened 2022-10-28
fix(idempotency): idempotent_function should support standalone falsy values

### LINKED ISSUE #1663: Feature request: Support `raise_on_no_idempotency_key` for `idempotent_function`
### Use case

Trying to use `idempotent_function` in an "optional" manner. Example:

```
import random

@idemp_f(data_keyword_argument="key")
def echo(key: str = None):
    return random.randint(1, 1000000)
```

When called like `echo(key=None)` will raise `RuntimeError`, even when `raise_on_no_idempotency_key` is set to False

### Solution/User Experience

Unless there is good reason not to respect the `IdempotencyConfig.raise_on_no_idempotency_key` value. Perhaps adding a similar parameter to `idempotent_function` like `raise_on_no_data_keyword_argument` or something to that effect.

### Alternative solutions

_No response_

### Acknowledgment

heitorlessa:
**Issue number:** #1663

## Summary

When using `idempotent_function` for any synchronous Python function (different from Lambda Handler), customers set which keyword argument we should use to hash data as an idempotency key.

We raise `RuntimeError` for cases where the keyword argument wasn't used correctly to prevent hashing the incorrect data. For example, function was called without the keyword argument (`function()`, instead of `function(data=data)`).

However, the current logic incorrectly checks the value available in the keyword argument, rather than the keyword argument was used. The following code leads to a `RuntimeError` when we should be serializing `None->null` just fine.

```python
...

import random

@idemp_f(data_keyword_argument="key")
def echo(key: str = None):
    return random.randint(1, 1000000)

echo(key=None)
```

### Changes

This PR corrects the logic and only raises `RuntimeError` if the keyword argument specified for hashing isn't in the list of keyword arguments available at the function call.

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/idempotency.py
+++ b/aws_lambda_powertools/utilities/idempotency/idempotency.py
@@ -134,13 +134,13 @@ def idempotent_function(
         if os.getenv(constants.IDEMPOTENCY_DISABLED_ENV):
             return function(*args, **kwargs)
 
-        payload = kwargs.get(data_keyword_argument)
-
-        if not payload:
+        if data_keyword_argument not in kwargs:
             raise RuntimeError(
                 f"Unable to extract '{data_keyword_argument}' from keyword arguments."
                 f" Ensure this exists in your function's signature as well as the caller used it as a keyword argument"
             )
+
+        payload = kwargs.get(data_keyword_argument)
 
         idempotency_handler = IdempotencyHandler(
             function=function,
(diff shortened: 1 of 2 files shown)


leandrodamascena (approved):
Nothing to change! LGTM!

PR #1670, opened 2022-10-28
fix(logger): fix unknown attributes being ignored by mypy

### LINKED ISSUE #1660: Static typing: Logger `__getattr__` definition causes mypy to not error when missing an attribute
### Static type checker used

mypy (project's standard)

3.9

latest

### Static type checker info

The issue here is a lack of output. Running `mypy` with no arguments on the file below does not produce any issues.

### Code snippet

```python

Logger().foobar()
```

### Possible Solution

Based on https://github.com/python/mypy/issues/13319, the preferred solution seems to be the following:

```python
from typing import TYPE_CHECKING

class Logger(logging.Logger):
    ...
    if not TYPE_CHECKING:
        def __getattr__(self, name):
            ...
```

This appears to resolve the issue on my machine.

nayaverdier:
**Issue number:** #1660

### Changes

Hides the `__getattr__` function in `Logger` from mypy and other type checkers

### User experience

Unknown attributes were not being flagged by mypy (e.g. `logger.foobar()` appeared as valid). Now, they appear as undefined attributes.

## Checklist

* [ ] Changes are documented

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -8,6 +8,7 @@ import sys
 import traceback
 from typing import (
     IO,
+    TYPE_CHECKING,
     Any,
     Callable,
     Dict,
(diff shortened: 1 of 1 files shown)


heitorlessa (approved):

heitorlessa:
thank you so much @nayaverdier for the quick turnaround - Making this part of today's release :)

Issue #1040, opened 2022-02-24
Feature request: utility function to get multiple parameters by name instead of path


hooverdc:
**Is your feature request related to a problem? Please describe.**

I have several functions in a severless application with several SSM parameters under the same path. Each function only needs access to some parameters and most functions will share some parameters but not all. I don't want functions to have access to parameters they don't require. I also don't want to call `get_parameter` for each parameter in order to keep cold start times down. I would like to be able to specify the parameter names and return them in one batch.

**Describe the solution you'd like**

A `get_parameters_by_name` function that accepts a list of strings of parameter names and returns a dict similar to the `get_parameters` function, but with the full parameter path as the key (presumably you could call this with parameters on different paths).

https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameters.html#get-parameters

**Describe alternatives you've considered**

Writing my own wrapper function that calls `get_parameters` from `boto3`.

**Additional context**
N/A

heitorlessa:
Hey @hooverdc, thanks for creating this feature request! I believe this is best handled by your own custom wrapper, or perhaps approaching this differently.
 
>  I don't want functions to have access to parameters they don't require. I also don't want to call get_parameter for each parameter in order to keep cold start times down. 

Without knowing the number of parameters you have, their size, and how you create them, have you considered creating an aggregate parameter with multiple values?

**Example**. This allows e2e/integ scenarios to make a single call and get everything they need, while still having separate parameters that can be useful for other purposes like deployment time.

```yaml
Resources:
    ResourcesMapParameter:
        Type: AWS::SSM::Parameter
        Properties:
            Name: !Sub "/${Stage}/service/trip_ingestion/test/config"
            Type: String
            Description: Service Map with common resources used for integ/e2e testing
            Value: !Sub |
                {
                    "MobileTripsBucket": "${MobileTripsBucket}",
                    "MobileTripsQueue": "${MobileTripsQueue}",
                    "TripIngestionFunction": "${TripIngestionFunction}",
                    "MobileTripsApiEndpoint": "https://${MobileTripsApi}.execute-api.${AWS::Region}.amazonaws.com/${MobileTripsApi.Stage}/trip",
                    "Stage": "${Stage}"
                }
```

Hope that helps!

Thanks

hooverdc:
[...]

PR #1678, opened 2022-11-02
feat(parameters): add get_parameters_by_name for SSM params in distinct paths

heitorlessa:
**Issue number:** #1040

## Summary

Today, customers can fetch multiple SSM parameters recursively by using a `path`. This happens server-side.

As described in the issue, customers may need to fetch multiple parameters in distinct paths. As of now, this requires additional boilerplate to get it done properly.

### Changes

This PR introduces `get_parameter_by_name` high-level function that accepts a dictionary containing:

* A parameter to fetch
* A dict optionally containing overrides for a parameter (e.g., `max_age`, `transform`, `decrypt`, etc.)

#### Non-related but necessary changes

* Adds `typing-extensions` as a dependency to provide strict typing for this feature. Depending on the `transform` value, the return type is precise.
* Refactored `transform_method` and `transform_value` to better support strict typing and related bugs
* Created `add_to_cache` method to support refactoring and handle `max_age=0, max_age=-1` scenarios we didn't cover before
* Exposed `has_not_expired_in_cache` method to ease testing

### User experience

```python
from typing import Any

from aws_lambda_powertools.utilities import get_parameters_by_name

parameters = {
  "/develop/service/commons/telemetry/config": {"max_age": 300, "transform": "json"},
  "/no_cache_param": {"max_age": 0},
  # inherit default values
  "/develop/service/payment/api/capture/url": {},
}

def handler(event, context):
    # This returns a dict with the parameter name as key
    response: dict[str, Any] = parameters.get_parameters_by_name(parameters=parameters, max_age=60)
    for parameter, value in response.items():
        print(f"{parameter}: {value}")
```

**Overrides fail-fast to support graceful error handling**

```python
from typing import Any

from aws_lambda_powertools.utilities import get_parameters_by_name

parameters = {
  "/develop/service/commons/telemetry/config": {"max_age": 300, "transform": "json"},
  # it would fail by default
  "/this/param/does/not/exist"
}

def handler(event, context):
    values: dict[str, Any] = parameters.get_parameters_by_name(parameters=parameters, raise_on_error=False)
    errors: list[str] = values.get("_errors", [])

    # Handle gracefully, since '/this/param/does/not/exist' will only be available in `_errors`
    if errors:
        ...

    for parameter, value in values.items():
        print(f"{parameter}: {value}")
```

## Checklist

* [x] Changes are documented
* [x] Test whether threads can be faster or not (Nope due to CPU slicing in Lambda + LWP)
* [x] Test whether async can be faster or not (Nope due to event loop setup + aioboto need)
* [x] Use `GetParameters` as default. Use `GetParameter` when `decrypt` is set on a per parameter
* [x] Functional test for options override
* [x] Functional test for partial failure

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/shared/functions.py
+++ b/aws_lambda_powertools/shared/functions.py
@@ -4,7 +4,7 @@ import logging
 import os
 import warnings
 from binascii import Error as BinAsciiError
-from typing import Optional, Union, overload
+from typing import Dict, Generator, Optional, Union, overload
 
 from aws_lambda_powertools.shared import constants
 
@@ -118,11 +118,6 @@ def powertools_debug_is_set() -> bool:
     return False
 
 
-def slice_dictionary(data, chunk_size: int):
-    # save CPU cycles if input is already small than chunk_size
-    if len(data) <= chunk_size:
-        yield data
-
-    data_iterator = iter(data)  # we don't know how big this is
+def slice_dictionary(data: Dict, chunk_size: int) -> Generator[Dict, None, None]:
     for _ in range(0, len(data), chunk_size):
-        yield {dict_key: data[dict_key] for dict_key in itertools.islice(data_iterator, chunk_size)}
+        yield {dict_key: data[dict_key] for dict_key in itertools.islice(data, chunk_size)}
--- a/aws_lambda_powertools/shared/functions.py
+++ b/aws_lambda_powertools/shared/functions.py
@@ -1,4 +1,5 @@
 import base64
+import itertools
 import logging
 import os
 import warnings
@@ -115,3 +116,13 @@ def powertools_debug_is_set() -> bool:
         return True
 
     return False
+
+
+def slice_dictionary(data, chunk_size: int):
+    # save CPU cycles if input is already small than chunk_size
+    if len(data) <= chunk_size:
+        yield data
+
+    data_iterator = iter(data)  # we don't know how big this is
+    for _ in range(0, len(data), chunk_size):
+        yield {dict_key: data[dict_key] for dict_key in itertools.islice(data_iterator, chunk_size)}
--- a/aws_lambda_powertools/utilities/feature_flags/appconfig.py
+++ b/aws_lambda_powertools/utilities/feature_flags/appconfig.py
@@ -15,8 +15,6 @@ from ... import Logger
 from .base import StoreProvider
 from .exceptions import ConfigurationStoreError, StoreClientError
 
-TRANSFORM_TYPE = "json"
-
 
 class AppConfigStore(StoreProvider):
     def __init__(
@@ -74,7 +72,7 @@ class AppConfigStore(StoreProvider):
                 dict,
                 self._conf_store.get(
                     name=self.name,
-                    transform=TRANSFORM_TYPE,
+                    transform="json",
                     max_age=self.cache_seconds,
                 ),
             )
--- a/aws_lambda_powertools/utilities/parameters/__init__.py
+++ b/aws_lambda_powertools/utilities/parameters/__init__.py
@@ -9,7 +9,7 @@ from .base import BaseProvider, clear_caches
 from .dynamodb import DynamoDBProvider
 from .exceptions import GetParameterError, TransformParameterError
 from .secrets import SecretsProvider, get_secret
-from .ssm import SSMProvider, get_parameter, get_parameters
+from .ssm import SSMProvider, get_parameter, get_parameters, get_parameters_by_name
 
 __all__ = [
     "AppConfigProvider",
@@ -22,6 +22,7 @@ __all__ = [
     "get_app_config",
     "get_parameter",
     "get_parameters",
+    "get_parameters_by_name",
     "get_secret",
     "clear_caches",
 ]
--- a/aws_lambda_powertools/utilities/parameters/appconfig.py
+++ b/aws_lambda_powertools/utilities/parameters/appconfig.py
@@ -9,6 +9,8 @@ from typing import TYPE_CHECKING, Any, Dict, Optional, Union
 import boto3
 from botocore.config import Config
 
+from aws_lambda_powertools.utilities.parameters.types import TransformOptions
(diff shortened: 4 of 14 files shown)


heitorlessa:
cc @peterschutt as this is the first PR that adds support for `typing-extensions` to make use of `Literal` for strict typing.

heitorlessa:
Two main questions to discuss in the original issue:

1.  Should we keep `parallel=` parameter when single thread is faster* in this context?
2. What should be the error strategy?
    * **(A)** Treat as non-atomic by catching the error, and include each failed param in a `errors` key along with the reason.
    * **(B)** Fail-fast by default with the option to override (`raise_on_failure`). When overridden, it falls back to A
    * **(C)** Something else

---

Did two tests within Lambda: 

1. 20 params to fetch leading to a timeout with `parallel=True` due to SSM throttling leading to exponential backoff even w/ 30s timeout
2. 10 params to fetch

**Quick results**

* **First trace**: 20 params in single and multi-threading with 128M memory size. The errors were due to throttling leading to exponential back-off to the point the operation timed out. Successful ones were with 10 params (no throttling).

* **Second trace**: Same 10 params but using with 1024M memory size

Without much further digging, this is largely due to CPU slicing in Lambda (one real core only at ~1.7G), combined with LWP cost and SSM throttling.

image

image

jplock:
- I could see this being abused by trying to fetch too many parameters, so if less than 10 keys to fetch, fetch serially, if more than 10, parallelize into batches of 10?
- I like option B with the `raise_on_failure` option
- From a usage standpoint, are people wanting to fetch separate specific keys like this, or do they really want to do fetch all child keys under a parent (like `/develop/service/*`)?

heitorlessa:
Thanks @jplock. Great to hear from you as usual.

For the first point, that's what we're gonna do with a small difference - if one enables decrypt as a default Or per parameter, we need to use GetParameter. Doing the latter in parallel can lead to timeouts easily, will do serially for decrypt calls instead.

> From a usage standpoint, are people wanting to fetch separate specific keys like this, or do they really want to do fetch all child keys under a parent (like /develop/service/*)?

The former. We already have a solution for the latter via get_parameters(path="/develop/service")

heitorlessa:
another TIL

Service model for `GetParameters` response requires a non-empty value for `InvalidParameters` even for a successful request - this makes functional testing with stubs impossible. 

Need to create an issue in Boto repo. Actual API returns an empty list.

```python
    ...
    stubber = stub.Stubber(provider.client)
    response = {
        "Parameters": [
            {
                "Name": dev_param,
                "Type": "String",
                "Value": "string",
                "Version": mock_version,
                "Selector": f"{dev_param}:{mock_version}",
                "SourceResult": "string",
                "LastModifiedDate": datetime(2015, 1, 1),
                "ARN": f"arn:aws:ssm:us-east-2:111122223333:parameter/{dev_param.removeprefix('/')}",
                "DataType": "string",
            },
            {
                "Name": prod_param,
                "Type": "String",
                "Value": "string",
                "Version": mock_version,
                "Selector": f"{prod_param}:{mock_version}",
                "SourceResult": "string",
                "LastModifiedDate": datetime(2015, 1, 1),
                "ARN": f"arn:aws:ssm:us-east-2:111122223333:parameter/{prod_param.removeprefix('/')}",
                "DataType": "string",
            },
        ],
        "InvalidParameters": [],  # NOTE: stub fails on validation despite not having any
    }

    expected_params = {"Names": param_names}
    stubber.add_response("get_parameters", response, expected_params)  # fail here
    stubber.activate()
```

**Validation error when adding response**

```python
E           botocore.exceptions.ParamValidationError: Parameter validation failed:
E           Invalid length for parameter InvalidParameters, value: 0, valid min length: 1
```

heitorlessa:
ASCII diagram to ease understanding of how we decide when to use one or both APIs to fetch parameters.

```
                           ┌────────────────────────┐
                       ┌───▶  Decrypt entire batch  │─────┐
                       │   └────────────────────────┘     │     ┌────────────────────┐
                       │                                  ├─────▶ GetParameters API  │
┌──────────────────┐   │   ┌────────────────────────┐     │     └────────────────────┘
│   Split batch    │───┼──▶│ No decryption required │─────┘
└──────────────────┘   │   └────────────────────────┘
                       │                                        ┌────────────────────┐
                       │   ┌────────────────────────┐           │  GetParameter API  │
                       └──▶│Decrypt some but not all│──────────▶├────────────────────┤
                           └────────────────────────┘           │ GetParameters API  │
                                                                └────────────────────┘
```

heitorlessa:
Merging after extensive tests and self-review - as Ruben and Leandro are on PTO

This PR made it clear that we have to prioritize refactoring Parameters entirely. Hit the following issues which delayed completion by nearly a week:

* `GetParameters` can decrypt all or no parameters
* `GetParameters` call is equal a single API call which makes it safer for API Throttling
* `GetParameters` is non-atomic which means errors are aggregated under `InvalidParameters`
* `GetParameters` can fail atomically when KMS key is invalid (or HTTP 500)
* Lacks of good typing revealed bugs in value transformation (`None`, `auto`, `Dict values`)
* `SSMProvider` and others have test side effects due to boto3 client initialization in super class when testing
* `SSM` also allows parameters without `/` which means we need to safeguard against `_errors` in case they're present when fetching parameters
* Boto stub for `GetParameters` incorrectly fails when `InvalidParameters` is an empty list. This meant making tests harder or artificially make an invalid param to support SDK data validation
* We didn't expose adding to cache and verifying whether param was in cache, requiring additional refactoring
* We were unnecessarily caching parameters with `max_age` null or negative
* We need a bunch of `overloads` due to the flexibility of `BaseProvider`, thus requiring `type-extensions` to get to an acceptable level of accuracy (excluding transform_value)
* Our `BaseProvider` nomenclature for `get` and `_get` like methods make it harder to implement complex logic like `get_parameters_by_name` which requires multiple levels of iterables

PR #1719, opened 2022-11-15
feat(streaming): add new s3 streaming utility

### LINKED ISSUE #1692: RFC: S3 Streaming utility

_No response_

New utility

### Summary

Implement a S3 Streaming utility so we can process S3 files that are bigger than memory, or temporary storage.

### Use case

As a user processing S3 files inside my Lambda, I would like to be able to consume big S3 files in a streaming fashion. I want Powertools to handle all the streaming and IO for me, so I can focus on writing the logic that handles the data.

Ideally, I would be able to enable plugins to further process the streaming data: for instance the data on S3 could be compressed, or could be formatted as JSON lines. I want Powertools to handle all the unpacking/deserialization for me. 

#### Current experience

Currently, one can tap into the the `StreamingBody` response from boto3 in order to stream bytes from S3:

```python
stream = s3.Object(bucket_name=bucket, key=key).get()["Body"]
```

One can also compose this stream with data transformations, like inflating a gzip stream:

```python
gz = GzipFile(raw=stream)
for line in gz:
   ....
```

However, this raw solutions has the following problems/limitations:
- The raw stream from boto3 is not seeakble (necessary for things like [`zipfile`](https://docs.python.org/3/library/zipfile.html))
- Poor DX when plugging multiple data transformations
- No transparent error handling for re-triable exceptions

### Proposal

Introduce a new utility called `streaming` and a new class `S3Object` that would implement the [`RawIOBase` interface](https://docs.python.org/3/library/io.html#io.RawIOBase).

```python
class S3Object(io.RawIOBase):
  def __init__(self, bucket: string, key: string, version_id: Optional[str], boto3_client: Optional[boto3.Client]):
```

The code would implement all the necessary methods to satisfy the `RawIOBase` protocol.

Additionally, create a plug-in system/middleware to plug additional data transformations on the stream. Common transformations could be included directly on the class constructor:

```python
S3Object(...., gunzip=True, json=True)
```

While still providing a way to implement custom data transformations:

```python
class CustomTransform(BaseTransform):
    def transform(self, source: io.BufferedIOBase) -> io.BufferedIOBase:
       ...

s3 = S3Object(...)
s3.add_transform(CustomTransform())
```

#### User experience

A user can use the utility from the Lambda handler:

```python
  s3_obj = S3Object(bucket=event.get("bucket"), key=event.get("key"), gunzip=True)
  for line in s3_obj:
    # do something with line
```

Under the hood, the utility would take care of fetching only the necessary bits as the data is consumed, and all the retrying caused by connection timeouts.

### Out of scope

- Staging the data into temporary storage (Lambda's Ephemeral storage, EFS, etc), before processing: can be added later if there's interest.
- Processing parallelization: it will increase the complexity of the solution and it can be added later. 

### Future work

Once the basics are in place, we could optimize even further the IO under the hood, by implementing read-head, async and parallel IO, etc. This could result in improved performance on future versions of Powertools, with no breaking API change.

### Potential challenges

Implementation is not trivial, so we need a good set of tests, including E2E tests.

_No response_

### Alternative solutions

When fetching an S3 object with boto3, there's a private accessor to the underlying `urllib3.response.HTTPResponse`. This could be used to stream the results, and could be enough if we're not implementing `seek`.

```python
s3_obj['Body']._raw_stream
<urllib3.response.HTTPResponse object at 0x1040e8460>
```

### Acknowledgment

rubenfonseca:
**Issue number:** #1692

## Summary

This PR adds a new utility to stream data from S3.

### Changes

New utility called `streaming` and a new `S3Object` class. Includes common data transformations and a way 
to implement custom data transformations and add them to the data pipeline.

### User experience

```python
from typing import Dict

from aws_lambda_powertools.utilities.streaming.s3 import S3Object


def lambda_handler(event: Dict[str, str], context: LambdaContext):
    s3 = S3Object(bucket=event["bucket"], key=event["key"], is_gzip=True)
    for line in s3:
        print(line)

```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

-----
[View rendered docs/utilities/streaming.md](https://github.com/rubenfonseca/aws-lambda-powertools-python/blob/feat/s3-streaming/docs/utilities/streaming.md)


mkdocs.yml
```diff
@@ -31,6 +31,7 @@ nav:
       - utilities/feature_flags.md
       - utilities/jmespath_functions.md
       - CloudFormation Custom Resources: https://github.com/aws-cloudformation/custom-resource-helper" target="_blank
+      - utilities/streaming.md
```
heitorlessa: do you see us having completely separate Streaming features like Event Handler REST vs GraphQL? I think not but asking anyway.

If so, we might want to change the structure now for an additional depth (`streaming/s3`, `streaming/lambda`)

tests/functional/streaming/test_s3_object.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+def test_s3_transform():
+    obj = S3Object(bucket="bucket", key="key")
+
+    new_obj = obj.transform(GzipTransform())
+    assert type(new_obj) is GzipFile
+
+
+def test_s3_transform_in_palce():
```
heitorlessa:
```suggestion
def test_s3_transform_in_place():
```

tests/functional/streaming/test_s3_object.py
```diff
@@ -0,0 +1,66 @@
...
+import pytest
+from botocore import stub
+from botocore.response import StreamingBody
+
+from aws_lambda_powertools.utilities.streaming import S3Object
+from aws_lambda_powertools.utilities.streaming._s3_seekable_io import _S3SeekableIO
+from aws_lambda_powertools.utilities.streaming.transformations import GzipTransform
+
+
+def test_s3_basic_stream():
```
heitorlessa, line 9: make a test where you try to read a file that cannot be fetched.

tests/functional/streaming/test_s3_object.py
```diff
@@ -0,0 +1,66 @@
...
+    s3_stub.add_response(
+        "get_object", {"Body": StreamingBody(raw_stream=io.BytesIO(payload), content_length=len(payload))}
+    )
+    s3_stub.activate()
+    obj = S3Object(bucket="bucket", key="key", boto3_s3_resource=s3_resource)
+
+    # WHEN you read some part of the object and then apply a transformation
+    assert obj.read(5) == b"hello"
+
+    # THEN it raises ValueError
```
heitorlessa: add more details on why the ValueError plz

aws_lambda_powertools/utilities/streaming/transformations/gzip.py
```diff
@@ -0,0 +1,9 @@
+from gzip import GzipFile
+from typing import IO
+
+from aws_lambda_powertools.utilities.streaming.transformations.base import BaseTransform
+
+
+class GzipTransform(BaseTransform):
```
heitorlessa, line 7: missing docstring to help aid customers understanding its usage and whether additional args can be used.

UX test: Customer imports it, tries to use it, what's the minimum they should know to be productive? they might not have access to the User docs.

aws_lambda_powertools/utilities/streaming/transformations/zip.py
```diff
@@ -0,0 +1,9 @@
+from typing import IO
+from zipfile import ZipFile
+
+from aws_lambda_powertools.utilities.streaming.transformations.base import BaseTransform
+
+
+class ZipTransform(BaseTransform):
```
heitorlessa, line 7: missing docstring to help aid customers understanding its usage and whether additional args can be used.

UX test: Customer imports it, tries to use it, what's the minimum they should know to be productive? they might not have access to the User docs.

aws_lambda_powertools/utilities/streaming/transformations/csv.py
```diff
@@ -0,0 +1,16 @@
+import csv
+import io
+from csv import DictReader
+from typing import IO
+
+from aws_lambda_powertools.utilities.streaming.transformations.base import BaseTransform
+
+
+class CsvTransform(BaseTransform):
```
heitorlessa, line 9: missing docstring to help aid customers understanding its usage and whether additional args can be used.

UX test: Customer imports it, tries to use it, what's the minimum they should know to be productive? they might not have access to the User docs.

aws_lambda_powertools/utilities/streaming/transformations/base.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    BaseTransform is the base class all data transformations need to implement.
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.args = args
+        self.kwargs = kwargs
+
+    @abstractmethod
+    def transform(self, input_stream: IO[bytes]) -> T:
```
heitorlessa, line 21: suuuuuper nice!

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+        return self.transformed_stream.__next__()
+
+    def __iter__(self):
+        return self.transformed_stream.__iter__()
+
+    def fileno(self) -> int:
+        raise NotImplementedError()
+
+    def flush(self) -> None:
+        raise NotImplementedError()
```
heitorlessa: Add brief docstring to quickly warn them that this is not available before they crash at runtime (for those with autocomplete but Mypy)

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+        return False
+
+    def truncate(self, size: Optional[int] = 0) -> int:
+        raise NotImplementedError()
+
+    def write(self, data: AnyStr) -> int:
+        raise NotImplementedError()
+
+    def writelines(self, lines: Iterable[AnyStr]) -> None:
+        raise NotImplementedError()
```
heitorlessa: Add brief docstring to quickly warn them that this is not available before they crash at runtime (for those with autocomplete but Mypy)

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+        in_place: bool, optional
+            Transforms the stream in place, instead of returning a new stream object. Defaults to false.
+
+        Returns
+        -------
+        T[bound=IO[bytes]], optional
+            If in_place is False, returns an IO[bytes] object representing the transformed stream
+        """
+        if self.tell() != 0:
+            raise ValueError(f"Cannot add transformations to a read object. Already read {self.tell()} bytes")
```
heitorlessa: As we have seek, is there any action they could do if they truly want it? if so, add at the end for an additionally helpful error.

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+            for transformation in self._data_transformations:
+                transformed_stream = transformation.transform(transformed_stream)
+
+            self._transformed_stream = transformed_stream
+
+        return self._transformed_stream
+
+    @overload
+    def transform(
+        self, transformations: Union[BaseTransform[T], Sequence[BaseTransform[T]]], in_place: Literal[True]
```
heitorlessa: In 3.7, we can now use `|` instead of Union if you prefer.

```
transformations: `BaseTransform[T] | Sequence[BaseTransform[T]`
```

you need an import per module you want to do that:

```python
```

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+    T,
+)
+
+if TYPE_CHECKING:
+    from mypy_boto3_s3 import S3ServiceResource
+
+
+class S3Object(IO[bytes]):
+    """
+    Seekable streamable S3 Object reader.
```
heitorlessa:
```suggestion
    Seekable and streamable S3 Object reader.
```

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+
+    S3Object implements the IO[bytes], backed by a seekable s3 streaming.
```
heitorlessa:
```suggestion
    S3Object implements the IO[bytes], backed by a seekable S3 streaming.
```

aws_lambda_powertools/utilities/streaming/s3_object.py
```diff
@@ -0,0 +1,235 @@
...
+        >>> print(line)
+
+    """
+
+    def __init__(
+        self,
+        bucket: str,
+        key: str,
+        version_id: Optional[str] = None,
+        boto3_s3_resource: Optional["S3ServiceResource"] = None,
```
heitorlessa: In 3.7, we no longer need that anymore if you add an import

```python
```

This makes static typing lazily evaluated at runtime. This also allows you to change it to the new syntax:

```python
boto3_client: S3ServiceResource | None = None,
```

aws_lambda_powertools/utilities/streaming/_s3_seekable_io.py
```diff
@@ -0,0 +1,187 @@
...
+        return self._s3_resource
+
+    @property
+    def s3_object(self) -> "Object":
+        """
+        Returns a boto3 S3Object
+        """
+        if self._s3_object is None:
+            if self.version_id is not None:
+                self._s3_object = self.s3_resource.ObjectVersion(
```
heitorlessa: how hard would it be to use the low level client `boto3.client("s3")` instead of the high-level one?

Reason: As of now, [boto Resource clients are not thread-safe but Clients](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/resources.html?highlight=multithreading#multithreading-or-multiprocessing-with-resources). 

Because we're dealing with streaming, I think that might not be a problem, so worth testing. I can see some developers trying to use `threading` when dealing with several objects since it's I/O bound and GIL isn't a problem.

[It is also completely possible for us to say](https://awslabs.github.io/aws-lambda-powertools-python/2.3.0/utilities/idempotency/#idempotent_function-decorator), this is not thread-safe due to AWS SDK not being thread-safe for higher level clients.

heitorlessa (changes requested):
superb! Added some initial comments before I have more bandwidth for a full review.

aws_lambda_powertools/utilities/streaming/s3_object.py
rubenfonseca: @heitorlessa The idea here is that once you start reading from a stream, you should not change the data transformations. We want to avoid the programming error of:

```
s3.transform(GzipTransform())
s3.readline()
s3.transform(CsvTransform())
```

which doesn't much sense, and has nothing to do with the possibility of seeking on the stream.

Do you have a better way to encode this guard?

heitorlessa (approved):

heitorlessa:
Pushed the final docs changes, including docs index and project readme to highlight as a new feature too.

PR #1724, opened 2022-11-16
feat(batch): add async_batch_processor for concurrent processing

### LINKED ISSUE #1708: Feature request: Async BatchProcessor (use case: slow processing of each item)
### Use case

I would like to process asynchronically the messages that reach a BatchProcessor, sometimes process messages from an SQS queue depends on HTTP calls or similar that can take some time, if all are done at the same time, it would not have an accumulated delay.

### Solution/User Experience

I have made a slight hack to allow this, extending the BatchProcessor class:

``` python
import asyncio
import sys
from typing import List, Tuple, Union

from aws_lambda_powertools.utilities.batch import BatchProcessor, SuccessResponse, FailureResponse


class AsyncBatchProcessor(BatchProcessor):

    async def _aprocess_record(self, record: dict) -> Union[SuccessResponse, FailureResponse]:
        """
        Process a record with instance's handler

        Parameters
        ----------
        record: dict
            A batch record to be processed.
        """
        data = self._to_batch_type(record=record, event_type=self.event_type, model=self.model)
        try:
            if self._handler_accepts_lambda_context:
                result = await self.handler(record=data, lambda_context=self.lambda_context)
            else:
                result = await self.handler(record=data)

            return self.success_handler(record=record, result=result)
        except Exception:
            return self.failure_handler(record=data, exception=sys.exc_info())

    async def aprocess(self) -> List[Tuple]:
        return list(await asyncio.gather(*[self._aprocess_record(record) for record in self.records]))
```

and the main code to run:

``` python
import asyncio
import json

from aws_lambda_powertools.utilities.batch import EventType
from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord

from aws_lambda_powertools import Tracer, Logger
from .async_batch_preprocessor import AsyncBatchProcessor

tracer = Tracer()
logger = Logger()

aprocessor = AsyncBatchProcessor(event_type=EventType.SQS)

a = 1


async def record_handler(record: SQSRecord):
    global a
    """
    Process here each record
    """
    payload: str = record.body
    if payload:
        item: dict = json.loads(payload)
        print(item)
        a += 1
        print(f'sleeping for {a} s')
        await asyncio.sleep(a)
        print('awaited!')
        # code code code...


@logger.inject_lambda_context
    return asyncio.run(alambda_handler(event, context))


async def alambda_handler(event, context: LambdaContext):
    batch = event["Records"]
    with aprocessor(records=batch, handler=record_handler):
        await aprocessor.aprocess()  # kick off processing, return list[tuple]
    return aprocessor.response()
```

### Alternative solutions

_No response_

### Acknowledgment

BakasuraRCE:
**Issue number:** [#1708](https://github.com/awslabs/aws-lambda-powertools-python/issues/1708)

## Summary

Add new `AsyncBatchProcessor` and `async_batch_processor` decorator to support async batch processing for a customer's record handler.

**Before**

```mermaid
classDiagram
    class BasePartialProcessor {
        <<interface>>
        +success_handler(record, result) SuccessResponse
        +failure_handler(record, exception) FailureResponse
        +process()
        +_prepare()
        +_clean()
        +_process_record_(record)
        +_async_process() List~Tuple~
        +__enter__()
        +__exit__()
        +__call__(records, handler, lambda_context)
    }

    class BatchProcessor {
        +Dict batch_response
        +response() Dict
        +_process_record()
        +_prepare()
        +_clean()
    }

    BasePartialProcessor <|-- BatchProcessor : implement
    note for BatchProcessor "implements entire interface and logic for batch processing"
```

**New**

```mermaid
classDiagram
    class BasePartialProcessor {
        <<interface>>
        +success_handler(record, result) SuccessResponse
        +failure_handler(record, exception) FailureResponse
        +process()
        +_prepare()
        +_clean()
        +_process_record_(record)
        +_async_process_record_(record)
        +_async_process() List~Tuple~
        +__enter__()
        +__exit__()
        +__call__(records, handler, lambda_context)
    }

    class BasePartialBatchProcessor {
        +Dict batch_response
        +response() Dict
        +_prepare()
        +_clean()
    }

    class BatchProcessor {
        +_process_record()
        +_async_process_record() NotImplementedError
    }

    class AsyncBatchProcessor {
        +_process_record() NotImplementedError
        +_async_process_record()
    }

    BasePartialProcessor <|-- BasePartialBatchProcessor : implement
    BasePartialBatchProcessor <|-- BatchProcessor : inherit
    BasePartialBatchProcessor <|-- AsyncBatchProcessor : inherit

    note for BasePartialBatchProcessor "implements shared logic for batch processing used for sync and async processing"
    note for BatchProcessor "only implements logic to call customers' record handler synchronously"
    note for AsyncBatchProcessor "only implements logic to call customers' record handler asynchronously"
```

### AsyncBatchProcessor class

New class to encapsulate async processing handlers by implementing `_async_process_record` interface. Same logic as `BatchProcessor` but async. It expects record handlers to be async functions now.

### async_batch_processor decorator

New decorator that instantiates `AsyncBatchProcessor`, also sync to prevent forcing customers to make their Handlers async causing side effects to existing middlewares. `AsyncBatchProcessor` handles sync->async operation transparently via closures.

### BasePartialBatchProcessor class

Contains shared logic for all processors be synchronous or asynchronous. By shared logic, this means success handler, failure handler, message ID collection from SQS, Kinesis, DynamoDB Streams, etc.

For a future major version, we can further decompose by creating more specialized classes, for example `KinesisStreamsBatchProcessor`. This will become more relevant when we support beyond Lambda runtime, where their logic and debugging will benefit from a clear separation.

### Async methods and ABCs

`BasePartialProcessor` base class add a new method `async_process` and a new async ABC `async_process_record()`.

* **`def async_process()`**. Synchronous function with an async closure to process records asynchronously. Despite the name, its outer function is sync to prevent forcing customers to make their Handlers async, otherwise it'd impact existing middlewares we're unaware of.
* **`async def _async_process_record()`**. Async version of `_process_record` abstract method

### BatchProcessor

We move all logic to `BasePartialBatchProcessor` to make it easier to split Sync vs Async processors.

Both processors `BatchProcessor` and `AsyncBatchProcessor` have the same logic, except one handles asyncIO when calling an async record handler.

For a future where we want to support a threaded version, we can create a specialized  `ThreadedBatchProcessor` class, inherit `BasePartialBatchProcessor`, and implement `_process_record`.

### User experience

image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


BakasuraRCE:
~@heitorlessa the code still is untested~

I didn't know where to put the new asynchronous decorator, so I created some tests in asynchronous version, and examples, let me know your comments

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -237,7 +255,58 @@ def batch_processor(
...
+    """
+    records = event["Records"]
+
+    with processor(records, record_handler, lambda_context=context):
+        await processor.async_process()
+
+    return await handler(event, context)
+
+
+class BaseBatchProcessorMixin(BasePartialProcessor):
```
heitorlessa: We can't use Mixin :/ as multiple inheritance will [impact our ability to use Mypyc later](https://github.com/awslabs/aws-lambda-powertools-python/issues/1708#issuecomment-1318341031). Could you use single inheritance instead, please?

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -237,7 +255,58 @@ def batch_processor(
...
+    -----------
+    * Async batch processors
+
+    """
+    records = event["Records"]
+
+    with processor(records, record_handler, lambda_context=context):
+        await processor.async_process()
+
+    return await handler(event, context)
```
heitorlessa: I've been thinking about this for the last hour, please feel free to correct me if I misread it.

I think we need to focus on **`processor`** being async while keeping the Lambda Handler synchronous. 

This means transforming **`async_batch_processor`** to **`def async_batch_processor`**, and use **`asyncio.run`** to prime and wait for the **`processor.async_process()`** coroutine to complete. Another benefit is making `async_lambda_handler` redundant.


**Why is that?**

1. **Changing the Lambda Handler to async requires an additional decorator.** This introduces an unnecessary overhead with multiple function calls, considering we have a decorator factory here (3fn) twice + decorated functions. This slows execution down, including [`trace_execution`](https://awslabs.github.io/aws-lambda-powertools-python/2.3.1/utilities/middleware_factory/#tracing-middleware-execution) feature will not work as expected and **might** block the event loop.
2. **Unforeseen issues with decorator stacking**. Customers often use other synchronous decorators like `@metrics.log_metrics`, `@logger`, `@tracer`,  `@validate`, and a mix of their own. A common issue is ordering and I'm 90% confident someone will put this in the wrong handler, for the Lambda handler.

tests/functional/test_utilities_batch.py
```diff
@@ -115,6 +119,17 @@ def handler(record):
     return handler
 
 
+@pytest.fixture(scope="module")
+def async_record_handler() -> Callable:
```
heitorlessa:
```suggestion
def async_record_handler() -> Callable[..., Awaitable[Any]]:
```

to make it easier to spot the return is a coroutine, not a sync return.

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -237,7 +255,58 @@ def batch_processor(
...
+        >>> processor = AsyncBatchProcessor(event_type=EventType.SQS)
+        >>>
+        >>> @async_lambda_handler
+        >>> @async_batch_processor(record_handler=async_record_handler, processor=processor)
+        >>> async def lambda_handler(event, context: LambdaContext):
+        >>>     return processor.response()
+
+    Limitations
+    -----------
+    * Async batch processors
```
heitorlessa:
```suggestion
    * Sync batch processors. Use `batch_processor` instead.
```

aws_lambda_powertools/asynchrony/base.py
```diff
@@ -0,0 +1,8 @@
+import asyncio
+
+from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
+
+
+@lambda_handler_decorator
+def async_lambda_handler(handler, event, context):
```
heitorlessa: See my comment on `async_batch_processor` as we can make this part redundant.

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -87,19 +88,6 @@ def _clean(self):
...
-        """
-        Process record with handler.
-        """
-        raise NotImplementedError()
-
-    def process(self) -> List[Tuple]:
-        """
-        Call instance's handler for each record.
-        """
-        return [self._process_record(record) for record in self.records]
```
heitorlessa: This will break custom processors created by customers, since `batch_processor` decorator will call `processor.process()` -- and TIL we didn't have tests for a custom processor, otherwise it'd have signaled it.

https://awslabs.github.io/aws-lambda-powertools-python/2.3.1/utilities/batch/#create-your-own-partial-processor

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -191,9 +179,39 @@ def failure_handler(self, record, exception: ExceptionInfo) -> FailureResponse:
...
+        """
+        Process record with handler.
+        """
+        raise NotImplementedError()
+
+    async def async_process(self) -> List[Tuple]:
+        """
+        Call instance's handler for each record.
+        """
+        return list(await asyncio.gather(*[self._async_process_record(record) for record in self.records]))
```
heitorlessa: comment only: This makes me think we should also have a `concurrent=True` in the synchronous version, where we use a Future and document its caveats on thread-safe boto3 stuff.

heitorlessa (changes requested):
Thank you once again for the heavy lifting - I like the overall direction! It even brought up that we should (separate PR) have a way to use a `ThreadPoolExecutor` as an opt-in feature like `concurrent=True` for the synchronous processor.

Major comments summarized:

* **Breaking change**. Removing `process()` from `BasePartialProcessor` breaks [custom processors](https://awslabs.github.io/aws-lambda-powertools-python/2.3.1/utilities/batch/#create-your-own-partial-processor). Turns out (TIL) we don't have a test for it hence why it wasn't caught up earlier.
* **Let's keep Lambda handler synchronous**. We can prevent a call stack slowdown, make `async_lambda_handler` redundant, and prevent a class of unforeseen issues for customers stacking decorators of different colours (sync/async) since ordering matters.
* **Single inheritance only**. As I mentioned earlier, we can't use multiple inheritance as that will impact our ability to use Mypyc later.

BakasuraRCE:
> Thank you once again for the heavy lifting - I like the overall direction! It even brought up that we should (separate PR) have a way to use a `ThreadPoolExecutor` as an opt-in feature like `concurrent=True` for the synchronous processor.
> 
> Major comments summarized:
> 
>     * **Breaking change**. Removing `process()` from `BasePartialProcessor` breaks [custom processors](https://awslabs.github.io/aws-lambda-powertools-python/2.3.1/utilities/batch/#create-your-own-partial-processor). Turns out (TIL) we don't have a test for it hence why it wasn't caught up earlier.
> 
>     * **Let's keep Lambda handler synchronous**. We can prevent a call stack slowdown, make `async_lambda_handler` redundant, and prevent a class of unforeseen issues for customers stacking decorators of different colours (sync/async) since ordering matters.
> 
>     * **Single inheritance only**. As I mentioned earlier, we can't use multiple inheritance as that will impact our ability to use Mypyc later.

Hey!

- Make the simple inheritance code forces to do things not so beautiful, such as loading the methods `_async_process_record` and `async_process` to the class `BasePartialProcessor`.
- So, I changed the `asyncio.run()` in async_batch_processor, I had BIG problems with that in another project.
- About the `concurrent=True`, following the best practices, it is best to have "atomic responsibilities" for methods/classes. It would be better to get a new class called `ThreadBatchProcessor` class with its corresponding `thread_batch_processor`, keeping this explicit, also solve the tend to confuse asynchrony with the concurrence

heitorlessa:
I'm glad you called it out - thank you for keeping us honest ;) 

Answering them

> Make the simple inheritance code forces to do things not so beautiful

I couldn't agree more. Because we optimize to run in a constrained environment (almost like embedded), we have to make these decisions which would be sub-optimal in different contexts. We're trading a more beautiful implementation with future speed gains that could save $ for customers.

> So, I changed the `asyncio.run()` in async_batch_processor, I had BIG problems with that in another project.

Mind sharing what the **big** problems were so we can learn from it? I'm new to asyncio in Python from a library authorship point of view - if there's a balance we can strike to avoid decorator stacking confusion and increased call stack while avoiding these scars you've earned, it'd be lovely!

> - About the `concurrent=True`, following the best practices, it is best to have "atomic responsibilities" for methods/classes.

I worded that poorly. What I meant for `concurrent=True` was an UX thinking, I didn't put any thought on the underlying implementation at the time of writing. You're absolutely right on `ThreadBatchProcessor` - way to go, as threading can get complex pretty quickly.

Since `batch_processor` decorator would simply call `ThreadBatchProcessor.process()` where the threading logic is, and it's also sync, I find `thread_batch_processor` mostly cognitive load when balancing semantics, intent and non-developers using it -- am I missing something here?

BakasuraRCE:
> Mind sharing what the big problems were so we can learn from it? I'm new to asyncio in Python from a library authorship point of view - if there's a balance we can strike to avoid decorator stacking confusion and increased call stack while avoiding these scars you've earned, it'd be lovely!

It is something that I still can't explain well, I have worked for a good time with this function in docker environments without memory or disk limitations and always following the documentation: `It should be used as a main entry point for asyncio programs, and should ideally only be called once.`

However, in the AWS Lambda environment, even with high limits, it simply fails when for some reason an unexpected error occurs(prematurely closing the main loop without allowing the error to spread up).

Imagine an error `423` in an API or an error `400` in the `boto3` behind an asynchronous wrapper with a thread underneath. In non-Lambda environments, the error is simply propagated up, and then the loop closes without complaints.

However, python's recommendation is to use `asyncio.run` instead of deep functions to create the main loop.

> Since batch_processor decorator would simply call ThreadBatchProcessor.process() where the threading logic is, and it's also sync, I find thread_batch_processor mostly cognitive load when balancing semantics, intent and non-developers using it -- am I missing something here?

Exactly, it is a matter of being explicit and not giving rise to confusion, it is also due to maintenance of the code, imagine in the future for some reason processing messages using the class `ThreadBatchProcessor`, it is not simply calling to `ThreadBatchProcessor.process()`.

The `batch_processor` would have to be modified in consequence to have synchronous logic and a multithread logic, this could be confusing to maintain, especially not intuitive.

I'm sorry for not having a more technical detail about asyncio, I wait for your thoughts :)

heitorlessa:
Aaah, that's highly likely due to the nature of container/process freezing in Lambda. There are similar side effects in NodeJS too.

Agreed on the unforeseeable nature of `thread_batch_processor` needing custom logic in the future. Let's go ahead with these. 

I'm going to re:Invent today and will be mostly away for a week, feel free to reach out to @rubenfonseca for anything.

We can help write the docs, do a thorough review, and contribute any possible gaps to make this quicker.

Thank you so much again for helping out with this (Gracias!!)

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -100,6 +101,26 @@ def process(self) -> List[Tuple]:
...
+    @abstractmethod
+    async def _async_process_record(self, record: dict):
+        """
+        Async process record with handler.
+        """
+        raise NotImplementedError()
+
+    def async_process(self) -> List[Tuple]:
+        """
+        Async call instance's handler for each record.
```
heitorlessa, line 126: For future maintenance, let's add a comment on the use of async closure to ease decorator chaining

examples/batch_processing/src/getting_started_async_batch_processor.py
```diff
@@ -0,0 +1,21 @@
+from aws_lambda_powertools.utilities.batch import async_batch_processor, EventType, AsyncBatchProcessor
+from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
+from aws_lambda_powertools.utilities.typing import LambdaContext
+
+
+async def async_record_handler(record: SQSRecord):
```
heitorlessa: np: what would be a realistic example here? maybe an async crawler?

As part of docs refactoring, we haven't reached Batch yet, we're trying to include more complete examples.

Example: https://awslabs.github.io/aws-lambda-powertools-python/2.4.0/core/tracer/#ignoring-certain-http-endpoints

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -100,6 +101,26 @@ def process(self) -> List[Tuple]:
...
+        """
+
+        async def async_process():
+            return list(await asyncio.gather(*[self._async_process_record(record) for record in self.records]))
+
+        # WARNING: Do not use "asyncio.run(async_process())", there are cases in which for some reason the main
+        # loop closes while using, causing the error "Event Loop is closed"
```
heitorlessa:
```suggestion
        # WARNING
        # Do not use "asyncio.run(async_process())" due to Lambda container thaws/freeze, otherwise we might get "Event Loop is closed"
        # Instead, get_event_loop() can also create one if a previous was erroneously closed
        # More: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html#runtimes-lifecycle-shutdown
```

More info for completion for future maintenance.

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -423,7 +378,7 @@ def _clean(self):
         if self._entire_batch_failed():
             raise BatchProcessingError(
                 msg=f"All records failed processing. {len(self.exceptions)} individual errors logged "
-                f"separately below.",
+                    f"separately below.",
```
heitorlessa: nice!

tests/functional/test_utilities_batch.py
```diff
@@ -639,3 +652,82 @@ def lambda_handler(event, context):
...
+        processed_messages = batch.async_process()
+
+    # THEN
+    assert processed_messages == [
+        ("success", first_record.body, first_record.raw_event),
+        ("success", second_record.body, second_record.raw_event),
+    ]
+
+    assert batch.response() == {"batchItemFailures": []}
+
```
heitorlessa, line 714: If you have bandwidth (or whoever maintainer frees up in January), let's add a test to prevent that regression we caught on custom processors?

For now, the simplest suffices so we can ensure we don't break customers who create custom processors.

Docs: https://awslabs.github.io/aws-lambda-powertools-python/2.4.0/utilities/batch/#create-your-own-partial-processor

heitorlessa (changes requested):
Last comments before winter holidays. 

In summary:

- **We need a summary of the changes in the PR body**. This helps future maintainers/contributors look up the reason for certain changes and for reviewer's understanding
- **Made comments on doc strings**. It will likely be foreign to other maintainers on the async closure so a more complete docstring helps remove ambiguity.
- **We need a more realistic async batch example**. This will ease writing docs and better emphasize the usefulness of this feature for other customers ;) 
- [Optional] **We need a test to prevent custom processor regression**. Earlier we manually caught that [custom processors use methods](https://awslabs.github.io/aws-lambda-powertools-python/2.4.0/utilities/batch/#create-your-own-partial-processor) that were removed as part of the initial PR. I'm pretty sure we will get sidetracked and not add after this PR is merged, so let's add it - whoever has the bandwidth in January.

Thank you so much again

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -100,6 +102,40 @@ def process(self) -> List[Tuple]:
...
+        # Detect environment and create a loop for each one
+        coro = async_process()
+        if os.environ.get('AWS_LAMBDA_RUNTIME_API'):
+            # Running in lambda server
+            loop = asyncio.get_event_loop()
+            task_instance = loop.create_task(coro)
+            return loop.run_until_complete(task_instance)
+        else:
+            # Running in another place like local tests
+            return asyncio.run(coro)
```
BakasuraRCE: @heitorlessa please check this, I get from Mangum that have more experience with async in lambdas

heitorlessa:
Back from holidays and our team offsite ends this week. I'll resume this PR next week so we can get it merged for our next release ;-)

Hope you had a great end of the year, and look forward to getting this to the finish line!

Thank you for the prolonged patience

aws_lambda_powertools/utilities/batch/base.py
heitorlessa: I need to read more on this. `AWS_LAMBDA_RUNTIME_API` might not be the best option.

heitorlessa: updated to TASK_ROOT as it's guaranteed to be in proper emulators too.

heitorlessa:
Pushed some changes to improve maintenance. **Next**: I'll create a summary of the changes in the PR before refactoring the doc examples, and create docs for this. If nothing else drags me over meetings and such, this should be in the next release **2.8.0** ;)

heitorlessa:
Pull request description changed to reflect all changes, and to make it easier for maintainers to reason what, why, and the effect of these changes.

Added two mermaid diagrams to more easily grasp it as a summary

---

**Before**

```mermaid
classDiagram
    class BasePartialProcessor {
        <<interface>>
        +success_handler(record, result) SuccessResponse
        +failure_handler(record, exception) FailureResponse
        +process()
        +_prepare()
        +_clean()
        +_process_record_(record)
        +_async_process() List~Tuple~
        +__enter__()
        +__exit__()
        +__call__(records, handler, lambda_context)
    }

    class BatchProcessor {
        +Dict batch_response
        +response() Dict
        +_process_record()
        +_prepare()
        +_clean()
    }

    BasePartialProcessor <|-- BatchProcessor : implement
    note for BatchProcessor "implements entire interface and logic for batch processing"
```


**New**

```mermaid
classDiagram
    class BasePartialProcessor {
        <<interface>>
        +success_handler(record, result) SuccessResponse
        +failure_handler(record, exception) FailureResponse
        +process()
        +_prepare()
        +_clean()
        +_process_record_(record)
        +_async_process_record_(record)
        +_async_process() List~Tuple~
        +__enter__()
        +__exit__()
        +__call__(records, handler, lambda_context)
    }

    class BasePartialBatchProcessor {
        +Dict batch_response
        +response() Dict
        +_prepare()
        +_clean()
    }

    class BatchProcessor {
        +_process_record()
        +_async_process_record() NotImplementedError
    }

    class AsyncBatchProcessor {
        +_process_record() NotImplementedError
        +_async_process_record()
    }

    BasePartialProcessor <|-- BasePartialBatchProcessor : implement
    BasePartialBatchProcessor <|-- BatchProcessor : inherit
    BasePartialBatchProcessor <|-- AsyncBatchProcessor : inherit

    note for BasePartialBatchProcessor "implements shared logic for batch processing used for sync and async processing"
    note for BatchProcessor "only implements logic to call customers' record handler synchronously"
    note for AsyncBatchProcessor "only implements logic to call customers' record handler asynchronously"
```

heitorlessa (approved):

heitorlessa:
ALL DONE 🎉 🚀 @BakasuraRCE -- I'll wait until Thursday for any last comments before we merge and push a new release ;)

I wanted to thank you one more time for this big effort. Since you last touched, I've made the following changes:

* [x] Documented API changes (new/refactoring) in the PR description
* [x] Removed the sync code snippet, and refactored the async code snippet to better show case this change
* [x] Created a new section in the docs, along with warnings when used with Tracer, and guidance for those new to AsyncIO
* [x] Fixed docstrings in classes (BatchProcessor and AsyncBatchProcessor) to ensure VSCode/PyCharm play nice when auto-completing docs

sthulb:
@BakasuraRCE Hey, I'm looking into writing a load test for this today/tomorrow to check perf :)

sthulb:
I created a simple load test. I created a SQS queue containing 5500 messages. When all of the messages were in the queue, I set the event source of my lambda to enabled. My event source had the following config:
- Concurrency: 2
- Batch size: 10000
- Max window: 60 secs

I ran two versions of the function (sync and async using the code in this PR). The sync version took on average 1.2 seconds, the async version took around 300ms. It’s a vast improvement. 

The function setup:
- Memory: 512
- Timeout: 10 secs

Issue #1791, opened 2022-12-20
Error `TypeError: object of type 'NoneType' has no len()` thrown when I have the following header set 'Access-Control-Allow-Origin': None


jasmarc:
I'm seeing `TypeError: object of type 'NoneType' has no len()` when I have the following header set:

```
'Access-Control-Allow-Origin': None
```

_Originally posted by @jasmarc in https://github.com/awslabs/aws-lambda-powertools-python/pull/1455#discussion_r1052828216_

leandrodamascena:
Hello @jasmarc! We merged the PR with the code to fix this issue and tomorrow we will release a new version in production. Thanks!

jasmarc:
Awesome! Thank you :)

PR #1793, opened 2022-12-20
fix(event_handlers): omit explicit None HTTP header values

leandrodamascena:
**Issue number:** #1791 

### Changes

Add a fix for when header is None and avoid iterator or len issues

### User experience

Although not recommended by the RFC that defines headers in the HTTP protocol, the user can force a header with a value of None. To avoid this problem the header which has value None will be omitted from the response

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/shared/headers_serializer.py
+++ b/aws_lambda_powertools/shared/headers_serializer.py
@@ -41,11 +41,14 @@ class HttpApiHeadersSerializer(BaseHeadersSerializer):
         # Duplicate headers are combined with commas and included in the headers field.
         combined_headers: Dict[str, str] = {}
         for key, values in headers.items():
+            # omit headers with explicit null values
+            if values is None:
+                continue
+
             if isinstance(values, str):
                 combined_headers[key] = values
             else:
-                if values:
-                    combined_headers[key] = ", ".join(values)
+                combined_headers[key] = ", ".join(values)
 
         return {"headers": combined_headers, "cookies": list(map(str, cookies))}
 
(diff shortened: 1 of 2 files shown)


aws_lambda_powertools/shared/headers_serializer.py
```diff
@@ -65,8 +66,9 @@ def serialize(self, headers: Dict[str, Union[str, List[str]]], cookies: List[Coo
             if isinstance(values, str):
                 payload[key].append(values)
             else:
-                for value in values:
-                    payload[key].append(value)
+                if values:
```
heitorlessa: when you accept a falsy value but not null, check for null otherwise a `X-Is-Admin: false` will be omitted here. For example, `if values is None`, would work in both cases.

aws_lambda_powertools/shared/headers_serializer.py
```diff
@@ -65,8 +66,9 @@ def serialize(self, headers: Dict[str, Union[str, List[str]]], cookies: List[Coo
...
+                    for value in values:
+                        payload[key].append(value)
```
heitorlessa: we can drop an additional for loop by using `list.extend`, as it accepts an Iterable and it'll add all elements at the end of the list. This drops the complexity

heitorlessa:
Last missing piece is docs. We should mention somewhere that custom headers with `None` value will be omitted. Pre-approving

heitorlessa (approved):

Issue #1772, opened 2022-12-09
Feature request: improve error messaging around IdempotencyPersistenceLayerError


cgoettel:
### Use case

I was trying to implement the basic example using Terraform to deploy the Lambda and DynamoDB. However, I had forgotten to include the policy for the Lambda to talk to DynamoDB and kept getting the following error:

```
[ERROR] IdempotencyPersistenceLayerError: Failed to save in progress record to idempotency store
Traceback (most recent call last):
  File "/var/task/aws_lambda_powertools/middleware_factory/factory.py", line 134, in wrapper
    response = middleware()
  File "/var/task/aws_lambda_powertools/utilities/idempotency/idempotency.py", line 79, in idempotent
    return idempotency_handler.handle()
  File "/var/task/aws_lambda_powertools/utilities/idempotency/base.py", line 97, in handle
    return self._process_idempotency()
  File "/var/task/aws_lambda_powertools/utilities/idempotency/base.py", line 116, in _process_idempotency
    raise IdempotencyPersistenceLayerError("Failed to save in progress record to idempotency store") from exc
```

Each error message is the same which isn't entirely helpful. So I dug through the code and could see that it was exception chaining, but that also wasn't particularly helpful. It wasn't until I tried to implement the basic example using AWS SAM that I was able to see the policy that I'd forgotten to attach.

### Solution/User Experience

Ideally, something in the error chain would receive a bad permissions error and that information would percolate up to the user. Every other time I deploy into AWS I get clear permissions errors and the reason I didn't even consider a permissions issue is because it was always so clear previously when that was it.

### Acknowledgment

- [X] This feature request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [X] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

mploski:
Hey @cgoettel, thanks for raising this issue. I agree we need to enrich the error message here to be more meaningful. I will start working on this in a few coming days (probably end of the week or beginning of the next one). Will update you about the progress

mploski:
Hey @cgoettel. I was able to reproduce mentioned issue and I can confirm your observation, that lambda emits logs in the shape you shared in the ticket itself. Let me share some details about it, just for future reference or interested readers :-)

[...]

PR #1813, opened 2023-01-05
fix(idempotency): Log nested exception message

mploski:
**Issue number: #1772

## Summary
Raise `IdempotencyPersistenceLayerError` exception with additional context taken from raised downstream exception 
### Changes

* Modify idempotency core logic to pass downstream exception to newly raised `IdempotencyPersistenceLayerError`
exception 
* Create base exception class that formats output visible in stdout and inherit from it in idempotency exceptions classes
Before this change:
```
[ERROR] IdempotencyPersistenceLayerError: ('Failed to save in progress record to idempotency store', ClientError('An error occurred (AccessDeniedException) when calling the PutItem operation: User: <REMOVED> is not authorized to perform: dynamodb:PutItem on resource: <REMOVED> because no identity-based policy allows the dynamodb:PutItem action'))
``` 

After:
```
[ERROR] IdempotencyPersistenceLayerError: Failed to save in progress record to idempotency store: (An error occurred (AccessDeniedException) when calling the PutItem operation: User: <REMOVED> is not authorized to perform: dynamodb:PutItem on resource:<REMOVED> because no identity-based policy allows the dynamodb:PutItem action)
```
This change is not mandatory. I'm happy to hear your opinion whether we should reformat exception message or not and what format would be the best

* Adjust tests

### User experience

Before then change all downstream exception details was hidden from user, making debugging harder.
Now downstream details are available
For details please follow: https://github.com/awslabs/aws-lambda-powertools-python/issues/1772

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/exceptions.py
+++ b/aws_lambda_powertools/utilities/idempotency/exceptions.py
@@ -3,23 +3,26 @@ Idempotency errors
 """
 
 
+from typing import Optional, Union
+
+
 class BaseError(Exception):
     """
     Base error class that overwrites the way exception and extra information is printed.
     See https://github.com/awslabs/aws-lambda-powertools-python/issues/1772
     """
 
-    def format_msg(self, *args: str) -> str:
-        full_msg = str(args[0])
-        if args[1:]:
-            full_msg += f': ({"".join(str(arg) for arg in args[1:])})'
-        return full_msg
+    def __init__(self, *args: Optional[Union[str, Exception]]):
+        self.message = str(args[0]) if args else ""
+        self.details = "".join(str(arg) for arg in args[1:]) if args[1:] else None
 
-    def __init__(self, *args: str):
-        if args:
-            super().__init__(self.format_msg(*args))
-        else:
-            super().__init__()
+    def __str__(self):
+        """
+        Return all arguments formatted or original message
+        """
+        if self.message and self.details:
+            return f"{self.message} - ({self.details})"
+        return self.message
 
 
 class IdempotencyItemAlreadyExistsError(BaseError):
--- a/aws_lambda_powertools/utilities/idempotency/exceptions.py
+++ b/aws_lambda_powertools/utilities/idempotency/exceptions.py
@@ -3,49 +3,68 @@ Idempotency errors
 """
 
 
-class IdempotencyItemAlreadyExistsError(Exception):
+class BaseError(Exception):
+    """
+    Base error class that overwrites the way exception and extra information is printed.
+    See https://github.com/awslabs/aws-lambda-powertools-python/issues/1772
+    """
+
+    def format_msg(self, *args: str) -> str:
+        full_msg = str(args[0])
+        if args[1:]:
+            full_msg += f': ({"".join(str(arg) for arg in args[1:])})'
+        return full_msg
+
+    def __init__(self, *args: str):
+        if args:
+            super().__init__(self.format_msg(*args))
+        else:
+            super().__init__()
+
+
+class IdempotencyItemAlreadyExistsError(BaseError):
     """
     Item attempting to be inserted into persistence store already exists and is not expired
     """
 
 
-class IdempotencyItemNotFoundError(Exception):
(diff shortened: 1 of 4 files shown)


aws_lambda_powertools/utilities/idempotency/exceptions.py
```diff
@@ -3,49 +3,68 @@
...
+            full_msg += f': ({"".join(str(arg) for arg in args[1:])})'
+        return full_msg
+
+    def __init__(self, *args: str):
+        if args:
+            super().__init__(self.format_msg(*args))
+        else:
+            super().__init__()
+
+
```
leandrodamascena, line 27: Hello @mploski! This is a huge improvement indeed. I remember the first time I needed to use the idempotency utility, I got an error due to permissions and spent a lot of time figuring out the reason for the "Failed to save record in progress to idempotency store" message. Creating a specific class to capture and return detailed errors makes life and adoption easier!!

I was wondering if we really need to instantiate the parent Exception class or if we could change this code to directly return the error message and make the code more clean. But it's just an idea to think about, I'm not sure and would just like to hear your side.

```python
class BaseError(Exception):
     """
    Base error class that overwrites the way exception and extra information is printed.
    See https://github.com/awslabs/aws-lambda-powertools-python/issues/1772
    """

    def __init__(self, *args: str):
        self.message = args[0] if args else ""
        self.args = args

    def __str__(self):
        """
        Return all arguments formatted or original message
        """
        if self.args:
            return f"{' - '.join(str(arg) for arg in self.args)}"
        return self.message
```

leandrodamascena:
I did all the tests with this code and simulated several scenarios and in all I was successful. Really a big improvement this @mploski. I just commented one file and I will use your feedback to learn more about Python :)

image

aws_lambda_powertools/utilities/idempotency/exceptions.py
mploski: Hey @leandrodamascena , thx for the feedback! 
I like your proposal and I refactored my code to follow your suggestion i.e to modify how exception is printed in __str__ dunder method. This approach also has additional benefit that we keep initial attributes ( in args list) in object itself so we can easily see how exception object was initialized during debugging if needed.

As for the super class calling. In theory it is suggested to call it as explained in Python Cookbook book: https://books.google.pl/books?id=f7wGeA71_eUC&pg=PA579&dq=To+illustrate+the+use+of+.args,+consider+this+interactive+session+with+the+built-in+RuntimeError+exception,+and+notice+how+any+number+of+arguments+can+be+used+with+the+raise+statement:&hl=en&sa=X&ved=2ahUKEwinm_Dy0bX8AhWMposKHQNHB3oQ6AF6BAgBEAI#v=onepage&q=To%20illustrate%20the%20use%20of%20.args%2C%20consider%20this%20interactive%20session%20with%20the%20built-in%20RuntimeError%20exception%2C%20and%20notice%20how%20any%20number%20of%20arguments%20can%20be%20used%20with%20the%20raise%20statement%3A&f=false. 
I dove deep more to see what happens when we call Exception __init__ method. It seems this behavior changed somewhere in newer python 3 releases and args list initialization was moved to __new__ dunder method so all needed stuff is initialized either way even without calling superclass init method ( as oppose to python 2.x and 3.0 for example) . So I think we can safely omit it. 
here is a  c code  for __new__ method in exception class in python 3 main branch: https://github.com/python/cpython/blob/3.7/Objects/exceptions.c#L47 - you can see that we copy args to instance attribute
and in Python 3.0: 
https://github.com/python/cpython/blob/v3.0/Objects/exceptions.c#L49 - you can see that this action is done only in __init__ not in __new__

Can I ask you to review it once again? :-)

leandrodamascena: Hello @mploski! Thank you very much for your feedback and all the links provided, it was very helpful to understand more deeply how the exception is handled and it arguments.

From my side everything is ok and we are ready to merge!

leandrodamascena (approved):
Approved.

PR #1853, opened 2023-01-23
feat(event_handlers) Add support for returning dict, status-code tuples from resolvers

### LINKED ISSUE #1845: Feature request: Return response and status code from APIGatewayRestResolver routes
### Use case

APIGatewayRestResolver currently supports 2 return values - either a Response object or a dict (see [_to_response](https://github.com/awslabs/aws-lambda-powertools-python/blob/d4feaedd42339899840b52ce61dd89d238fe7651/aws_lambda_powertools/event_handler/api_gateway.py#L731))

However, when returning a dict, the status code is hardcoded to 200

It is a common use case to want to return other status codes, eg. a 201, alongside a dict of the created object

### Solution/User Experience

Flask handles this by allowing a tuple to be returned with the first value being the response dict and the second being a status code (they also allow returning headers but I'm not pushing for that)

I currently use a subclassed resolver to add this functionality with a _to_response like this:
```
def _to_response(self, result: Union[Dict, Response, Tuple]) -> Response:
        if isinstance(result, tuple):
            response, code = result
            return Response(
                status_code=code,
                content_type=content_types.APPLICATION_JSON,
                body=self._json_dump(response),
            )
        else:
            return super()._to_response(result)
```

The flask implementation looks like this with additional validation and the header handling
```
  if isinstance(rv, tuple):
            len_rv = len(rv)

            # a 3-tuple is unpacked directly
            if len_rv == 3:
                rv, status, headers = rv  # type: ignore[misc]
            # decide if a 2-tuple has status or headers
            elif len_rv == 2:
                if isinstance(rv[1], (Headers, dict, tuple, list)):
                    rv, headers = rv
                else:
                    rv, status = rv  # type: ignore[assignment,misc]
            # other sized tuples are not allowed
            else:
                raise TypeError(
                    "The view function did not return a valid response tuple."
                    " The tuple must have the form (body, status, headers),"
                    " (body, status), or (body, headers)."
                )
```

Implementing this would allow something like this:
```
@app.post("/entity/")
def create_entity():
    data = router.current_event.json_body
    new_entity = Entity(**data)
    return new_entity.asdict(), 201
```

### Alternative solutions

_No response_

### Acknowledgment

theoutsider24:
**Issue number:**
#1845

### Changes

The response handling in api gateway has been extended to handle dict, status-code tuples

### User experience

Endpoints with non-200 status code no longer need to manually build Response objects, they can simply return a dict and status code

## Checklist

* [x] Changes are documented

<details>
No

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -735,13 +735,13 @@ class ApiGatewayResolver(BaseRouter):
 
         - Dict[str, Any]: Rest api response with just the Dict to json stringify and content-type is set to
           application/json
-        - Tuple[dict, str]: Same dict handling as above but with the option of including a status code
+        - Tuple[dict, int]: Same dict handling as above but with the option of including a status code
         - Response: returned as is, and allows for more flexibility
         """
-        status_code = 200
+        status_code = HTTPStatus.OK
         if isinstance(result, Response):
             return result
-        elif isinstance(result, tuple) and len(tuple) == 2:
+        elif isinstance(result, tuple) and len(result) == 2:
             # Unpack result dict and status code from tuple
             result, status_code = result
 
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -728,21 +728,26 @@ class ApiGatewayResolver(BaseRouter):
 
         return None
 
-    def _to_response(self, result: Union[Dict, Response]) -> Response:
+    def _to_response(self, result: Union[Dict, Tuple, Response]) -> Response:
         """Convert the route's result to a Response
 
-         2 main result types are supported:
+         3 main result types are supported:
 
         - Dict[str, Any]: Rest api response with just the Dict to json stringify and content-type is set to
           application/json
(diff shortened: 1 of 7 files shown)


rubenfonseca (changes requested):
@theoutsider24 thank you for opening this PR!

Two asks:
- can you please try to fix the tests? I believe we need some type updates to the tests
- would you mind adding an update to the docs with the new use case?

We would love if you could help us with this, but otherwise we would be happy to assist :)

theoutsider24:
@rubenfonseca Thanks for the comments! 

I've fixed that issue with the type-hinting
Regarding the docs - where do you think the most appropriate place would be to mention it? I couldn't see a place that seemed obvious so I just included it in some existing examples that seemed relevant

rubenfonseca (approved):

leandrodamascena:
Thank you so much for this PR @theoutsider24! Ruben updated the documentation and we merged it!

theoutsider24:
That's awesome! Thanks so much for your help guys, really pleased I was able to contribute and I look forward to using the feature myself 💪

rubenfonseca:
This is now released under [2.7.0](https://github.com/awslabs/aws-lambda-powertools-python/releases/tag/v2.7.0) version!

PR #1861, opened 2023-01-25
fix(tests): make sure multiple e2e tests run concurrently

### LINKED ISSUE #1860: Bug: E2E tests are currently not working
### Expected Behaviour

E2E tests are supposed to pass on every commit on the `develop` branch.

### Current Behaviour

E2E tests are failing for a long time.

This is being caused by having [more than one test file for some of the utilities](https://github.com/awslabs/aws-lambda-powertools-python/tree/ffbcdcb435b8636c5614298a257e3b01c6415e4e/tests/e2e/event_handler). This causes problems as pytest tries to run the same utility at the same time, causing [CDK to clash and erroring out](https://github.com/awslabs/aws-lambda-powertools-python/actions/runs/3998001399/jobs/6860071178).

### Code snippet

```python
`make e2e-test`
```

### Possible Solution

There are two possible solutions:

* make sure each directory only has one test file:
	* PROS: more parallelism
	* CONS: sharing infrastructure between the same utility becomes harder, possibly leading to dupplication 

* change e2e infrastructure to distribute tests per directory instead of per file

### Steps to Reproduce

`make e2e-test` is currently failing on `develop`

latest

3.9

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #1860

### Changes

Fix E2E tests by marking all tests in the same folder with the same `xdist_group`.

### User experience

Before this change, tests would randomly fail, as a test would try to use infrastructure that was already being teared down.
With this change, all related tests are scheduled together, thus avoiding the problem.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -23,6 +23,7 @@ def lambda_layer_build(tmp_path_factory: pytest.TempPathFactory, worker_id: str)
 
     layer = LocalLambdaPowertoolsLayer()
     yield from call_once(
+        id="lambda_layer",
         task=layer.build,
         tmp_path_factory=tmp_path_factory,
         worker_id=worker_id,
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,4 +1,5 @@
 import pytest
+from xdist.scheduler import LoadGroupScheduling
 
 from tests.e2e.utils.infrastructure import call_once
 from tests.e2e.utils.lambda_layer.powertools_layer import LocalLambdaPowertoolsLayer
@@ -27,3 +28,11 @@ def lambda_layer_build(tmp_path_factory: pytest.TempPathFactory, worker_id: str)
         tmp_path_factory=tmp_path_factory,
         worker_id=worker_id,
     )
+
+
+# Hook to use the xdist_group to decide how to schedule tests among the different workers.
+# Each test needs to be marked with @pytest.mark.xdist_group(name=...). This way, all
+# the tests on each directory will be scheduled on the same worker, thus sharing the
+# infrastructure.
+def pytest_xdist_make_scheduler(config, log):
+    return LoadGroupScheduling(config, log)
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -23,7 +23,6 @@ def lambda_layer_build(tmp_path_factory: pytest.TempPathFactory, worker_id: str)
 
     layer = LocalLambdaPowertoolsLayer()
     yield from call_once(
-        job_id="lambda_layer",
         task=layer.build,
         tmp_path_factory=tmp_path_factory,
         worker_id=worker_id,
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -23,7 +23,7 @@ def lambda_layer_build(tmp_path_factory: pytest.TempPathFactory, worker_id: str)
 
     layer = LocalLambdaPowertoolsLayer()
     yield from call_once(
-        id="lambda_layer",
+        job_id="lambda_layer",
         task=layer.build,
         tmp_path_factory=tmp_path_factory,
         worker_id=worker_id,
--- a/tests/e2e/conftest.py
+++ b/tests/e2e/conftest.py
@@ -1,5 +1,4 @@
 import pytest
-from xdist.scheduler import LoadGroupScheduling
 
 from tests.e2e.utils.infrastructure import call_once
 from tests.e2e.utils.lambda_layer.powertools_layer import LocalLambdaPowertoolsLayer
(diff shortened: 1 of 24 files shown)


tests/e2e/conftest.py
```diff
@@ -27,3 +28,11 @@ def lambda_layer_build(tmp_path_factory: pytest.TempPathFactory, worker_id: str)
...
         worker_id=worker_id,
     )
+
+
+# Hook to use the xdist_group to decide how to schedule tests among the different workers.
+# Each test needs to be marked with @pytest.mark.xdist_group(name=...). This way, all
+# the tests on each directory will be scheduled on the same worker, thus sharing the
+# infrastructure.
+def pytest_xdist_make_scheduler(config, log):
+    return LoadGroupScheduling(config, log)
```
heitorlessa: Since we're not customizing `load-group`, do we need it here **vs** CLI argument?

tests/e2e/event_handler/conftest.py
```diff
@@ -3,7 +3,7 @@
 from tests.e2e.event_handler.infrastructure import EventHandlerStack
 
 
-@pytest.fixture(autouse=True, scope="module")
+@pytest.fixture(autouse=True, scope="package")
```
heitorlessa, line 6: TIL - doesn't `module` cover it as we don't have sub-modules inside?

tests/e2e/parameters/infrastructure.py
```diff
@@ -27,7 +27,7 @@ def create_resources(self):
             iam.PolicyStatement(
                 effect=iam.Effect.ALLOW,
                 actions=[
-                    "ssm:GetParameter",
+                    "ssm:GetParameters",
```
heitorlessa, line 30: great catch!

tests/e2e/utils/functions.py
```diff
@@ -1,14 +1,27 @@
...
+import time
+from concurrent.futures import Future, ThreadPoolExecutor
+from typing import List
 
 from tests.e2e.utils import data_fetcher  # noqa F401
 
 
 def execute_lambdas_in_parallel(function_name: str, lambdas_arn: list, arguments: str):
+    def f(fname: str, farn: str, fargs: str):
+        return eval(fname)(farn, fargs)
```
heitorlessa: docstring and if possible a more descriptive function name

tests/e2e/utils/functions.py
```diff
@@ -1,14 +1,27 @@
...
+
     result_list = []
     with ThreadPoolExecutor() as executor:
-        running_tasks = executor.map(lambda exec: eval(function_name)(*exec), [(arn, arguments) for arn in lambdas_arn])
+        running_tasks: List[Future] = []
+        for arn in lambdas_arn:
+            # Sleep 0.5, 1, 1.5, ... seconds between each invocation. This way
+            # we can guarantee that lambdas are executed in parallel, but they are
```
heitorlessa, line 14: not sure I follow - how will they be called in parallel if you're adding them to the executor within a 0.5s interval? What was the issue before?

heitorlessa (changes requested):
great work!! Some questions so I can become familiar with certain decisions.

tests/e2e/event_handler/conftest.py
rubenfonseca: Running pytest with `--collect-only` shows the following output:

```
<Package parameters>
  <Module test_appconfig.py>
    <Function test_get_parameter_appconfig_freeform>
  <Module test_ssm.py>
    <Function test_get_parameters_by_name>
```

So having it on the module level was instantiating the fixture twice for a given folder

tests/e2e/conftest.py
rubenfonseca: Reason is that if I manually run e2e tests with `pytest tests/e2e/...` the group scheduling was not loaded. Only if I ran all the tests with `make e2e-tests`.

This way group scheduling will always be used for e2e tests, no matter how we run pytest.

tests/e2e/utils/functions.py
rubenfonseca: Use case: we want to fire two lambdas at the same time for the idempotency test. Each Lambda will block for 3 seconds. We want the first lambda to block for 3 seconds, and the second Lambda to throw with "execution is already running".

Before this change, both Lambdas will be fired at the same time, and due to networking jitter and other factors, we couldn't assume the order that they will actually be run. So in this case we guarantee that, while they will be called in parallel, each successive invocation will be slightly shifted in time to make sure we guarantee the output order. Does this make sense?

tests/e2e/utils/functions.py
rubenfonseca: I've refactored this to me more clear

rubenfonseca:
@heitorlessa addressed your questions and refactored some code, please let me know your thoughts

tests/e2e/conftest.py
heitorlessa: Ahh of course! That solves step-through debugging too.

Thanks!

tests/e2e/event_handler/conftest.py
heitorlessa: TIL I misunderstood Pytest fixtures for a long time then, thinking it'd match Python module.

Thank you!

tests/e2e/utils/functions.py
heitorlessa: It does - thank you!!

For the future (not in this PR), I'd suggest renaming or having a parameter for that purpose so we don't accidentally use it for another use case that might incur side effects in assertion (e.g., execution finishing fast enough).

heitorlessa (approved):

Issue #1877, opened 2023-01-31
Maintenance: E2E tests are flaky


rubenfonseca:
### Summary

After the last batch of fixes, the E2E tests are still flaky.

For instance [awslabs/aws-lambda-powertools-python/actions/runs/4047838730/jobs/6962344909](https://github.com/awslabs/aws-lambda-powertools-python/actions/runs/4047838730/jobs/6962344909) failed with a logging not getting enough messages.

Other times, it seems the whole setup fails for unknown reasons: [awslabs/aws-lambda-powertools-python/actions/runs/4047900294](https://github.com/awslabs/aws-lambda-powertools-python/actions/runs/4047900294)

### Why is this needed?

To have confidence in our E2E tests

### Which area does this relate to?

Tests

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

PR #1878, opened 2023-01-31
fix(tests): make logs fetching more robust

rubenfonseca:
**Issue number:** #1877

### Changes

This PR adds an optional field to the logger fetcher, where we could indicate how many 
lines of logging we are expecting. If we fail to fetch that number, we use the existing
retry mechanism to try multiple times, before bailing out.

### User experience

This should help with some of the flaky e2e tests.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/tests/e2e/utils/data_fetcher/logs.py
+++ b/tests/e2e/utils/data_fetcher/logs.py
@@ -122,6 +122,7 @@ class LogFetcher:
 def get_logs(
     function_name: str,
     start_time: datetime,
+    expected_number_of_logs: Optional[int],
     filter_expression: Optional[str] = None,
     log_client: Optional[CloudWatchLogsClient] = None,
 ) -> LogFetcher:
@@ -133,6 +134,8 @@ def get_logs(
         Name of Lambda function to fetch logs for
     start_time : datetime
         Start date range to filter traces
+    expected_number_of_logs : Optional[int]
+        Retry fetching logs until this number of log lines are obtained
     log_client : Optional[CloudWatchLogsClient], optional
         Amazon CloudWatch Logs Client, by default boto3.client('logs)
     filter_expression : Optional[str], optional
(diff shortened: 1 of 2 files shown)


tests/e2e/utils/data_fetcher/logs.py
```diff
@@ -133,6 +134,8 @@ def get_logs(
         Name of Lambda function to fetch logs for
     start_time : datetime
         Start date range to filter traces
+    expected_number_of_logs : Optional[int]
```
heitorlessa:
```suggestion
    minimum_log_entries : Optional[int]
```

align with `TraceFetcher(minimum_traces=1)`, then we can have as 1. If you're using Log Entries you're expecting at least one entry, otherwise set it explicitly.

tests/e2e/utils/data_fetcher/logs.py
```diff
@@ -143,6 +146,11 @@ def get_logs(
     LogFetcher
         LogFetcher instance with logs available as properties and methods
     """
-    return LogFetcher(
+    log_fetcher = LogFetcher(
         function_name=function_name, start_time=start_time, filter_expression=filter_expression, log_client=log_client
     )
+
+    if expected_number_of_logs is not None and len(log_fetcher) < expected_number_of_logs:
```
heitorlessa: after that change, this is simplified with and easier to read with:

```suggestion
    if minimum_log_entries < len(log_fetcher):
```

heitorlessa (changes requested):
UX change and complexity baseline suggestion.

Thanks for spotting this, we did the same in `TraceFetcher` before after hitting eventual consistency too.

heitorlessa (approved):
🚀🚀🚀

PR #1880, opened 2023-02-01
feat(metrics): add default_dimensions to single_metric

### LINKED ISSUE #1859: single_metric() to inherit default dimensions
### Use case

At the root level I define default metric which is environment name in my case.

```
metrics = Metrics()
metrics.set_default_dimensions(Environment=ENVIRONMENT)
```

This is all fine and `Environment` gets propagated to all metrics. However, when `single_metric()` is used this default dimension is ignored. This is what I do to overcome that.
```
metrics = Metrics()

with single_metric(
    name="Metric1", unit=MetricUnit.Milliseconds, value=elapsed_ms
) as this_metric:
    this_metric.add_dimension(name="TableName", value=self.table_name)


    for name, value in metrics.default_dimensions.items():
        this_metric.add_dimension(name=name, value=value)
```

or it should be possible with `this_metric.default_dimensions = metrics.default_dimensions`, I did not tried this.

This way all the default dimensions are propagated with a single metric. 

### Solution/User Experience

When creating a `SingleMetric` with `single_metric` context manager default dimensions should be copied to a SingleMetric instance.

I can imagine the situation when you don't want to propagate default dimensions to the SingleMetric. 

To keep existing usage untouched I suppose the default behavior should be to not propagate default dimensions. So the solution can be the additional keyword argument `inherit_default_dimensions: bool = False`
```
with single_metric("Name", MetricUnit.Count, 1, inherit_default_dimensions=True) as this_metric:
    this_metric.add_dimension(name="Dim1", value="Val")
```

### Alternative solutions

_No response_

### Acknowledgment

prudnikov:
**Issue number:** #1859

### Changes

`default_dimensions` keyword argument was added to `single_metric` context manager. It accepts custom dimensions dictionary as well as `metrics.default_dimensions`.

Initial I proposed `inherit_default_dimensions: bool = False`, bug when started implementing decided to go with the similar to `@metrics.log_metrics` approach with specifying `default_dimensions` keyword argument.

### User experience

It was not possible to reuse default dimensions. One of the common default dimensions is Environment, once it is defined on Metrics instance it can be reused in SingleMetric as well. It is also possible to provide any other dimensions as value to `default_dimensions`.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


rubenfonseca (approved):
This is a stellar PR, thank you so much for the great work! I believe the new solution is much simpler and easy to understand. Again, contributions like this one always make our day! Thanks!

PR #1930, opened 2023-02-15
fix(logger): support exception and exception_name fields at any log level

### LINKED ISSUE #1923: Bug: Allow reporting entry logs with custom exception and exception_name
### Expected Behaviour

Due to [internal logic](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/logging/formatter.py#L155) `exception` and `exception_name` always get overridden.

The following statement should serialize the values provided: 
```
logger.info("Report", extra={"exception": "Oops!  That was not valid", "exception_name": "ValueError" })
```

```
logger.append_keys(exception="Oops!  That was not valid", exception_name="ValueError")
logger.info("Report")
```

and produce an output like:
```
{
        "level": "INFO",
        "message": "Report",
        "exception": "Oops!  That was not valid",
        "exception_name": "ValueError",
}
```

### Current Behaviour

Due to  [internal logic](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/logging/formatter.py#L155) these field get overridden with None:

```
{
        "level": "INFO",
        "message": "Report"
}
```

### Code snippet

```python

logger = Logger()

logger.info("Report", extra={"exception": "Oops!  That was not valid", "exception_name": "ValueError"})
```

### Possible Solution

Checking if these fields already exist on the `formatted_log` instead of always calling `_extract_log_exception` would fix this issue.

### Steps to Reproduce

See code snippet section

latest

3.9

### Packaging format used

PyPi

### Debugging logs

_No response_

iago1460:
**Issue number:** #1923 

## Summary
Providing fix for reporting entry logs with custom exception and exception_name fields 

### Changes

Changed format log to honor `exception` and `exception_name` fields

### User experience

given:

```python
logger.info("Report", extra={"exception": "Oops!  That was not valid", "exception_name": "ValueError" })
```

before:

```json
{
        "level": "INFO",
        "message": "Report"
}
```

after:

```json
{
        "level": "INFO",
        "message": "Report",
        "exception": "Oops!  That was not valid",
        "exception_name": "ValueError",
}
```

## Checklist

* [ ] Changes are documented

<details>
no
</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -152,7 +152,9 @@ class LambdaPowertoolsFormatter(BasePowertoolsFormatter):
         """Format logging record as structured JSON str"""
         formatted_log = self._extract_log_keys(log_record=record)
         formatted_log["message"] = self._extract_log_message(log_record=record)
-        formatted_log["exception"], formatted_log["exception_name"] = self._extract_log_exception(log_record=record)
+        extracted_exception, extracted_exception_name = self._extract_log_exception(log_record=record)
+        formatted_log["exception"] = formatted_log.get("exception", extracted_exception)
+        formatted_log["exception_name"] = formatted_log.get("exception_name", extracted_exception_name)
         formatted_log["xray_trace_id"] = self._get_latest_trace_id()
         formatted_log = self._strip_none_records(records=formatted_log)
 
--- a/aws_lambda_powertools/logging/formatter.py
+++ b/aws_lambda_powertools/logging/formatter.py
@@ -152,6 +152,8 @@ class LambdaPowertoolsFormatter(BasePowertoolsFormatter):
         """Format logging record as structured JSON str"""
         formatted_log = self._extract_log_keys(log_record=record)
         formatted_log["message"] = self._extract_log_message(log_record=record)
+        # exception and exception_name fields can be added as extra key
+        # in any log level, we try to extract and use them first
         extracted_exception, extracted_exception_name = self._extract_log_exception(log_record=record)
         formatted_log["exception"] = formatted_log.get("exception", extracted_exception)
         formatted_log["exception_name"] = formatted_log.get("exception_name", extracted_exception_name)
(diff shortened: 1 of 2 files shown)


aws_lambda_powertools/logging/formatter.py
```diff
@@ -152,7 +152,11 @@ def format(self, record: logging.LogRecord) -> str:  # noqa: A003
         """Format logging record as structured JSON str"""
         formatted_log = self._extract_log_keys(log_record=record)
         formatted_log["message"] = self._extract_log_message(log_record=record)
-        formatted_log["exception"], formatted_log["exception_name"] = self._extract_log_exception(log_record=record)
+        # exception and exception_name fields can be added as extra key
+        # in any log level, we try to extract and use them first
+        extracted_exception, extracted_exception_name = self._extract_log_exception(log_record=record)
+        formatted_log["exception"] = formatted_log.get("exception", extracted_exception)
+        formatted_log["exception_name"] = formatted_log.get("exception_name", extracted_exception_name)
```
leandrodamascena, line 159: I liked this solution. We could use an `if condition` to test if the keys already exist, and if not, try to extract from the exception, but that works fine too. Just out of curiosity, I ran this solution and the solution using `if condition` and we didn't get any time penalty running this. The Logger utility is widely used by our customers, so every millisecond counts in its execution.

leandrodamascena (approved):
LGTM :)

aws_lambda_powertools/logging/formatter.py
iago1460: cheers, adding an `if condition` was my first idea, but I discard it because I had to make the decision if I should check just one or both keys and what to do if only one of the keys was provided. 

Also this approach is not affecting existing performance since `_extract_log_exception` was always called before, but please feel free to improve it further. Regards

tests/functional/test_logger.py
```diff
@@ -251,6 +251,21 @@ def test_logger_append_duplicated(stdout, service_name):
...
 
 
+def test_logger_honors_given_exception_keys(stdout, service_name):
+    # GIVEN Logger is initialized with exception and exception_name fields
+    logger = Logger(
+        service=service_name, stream=stdout, exception="exception_value", exception_name="exception_name_value"
+    )
+
+    # WHEN log level info
+    logger.info("log")
```
heitorlessa: Suggestion: add the exception fields here so it's easier to scan and understand what's being tested.

logger.info("log", exception=..., exception_name=...)

leandrodamascena: Good catch @heitorlessa! Thank you!

PR #1934, opened 2023-02-16
feat(batch): add support to SQS FIFO queues (SqsFifoPartialProcessor)

### LINKED ISSUE #1927: Feature request: Handle SQS FIFO queues with the Bach Processing utility
### Discussed in https://github.com/awslabs/aws-lambda-powertools-python/discussions/1913

<sup>Originally posted by **whoDoneItAgain** February  8, 2023</sup>
I've been reading over the documentation and I cant find an answer.

Does Powertools stop processing items after a failed and marks the failed plus all subsequent as failed in its response back to SQS?

Ex. r1,r2,r3,r4,r5 is sent. function fails on r3. ReportBatchItemFailures includes r3 as well as r4 and r5.

</div>

rubenfonseca:
**Issue number:** #1927

### Changes

This PR adds support for SQS FIFO queues on the Batch Processing utility.

From [docs](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#services-sqs-batchfailurereporting):

> If you're using this feature with a FIFO queue, your function should stop processing messages after the first failure and return all failed and unprocessed messages in `batchItemFailures`. This helps preserve the ordering of messages in your queue.

### User experience

Before this change, using the `BatchProcessor` with an SQS FIFO would result into undefined behaviour, as we were still continuing to process records after the first failure.

Using the new specialized class `SqsFifoPartialProcessor` makes the utility safe to use with SQS FIFO.

carbon (37)

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
+++ b/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
@@ -1,0 +1,101 @@
+import sys
+from typing import List, Optional, Tuple, Type
+
+from aws_lambda_powertools.utilities.batch import BatchProcessor, EventType
+from aws_lambda_powertools.utilities.parser.models import SqsRecordModel
+
+#
+# type specifics
+#
+has_pydantic = "pydantic" in sys.modules
+
+# For IntelliSense and Mypy to work, we need to account for possible SQS subclasses
+# We need them as subclasses as we must access their message ID or sequence number metadata via dot notation
+if has_pydantic:
+    BatchTypeModels = Optional[Type[SqsRecordModel]]
+
+
+class SQSFifoCircuitBreakerError(Exception):
+    """
+    Signals a record not processed due to the SQS FIFO processing being interrupted
+    """
+
+    pass
+
+
+class SQSFifoPartialProcessor(BatchProcessor):
+    """Specialized BatchProcessor subclass that handles FIFO SQS batch records.
+
+    As soon as the processing of the first record fails, the remaining records
+    are marked as failed without processing, and returned as native partial responses.
+
+    Example
+    _______
+
+    ## Process batch triggered by a FIFO SQS
+
+    ```python
+    import json
+
+    from aws_lambda_powertools import Logger, Tracer
+    from aws_lambda_powertools.utilities.batch import SQSFifoPartialProcessor, EventType, batch_processor
+    from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
+    from aws_lambda_powertools.utilities.typing import LambdaContext
+
+
+    processor = SQSFifoPartialProcessor()
+    tracer = Tracer()
+    logger = Logger()
+
+
+    @tracer.capture_method
+    def record_handler(record: SQSRecord):
+        payload: str = record.body
+        if payload:
+            item: dict = json.loads(payload)
+        ...
+
+    @logger.inject_lambda_context
+    @tracer.capture_lambda_handler
+    @batch_processor(record_handler=record_handler, processor=processor)
+    def lambda_handler(event, context: LambdaContext):
+        return processor.response()
+    ```
+    """
+
+    circuitBreakerError = SQSFifoCircuitBreakerError("A previous record failed processing.")
+
+    def __init__(self, model: Optional["BatchTypeModels"] = None):
+        super().__init__(EventType.SQS, model)
+
+    def process(self) -> List[Tuple]:
+        result: List[Tuple] = []
+
+        for i, record in enumerate(self.records):
+            """
+            If we have failed messages, it means that the last message failed.
+            We then short circuit the process, failing the remaining messages
+            """
+            if self.fail_messages:
+                return self._short_circuit_processing(i, result)
+
+            """
+            Otherwise, process the message normally
+            """
+            result.append(self._process_record(record))
+
+        return result
+
+    def _short_circuit_processing(self, first_failure_index: int, result: List[Tuple]) -> List[Tuple]:
+        remaining_records = self.records[first_failure_index:]
+        for remaining_record in remaining_records:
+            data = self._to_batch_type(record=remaining_record, event_type=self.event_type, model=self.model)
+            result.append(
+                self.failure_handler(
+                    record=data, exception=(type(self.circuitBreakerError), self.circuitBreakerError, None)
+                )
+            )
+        return result
+
+    async def _async_process_record(self, record: dict):
+        raise NotImplementedError()
--- a/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
+++ b/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
@@ -12,7 +12,7 @@ class SQSFifoCircuitBreakerError(Exception):
     pass
 
 
-class SQSFifoPartialProcessor(BatchProcessor):
+class SqsFifoPartialProcessor(BatchProcessor):
     """Process native partial responses from SQS FIFO queues.
 
     Stops processing records when the first record fails. The remaining records are reported as failed items.
@@ -26,12 +26,12 @@ class SQSFifoPartialProcessor(BatchProcessor):
     import json
 
     from aws_lambda_powertools import Logger, Tracer
-    from aws_lambda_powertools.utilities.batch import SQSFifoPartialProcessor, EventType, batch_processor
+    from aws_lambda_powertools.utilities.batch import SqsFifoPartialProcessor, EventType, batch_processor
     from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
     from aws_lambda_powertools.utilities.typing import LambdaContext
 
 
-    processor = SQSFifoPartialProcessor()
+    processor = SqsFifoPartialProcessor()
     tracer = Tracer()
     logger = Logger()
 
--- a/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
+++ b/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
@@ -1,18 +1,7 @@
-import sys
-from typing import List, Optional, Tuple, Type
+from typing import List, Optional, Tuple
 
 from aws_lambda_powertools.utilities.batch import BatchProcessor, EventType
-from aws_lambda_powertools.utilities.parser.models import SqsRecordModel
-
-#
-# type specifics
-#
-has_pydantic = "pydantic" in sys.modules
-
-# For IntelliSense and Mypy to work, we need to account for possible SQS subclasses
-# We need them as subclasses as we must access their message ID or sequence number metadata via dot notation
-if has_pydantic:
-    BatchTypeModels = Optional[Type[SqsRecordModel]]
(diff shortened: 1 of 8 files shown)


aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
+import sys
+from typing import List, Optional, Tuple, Type
+
+from aws_lambda_powertools.utilities.batch import BatchProcessor, EventType
+from aws_lambda_powertools.utilities.parser.models import SqsRecordModel
```
heitorlessa: this is gonna fail if it doesn't have pydantic. Let's move this entire logic to a `types.py` as it's spans beyond `base.py` now

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+class SQSFifoCircuitBreakerError(Exception):
+    """
+    Signals a record not processed due to the SQS FIFO processing being interrupted
+    """
+
+    pass
+
+
+class SQSFifoPartialProcessor(BatchProcessor):
+    """Specialized BatchProcessor subclass that handles FIFO SQS batch records.
```
heitorlessa: Shorten that. Focus on explain the FIFO element (less on subclassing, that's internal)

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+
+    As soon as the processing of the first record fails, the remaining records
```
heitorlessa: We can do better here - let's shorten it and explain in layman terms what this all means.

Always think of a non-developer persona reading this for the first time to understand whether they should use it or not. Once our new API reference is live, this will be the headline.

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+    ```python
+    import json
+
+    from aws_lambda_powertools import Logger, Tracer
+    from aws_lambda_powertools.utilities.batch import SQSFifoPartialProcessor, EventType, batch_processor
+    from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
+    from aws_lambda_powertools.utilities.typing import LambdaContext
+
+
+    processor = SQSFifoPartialProcessor()
```
heitorlessa: will need change once class is renamed

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+        super().__init__(EventType.SQS, model)
+
+    def process(self) -> List[Tuple]:
+        result: List[Tuple] = []
+
+        for i, record in enumerate(self.records):
+            """
+            If we have failed messages, it means that the last message failed.
+            We then short circuit the process, failing the remaining messages
+            """
```
heitorlessa: that's interesting take, TIL! I wonder if these should be `#` or overwrite the docstring for `def process` since it differs from the typical `BatchProcessor::process()`

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+
+        return result
+
+    def _short_circuit_processing(self, first_failure_index: int, result: List[Tuple]) -> List[Tuple]:
+        remaining_records = self.records[first_failure_index:]
+        for remaining_record in remaining_records:
+            data = self._to_batch_type(record=remaining_record, event_type=self.event_type, model=self.model)
+            result.append(
+                self.failure_handler(
+                    record=data, exception=(type(self.circuitBreakerError), self.circuitBreakerError, None)
```
heitorlessa: Since the tuple creation is static (type(self.circuitBreakerError), self.circuitBreakerError, None), create it earlier where `_short_circuit_processing` can access it - same way you micro-optimized the `circuitBreakerError` as a class attr.

Creating these tuples in a large data pipeline will use unnecessary memory. It'll also have a nice side-effect of making it easier to read code:

```python
result.append(self.failure_handler(record=data, exception=self.circuit_break_exc)
```

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+            """
+            If we have failed messages, it means that the last message failed.
+            We then short circuit the process, failing the remaining messages
+            """
+            if self.fail_messages:
+                return self._short_circuit_processing(i, result)
+
+            """
+            Otherwise, process the message normally
+            """
```
heitorlessa:
```suggestion
	   # Otherwise, process the message normally
```

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
```diff
@@ -0,0 +1,101 @@
...
+            result.append(self._process_record(record))
+
+        return result
+
+    def _short_circuit_processing(self, first_failure_index: int, result: List[Tuple]) -> List[Tuple]:
```
heitorlessa, line 81: Q: do we need a docstring? It's clear to me, is it for @leandrodamascena ?

heitorlessa (changes requested):
This is great!! Made some initial comments and spotted two issues. After lunch, I'll look over docs and tests.

Thank you @rubenfonseca <3

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
rubenfonseca: @heitorlessa what would you suggest the new class name would be? I believe I missed that on the review

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
rubenfonseca: Added a docstring to make it clearer

aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
heitorlessa: ahh it got lost somehow - I suggested either `SqsFifoPartialProcessor` or `SqsFifoProcessor`. `SQS` isn't aligned with other parts of the code.

aws_lambda_powertools/utilities/batch/types.py
```diff
@@ -0,0 +1,24 @@
...
+        KinesisDataStreamRecord as KinesisDataStreamRecordModel,
+    )
+    from aws_lambda_powertools.utilities.parser.models import SqsRecordModel
+
+    BatchTypeModels = Optional[
+        Union[Type[SqsRecordModel], Type[DynamoDBStreamRecordModel], Type[KinesisDataStreamRecordModel]]
+    ]
+    BatchSqsTypeModel = Optional[Type[SqsRecordModel]]
+else:
+    BatchTypeModels = Optional  # type: ignore
```
heitorlessa: did you mean `BatchTypeModels = "BatchTypeModels"` like we had previously?

docs/utilities/batch.md
```diff
@@ -347,6 +347,24 @@ Processing batches from SQS works in four stages:
     }
     ```
 
+#### FIFO queues
+
+If you're using this feature with a FIFO queue, you should use the `SQSFifoPartialProcessor` class instead. We will
```
heitorlessa:
```suggestion
When using [SQS FIFO queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html){target="_blank"}, use the `SQSFifoPartialProcessor` class instead. We will
```

tests/functional/test_utilities_batch.py
```diff
@@ -40,7 +42,7 @@
 def sqs_event_factory() -> Callable:
     def factory(body: str):
         return {
-            "messageId": "059f36b4-87a3-44ab-83d2-661975830a7d",
+            "messageId": str(uuid.uuid4()),
```
heitorlessa:
```suggestion
            "messageId": f"{uuid.uuid4()}",
```

tests/functional/test_utilities_batch.py
```diff
@@ -117,6 +119,17 @@ def handler(record):
...
 
+@pytest.fixture(scope="module")
+def sqs_fifo_record_handler() -> Callable:
+    def handler(record):
+        body = record["body"]
+        if "fail" in body:
+            raise Exception("Failed to process record.")
+        return body
+
+    return handler
```
heitorlessa: naive question, do we need it given it's the exact same logic as `record_handler` fixture?

tests/functional/test_utilities_batch.py
```diff
@@ -654,6 +667,48 @@ def lambda_handler(event, context):
...
+
+    # THEN
+    assert result["batchItemFailures"] == []
+
+
+def test_sqs_fifo_batch_processor_middleware_with_failure(sqs_event_factory, sqs_fifo_record_handler):
+    # GIVEN
+    first_record = SQSRecord(sqs_event_factory("success"))
+    second_record = SQSRecord(sqs_event_factory("fail"))
+    # this would normally suceed, but since it's a FIFO queue, it will be marked as failure
```
heitorlessa: loved this comment. thank you for your great attention to detail @rubenfonseca

aws_lambda_powertools/utilities/batch/types.py
rubenfonseca: Yes that makes much more sense, thank you!

docs/utilities/batch.md
rubenfonseca: Much better, thanks!

tests/functional/test_utilities_batch.py
rubenfonseca: Absolutely right. I created it in the beginning because I thought it would be different, but it's not :) Changing it

docs/utilities/batch.md
```diff
@@ -349,8 +349,8 @@ Processing batches from SQS works in four stages:
 
 #### FIFO queues
 
-If you're using this feature with a FIFO queue, you should use the `SQSFifoPartialProcessor` class instead. We will
-stop processing messages after the first failure, and return all failed and unprocessed messages in `batchItemFailures`.
+When using [SQS FIFO queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html){target="_blank"},
+We will stop processing messages after the first failure, and return all failed and unprocessed messages in `batchItemFailures`.
```
heitorlessa:
```suggestion
When using [SQS FIFO queues](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html){target="_blank"}, we will stop processing messages after the first failure, and return all failed and unprocessed messages in `batchItemFailures`.
```

heitorlessa (approved):

Issue #1968, opened 2023-03-01
Bug: Idempotency DynamoDB _put_record method overwrites static_pk_value


Tankanow:
### Expected Behaviour

When I pass a `static_pk_value` value into the `DynamoDBPersistenceLayer`, I expect that value to be used in the Primary Key.

### Current Behaviour

The `PK` value is overwritten with `data_record.idempotency_key`.

### Code snippet

```python
[This test](https://github.com/Tankanow/aws-lambda-powertools-python/pull/1/files) shows the problem. Running this test results in this failure:


E           aws_lambda_powertools.utilities.idempotency.exceptions.IdempotencyPersistenceLayerError: Failed to save in progress record to idempotency store - (Error getting response stub for operation PutItem: Expected parameters:
E           {'ConditionExpression': 'attribute_not_exists(#id) OR #expiry < :now OR '
E                                   '(#status = :inprogress AND '
E                                   'attribute_exists(#in_progress_expiry) AND '
E                                   '#in_progress_expiry < :now_in_millis)',
E            'ExpressionAttributeNames': {'#expiry': 'expiration',
E                                         '#id': 'id',
E                                         '#in_progress_expiry': 'in_progress_expiration',
E                                         '#status': 'status'},
E            'ExpressionAttributeValues': {':inprogress': {'S': 'INPROGRESS'},
E                                          ':now': {'N': <ANY>},
E                                          ':now_in_millis': {'N': <ANY>}},
E            'Item': {'expiration': {'N': <ANY>},
E                     'id': {'S': 'static-value'},
E                     'in_progress_expiration': {'N': <ANY>},
E                     'sk': {'S': 'test-func.functional.idempotency.test_idempotency.test_idempotent_lambda_compound_static_pk_value_has_correct_pk.<locals>.lambda_handler#be4755f20c4e6f94f43bc3f5849a26fb'},
E                     'status': {'S': 'INPROGRESS'}},
E            'TableName': 'TEST_TABLE'},
E           but received:
E           {'ConditionExpression': 'attribute_not_exists(#id) OR #expiry < :now OR '
E                                   '(#status = :inprogress AND '
E                                   'attribute_exists(#in_progress_expiry) AND '
E                                   '#in_progress_expiry < :now_in_millis)',
E            'ExpressionAttributeNames': {'#expiry': 'expiration',
E                                         '#id': 'id',
E                                         '#in_progress_expiry': 'in_progress_expiration',
[...]

PR #1899, opened 2023-02-06
fix(idempotency): make idempotent_function decorator thread safe

mploski:
**Issue number**: https://github.com/awslabs/aws-lambda-powertools-python/issues/1718

## Summary

PR modifies idempotency feature to make sure decorating user function with idempotency decorator is thread safe
### Changes

> Please provide a summary of what's being changed

PR modifies idempotency feature to use low level boto3 dynamodb client which is thread safe. It also adds other features/fixes as:

- End2end test to ensure function decorated with idempotency features can be used in multiple parallel threads
- Fix pylance type checking in an if condition
- Add [type](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/dynamodb.html) to DynamoDB conditions/items in Idempotency utility and functional tests
- Fix some other small stylistic errors

### User experience

> Please share what the user experience looks like before and after this change

User experience stays the same, but user doesn't need to worry about thread safety
## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [ ] Changes are documented
* [ ] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #1970, opened 2023-03-01
fix(idempotency): revert dict mutation that impacted static_pk_value feature

heitorlessa:
**Issue number:** #1968

> **NOTE**: This is literally a contribution from @Tankanow. Had to cut a PR to get it merged quicker due to the regression.

### Changes

Reverts a regression added in #1899 when making DynamoDB client thread-safe.

`static_pk_value` is used by customers using Single Table design pattern with DynamoDB, where a composite key is used with a static value for its "primary key" attr. This change reverts a dictionary mutation that discarded the value from `static_pk_value` and used `key_id` instead.

In simple terms, same effect as:

```python
a = {"my_key": "must_be_static"}
b = {
    **a,
    "my_key": "this is the final value"  # this shouldn't exist as it'll always win last insertion
}
``` 

### User experience

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/tests/functional/idempotency/conftest.py
+++ b/tests/functional/idempotency/conftest.py
@@ -216,6 +216,13 @@ def persistence_store_compound(config):
 
 
 @pytest.fixture
+def persistence_store_compound_static_pk_value(config, static_pk_value):
+    return DynamoDBPersistenceLayer(
+        table_name=TABLE_NAME, boto_config=config, key_attr="id", sort_key_attr="sk", static_pk_value=static_pk_value
+    )
+
+
+@pytest.fixture
 def idempotency_config(config, request, default_jmespath):
     return IdempotencyConfig(
         event_key_jmespath=request.param.get("event_key_jmespath") or default_jmespath,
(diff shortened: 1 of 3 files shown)


tests/functional/idempotency/conftest.py
```diff
@@ -246,3 +253,58 @@ def _func_echo_decoder(self, value):
...
     return mock.MagicMock()
+
+
+@pytest.fixture
+def static_pk_value():
+    return "static-value"
+
+
+@pytest.fixture
+def expected_params_update_item_compound_key_static_pk_value(
```
Tankanow, line 264: @heitorlessa, I made a slight update to reduce fixture copy/pasta: https://github.com/Tankanow/aws-lambda-powertools-python/pull/1

Tankanow: If you want, I can submit an actual PR now or you can just update this one.

Tankanow: @heitorlessa, thanks for migrating this over. I should have just submitted a PR in this repo. I didn't do that originally b/c I wanted to show the regression without cluttering the PR list here.

heitorlessa: I did that the minute you pushed to your fork ;) I'm gonna merge this now so I can go to sleep.

heitorlessa: no worries at all @Tankanow. I'm glad you took the time to create test because that would've slowed us down by an hour otherwise. It was pretty efficient to test the regression - in fact, we never had tests for that specific feature

Tankanow (commented):
I don't have the power, but FWIW, Approved. 😄

heitorlessa:
Added co-authorship to the merge commit but as your email isn't public the no-reply might not credit it correctly (I'll do in the release regardless)

https://github.com/awslabs/aws-lambda-powertools-python/commit/ca9686d23ffc4b43691d666f07f949abe46b36a8

PR #1982, opened 2023-03-05
feat(parser): support for S3 Event Notifications via EventBridge

### LINKED ISSUE #1656: Feature request: Support S3 Event Notifications Model for Event Bridge
### Use case

With the introduction of S3 Event Notifications in Event Bridge [here](https://aws.amazon.com/blogs/aws/new-use-amazon-s3-event-notifications-with-amazon-eventbridge/) we need the ability to quickly unpack the event payload from the Event Bridge Envelope.

The structure of the message is described [here](https://docs.aws.amazon.com/AmazonS3/latest/userguide/ev-events.html).

### Solution/User Experience

It would suffice to extend the s3 models in the the parser [module](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parser/models/s3.py).

### Alternative solutions

Have the module instantiated at the project level and just use the envelope unpacking mechanism.

### Acknowledgment

ivica-k:
**Issue number:** #1656 

### Changes

adds support for S3 event notifications through EventBridge

### User experience

```python
from aws_lambda_powertools.utilities.parser import parse, envelopes
from aws_lambda_powertools.utilities.parser.models import S3EventNotificationEventBridgeDetailModel

from aws_lambda_powertools.utilities.data_classes import S3EventBridgeEvent

def lambda_handler(event: dict, context):
    dataclass_event = S3EventBridgeEvent(event)
    parsed_event: S3EventNotificationEventBridgeDetailModel = parse(
        model=S3EventNotificationEventBridgeDetailModel,
        envelope=envelopes.EventBridgeEnvelope,
        event=event
    )

    print(f"From dataclass: {dataclass_event.account}")
    print(f"From dataclass: {dataclass_event.detail.object.etag}")
    print(f"From dataclass: {dataclass_event.detail.bucket.name}")

    print(f"From pydantic model: {parsed_event.object.etag}")
    print(f"From pydantic model: {parsed_event.bucket.name}")


if __name__ == "__main__":
    event = {
        "version": "0",
        "id": "17793124-05d4-b198-2fde-7ededc63b103",
        "detail-type": "Object Created",
        "source": "aws.s3",
        "account": "111122223333",
        "time": "2021-11-12T00:00:00Z",
        "region": "ca-central-1",
        "resources": [
            "arn:aws:s3:::DOC-EXAMPLE-BUCKET1"
        ],
        "detail": {
            "version": "0",
            "bucket": {
                "name": "DOC-EXAMPLE-BUCKET1"
            },
            "object": {
                "key": "example-key",
                "size": 5,
                "etag": "b1946ac92492d2347c6235b4d2611184",
                "version-id": "IYV3p45BT0ac8hjHg1houSdS1a.Mro8e",
                "sequencer": "617f08299329d189"
            },
            "request-id": "N4N7GDK58NMKJ12R",
            "requester": "123456789012",
            "source-ip-address": "1.2.3.4",
            "reason": "PutObject"
        }
    }

    lambda_handler(event=event, context=None)
```

Prints:

```python
From dataclass: 111122223333
From dataclass: b1946ac92492d2347c6235b4d2611184
From dataclass: DOC-EXAMPLE-BUCKET1
From pydantic model: b1946ac92492d2347c6235b4d2611184
From pydantic model: DOC-EXAMPLE-BUCKET1
```

### Notes

I made this PR while code changes are half done to make the discussion about further changes easier. Tests and documentation will be added later, once these functional changes are deemed good.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgement


heitorlessa:
hey @ivica-k - THANK YOU, excited to review your first PR shortly. I've just added `python` to your code snippets to enable syntax highlighting.

heitorlessa:
hey @ivica-k - reviewing it now... let's break this into two PRs: one for Parser and one for Event Source Data Classes.

aws_lambda_powertools/utilities/data_classes/__init__.py
```diff
@@ -16,7 +16,7 @@
 from .kinesis_firehose_event import KinesisFirehoseEvent
 from .kinesis_stream_event import KinesisStreamEvent
 from .lambda_function_url_event import LambdaFunctionUrlEvent
-from .s3_event import S3Event
+from .s3_event import S3Event, S3EventBridgeEvent
```
heitorlessa: Let's break this into a separate PR. That way we have two explicit PRs we can look back on those two features for distinct utilities.

aws_lambda_powertools/utilities/data_classes/s3_event.py
```diff
@@ -16,6 +19,138 @@ def source_ip_address(self) -> str:
...
+        """Get the source IP address of S3 request. Only present for events triggered by an S3 request."""
+        return self["source-ip-address"]
+
+    @property
+    def reason(self) -> Optional[str]:
+        """Get the reason for the S3 notification.
+
+        For 'Object Created events', the S3 API used to create the object: `PutObject`, `POST Object`, `CopyObject`, or
+        `CompleteMultipartUpload`. For 'Object Deleted' events, this is set to `DeleteObject` when an object is deleted
+        by an S3 API call, or 'Lifecycle Expiration' when an object is deleted by an S3 Lifecycle expiration rule.
```
heitorlessa: NICE touch!

heitorlessa (changes requested):
Looks great! The only changes we'll need to ensure this works as expected are:

* A real event generated under `tests/events` (replace accountID/resource name with dummy ones) - [S3 Object Notification Example](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/tests/events/s3Event.json)
* Test for Parser to guarantee mandatory vs optional field work as expected and don't fail at runtime with validation error: [S3 Model test example](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/tests/functional/parser/test_s3.py)

Lemme know if you need any help with it!

aws_lambda_powertools/utilities/data_classes/__init__.py
ivica-k: Will do, probably over the weekend though.

ivica-k: This PR now only contains changes related to the parser utility. Dataclasses event will be a separate PR, as you suggested.

ivica-k: Created PR for dataclasses #1990

heitorlessa:
There's only one medium change to make and I'm doing that now to speed up the merge - `S3EventNotificationEventBridgeModel` is the parent and preferred piece we document and test , because the `envelope` will throw away EventBridge metadata a customer might want to have.

`S3EventNotificationEventBridgeDetailModel` would be similar to `SqsRecordModel`, where you want to dig in where the payload is only, discarding everything else (e.g., `SqsModel`). 

With that in mind (if it's any clearer), we first test the parent model `S3EventNotificationEventBridgeModel`, and then do a specific one for its associated envelope.

I've changed the exports and the docs, I'll work on the tests.

PS: Parser testing suite is something I'd love to redo entirely

heitorlessa:
**What I've changed**

* **Added tests `_no_envelope` and replaced hardcoded data from event with the actual raw event** - this is something I wish we've done before in the Parser's test suite, as it's brittle if we ever generate events to test contracts on a regular basis
* **Fixed field type in `S3EventNotificationEventBridgeModel`** - I caught this when doing the tests without envelope, since `detail = S3EventNotificationEventBridgeModel` doesn't allow for Pydantic to [recursively instantiate and validate nested models](https://seraf.dev/nested-pydantic-model-from-dict-f332f277881). It's now `detail: S3EventNotificationEventBridgeDetailModel`.
* **Removed `None` tests from new `no_envelope` tests** - Pydantic will raise a validation error if a field is typed as required but it's missing in the payload, so we don't need to test as any discrepancy will result in a validation error. That is, fields with `Optional[]` don't need a specific assertion.


**Last items to review on Monday**

* I'm almost sure we don't need those tests that use an Envelope, but I'm dizzy from meetings and can't make that call today. 
  * **Why**: The reason for Envelope is for customers to bring their own custom model, so we'd apply their model in a particular field (`detail`) to which we'd discard all parent metadata (e.g., `event.id`) and return their Model already initialized and validated. Customers won't be able to bring their own models in this particular event per se as it's controlled by S3. If they could, they could still use the `EventBridgeEnvelope` which is already tested elsewhere.
* Fix mypy typing on `Type[Model]`.

If we don't speak again, have a fantastic weekend @ivica-k !!

ivica-k:
Damn, it would have been easier for you to make the whole PR from scratch than to fix my things...

Thank you for the great explanation of changes and have a great weekend :)

If it makes any difference, I agree with your point about testing the envelope; as you said, `EventBridgeEnvelope` is already test and working.

heitorlessa:
Not at all @ivica-k - getting the events, the structure, and kick starting everything is paramount... we can't thank you enough ;-)

We'll finish on Monday and move to the data classes to see if we can fix that super odd GH Actions checkout issue.

🙏

ivica-k:
Hey @heitorlessa I feel bad that you're doing all of these commits; is there a chance I could participate at this point?

heitorlessa:
Let me ping you on Discord and we can do it together - I didn't really need do this https://github.com/awslabs/aws-lambda-powertools-python/pull/1982/commits/a50116c1f7dad3ae5aa9b5a4594a4a279147c079, but I figured this is the best opportunity to start paying some tech debt I always wanted to (I don't like the way tests are done in Parser).

There's one last thing I'm doing now which is making sure Mypy works.

heitorlessa:
Okay finished, mind having a review to see if I missed anything?

**What have I done**?

* Replaced the original tests with `_no_envelope` tests I added
* [Paid an old typing tech debt](https://github.com/awslabs/aws-lambda-powertools-python/pull/1982/commits/f8d49e0ba6f79af40c680d33e5f102ffb62a9615) with `EventBridgeModel` which mypy detected in this PR - `S3EventNotificationEventBridgeModel` overrides `detail` with a model, which we should've addressed ages ago. A lesson I learned here was there's a surprising number of customers using Pydantic but not Mypy, otherwise there would've been an issue created related to `EventBridgeModel`.
* [Paid one of the oldest tech debt](https://github.com/awslabs/aws-lambda-powertools-python/pull/1982/commits/a50116c1f7dad3ae5aa9b5a4594a4a279147c079) in migrating functional tests to unit tests. Ideally, I should've created multiple issues for this, but I had 10 spare minutes and I thought this would be the perfect opportunity to do now.
    - Tech debt background: Parser's functional test use `event_parser` and standalone functions to do assertions. This should be decoupled to ease writing tests + perf - unit tests to test model validation, functional tests to test `parse`, `event_parser` with/without envelope for any model. **HELP**: This is an area I'd love to get help as there's plenty to decouple here in the codebase. Similarly to event source data classes that you're also working on.

**Special note**. Never feel bad about this. We wouldn't be this far without your help. There's also so many competing priorities that if you hadn't started this PR we wouldn't even touch this feature until later in the year, even more so that I'm so happy we paid two tech debts that will pave the road for a better code base. Typically, if there's an issue related I'd ask any contributor to kindly factor these in, but we haven't created an issue to cover these tech debts hence why I did it.

heitorlessa (approved):

heitorlessa:
Explanation on the typing (in case you aren't familiar with mypy)

https://github.com/ivica-k/aws-lambda-powertools-python/blob/feat(event-source)-1656-s3-event-eventbridge/aws_lambda_powertools/utilities/parser/types.py#L17

```python
AnyInheritedModel = Union[Type[BaseModel], BaseModel]
RawDictOrModel = Union[Dict[str, Any], AnyInheritedModel]
```

`Type[BaseModel]` allows us to accept any model that inherits a generic model, while `BaseModel` allows any unbounded model (those not initialized).

```python
class EventBridgeModel(BaseModel):
    ...
    detail: RawDictOrModel
```

If a customer now wants to bring their own model, this won't trigger a Mypy issue anymore:

```python
class Order(BaseModel):
    payment_id: str


class OrderEventBridge(EventBridgeModel):
    detail: Order
```

What's even better... a customer can even use the superset `Json` type from Pydantic to deserialize JSON string into a Dict and then to their Model without any mypy complaints too

```python
class Order(BaseModel):
    payment_id: str


class OrderEventBridge(EventBridgeModel):
    detail: Json[Order]
```

heitorlessa:
@ivica-k thanks for jumping on a call with me - as we've just agreed, I'll merge this now as soon as CI checks pass, and we can move on to the data class PR one.

PR #1998, opened 2023-03-10
feat(event-handler): add appsync batch resolvers

### LINKED ISSUE #1303: Feature request: support BatchInvoke when using AppSyncResolver
### Use case

Currently, an event is expected to be of type ```Dict[str,Any]```. When using 'BatchInvoke' in a direct lambda resolver, the event becomes type ```List[Dict[str,Any]]```. Hence, it is not possible to route the event to the expected resolver. 

A use case is given by the  [official dev guide](https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-lambda-resolvers.html). Without the ability to use 'BatchInvoke' users will run into n + 1 problems, when using appsync-GraphQl.

### Solution/User Experience

I am not 100% sure how a good solution would look like. Maybe it would be possible to use the ```AppSyncResolverEvent``` to handle one individual event, or a list of events. 
Given the described use case, the events should usually only diverge in the source.

### Alternative solutions

_No response_

### Acknowledgment

mploski:
**Issue Number:** https://github.com/awslabs/aws-lambda-powertools-python/issues/1303

## Summary
Support Appsync batching mechanism for Lambda Resolvers 
### Changes

* Modify Appsync BaseRouter to accept List of AppSyncResolverEventT events alongside singular event
* Store List of AppSyncResolverEventT events in BaseRouter.current_event
* Add Functional test covering batch processing
* Add aws-cdk-aws-appsync-alpha cdk package
* Add end2end tests that
   - deploys AppSync API with lambda resolver and specific schema
   -  Validates different resolvers including batching mechanism

### User experience

If batch processing is used for specific method, user can access AppSyncResolver events directly in their annotated method and do appropriate processing. Appsync resolver calls method for every event from the coming list of events.

#### Processing Batch items individually
```python
from typing import List, Optional

from aws_lambda_powertools.event_handler import AppSyncResolver
from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent

app = AppSyncResolver()


posts = {
    "1": {
        "post_id": "1",
        "title": "First book",
        "author": "Author1",
        "url": "https://amazon.com/",
        "content": "SAMPLE TEXT AUTHOR 1",
        "ups": "100",
        "downs": "10",
    },
    "2": {
        "post_id": "2",
        "title": "Second book",
        "author": "Author2",
        "url": "https://amazon.com",
        "content": "SAMPLE TEXT AUTHOR 2",
        "ups": "100",
        "downs": "10",
    },
    "3": {
        "post_id": "3",
        "title": "Third book",
        "author": "Author3",
        "url": None,
        "content": None,
        "ups": None,
        "downs": None,
    }
}

posts_related = {
    "1": [posts["2"]],
    "2": [posts["3"], posts["1"]],
    "3": [posts["2"], posts["1"]],
}


class Post(BaseModel):
    post_id: str
    author: str
    title: str
    url: str
    content: str
    ups: str
    downs: str

@app.batch_resolver(type_name="Post", field_name="relatedPosts")
def related_posts(event: AppSyncResolverEvent) -> Optional[list]:
    return posts_related[event.source["post_id"]] if event.source else None


def lambda_handler(event, context: LambdaContext) -> dict:

```

#### Processing Batch items at once

```python
from typing import Any, List

from aws_lambda_powertools.event_handler import AppSyncResolver
from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent

logger = Logger()
app = AppSyncResolver()


posts_related = {
    "1": {"title": "post1"},
    "2": {"title": "post2"},
    "3": {"title": "post3"},
}


@app.batch_resolver(type_name="Query", field_name="relatedPosts")
def related_posts(event: List[AppSyncResolverEvent]) -> List[Any]:
    results = []

    for record in event:  # (1)!
        post_id = record.arguments.get("post_id")
        try:
            results.append(posts_related[post_id] if post_id else None)
            # Add other logic here
        except Exception:
            logger.error("Error processing record", post_id=post_id)

    return results


def lambda_handler(event, context: LambdaContext) -> dict:
```

Under the hood, Router:
1. Recognize we got a list of events so it assumes request coming from appsync batch operations.
2. It registers methods that we want to use to handle batch events. We use for it brand new resolver called `batch_resolver`.
3. Router finds this method bases on `type_name` and `field_name`  and calls it for every event in the list. It passes the event itself to it as parameter argument. Optionally it pass user specified parameters if they are present - they are taken from the inside of the event itself. We pass user specified parameters to keep similar UX as with base `resolver`.
4. Once all events are processed by decorated method, router combines the output from multiple method calls in a form of a list and send it back to the appsync.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -10,7 +11,7 @@
 
 
 class BaseRouter:
-    current_event: AppSyncResolverEventT  # type: ignore[valid-type]
+    current_event: Union[AppSyncResolverEventT, List[AppSyncResolverEventT]]  # type: ignore[valid-type]
```
heitorlessa: I think this requires a new specialized class: `AppSyncBatchResolver`. 

**Why**

* `current_event` will trip people using Mypy off, as it'd be difficult to know when it's a single event vs a Batch event
* Partial error handling - a customer can return individual errors (`None`) as part of its response instead of failing flat: https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-lambda-resolvers.html#returning-individual-errors
    - We could argue that a customer would have to know this and do a try/except themselves, or we could also make it easier.

**Alternative**

We could introduce a new field: `current_batch`. This means always make sure it is populated with a single list item even if it's not a Batch Processor thing, so it makes it easier for customers migrating from Single Invoke to Batch Invoke (and vice-versa).

However, there's still a gap in how we can make it easier for customers wanting to fail gracefully in a Batch Invoke.

@cponfick-bhs, @thenamanpatwari - we'd love to hear your feedback to help provide an optimal UX for everyone.

heitorlessa (changes requested):

aws_lambda_powertools/event_handler/appsync.py
mploski: Thanks @heitorlessa  for the suggestion :-) 
I agree it might be cumbersome for the users to know what's happening. The basic idea was to do this as transparent as possible to the user and discover automatically if we deal with batch/non-batch without a need to annotate method differently. 


I like the idea of adding partial error handling, although I would add it as an feature in next step ( and think if we want to make it configurable and/or return user specific error). In this particular PR I would focus on making batch invoke operational as right now we're failing with the ugly exception.

The Alternative approach could work, but it seems like a breaking change if we remove current_event attribute OR additional effort to deprecate it later.

To sum up: I will pursue a design with AppSyncBatchResolver to make solution more explicit + update the documentation, and as a next step we can add task to add partial failure feature into the resolver

heitorlessa: Absolutely agree on partial error in a separate PR - should've clarified that upfront!

oh, I wouldn't remove the `current_event`, I'd add `current_batch` **in addition** to `current_event` as a breaking a change is off the cards :) 

If you're going to work on `AppSyncBatchResolver`, we'll still need `current_event` because a customer might have a combination of Single and Batch resolvers, yet a single `return app.resolve()` in the Lambda handler - if that makes sense?

mploski: Hmm, since we still will use AppSyncResolver as our entrypoint (to have single app.resolve()) maybe we don't need specialized AppSyncBatchResolver class. instead we could add additional attribute into `app.resolve` inidicating what we expect  i.e `@app.resolver(type_name="Query", field_name="getPost", batch=True)` or we can have specialised method for batch processing i.e `@app.batch_resolver(type_name="Query", field_name="getPost")`.

For latter option, we have later a space to extend it with additional logic -partial error handling for example.
In both cases user will know what to expect + we have additional separation for batch/non-batch routing. What do you think?

heitorlessa: I think we're onto something with `@app.batch_resolver` - allow me to summarize the call we've just had:

> **There are two fundamental problems**

1. **Correctly handling `app.current_event` when it's a Batch Invoke**. Still unsure what the value should be, whether it's a empty `AppSyncResolverEvent`, a new `AppSyncResolverEvents` to correctly identify a `list[AppSyncResolverEvent]`.
2. **Making explicit enough for the authoring experience to not rely on `app.current_event`**, since that wouldn't be correct - even if we add an Union type, this will degrade the experience for customers not using batch invoke.

What we've discussed is focusing on the second problem first by having a similar experience to the Batch Processing feature, by introducing a `@app.batch_resolver` instead of a newly specialized Class as that would confuse customers and migrating to it would be harder.

This means inverting the authoring experience for customers to only care about a single event like they do today, and we handle the Batch characteristics of it -- error handling, one event at a time, merge results into a list, partial failures, etc. 

This even opens the door to have additional tracing for every invocation so they know which field took the longest, easily with `@tracer.capture_method`.

**Example**

```python
from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent

app = AppSyncResolver()

@app.batch_resolver(type_name="Query", field_name="listTodos")
def list_todos(event: AppSyncResolverEvent) -> Todo:
    return Todo() # ValueError.... [{Todo1}, None, {Todo3}]


```

mploski: hey @heitorlessa I applied discussed changes. Please take a look at this one more time :-) Once we get to the conclusion, I want also to modify documentation. If you can point me to the best place where we want to add it ( examples? docstrings?) that would be great.

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -77,16 +98,20 @@ def common_field() -> str:
...
+        context: LambdaContext,
+        data_model: Type[AppSyncResolverEvent] = AppSyncResolverEvent,
     ) -> Any:
         """Resolve field_name
 
         Parameters
         ----------
-        event : dict
+        event : dict | List[dict]
             Lambda event
```
heitorlessa: now that we have a `List[dict]` too, add more context that the event also support BatchInvoke.

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -1,21 +1,32 @@
...
+from typing import Any, Callable, List, Optional, Type, Union
 
 from aws_lambda_powertools.utilities.data_classes import AppSyncResolverEvent
 from aws_lambda_powertools.utilities.typing import LambdaContext
 
 logger = logging.getLogger(__name__)
 
-AppSyncResolverEventT = TypeVar("AppSyncResolverEventT", bound=AppSyncResolverEvent)
 
+class RouterContext:
```
heitorlessa: Nice!! Worth checking whether a `Router` instance still gets the Context injected correctly (in addition to AppSyncResolver) - our functional tests should cover it already.

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -28,23 +39,33 @@ def resolver(self, type_name: str = "*", field_name: Optional[str] = None):
...
+        def register(func):
+            logger.debug(f"Adding batch resolver `{func.__name__}` for field `{type_name}.{field_name}`")
+            self._batch_resolvers[f"{type_name}.{field_name}"] = {"func": func}
+            return func
 
-class AppSyncResolver(BaseRouter):
+        return register
+
+
+class AppSyncResolver(ResolverRegistry, RouterContext):
```
heitorlessa: favour composition instead. This will bite us when we use `mypyc` to compile pure python code like these for speed, since multi-inheritance/mixins aren't a thing ... among other typical issues with it.

heitorlessa (dismissed):
Thank you so much for the change, that's much better! Made one comment about favouring composition instead of multiple inheritance (it'll bite us!).

One super helpful change would be to update PR description with a sample UX, so that customers could easily spot what's coming, we could reuse in our docs, and release notes :-)

I'll do a proper review next week

aws_lambda_powertools/event_handler/appsync.py
mploski: Correct, functional tests have it covered and it works as expected

aws_lambda_powertools/event_handler/appsync.py
mploski: Refactored fully to use composition. Broke the dependencies between classes by introducing interfaces

mploski:
> Thank you so much for the change, that's much better! Made one comment about favouring composition instead of multiple inheritance (it'll bite us!).
> 
> One super helpful change would be to update PR description with a sample UX, so that customers could easily spot what's coming, we could reuse in our docs, and release notes :-)
> 
> I'll do a proper review next week

@heitorlessa Did further refactor to address things we discussed (UX + design). Let me know if that's ok

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -1,21 +1,78 @@
...
+        """Append key=value data as routing context"""
+        self._context.update(**additional_context)
+
+    @context.deleter
+    def context(self):
+        """Resets routing context"""
+        self._context.clear()
+
+
+class IResolverRegistry(ABC):
```
heitorlessa: Perhaps you were programming in another language and forgot this?

We don't do that here ;-) /s

mploski: Eh, I used different convention I'm using in different project ;-) Python doesn't have agreed convention for naming interfaces ( at least there is no mention of this in PEP8) and thats what happens :-) Anyway, will adjust it to our own convention. Thx!

mploski: hey! Pushed one more commit to the branch. It address 3 things:
* Refactor interface naming
* Decided to change dependencies between Router and AppSyncResolver. Since it seems they are heavy dependent, have the same public interfaces it makes sense to use inheritance here. This way I dont need to copy the same interfaces between two classes as before
* Simplifying `_call_batch_resolver` method and remove groupby
cc @leandrodamascena

leandrodamascena:
Hello @mploski! I'm still working on validating this PR and creating tests with as many variations as we have to make sure it's ok. So far I haven't found any problems.
We're releasing a release today to fix an emergency bug we have in Codepipeline's EventSource, but I hope to include this PR in the next PR.

Thanks as always for putting in the effort to make this project even better. :rocket:

leandrodamascena:
Hi @mploski!! 

I tested the code in different ways and with different resolvers and fields and everything worked exactly as expected. I spent a few days doing exploratory tests and found nothing that failed. I also ran the end2end tests and checked the features and everything was as expected.

My questions/next steps at this point are:

1 - Whether we will have a problem in the end2end test when running the different versions of Python in parallel. I haven't used the `aws-cdk.aws-appsync-alpha` constructor yet, and I don't know if there are fields that need to be unique across stacks. I know we may have two APIs in AppSync with the same name, but we need to check.

2 - For me the code makes sense and we don't have a break in the current functionality, but I'd like a double-check here @heitorlessa! Just to make sure I'm not missing anything. 

3 - We need to adjust the documentation to explain that we now support batch calling and how the user can use it.

But again, great job @mploski. Every PR I interact with you, I learn a lot of new things and improve my skills! Thank you for your patience and we will merge this PR soon :)

thenamanpatwari:
Based on my understanding of the User Experience description of this PR, I am not fond of using AppSyncResolver to iteratively call my python code for every event passed in the batch request. In the proposed setup, I no longer see the benefits of using AppSync Batch Resolvers. Please correct me if I am mistaken. 

To illustrate, the Lambda function related_posts() in the provided example uses the "post_id" field from the posts_related dictionary for each event passed in AppSyncResolverEvent. In real-world scenarios, the in-memory lookup of posts_related dictionary is substituted with a database call. We use GraphQL to batch these database calls, particularly if the "post_id" field is expected to be repeated across multiple resolver calls. While the caching ability of AppSync is helpful, the TTL is not very effective (and expensive), and the initial call to populate the AppSync cache can be very slow.

As a user of AppSync Batch Resolvers, I would like to have more control over how my lambda resolvers group responses to Batch Events. This allows us to solve the [GraphQL n+1 problem](https://hygraph.com/blog/graphql-n-1-problem)

The example provided shows how we use Lambda functions to respond to AppSync Batch requests, with the posts_batch_resolver() function being a key part of the setup.


```python
@router.resolver(type_name="Post", field_name="relatedPosts")
@cdk_gql.appsync_batch(2000)
@tracer.capture_method
def posts_batch_resolver() -> List[List[Optional[Post]]]:
   # Extract IDs from each request in Batch Call. This will be used for the DB Batch call
    post_ids = batch_util.get_required_ids(event_list=router.current_event.event_list, fieldname="post_id")   

    # Remove duplicate IDs across all events. This reduces redundant lookups for the DB call
    unique_ids = list(set(post_ids))

    # Make single DB call to retrieve data for all IDs
    unique_dict = dao.get_relatedposts_from_id_list(id_list= unique_ids)

    # Build response
    to_return = []
    for _id in post_ids:
        to_return.append({'data':unique_dict[_id] if _id in unique_dict else None})

    return to_return
```

This is the base Lambda function that sets up the Router and resolvers the AppSync Request..


```python

@logger.inject_lambda_context(correlation_id_path=correlation_paths.APPSYNC_RESOLVER)
    """ This is the base Lambda handler for all GraphQL reoslvers. """
    if isinstance(event, list):
        return app.resolve(event, context, data_model=batch_util.BatchRequest)
```


This is the batch_util file

```python
from typing import List
from aws_lambda_powertools.utilities.data_classes.appsync_resolver_event import (
    AppSyncResolverEvent,
)

class BatchRequest(AppSyncResolverEvent):

    def __init__(self, event_list):
        super().__init__(event_list[0])

        self._event_list : List[AppSyncResolverEvent] = []
        for event in event_list:
            self._event_list.append(AppSyncResolverEvent(event))

    @property
    def event_list(self) -> dict:
        return self._event_list

def get_required_ids(event_list: List[AppSyncResolverEvent], fieldname: str) -> List[str]:
    ids = []
    for event in event_list:
        ids.append(event.source[fieldname])
    return ids
```



Please let me know if this makes sense.

Btw `@cdk_gql.appsync_batch(2000)`  is a custom decorator we use to auto-generate `aws-cdk-lib/aws-appsync` CDK code  for creating resolvers. I would love to discuss this use case in a separate ticket.

leandrodamascena:
Thanks for the feedback to improve the user experience on this PR @thenamanpatwari. I believe that a better way to resolve this would be to schedule a meeting to discuss this point and understand how to implement this feature.

Do you have an agenda for a meeting this coming week? @mploski @heitorlessa @thenamanpatwari.

Thank you

thenamanpatwari:
I'd love to connect and figure this out.
I am available either on May 11 or May 18 around noon CST

mploski:
> Thanks for the feedback to improve the user experience on this PR @thenamanpatwari. I believe that a better way to resolve this would be to schedule a meeting to discuss this point and understand how to implement this feature.
> 
> Do you have an agenda for a meeting this coming week? @mploski @heitorlessa @thenamanpatwari.
> 
> Thank you

Happy to connect and discuss it in details.  Im on PTO 11-18 May. Im ready to jump in the call before or after.

thenamanpatwari:
Let us talk after your PTO on May 22. Does noon CST work?
Can you please setup a Chime room? I am not really sure how I should share my email address with you.

heitorlessa:
Naman - could you send an email to aws-lambda-powertools-feedback at amazon dot com?

We can then reply with a link with multiple slots you can choose to make this easier.

Apologies for the back and forth on this

thenamanpatwari:
Email sent. 
No issues on the back-and-forth, I am just glad this is moving forward. I want to maintain as little code as possible haha.

leandrodamascena:
Hi @thenamanpatwari! Thank you so much for meeting and for providing useful information and data points to help us deliver a better experience for customers. For now, it's our homework to review all the details you've given us and think about how to take the next steps. Probably next week we will have news about this and we will update here so we can decide the next steps in this PR.

Thanks for supporting Powertools and taking the time to tell us about your experience.

joseph-wortmann:
Very nice stuff! We're going to begin using this library to speed up our Lambda work.

I implemented my own [appsync-router](https://github.com/QuiNovas/appsync-router) package a few years ago. Some take aways from the batch processing portion (we use batch invocation quite heavily):

- For most batch invocations, as the datasource developer I don't (and shouldn't) care that It is a batch invocation. The resolver/router library should handle this for me.
- I should be able to control whether the methods fulfilling the invocations are run concurrently or sequentially (my stuff only does this at the router level, an improvement would be at the field level). Sequentially can be pretty bad for large batches that require individual DB or HTTP resolution.
- For some batch invocations, the ability to deliver the whole batch at once (as detailed above) is really nice. That was not part of what I built, but I can definitely see a use, especially if you can make one data access call to resolve the whole batch.

I'm thinking about using your resolver in going-forward projects. Also thinking about contributing, if you'll have me. We do a LOT of Lambda development in Python (actually, most of our work).

heitorlessa:
Just had a call with @mploski and @leandrodamascena ... @leandrodamascena will comment with an actionable plan here soon.

After reading this conversation and learning what Michal and Leandro discussed with @thenamanpatwari, there's another way to improve this with Partial failure + all-at-once processing (Asyncio) while giving power users more flexibility to prevent a N+1.

leandrodamascena:
Hello everyone. The next steps in this PR are as follows:

1 - We will enhance the user experience by handling partial failures in batch processing. This improvement will simplify the handling of failed records and reduce the need for manual implementation of error-handling records.

2 - We will introduce the ability to process records async, similar to [BatchProcessor utility](https://awslabs.github.io/aws-lambda-powertools-python/2.16.2/utilities/batch/#processing-messages-asynchronously). For batch resolvers defined as async functions, we'll call with all items at the same time (asyncio.gather), since the intent is clear it can be processed concurrently safely.

@mploski we need to work to add these features and create the documentation.

@thenamanpatwari, your use case is highly valuable to us, and we don't want to miss the opportunity to implement this advanced feature for batch processing with AppSync. We kindly request you open a [Feature Request issue](https://github.com/awslabs/aws-lambda-powertools-python/issues/new?assignees=&labels=feature-request%2Ctriage&projects=&template=feature_request.yml&title=Feature+request%3A+TITLE) so that we can prioritize and support this feature. If possible, please consider submitting a PR with the code for us to merge into the repository. It is very important for us to have contribs/PR from our users.

Thank you all for your patience. Indeed, this implementation has involved complex discussions and considerations.

thenamanpatwari:
Hey @leandrodamascena 

Thank you for the update. I have been playing around with[ SQS Batch Processing feature of Lambda Power Tools ](https://docs.powertools.aws.dev/lambda/python/2.16.2/utilities/batch/#processing-messages-asynchronously) and I understand the reasoning behind your use-case. I think there are situations when this feature would be useful for AppSync. However this feature does not solve the GraphQL n+1 problem.

I accept your reasoning of implementing the advanced "batch" batch processing feature as an add-on after your initial implementation for batch processing. I will wait for this PR to get merged and then will either raise a new ticket or consider doing a PR myself along with the help of @joseph-wortmann

heitorlessa:
Sounds like a plan!

@mploski @leandrodamascena whats the latest here?

What can I do to get this over to the finish line?

thenamanpatwari:
We have identified some really cool use-cases for this feature. Eagerly looking forward to its successful implementation.

leandrodamascena:
Hey everyone! I had a meeting with @mploski to complete the changes to this PR and we agree to add the `Partial Error Handling` to a separate PR. The reason we decided this is because returning the error to AppSync requires the customer to [change the mapping template](https://docs.aws.amazon.com/appsync/latest/devguide/tutorial-lambda-resolvers.html#returning-errors) to handle the error and it's done. 

We'll split the PRs to make code review easier and allow us to create consistents examples of this. This PR has been open for a long time and adding the `Partial Error Handling` now would take much longer to merge.

Missing in this PR:
- [ ] Documentation

thenamanpatwari:
Hey team, any updates?
Building a feature currently where this capability would be insanely helpful

rubenfonseca:
> Hey team, any updates? Building a feature currently where this capability would be insanely helpful

Hi @thenamanpatwari thank you for the ping! We are currently working on two major features that will be released in the beginning of January, and immediately after that we'll look at finalising this one. So our current target to release this feature is the end of January.

leandrodamascena:
Hey everyone! I appreciate your patience in waiting for this new feature. I'm working on refactoring some parts of this code and we hope to release it in the next version scheduled for February 2nd.

In case of any update, I'll let you know.

leandrodamascena:
Hey @rubenfonseca! This PR is ready for review!

Thanks

leandrodamascena:
Hey, everyone! We will have an internal meeting with the AppSync Service team to discuss error handling when consuming an AppSync batch process. We want to hear from them about use cases and what challenges customers are facing.

After this meeting, I will update this PR. The expectation of having this feature in the next release remains, but I will update this PR if we have any additional news from this meeting.

leandrodamascena:
The meeting with the AppSync team has been rescheduled for Friday the 2nd and we will not have time to include this feature in this release. We plan to include it in the other release in 1 or 2 weeks.

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -28,69 +261,109 @@ def resolver(self, type_name: str = "*", field_name: Optional[str] = None):
...
+        return {"preferredPaymentMethod": method}
+
+    @app.resolver(field_name="commonField")
+    def common_field() -> str:
+        # Would match all fieldNames matching 'commonField'
+        return str(uuid.uuid4())
+    ```
+    """
+
+    def __init__(self, raise_error_on_failed_batch: bool = False):
```
heitorlessa:
```suggestion
    def __init__(self):
```

Move this to the `batch_resolver` decorator and name it `raise_on_error=True`.

**Why**

This will allow customers to have fine-grained control. For example, they might have more than one batch operations, where one could fail gracefully (partial error) while critical ones should fail upon the first failure.

This solves both. As a bonus, it shortens the flag, makes it more explicit, and it decouples from Direct Invocation ;)

heitorlessa (changes requested):
UX change to make this more powerful ;)

leandrodamascena:
Hey @heitorlessa! Thanks for your suggestion about implementing fine-grained exception control per route. It makes a lot more sense and add flexibility when handling exceptions in specific routes.

I made the changes and pls you can review now.

heitorlessa:
Thank you Le! self-assigning and looking into the afternoon to get the final doc touches.

heitorlessa:
Note to self before merging -- `raise_on_error` flag should only be available in the Batch resolver, not direct resolver.

heitorlessa:
I'm gonna need more time, I've just noticed we don't have a SAM example for how to configure a [Batch Resolver integration](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-property-graphqlapi-resolver.html#sam-property-graphqlapi-resolver-examples).

I'll chat with @rubenfonseca whether we make a patch release to drop py3.7 today, and then end of next week release this feature with docs + UX corrections.

---

> ![NOTE]
> I'm taking the time to refresh parts of the doc as we haven't improved this section after new writing learnings

For example, a new intro diagram

```mermaid
stateDiagram-v2
    direction LR
    EventSource: AWS Lambda Event Sources
    EventHandlerResolvers: AWS AppSync Direct invocation<br/><br/> AWS AppSync Batch invocation
    LambdaInit: Lambda invocation
    EventHandler: Event Handler
    EventHandlerResolver: Route event based on GraphQL type/field keys
    YourLogic: Run your registered resolver function
    EventHandlerResolverBuilder: Adapts response to Event Source contract
    LambdaResponse: Lambda response

    state EventSource {
        EventHandlerResolvers
    }

    EventHandlerResolvers --> LambdaInit

    LambdaInit --> EventHandler
    EventHandler --> EventHandlerResolver

    state EventHandler {
        [*] --> EventHandlerResolver: app.resolve(event, context)
        EventHandlerResolver --> YourLogic
        YourLogic --> EventHandlerResolverBuilder
    }

    EventHandler --> LambdaResponse
```

heitorlessa:
testing automation as I resume working on this today 🎉

heitorlessa:
Progress before merging

* [ ] Refine docs
  * [ ] Mention of `current_batch_event` over `current_event`
  * [ ] Exception handling
  * [ ] Correct branding (AppSync vs Appsync)
  * [ ] Note that we use `gather` for high concurrency in async batch
  * [ ] Add proper GraphQL Schema to best explain N+1 issue
  * [ ] Improve code snippet with a realistic use case ^
* [ ] Get clarification about `InconsistentPayload` exception (when can this happen?)
  * If it is expected, document it
* [ ] Review tests
  * [ ] Update tests with an actual batch schema example (it's reusing single resolver data)
  * [x] Make async batch processing functions async (some are sync causing false positives)
  * [x] Learn about batch tests (+confirm use cases coverage)
* [x] Revisit `call`, `async_call`, etc. levels of abstraction
  * [x] batch_resolver
    * [x] Enrich docstring with call flow/chain
    * [x] Exception names (`Error` suffix)
    * [x] Simplify argument name
    * [x] Reduce redundant docstrings on exceptions, keep at the top only when DRY is used
    * [x] Review whether we have debugging statements
    * [x] Create `try/catch` once instead of every loop iteration
    * [x] Reduce async call stack by using `asyncio.gather(return_exceptions=True)`
  * [x] single_resolver
    * [x] Exceptions (verify if `ValueError` is new or before)
    * [x] add debugging
* [x] `raise_on_error` on batch_resolver decorator only (edit: it was correct, got confused by new layers of abstraction)
* [x] Prefix `async` instead of async in the middle (`batch_async` vs `async_batch`)
* [x] Undo composition where it increases call stack
  * [x] Context
  * [x] Move router related logic to `router.py` to ease maintenance and grasp levels of indirection
* [x] Revisit Registry
  * [x] Use explicit methods over getter/setters
* [x] Rename `BasePublic` to `BaseRouter` (explicitness on intention/use)

heitorlessa:
self reminder to chat with Leandro about `return_exceptions` and call stacks.

image

heitorlessa:
Finished reviewing and last bits of refactoring, docstrings, etc. -- I'm leaving the [last bits](https://github.com/aws-powertools/powertools-lambda-python/pull/1998#issuecomment-1959349360) to you @leandrodamascena so we can merge this.

In summary:

* Docs need improvement -- customers will appreciate a more realistic use case, a schema, etc.
* `InconsistentPayloadError` -- when would this happen? If it can happen due to misconfiguration, explain in the docs
* Tests need minor improvement -- tests are reusing single resolver data, we need a more specific schema (it'll shorten the next maintainer cognitive cycle)
* `batch_resolver` decorator -- missing a docstring with an example to improve UX; created an example in `@resolver` that can be used as inspiration

buggy:
> Router finds this method bases on type_name and field_name and calls it for every event in the list. It passes the event itself to it as parameter argument. Optionally it pass user specified parameters if they are present - they are taken from the inside of the event itself. We pass user specified parameters to keep similar UX as with base resolver.

The problem with having the router call a method for each event then assembling the results is that you lose efficiently and simply move the N+1 problem further down the stack.

For example: You end up running multiple small queries against a database, sometimes requesting the same data, rather than de-depulicate ID's then run a single query to fetch all data.

Your Lambda function no longer has the N+1 problem because you moved it to the database or where ever you're fetching data

heitorlessa:
100% Rich - thank you for catching this early.

We were planning to call the method with the entire batch and build the
result later as a follow up. However, we missed the bigger picture
altogether as this suffered many changes.

I wonder if the value is only routing to a method. The graceful error
handling could be done by the method we route to since they control the
result builder now.

Thoughts?

heitorlessa:
Re-sending as formatting got lost when replying via email.

100% Rich - thank you for catching this early. We were planning to call the method with the entire batch once and build the result later as a follow up. 

However, we missed the bigger picture altogether as this suffered many changes and reviews 🤦 -- with the entire batch, you could filter duplicates and call DB fewer times or once. 

I wonder though - is there any other value add we could provide besides routing?

Graceful error handling is lost as we lose successful/failure visibility. I thought of a new mechanism to invert dependency -- the called method tells us what worked, and upon exception we handle it.

As for the result builder, we'd merely propagate through hence me questioning the value add besides routing. In the future the value could be data validation like we do in REST. 

Thoughts?

leandrodamascena:
> * `InconsistentPayloadError` -- when would this happen? If it can happen due to misconfiguration, explain in the docs

Hello @heitorlessa! As we discussed yesterday and agreed, I removed this check to allow more use cases and prevent us from blocking someone.

I'm working on the other items.

heitorlessa:
can I have a quick update on this, please?

heitorlessa:
@leandrodamascena when you're back on Friday, could you schedule a review with @sthulb and I so we can schedule it for release either next Thursday (14th) or the 28th?

leandrodamascena:
Hey @heitorlessa! Ready for the final review.

tests/e2e/event_handler_appsync/test_appsync_resolvers.py
```diff
@@ -110,3 +110,41 @@ def test_appsync_get_related_posts_batch(appsync_endpoint, appsync_access_key):
         assert post["post_id"] in related_posts_ids
     for post in data["getPost"]["relatedPostsAsync"]:
         assert post["post_id"] in related_posts_ids
+
+
+@pytest.mark.xdist_group(name="event_handler")
+def test_appsync_get_related_posts_batch_with_aggregate(appsync_endpoint, appsync_access_key):
```
heitorlessa, line 129: could you break the assertions for sync first, then async, and add more details on GIVEN/WHEN as we're doing two tests here, plz?

tests/functional/event_handler/required_dependencies/appsync/test_appsync_batch_resolvers.py
```diff
@@ -942,3 +649,181 @@ async def get_locations2(event: AppSyncResolverEvent, name: str) -> str:
...
+                "id": [3, 4],
+            },
+        },
+    ]
+
+    app = AppSyncResolver()
+
+    # WHEN the batch resolver for the listLocations field is defined
+    @app.batch_resolver(field_name="listLocations")
+    def create_something(event: List[AppSyncResolverEvent]) -> Optional[list]:  # noqa AA03 VNE003
```
heitorlessa: I think you meant `list` instead of `Optional[list]`, otherwise this would fail in your exception.

I think you also have a test for that type of scenario too (something that doesn't return)

tests/functional/event_handler/required_dependencies/appsync/test_appsync_batch_resolvers.py
```diff
@@ -942,3 +649,181 @@ async def get_locations2(event: AppSyncResolverEvent, name: str) -> str:
...
+    # THEN the resolver should correctly process the batch of queries
+    result = app.resolve(event, LambdaContext())
+    assert result == [appsync_event["source"]["id"] for appsync_event in event]
+
+    assert app.current_batch_event and len(app.current_batch_event) == len(event)
+    assert not app.current_event
+
+
+def test_resolve_batch_processing_with_aggregate_and_returning_a_non_list():
+    # GIVEN a list of events representing GraphQL queries for listing locations
```
heitorlessa, line 779: could you do one last pass on GIVEN/WHEN/THEN so it matches the tests (what we're testing) etc, plz plz?

tests/functional/event_handler/required_dependencies/appsync/test_appsync_batch_resolvers.py
```diff
@@ -942,3 +649,181 @@ async def get_locations2(event: AppSyncResolverEvent, name: str) -> str:
...
+            },
+        },
+    ]
+
+    app = AppSyncResolver()
+
+    # WHEN the batch resolver for the listLocations field is defined
+    @app.async_batch_resolver(field_name="listLocations")
+    async def create_something(event: List[AppSyncResolverEvent]) -> Optional[list]:  # noqa AA03 VNE003
+        return event[0].source.get("id") if event[0].source else None
```
heitorlessa, line 833: idea: have one test with where you don't return anything, simply process whatever. It's something our customers would end up accidentally having too.

heitorlessa (changes requested):
a few changes in GIVEN/WHEN/THEN and usage of `response[0]`. 

I'll run the docs locally this afternoon for a proper review!

tests/functional/event_handler/required_dependencies/appsync/test_appsync_batch_resolvers.py
leandrodamascena: Yeah! Done

tests/functional/event_handler/required_dependencies/appsync/test_appsync_batch_resolvers.py
leandrodamascena: Yeah! Done

tests/e2e/event_handler_appsync/test_appsync_resolvers.py
leandrodamascena: Done

heitorlessa (dismissed):
NOW IT IS COMPLETED! SHIP!

```mermaid
stateDiagram-v2
    direction LR
    LambdaInit: Lambda invocation
    EventHandler: Event Handler
    EventHandlerResolver: Route event based on GraphQL type/field keys
    Client: Client query (listPosts)
    YourLogic: Run your registered resolver function
    EventHandlerResolverBuilder: Verifies response is a list
    AppSyncBatchPostsResolution: query listPosts
    AppSyncBatchPostsItems: get all posts data <em>(id, title, relatedPosts)</em>
    AppSyncBatchRelatedPosts: get related posts <em>(id, title, relatedPosts)</em>
    AppSyncBatchAggregate: aggregate batch resolver event
    AppSyncBatchLimit: reached batch size limit
    LambdaResponse: Lambda response

    Client --> AppSyncBatchResolverMode
    state AppSyncBatchResolverMode {
        [*] --> AppSyncBatchPostsResolution
        AppSyncBatchPostsResolution --> AppSyncBatchPostsItems
        AppSyncBatchPostsItems --> AppSyncBatchRelatedPosts: <strong>N additional queries</strong>
        AppSyncBatchRelatedPosts --> AppSyncBatchRelatedPosts
        AppSyncBatchRelatedPosts --> AppSyncBatchAggregate
        AppSyncBatchRelatedPosts --> AppSyncBatchAggregate
        AppSyncBatchRelatedPosts --> AppSyncBatchAggregate
        AppSyncBatchAggregate --> AppSyncBatchLimit
    }

    AppSyncBatchResolverMode --> LambdaInit: 1x Invoke with N events
    LambdaInit --> EventHandler

    state EventHandler {
        [*] --> EventHandlerResolver: app.resolve(event, context)
        EventHandlerResolver --> YourLogic
        YourLogic --> EventHandlerResolverBuilder
        EventHandlerResolverBuilder --> LambdaResponse
    }
```

leandrodamascena (approved):
APPROVED @heitorlessa! ❤️

heitorlessa (approved):

sthulb (approved):

thenamanpatwari:
A critical bug has been observed for this feature - https://github.com/aws-powertools/powertools-lambda-python/issues/6251

PR #2004, opened 2023-03-13
fix(typing): swap NoReturn with None for methods with no return value

### LINKED ISSUE #2002: `BaseSegment` defines `put_annotation` as having return type `NoReturn`
### Expected Behaviour

The return type should be `None`. `NoReturn` means that the function never returns (e.g. it always throws an exception) and makes type checkers (e.g. mypy) mark all code following `put_annotation` as unreachable.

Also applies to `put_metadata`, `patch`, and `patch_all`.

### Current Behaviour

These methods breaks type checking in any modules they're used because all code following a call to e.g. `put_annotation` is marked as unreachable.

### Code snippet

```python
with Tracer().provider.in_subsegment('test') as subsegment:
    subsegment.put_metadata('x', 1)
    print("hi")  # marked as unreachable
```

### Possible Solution

Change the return type to `None`.

### Steps to Reproduce

Use the code snippet above and run `mypy` on the file.

latest

3.9

### Packaging format used

PyPi

### Debugging logs

_No response_

tibbe:
**Issue number:** #2002

## Summary

NoReturn are for methods that never return (e.g. always raise exceptions). Using NoReturn causes e.g. mypy to think all code following the call to the NoReturn method to be unreachable.

### Changes

The

- `put_annotation`
- `put_metadata`
- `patch`
- `patch_all`

methods of `BaseSegment` now have the correct return type for methods that return not (but do in fact return).

### User experience

Before this change using any of the above methods would lead to type errors/unreachable code errors when used with e.g. mypy.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/tracing/base.py
+++ b/aws_lambda_powertools/tracing/base.py
@@ -2,7 +2,7 @@ import abc
 import numbers
 import traceback
 from contextlib import contextmanager
-from typing import Any, Generator, List, NoReturn, Optional, Sequence, Union
+from typing import Any, Generator, List, Optional, Sequence, Union
 
 
 class BaseSegment(abc.ABC):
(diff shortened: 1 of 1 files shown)


rubenfonseca (approved):

rubenfonseca:
@tibbe thank you so much for reporting and fixing this so quickly! We really appreciate this, and we will release this to the public later this week

PR #2073, opened 2023-03-30
docs(idempotency): fixes to testing your code section

### LINKED ISSUE #2072: Docs: Fix local tests example in the Idempotency Testing Your Code section

Improve the documentation by changing the example to be more realistic.

Idempotency

### How can we improve?

Changing documentation

### Got a suggestion in mind?

_No response_

### Acknowledgment

leandrodamascena:
**Issue number:** #2072 

## Summary

Fix local test example

### Changes

 - Changed documentation

### User experience

Before
image

After
image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


rubenfonseca (approved):

docs/utilities/idempotency.md
```diff
@@ -1115,10 +1115,10 @@ To test with [DynamoDB Local](https://docs.aws.amazon.com/amazondynamodb/latest/
         return LambdaContext()
 
     def test_idempotent_lambda(lambda_context):
-        # Create our own Table resource using the endpoint for our DynamoDB Local instance
-        resource = boto3.resource("dynamodb", endpoint_url='http://localhost:8000')
-        table = resource.Table(app.persistence_layer.table_name)
-        app.persistence_layer.table = table
+        # Configure the boto3 to use the endpoint for the DynamoDB Local instance
+        resource = boto3.client("dynamodb", endpoint_url='http://localhost:8000')
```
heitorlessa: have you tested it? sounds like we're missing something here - we're creating a low-level client, not resource client anymore, but then we don't use it... and we're using a hardcoded table name instead of using the previously registered table name in the persistency layer.

leandrodamascena: I tested it and was able to reproduce it in my local environment, even when it fails because I added a table that doesn't exists. But yes, it might be good if you double-check.

I added the hardcoded name to demonstrate to the customer that it can be changed by setting a new value for the variable. Doesn't it make sense for us to keep this or even get it from the current `app.persistence_layer.table_name` variable? I think just removing the hardcoded we don't show the real use of this.

heitorlessa: Definitely get from the current `table_name` variable (test to be sure). Reason being is that we're trying to demonstrate how customers would test themselves - if they use Serverless framework for example, a plugin would automatically create a DynamoDB table with the name of the table they're using in Lambda.

If we show it hard coded, it could lead to confusion - `Wait? Do I have to use a different table name? Should I update my DynamoDB Local table bootstrap script?`.

Docs point of view: We're testing `app.py` who already set the table name, `DynamoDBPersistenceLayer(table_name="idempotency")`. From a testing point of view, you'd want to use the same resource defined there, so you don't get a drift. If you want to set a separate one, a good practice would be to actually set the table name in `app.py` using env var which a test could inject before importing it :) 

Hope that makes sense.

leandrodamascena: > Definitely get from the current table_name variable (test to be sure). Reason being is that we're trying to demonstrate how customers would test themselves - if they use Serverless framework for example, a plugin would automatically create a DynamoDB table with the name of the table they're using in Lambda.

Done. I added a comment to explain that the user can change the variable value and use other value for the table name.

> If we show it hard coded, it could lead to confusion - Wait? Do I have to use a different table name? Should I update my DynamoDB Local table bootstrap script?.

Yes, I totally agree with your explanation.

> Docs point of view: We're testing app.py who already set the table name, DynamoDBPersistenceLayer(table_name="idempotency"). From a testing point of view, you'd want to use the same resource defined there, so you don't get a drift. If you want to set a separate one, a good practice would be to actually set the table name in app.py using env var which a test could inject before importing it :)

Including this as an environment variable might confuse users that this environment variable is something from the idempotency provider and not specific to this test, no? I mean, for example, we have a `POWERTOOLS_IDEMPOTENCY_DISABLED` variable that is global to the provider and it makes sense to set it via an environment variable. I'm really not sure, I'm just bringing one more thought to the table.

leandrodamascena:
I am moving the comment from issue to PR. 
Refer: https://github.com/awslabs/aws-lambda-powertools-python/issues/2072#issuecomment-1494339573

> Heitor Lessa:
only last comment before merging is that this part shouldn't work, please double check in case I'm wrong: [#2073 (files)](https://github.com/awslabs/aws-lambda-powertools-python/pull/2073/files#diff-e612f95a3dbd33cabab2e49899653b8cc5d5e809bfad45aa54cdee41d22e5808R1253)
Reason. Unless AWS SDK changed, the DynamoDB client wouldn't connect to DynamoDB Local because the client isn't being replaced, therefore it'd use the real DynamoDB external endpoint

Yeah @heitorlessa, you're right! I was missing it because in my local tests, I was configuring the boto3.session in the `app.py` file and I forgot to remove it. My bad. I already sent the new code.

heitorlessa:
@leandrodamascena created PR exposing the SDK Client **and** accepting a custom SDK client so customers can do this more cleanly.

I'll let you review, once merged, then we can finish this one.

https://github.com/awslabs/aws-lambda-powertools-python/pull/2087

heitorlessa (approved):

PR #2108, opened 2023-04-11
feat(parser): add support for SQS-wrapped S3 event notifications

### LINKED ISSUE #2078: Feature request: Improve support for SQS-wrapped S3 event notifications
### Use case

_Originally discussed in https://github.com/awslabs/aws-lambda-powertools-python/issues/1656#issuecomment-1384670725._

S3 event notifications can be sometimes be ingested into Lambda via an intermediary such as an SQS queue (i.e. Lambda event source), for various architectural reasons - batching, retries, etc.. However, from the Lambda function's perspective, the intermediary might not be too important; what's important is the S3 event notification itself.

With the current Powertools built-ins, it is possible to parse the [S3 event notification data structure](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html) out of the [SQS message data structure](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html#example-standard-queue-message-event), but this requires some awkward boilerplate code to chain the two data structures together (mostly due to the JSON-formatted wrapper in the middle):

```py
from aws_lambda_powertools.utilities.batch import BatchProcessor, EventType, batch_processor
from aws_lambda_powertools.utilities.parser.models import SqsRecordModel, S3Model
from pydantic import Json

class SqsS3EventNotificationModel(SqsRecordModel):
  body: Json[S3Model]

processor = BatchProcessor(event_type=EventType.SQS, model=SqsS3EventNotificationModel)

def record_handler(record: SqsS3EventNotificationModel):
  for s3_record in record.body.Records:
    do_something_useful(s3_record)

@batch_processor(record_handler=record_handler, processor=processor)
def handler(event, context):
  return processor.response()
```

### Solution/User Experience

A simple quick-win to the user-experience would be to make a new data model available as a built-in:

```py
from aws_lambda_powertools.utilities.parser.models import SqsRecordModel, S3Model
from pydantic import Json

class SqsS3EventNotificationModel(SqsRecordModel):
  body: Json[S3Model]
```

This would be compatible with the existing framework and could be used as just another model, e.g.:

```py
from aws_lambda_powertools.utilities.batch import BatchProcessor, EventType, batch_processor

processor = BatchProcessor(event_type=EventType.SQS, model=SqsS3EventNotificationModel)

def record_handler(record: SqsS3EventNotificationModel):
  for s3_record in record.body.Records:
    do_something_useful(s3_record)

@batch_processor(record_handler=record_handler, processor=processor)
def handler(event, context):
  return processor.response()
```

**Note**: see [this comment about Mypy compatibility](https://github.com/awslabs/aws-lambda-powertools-python/issues/1656#issuecomment-1409922359), although I believe it may no longer be a concern due to a [recent typing improvement](https://github.com/awslabs/aws-lambda-powertools-python/pull/1982/commits/f8d49e0ba6f79af40c680d33e5f102ffb62a9615). 

### Alternative solutions

If there's a way to skip the intermediate handler function, that would be even nicer (but see trade-off below).

In other words, rather than needing to define a `def record_handler(record: SqsS3EventNotificationModel)` that iterates through the `record.body.Records`, it would be nice to just directly define a `def record_handler(record: S3Model)` and let the processor unwrap everything instead.

This would keep the business logic clean, and not having to care whether it came via SQS or EventBridge, etc..

However, the trade-off to this even simpler interface would be that any event metadata (from both the SQS event _and_ the S3 event notification) would be unavailable.

### Acknowledgment

theipster:
**Issue number:** #2078

### Changes

Exposes a new data model for SQS-wrapped S3 event notifications.

### User experience

Please see #2078 description.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


theipster:
@leandrodamascena Thanks for [your guidance](https://github.com/awslabs/aws-lambda-powertools-python/issues/2078#issuecomment-1503407043). Do these latest changes look roughly sane to you? 🙂 

I've cheekily borrowed the `S3Model` assertions, seeing as the primary concern of `SqsS3EventNotificationModel` is more about chaining `SqsRecordModel` and `S3Model` together, rather than the properties themselves.

leandrodamascena:
> @leandrodamascena Thanks for [your guidance](https://github.com/awslabs/aws-lambda-powertools-python/issues/2078#issuecomment-1503407043). Do these latest changes look roughly sane to you? slightly_smiling_face
> 
> I've cheekily borrowed the `S3Model` assertions, seeing as the primary concern of `SqsS3EventNotificationModel` is more about chaining `SqsRecordModel` and `S3Model` together, rather than the properties themselves.

Hi @theipster! Surely these things make sense. I didn't get a chance to see why the CI is crashing, it looks like `mypy` complaining about types. But don't worry, I always suffer from `mypy` errors also :cry: .

I will look into this today and we can work together to complete it and release in the next Powertools version. So far, very good job :clap:!

leandrodamascena:
Hi @theipster! I'm trying to send commits to this PullRequest and I can't because it looks like you didn't allow the maintainers to do that.

There is an option in this pull request that must be checked to allow this. Can you enable it, please?
image

leandrodamascena:
@theipster hi!! When you have time, could you pls check my last comment?

Here's the patch I'm looking to apply to this PR to test things out: https://gist.github.com/leandrodamascena/69cad79e01fcff4697c3672f5db14f67

theipster:
@leandrodamascena Apologies for the late reply. I've now enabled this setting; thanks for your patch.

Is there any reason why mypy requires `Sequence[]` here, whereas it's fine with using `List[]` elsewhere, e.g. for [`S3Model`](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parser/models/s3.py#L114-L115)? Is there some kind of exception in place for `S3Model`?

leandrodamascena:
> @leandrodamascena Apologies for the late reply. I've now enabled this setting; thanks for your patch.
> 
> Is there any reason why mypy requires `Sequence[]` here, whereas it's fine with using `List[]` elsewhere, e.g. for [`S3Model`](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/aws_lambda_powertools/utilities/parser/models/s3.py?rgh-link-date=2023-04-25T02%3A03%3A16Z#L114-L115)? Is there some kind of exception in place for `S3Model`?

Hi @theipster no worries! Thanks for enabling this.

The problem is not in the `S3Model` and yes, the code works as expected. But to avoid issues with different data types, we use mypy, a library for static type checking. This helps us identify potential bugs during the development process and make the code more reliable. Because of this, the code seems to work but when mypy checks the types it encounters some issues.

The problem we're encountering is due to the `Records` field in the parent class, `SqsModel`, which is set as a `List`. Since `List` is an invariant data type, we're running into issues when trying to override this field with a str type `Records` field in the `SqsS3EventNotificationModel` class. As a result, mypy is complaining about this.

I changed the field type to `Sequence` in `SqsModel` and now the errors have stopped. `Sequence` is a more generic data type and `List` is a subclass of it.

I don't know if I made myself very clear about this, but I leave some references here:

https://blog.daftcode.pl/covariance-contravariance-and-invariance-the-ultimate-python-guide-8fabc0c24278

https://mypy.readthedocs.io/en/stable/common_issues.html#variance

tests/functional/parser/test_sqs_s3_event_notification.py
```diff
@@ -0,0 +1,43 @@
...
+    for record in sqs_event_dict["Records"]:
+        record["body"] = json_serialize(s3_event_notification_dict)
+
+    parsed_event: SqsS3EventNotificationModel = handle_sqs_json_body_containing_s3_notifications(
+        sqs_event_dict, LambdaContext()
+    )
+
+    assert len(parsed_event.Records) == 2
+    for parsed_sqs_record in parsed_event.Records:
+        assert_s3(parsed_sqs_record.body)
```
leandrodamascena: I would like an opinion here @rubenfonseca @heitorlessa. I think this test is ok, do you see another way to optimize this?

rubenfonseca: Looks ok to me

docs/utilities/parser.md
```diff
@@ -175,6 +175,7 @@ Parser comes with the following built-in models:
 | **LambdaFunctionUrlModel**              | Lambda Event Source payload for Lambda Function URL payload                  |
 | **KafkaSelfManagedEventModel**          | Lambda Event Source payload for self managed Kafka payload                   |
 | **KafkaMskEventModel**                  | Lambda Event Source payload for AWS MSK payload                              |
+| **SqsS3EventNotificationModel**         | Lambda Event Source payload for SQS-wrapped S3 event notifications           |
```
heitorlessa:
```suggestion
| **SqsS3EventNotificationModel**         | Lambda Event Source payload for S3 event notifications wrapped in SQS event (S3->SQS)           |
```

suggestion to make it easier to parse at a glance (direction of travel)

tests/functional/parser/test_sqs_s3_event_notification.py
heitorlessa: This is a good opportunity to move to the new unit tests since all we should care here is whether the new Model can be instantiated correctly (along with its ramifications, JSON deserialization etc.).

Following this, all you'd need to do here is:

* Create a new S3->SQS JSON event file under `tests/events` 
* Create a new test at `tests/unit/parser/test_s3.py`
* Test whether the model can be initialized without any problem (we already test both SQS and S3 Models)
* Test whether an invalid JSON in the S3 data will lead to a model validation failure

Previously, we were shoving everything under `tests/functional` where we have a lot of boilerplate like `@event_parser` and these functions, when in fact we're merely testing the Model since we already have separate tests for the `@event_parser`... which in this case is a proper `functional` test.

Hope that makes sense ;)

leandrodamascena: It makes perfect sense @heitorlessa! I changed the files as per this feedback

aws_lambda_powertools/utilities/parser/models/__init__.py
```diff
@@ -50,6 +50,10 @@
     S3Model,
     S3RecordModel,
 )
+from .s3_event_notification import (
+    S3SqsEventNotificationModel,
+    S3SqsEventNotificationRecordModel,
```
leandrodamascena, line 55: Hey @theipster, just wanted to give you a heads-up that I changed the class names to make them clearer. I used the prefix `S3Sqs` instead of `SqsS3` since the event comes from S3 to SQS.

docs/utilities/parser.md
```diff
@@ -156,25 +156,26 @@ def my_function():
...
+| **KinesisFirehoseModel**                | Lambda Event Source payload for Amazon Kinesis Firehose                               |
+| **LambdaFunctionUrlModel**              | Lambda Event Source payload for Lambda Function URL payload                           |
+| **S3EventNotificationEventBridgeModel** | Lambda Event Source payload for Amazon S3 Event Notification to EventBridge.          |
+| **S3Model**                             | Lambda Event Source payload for Amazon S3                                             |
+| **S3ObjectLambdaEvent**                 | Lambda Event Source payload for Amazon S3 Object Lambda                               |
+| **S3SqsEventNotificationModel**         | Lambda Event Source payload for S3 event notifications wrapped in SQS event (S3->SQS) |
+| **SesModel**                            | Lambda Event Source payload for Amazon Simple Email Service                           |
+| **SnsModel**                            | Lambda Event Source payload for Amazon Simple Notification Service                    |
+| **SqsModel**                            | Lambda Event Source payload for Amazon SQS                                            |
```
leandrodamascena, line 178: I took the opportunity to organize the names of the models in alphabetical order.

leandrodamascena:
Hi @theipster, congrats on this amazing job, you've done most of the work.
I just made some small changes to bring closer the developer experience we have and the project architecture.

Thank you very much for your time dedicated to this PR.  Do you have any considerations before we make the last review and merge this PR?

theipster:
@leandrodamascena Thank you for your support in making this go much more smoothly!

Nothing to add from my side, LGTM ✅

tests/events/sqsS3Event.json
rubenfonseca, line 1: Should we rename this file to `s3SqsEvent.json` for consistency?

tests/unit/parser/test_s3.py
```diff
@@ -105,3 +110,33 @@ def test_s3_eventbridge_notification_object_restore_completed_event():
     assert model.detail.requester == raw_event["detail"]["requester"]
     assert model.detail.restore_expiry_time == raw_event["detail"]["restore-expiry-time"]
     assert model.detail.source_storage_class == raw_event["detail"]["source-storage-class"]
+
+
+def test_s3_sqs_event_notification():
+    raw_event = load_event("sqsS3Event.json")
```
rubenfonseca: Change this line too

tests/unit/parser/test_s3.py
```diff
@@ -105,3 +110,33 @@ def test_s3_eventbridge_notification_object_restore_completed_event():
...
+    assert model.Records[0].body.Records[0].eventVersion == body["Records"][0]["eventVersion"]
+    assert model.Records[0].body.Records[0].eventSource == body["Records"][0]["eventSource"]
+    assert model.Records[0].body.Records[0].eventTime == datetime.fromisoformat(
+        body["Records"][0]["eventTime"].replace("Z", "+00:00")
+    )
+    assert model.Records[0].body.Records[0].eventName == body["Records"][0]["eventName"]
+
+
+def test_s3_sqs_event_notification_body_invalid_json():
+    raw_event = load_event("s3Event.json")
```
rubenfonseca: Change this line too

rubenfonseca (changes requested):

tests/unit/parser/test_s3.py
leandrodamascena: This test was wrong. I was using a S3 JSON file to simulate an invalid body, but it is supposed to use an invalid json body (like a string).

Fixed.

tests/unit/parser/test_s3.py
leandrodamascena: Fixed

tests/events/sqsS3Event.json
leandrodamascena: Fixed

tests/unit/parser/test_s3.py
theipster: If you compare against my original code in commit 251d8ba, I used `sqsEvent.json`, which is another way to solve this problem.

This is also why I originally named the classes `SqsS3...` (because the SQS payload literally wraps the S3 payload) rather than `S3Sqs...` (the direction of traffic). They're simply different ways of conceptualising the same scenario, neither is particularly right or wrong I guess!

leandrodamascena: You're right @theipster, I used the wrong words to explain why I changed the test. It's not a question of being right or wrong, it's just different ways of solving the same question. 

Thank you very much for this feedback, it will help me to avoid mistakes like this in the future.

rubenfonseca (approved):
LGTM

PR #2122, opened 2023-04-13
chore(ci): add support for x86-64 regions only

### LINKED ISSUE #1950: Maintenance: Adding new AWS Regions to the Lambda Layers Deployment
### Summary

AWS recently launched new regions: 

Asia Pacific (Melbourne)
ap-southeast-4
Asia Pacific (Hyderabad)
ap-south-2
Europe (Spain)
eu-south-2
Europe (Zurich)
eu-central-2
Middle East (UAE)
me-central-1

[AWS Global Infrastructure Map](https://aws.amazon.com/about-aws/global-infrastructure/)

According to this [AWS blog post](https://aws.amazon.com/blogs/aws/aws-lambda-functions-powered-by-aws-graviton2-processor-run-your-functions-on-arm-and-get-up-to-34-better-price-performance/), graviton2 (arm64) is not yet available in Spain and Zurich.

### Why is this needed?

Deploy Lambda Layers to these new regions.

_No response_

### Solution

_No response_

### Acknowledgment

rubenfonseca:
**Issue number:** #1950

### Changes

This PR changes our Layer and Canary stack to support regions that only support x86_64 Lambdas. We couldn't find an easy and reliable way to
detect dinamically if a given region supports ARM64 Lambdas, so instead we maintain the map manually and indicate which regions don't support it.

This makes heavy use of CDK/CloudFormation Parameters and Conditions, so the same CloudFormation template can be used across all regions.

### User experience

After this change, we will generate Lambda Layers for regions that only support x86_64 Lambdas. We will need to update the documentation when the 
first Layer version is generated, and validate that the script to update the layer version in the docs works with the single-architecture regions.

### Regions that will be added by this PR

- ap-south-2
- ap-southeast-4
- eu-central-2
- eu-south-2
- me-central-1

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/layer/layer/layer_stack.py
+++ b/layer/layer/layer_stack.py
@@ -39,7 +39,7 @@ class Layer(Construct):
         powertools_version: str,
         architecture: Optional[Architecture] = None,
         **kwargs
-    ):
+    ) -> None:
         super().__init__(scope, construct_id, **kwargs)
 
         layer = LambdaPowertoolsLayer(
--- a/layer/layer/layer_stack.py
+++ b/layer/layer/layer_stack.py
@@ -1,39 +1,58 @@
-from aws_cdk import CfnOutput, RemovalPolicy, Stack
+from typing import Optional
+
+import jsii
+from aws_cdk import (
+    Aspects,
+    CfnCondition,
+    CfnOutput,
+    CfnParameter,
+    CfnResource,
+    Fn,
+    IAspect,
+    RemovalPolicy,
+    Stack,
+)
 from aws_cdk.aws_lambda import Architecture, CfnLayerVersionPermission
 from aws_cdk.aws_ssm import StringParameter
 from cdk_aws_lambda_powertools_layer import LambdaPowertoolsLayer
 from constructs import Construct
 
 
-class LayerStack(Stack):
+@jsii.implements(IAspect)
+class ApplyCondition:
+    def __init__(self, condition: CfnCondition):
+        self.condition = condition
+
+    def visit(self, node):
+        if isinstance(node, CfnResource):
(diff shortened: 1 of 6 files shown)


rubenfonseca:
@leandrodamascena unfortunately there's no easy to way to test this. So please check the code to see if it makes sense, and after we merge we can try a couple of workflows to see if any change is needed.

layer/layer/layer_stack.py
```diff
@@ -1,39 +1,58 @@
...
         scope: Construct,
         construct_id: str,
+        layer_version_name: str,
         powertools_version: str,
-        ssm_paramter_layer_arn: str,
-        ssm_parameter_layer_arm64_arn: str,
+        architecture: Optional[Architecture] = None,
         **kwargs
-    ) -> None:
+    ):
```
leandrodamascena: Just out of curiosity. Any reason to remove the `None` here?

leandrodamascena:
> @leandrodamascena unfortunately there's no easy to way to test this. So please check the code to see if it makes sense, and after we merge we can try a couple of workflows to see if any change is needed.

The code makes perfect sense to me, but like you said: there's no way to test it and just looking at the code I have no idea if we need to change anything else.
So, I follow you: let's merge this PR and check for possible changes. I'll approve.

leandrodamascena (approved):
Approved.

layer/layer/layer_stack.py
rubenfonseca: Sorry no, I'll re-add it to make it more explicit.

Issue #2156, opened 2023-04-21
Pydantic required when using BatchProcessor utiltiies


walmsles:
### Expected Behaviour

I have installed aws-lambda-powertools without any optional dependencies and am not using any parser features.   I expect to be able to use Batch Processor without any errors.

When including BatchProcessor to process SQS Records (without Models) batch/base.py imports "ValidationError" from parser which requires the "pydantic" dependency and breaks the optional constraint.

### Current Behaviour

ModuleNotFoundError: No module named 'pydantic'

### Code snippet

```python
import json

from aws_lambda_powertools import Logger, Tracer
from aws_lambda_powertools.utilities.batch import BatchProcessor, EventType, process_partial_response
from aws_lambda_powertools.utilities.parser.models import SqsRecordModel
from aws_lambda_powertools.utilities.typing import LambdaContext
from aws_lambda_powertools.utilities.parser import BaseModel
from aws_lambda_powertools.utilities.parser.types import Json


class Order(BaseModel):
    item: dict

class OrderSqsRecord(SqsRecordModel):
    body: Json[Order]  # deserialize order data from JSON string

processor = BatchProcessor(event_type=EventType.SQS, model=OrderSqsRecord)
tracer = Tracer()
logger = Logger()


@tracer.capture_method
def record_handler(record: OrderSqsRecord):
    return record.body.item

@logger.inject_lambda_context
@tracer.capture_lambda_handler
def lambda_handler(event, context: LambdaContext):
    return process_partial_response(event=event, record_handler=record_handler, processor=processor, context=context)
```

### Possible Solution

remove line 29.  Change Exception handling to use String Comparison instead to detect ValidationError so works whether Pydantic is available or not? 

e.g.
Instead of:
```python
        except ValidationError:
            return self._register_model_validation_error_record(record)
        except Exception:
            return self.failure_handler(record=data, exception=sys.exc_info())
```

Perhaps:
```python
        except Exception as exc:
            if exc.__class__.__name__ == "ValidationError":
                return self._register_model_validation_error_record(record)
            return self.failure_handler(record=data, exception=sys.exc_info())
```

### Steps to Reproduce

1. install aws-lambda-powertools with no optional dependecies
2. Take code snippet from BatchProcessor page for SQS
3. Try and run it using Python
4. GET Module IMport Error

### AWS Lambda Powertools for Python version

latest

### AWS Lambda function runtime

3.9

### Packaging format used

PyPi

### Debugging logs

[...]

PR #2157, opened 2023-04-21
fix(batch): resolve use of ValidationError in batch

walmsles:
**Issue number:** #2156 

## Summary
Remove ValidationError import to remove implicit pydantic dependency

### Changes

Removed use of "ValidationError" class in batch.py.  Replaced instead with slightly nastier string class name comparison to detect the ValidationError.  Intention of code remain in tact in that the ValidationError will be correctly handled when pydantic model validation fails.

### User experience

Before:
[ERROR] Runtime.ImportModuleError: Unable to import module 'app': No module named 'pydantic'

After:

## Checklist

* [x] Changes are documented

<details>
No not a breaking change - reversibng a breaking change

**RFC issue number**: Nil

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/batch/base.py
+++ b/aws_lambda_powertools/utilities/batch/base.py
@@ -26,7 +26,6 @@ from aws_lambda_powertools.utilities.data_classes.kinesis_stream_event import (
     KinesisStreamRecord,
 )
 from aws_lambda_powertools.utilities.data_classes.sqs_event import SQSRecord
-from aws_lambda_powertools.utilities.parser import ValidationError
 from aws_lambda_powertools.utilities.typing import LambdaContext
 
 logger = logging.getLogger(__name__)
@@ -496,9 +495,10 @@ class BatchProcessor(BasePartialBatchProcessor):  # Keep old name for compatibil
                 result = self.handler(record=data)
 
             return self.success_handler(record=record, result=result)
-        except ValidationError:
-            return self._register_model_validation_error_record(record)
-        except Exception:
+        except Exception as exc:
+            if exc.__class__.__name__ == "ValidationError":
+                return self._register_model_validation_error_record(record)
+
             return self.failure_handler(record=data, exception=sys.exc_info())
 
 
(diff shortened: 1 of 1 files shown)


heitorlessa:
cc @leandrodamascena @rubenfonseca

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -634,7 +639,8 @@ async def _async_process_record(self, record: dict) -> Union[SuccessResponse, Fa
                 result = await self.handler(record=data)
 
             return self.success_handler(record=record, result=result)
-        except ValidationError:
-            return self._register_model_validation_error_record(record)
-        except Exception:
+        except Exception as exc:
+            if exc.__class__.__name__ == "ValidationError":
+                return self._register_model_validation_error_record(record)
+
```
leandrodamascena: I wonder if the same solution should be applied here @heitorlessa

leandrodamascena (changes requested):

aws_lambda_powertools/utilities/batch/base.py
heitorlessa: definitely -> done: https://github.com/awslabs/aws-lambda-powertools-python/pull/2157/commits/78284f72371eb8cc5f4b9e55a708a8c298679039

leandrodamascena: Thanks!

heitorlessa:
Changed solution slightly so we have better assurances ahead of Pydantic V2 by depending on a public contract in the Exception -- if that exception name changes or gets subclassed it won't break it. To top it over, we triple check by verifying it came from the actual model a customer gave us.


---

```python
        except Exception as exc:
            # NOTE: Pydantic is an optional dependency, but when used and a poison pill scenario happens
            # we need to handle that exception differently.
            # We check for a public attr in validation errors coming from Pydantic exceptions (subclass or not)
            # and we compare if it's coming from the same model that trigger the exception in the first place
            model = getattr(exc, "model", None)
            if model == self.model:
                return self._register_model_validation_error_record(record)
```

aws_lambda_powertools/utilities/batch/base.py
walmsles: Yes there are 2 places

heitorlessa (approved):

leandrodamascena:
Approved!

PR #2161, opened 2023-04-24
feat(ci): dispatch GitHub analytics action

### LINKED ISSUE #2160: Maintenance: Add a GitHub analytics dispatcher
### Why is this needed?

In order to fetch repository analytics in a secure way, we need to vend the GitHub action token to a synchronous Lambda function that will crawl the GitHub API.

Other

### Solution

We need to create a GitHub action that will assume an OIDC role and call a dispatcher function with the GitHub token.

### Acknowledgment

- [X] Should this be considered in other Lambda Powertools languages? i.e. [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

rubenfonseca:
**Issue number:** #2160

### Changes

This PR adds a new GitHub action that runs every hour. The action assumes an OIDC role and calls a Lambda Function that dispatches all the crawling functions synchronously.

### User experience

After this change, we should see fresh analytics every hour.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


.github/workflows/reusable_dispatch_analytics.yml
```diff
@@ -3,10 +3,40 @@ name: Reusable dispatch analytics
 on:
   workflow_dispatch:
 
+  schedule:
```
heitorlessa: Perhaps split into a dedicated scheduled workflow so reusable remains a dispatch that can be called like other reusables?

.github/workflows/reusable_dispatch_analytics.yml
```diff
@@ -3,10 +3,40 @@ name: Reusable dispatch analytics
...
+  issues: read
+  discussions: read
+  packages: read
+  pages: read
+  pull-requests: read
+  repository-projects: read
+  security-events: read
+  statuses: read
+
 jobs:
```
heitorlessa: Can we run in an environment do secrets are encrypted from forks?

You never know if we ever accidentally make a mistake in the IAM OIDC and allow any repo. This will prevent this attack vector.

heitorlessa (changes requested):
Two tiny changes. fantastic work, Ruben! 

It's quite a genius move to do this and cut down all the infra required to exchange tokens and whatnot per repo

.github/workflows/reusable_dispatch_analytics.yml
rubenfonseca: I've moved this to a normal workflow, and removed the reusable part.

.github/workflows/reusable_dispatch_analytics.yml
rubenfonseca: Done, created a new "analytics" environment and moved the secrets there. I've removed the old secrets too.

rubenfonseca: Please note that the IAM OIDC would not be assumable since there's a condition to only be assumable by this repository too. In any case, it's better this way.

heitorlessa (approved):

dreamorosi:
Hey folks, do we need something like this in the TS repo?

rubenfonseca:
@dreamorosi yes Andrea, I will send instructions and open PR for the rest of the runtimes during the day!

PR #2171, opened 2023-04-27
feat(metrics): add flush_metrics() method to allow manual flushing of metrics

### LINKED ISSUE #2109: Feature request: Add A method to manually flush metrics
### Use case

When not running within a standard Lambda handler (in my case, running within the Lambda Web Adapter), the @log_metrics decorator does not function. It is expecting the standard Lambda event,context parameters to be passed in.

Would be helpful to have a function that you could manually call before returning from a function to flush the metrics out to stdout so that they get logged.

### Solution/User Experience

Based on the decorator code, I built the is as a function in my code that flushes the metrics, but would be more resilient if it were in the class.

`def flush_metrics(a=None):
    m = metrics.serialize_metric_set()
    metrics.clear_metrics()
    print(json.dumps(m, separators=(",", ":")))    
    return a`

Note the parameter (a) there is not used in the function at all, and is simply returned to the caller. This was done so that I could wrap the "flush" and return into one line such as:

`    return flush_metrics({"message": f"Count: {count}"})`

### Alternative solutions

_No response_

### Acknowledgment

rubenfonseca:
**Issue number:** #2109

### Changes

This PR adds a new `flush_metrics` to the metrics utility to manually flush the metrics.

carbon (1)

### User experience

Before this PR it was hard for users of the metrics utility to flush the metrics when not running on the Lambda runtime. This simple refactor enables the users
quick access to manually flushing the metrics.

image

## Checklist

- [x] Changes are documented
- [ ] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/examples/metrics/src/flush_metrics.py
+++ b/examples/metrics/src/flush_metrics.py
@@ -1,5 +1,3 @@
-import json
-
 from aws_lambda_powertools import Metrics
 from aws_lambda_powertools.metrics import MetricUnit
 from aws_lambda_powertools.utilities.typing import LambdaContext
@@ -9,6 +7,4 @@ metrics = Metrics()
 
 def lambda_handler(event: dict, context: LambdaContext):
     metrics.add_metric(name="SuccessfulBooking", unit=MetricUnit.Count, value=1)
-    your_metrics_object = metrics.serialize_metric_set()
-    metrics.clear_metrics()
-    print(json.dumps(your_metrics_object))
+    metrics.flush_metrics()
--- a/examples/metrics/src/flush_metrics.py
+++ b/examples/metrics/src/flush_metrics.py
@@ -5,6 +5,12 @@ from aws_lambda_powertools.utilities.typing import LambdaContext
 metrics = Metrics()
 
 
+def book_flight(flight_id: str, **kwargs): 
+    # logic to book flight
+    ...
+    metrics.add_metric(name="SuccessfulBooking", unit=MetricUnit.Count, value=1)
+
+
 def lambda_handler(event: dict, context: LambdaContext):
     try:
         metrics.add_metric(name="SuccessfulBooking", unit=MetricUnit.Count, value=1)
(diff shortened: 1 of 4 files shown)


examples/metrics/src/flush_metrics.py
rubenfonseca, line 1: This example file was not being used in the documentation, so I just changed the content and started using it.

leandrodamascena (commented):
Another excellent job Ruben!!

I'm wondering if we should do the same as the typescript and move the "Flushing metrics manually" topic to the "Flushing Metrics" menu. What do you think about?

Reference:
https://awslabs.github.io/aws-lambda-powertools-typescript/latest/core/metrics/#manually

image

rubenfonseca:
Peer-review feedback:

- remove single metric example
- expand the example with try/catch to handle uncaught exceptions when manually flushing metrics

examples/metrics/src/flush_metrics.py
```diff
@@ -6,5 +6,7 @@
 
 
 def lambda_handler(event: dict, context: LambdaContext):
```
heitorlessa, line 14:
```suggestion
def book_flight(flight_id: str, **kwargs): 
    # logic to book flight
    ...
    metrics.add_metric(name="SuccessfulBooking", unit=MetricUnit.Count, value=1)


def lambda_handler(event: dict, context: LambdaContext):
```

examples/metrics/src/flush_metrics.py
```diff
@@ -6,5 +6,7 @@
...
-    metrics.add_metric(name="SuccessfulBooking", unit=MetricUnit.Count, value=1)
-    metrics.flush_metrics()
+    try:
+        metrics.add_metric(name="SuccessfulBooking", unit=MetricUnit.Count, value=1)
```
heitorlessa:
```suggestion
        book_flight(flight_id=event.get("flight_id", ""))
```

heitorlessa (changes requested):
dummy logic to try/finally makes more sense

leandrodamascena (approved):
Approved :)
image

PR #2183, opened 2023-05-01
feat(logger): add DatadogLogFormatter and observability provider

### LINKED ISSUE #1433: Feature request: Add easy support for Datadog and possibly other observability solutions
### Use case

I have played around with the Datadog and the AWS Powertools Lambda layers and as one that needs to integrate with Datadog, the Datadog Lambda layer is a good choice for getting that integration set up flawlessly.

But I also really enjoy some of the features that AWS Powertools have incorporated into their Lambda layer and as a developer I find it to be a very useful tool. Last time I checked (a while ago), the AWS Powertools layer was more lightweight as well.

Sadly, because both layers use similar packages (boto3 for example among other things), I believe they are not exactly compatible with each other. And either way, adding both would make our Lambdas' start times much worse as both layers are not exactly light either.

I have one collection of Lambdas using the Datadog layer and another collection using Poewrtools. I have thus noticed some of the differences that make Powertools tricky to easily integrate with Datadog despite being the more useful tool purely from a developer's perspective:
* the log format is slightly different and it does not allow Datadog to easily query logs by things like request ID. It would be nice to be able to specify a Datadog log format that we can easily use within Powertools without having to create a custom one ourselves. I believe compatibility can be easily achieved here and all we need is a special logger format that mimics the one used by Datadog
* Datadog does not ingest AWS CloudWatch metrics. I believe the easiest way to do a custom metric that gets interpreted by Datadog is to send logs in this format `{"m": "Metric name", "v": "Metric value", "e": "Unix timestamp (seconds)", "t": "Array of tags"}`. I know this is different from the embedded metrics format which AWS already provides - https://aws.amazon.com/about-aws/whats-new/2019/11/amazon-cloudwatch-launches-embedded-metric-format but maybe we can have a way to support both based on some configuration setting? Sadly the metrics that Powertools provides at the moment are not easily ingestible into Datadog
* I have not tested it sufficiently, but if there is some incompatibility with the way Datadog handles traces, maybe someone else that has that knowledge can keep in mind that some adaptations would be useful there too? From what I can understand, there should be no problems there as traces are ingested straight from XRay but I have sadly under-utilised that functionality

### Solution/User Experience

Provide a configuration or an option to define the format when setting up loggers, metrics and traces, which would allow for better integration with other ovservability solutions other than AWS' CloudWatch and XRay

### Alternative solutions

It may be possible to do all these already. Maybe a separate package that adds this compatibility can be created that works well with Powertools, as long as Powertools already has easy methods to manipulate its behaviour in the required ways to allow for the requested integration.

### Acknowledgment

### LINKED ISSUE #2014: RFC: Support for external observability providers - Logging

https://github.com/awslabs/aws-lambda-powertools-python/issues/1433

Logger

### Summary

This RFC is the first of three that will detail how the Logger, Metrics, and Tracer utilities can be extended to support a bring-your-own-observability-provider format for customers who want to use platforms other than AWS CloudWatch.

This RFC will be specifically for the Logger. Currently, in the Logger for Python (and Typescript), customers can bring their own [Logging Formatter](https://awslabs.github.io/aws-lambda-powertools-python/2.10.0/core/logger/#bring-your-own-formatter) and have the option to provide a configuration option to define the output format when they're setting up their Logger in order to allow for better integration with observability solutions other than CloudWatch.

Since this is already quite powerful, we are seeking to improve the existing mechanism by offering out of the box support for Datadog and a few other providers (TBD) for the Logger Formatter.

Metrics RFC: https://github.com/awslabs/aws-lambda-powertools-python/issues/2015
Tracer RFC: https://github.com/awslabs/aws-lambda-powertools-python/issues/2030

### Use case

The primary use case for this utility would be for developers who want a simplified way to integrate the Logger Formatter with the most popular third party observability providers (ie. DataDog).

### Proposal

**Current Logger Formatter experience**

Below is a simple snippet of how the Logger Formatter currently works:

```python

class CustomFormatter(LambdaPowertoolsFormatter):
    def serialize(self, log: dict) -> str:
        """Serialize final structured log dict to JSON str"""
        log["event"] = log.pop("message")  # rename message key to event
        return self.json_serializer(log)  # use configured json serializer
        logger = Logger(service="payment", logger_formatter=CustomFormatter())logger.info("hello")

logger = Logger(service="payment", logger_formatter=CustomFormatter())
logger.info("hello")
```

> Output

```json
{
    "level": "INFO",
    "location": "hello_name:8",
    "event": "Request from Alice received",
    "timestamp": "2023-03-21 13:25:36,129+0100",
    "service": "service_undefined"
}
```

**Logger Formatter proposal**

For this new utility, we propose a new parameter to the existing logger that developers can use to specify which observability provider they would like to receive their logs. The below code snippet is a rudimentary look at how this utility can be used and how it will function. Out of the box, we will support DataDog and a few other providers (TBD).

```python
from aws_lambda_powertools.logging.providers import DatadogLogFormatter
    
logger = Logger(service="payment", logger_formatter=DatadogLogFormatter())
logger.info("hello")
```
> Output
```json
{
     "datetime": "2021-11-22 15:32:02,145", 
     "level": "INFO", 
     "message": "Request from Alice received",
     "requestId": "9b1deb4d-3b7d-4bad"
}
```

**Custom logger usage**

If the customer would like to use another observability provider, or define their own logger functions, we will define an interface that the customer can implement and pass in to the Logger class.

### Out of scope

Sending logs from Powertools to the customer's desired observability platform will not be in the scope of this project. The implementation should only support modifying the output of the Logger so that the customer can push them to their platform of choice using the [Lambda Handler](https://awslabs.github.io/aws-lambda-powertools-python/2.1.0/core/logger/#bring-your-own-handler).

### Potential challenges

We need to determine which platforms we want to support out-of-the-box (apart from Datadog).

We will have to integrate with (and thus, have a dependency on) Datadog and any other platforms we decide to support out-of-the-box.

### Acknowledgment

heitorlessa:
**Issue number:** #2014

## Summary

This PR add a new `DatadogLogFormatter` and documents the idea of Observability Provider.

### Changes

### User experience

image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/logging/formatters/datadog.py
+++ b/aws_lambda_powertools/logging/formatters/datadog.py
@@ -65,13 +65,13 @@ class DatadogLogFormatter(LambdaPowertoolsFormatter):
             Key-value to persist in all log messages
         """
         super().__init__(
-            json_serializer,
-            json_deserializer,
-            json_default,
-            datefmt,
-            use_datetime_directive,
-            log_record_order,
-            utc,
-            use_rfc3339,
+            json_serializer=json_serializer,
+            json_deserializer=json_deserializer,
+            json_default=json_default,
+            datefmt=datefmt,
+            use_datetime_directive=use_datetime_directive,
+            log_record_order=log_record_order,
+            utc=utc,
+            use_rfc3339=use_rfc3339,
             **kwargs,
         )
--- a/aws_lambda_powertools/logging/formatters/datadog.py
+++ b/aws_lambda_powertools/logging/formatters/datadog.py
@@ -15,9 +15,55 @@ class DatadogLogFormatter(LambdaPowertoolsFormatter):
         use_datetime_directive: bool = False,
         log_record_order: list[str] | None = None,
         utc: bool = False,
-        use_rfc3339: bool = True,
+        use_rfc3339: bool = True,  # NOTE: The only change from our base formatter
         **kwargs,
     ):
+        """Datadog formatter to comply with Datadog log parsing
+
+        Changes compared to the default Logger Formatter:
+
+        - timestamp format to use RFC3339 e.g., "2023-05-01T15:34:26.841+0200"
+
+
+        Parameters
+        ----------
+        log_record_order : list[str] | None, optional
+            _description_, by default None
+
+        Parameters
+        ----------
+        json_serializer : Callable, optional
+            function to serialize `obj` to a JSON formatted `str`, by default json.dumps
+        json_deserializer : Callable, optional
+            function to deserialize `str`, `bytes`, bytearray` containing a JSON document to a Python `obj`,
+            by default json.loads
+        json_default : Callable, optional
+            function to coerce unserializable values, by default str
+
+            Only used when no custom JSON encoder is set
+
+        datefmt : str, optional
+            String directives (strftime) to format log timestamp.
+
+            See https://docs.python.org/3/library/time.html#time.strftime or
+        use_datetime_directive: str, optional
+            Interpret `datefmt` as a format string for `datetime.datetime.strftime`, rather than
+            `time.strftime` - Only useful when used alongside `datefmt`.
+
+            See https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior . This
+            also supports a custom %F directive for milliseconds.
+
+        log_record_order : list, optional
+            set order of log keys when logging, by default ["level", "location", "message", "timestamp"]
+
+        utc : bool, optional
+            set logging timestamp to UTC, by default False to continue to use local time as per stdlib
+        use_rfc3339: bool, optional
+            Whether to use a popular dateformat that complies with both RFC3339 and ISO8601.
+            e.g., 2022-10-27T16:27:43.738+02:00.
+        kwargs
+            Key-value to persist in all log messages
+        """
         super().__init__(
             json_serializer,
             json_deserializer,
--- a/aws_lambda_powertools/logging/formatters/datadog.py
+++ b/aws_lambda_powertools/logging/formatters/datadog.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+from typing import Any, Callable
+
+from aws_lambda_powertools.logging.formatter import LambdaPowertoolsFormatter
+
+
+class DatadogLogFormatter(LambdaPowertoolsFormatter):
+    def __init__(
+        self,
+        json_serializer: Callable[[dict], str] | None = None,
+        json_deserializer: Callable[[dict | str | bool | int | float], str] | None = None,
+        json_default: Callable[[Any], Any] | None = None,
+        datefmt: str | None = None,
+        use_datetime_directive: bool = False,
+        log_record_order: list[str] | None = None,
+        utc: bool = False,
+        use_rfc3339: bool = True,
+        **kwargs,
+    ):
+        super().__init__(
+            json_serializer,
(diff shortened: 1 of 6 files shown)


aws_lambda_powertools/logging/formatters/datadog.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(
+        self,
+        json_serializer: Callable[[dict], str] | None = None,
+        json_deserializer: Callable[[dict | str | bool | int | float], str] | None = None,
+        json_default: Callable[[Any], Any] | None = None,
+        datefmt: str | None = None,
+        use_datetime_directive: bool = False,
+        log_record_order: list[str] | None = None,
+        utc: bool = False,
+        use_rfc3339: bool = True,
```
rubenfonseca: Not change required, just a question: the only difference between this constructor and the base LambdaPowertoolsFormatter is this `use_rfc3399` boolean that defaults to `true`, right?

heitorlessa:
Yeees! It’s rather silly isn’t it ;-)

aws_lambda_powertools/logging/formatters/datadog.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import annotations
+
+from typing import Any, Callable
+
+from aws_lambda_powertools.logging.formatter import LambdaPowertoolsFormatter
+
+
+class DatadogLogFormatter(LambdaPowertoolsFormatter):
+    def __init__(
```
rubenfonseca, line 9: Can we add some docstring explaining why this exist and what the small difference is to the base class?

aws_lambda_powertools/logging/formatters/datadog.py
rubenfonseca:
```suggestion
        use_rfc3339: bool = True, // This is the only change from the base class
```

or something similar :) it helps future readers to spot the difference

rubenfonseca (changes requested):
just some small changes, good stuff!

aws_lambda_powertools/logging/formatters/datadog.py
heitorlessa: you're the best! Doing that now!

heitorlessa: Done. Added a specific list so we can easily append if future changes are required - lemme know if that addresses it.

rubenfonseca (approved):
Looks great!

heitorlessa:
we shall wait until GitHub or PyPi behaves well again ;)

aws_lambda_powertools/logging/formatters/datadog.py
jplock: Would there be any downsides to making `use_rfc3339` default to true if it were communicated well enough in release notes or the next major version?

hscheib:
Figure I would give a community member thumbs up. I copied the `DatadogLogFormatter` into my lambda to test and logs started showing in Datadog as expected. 🚀  Thanks for the change!

rubenfonseca (approved):

heitorlessa:
> Figure I would give a community member thumbs up. I copied the `DatadogLogFormatter` into my lambda to test and logs started showing in Datadog as expected. 🚀 Thanks for the change!

Wonderful to hear, please keep up shouting out more :)

PR #2194, opened 2023-05-02
feat(metrics): support to bring your own metrics provider

### LINKED ISSUE #1433: Feature request: Add easy support for Datadog and possibly other observability solutions

### LINKED ISSUE #2015: RFC: Support for external observability providers - Metrics

https://github.com/awslabs/aws-lambda-powertools-python/issues/1433
https://github.com/awslabs/aws-lambda-powertools-python/issues/2014

Metrics

#### Problem

Customers has been using Lambda Powertools alongside with third party observability provider for Logging, metrics and tracing. Focusing on metrics part (this RFC is a part of support for observability provider. For Logging please check out [this RFC](https://github.com/awslabs/aws-lambda-powertools-python/issues/2014)), while Powertools for AWS Lambda (Powertools) provided powerful and easy to use metrics class. However,  Powertools only supports [AWS CloudWatch Embedded Metric Format](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html) (EMF), which is unfriendly to third party observability provider, bringing hardship to our customers when trying to ingest this format into other observability solutions, such as DataDog.

#### Goal

Goal for this RFC is to enable third party observability provider like Datadog, NewRelic, Lumigo, etc. to create their own Powertools specific metric provider and offer to their own customers. e.g., [DataDog metric format.](https://github.com/DataDog/datadog-lambda-python/blob/main/datadog_lambda/metric.py#L77) And if customer eventually want to add additional value or add features themselves. They can easily do so by overriding a provider too. 

### Use case

Typical use case of this is utility would be for customers who use Lambda Powertools to collect metrics and want to have them ingested into a third party observability provider. 

#### Current metric experience 

The current Powertools’ metrics utility creates custom metrics asynchronously by logging metrics to standard output following Amazon CloudWatch Embedded Metric Format (EMF), and these metrics can then be seen in the CloudWatch console. This utility can aggregate up to 100 metrics using a single CloudWatch EMF object (which is in JSON format).

```python


# Current experience in using metrics should not change, i.e. these metrics would still output to CW
metrics = Metrics()

@metrics.log_metrics # ensures metrics are flushed upon request completion/failure

# JSON output
#     "_aws": {
#         "Timestamp": 1656685750622,
#         "CloudWatchMetrics": [
#             {
#                 "Namespace": "ServerlessAirline",
#                 "Dimensions": [
#                     [
#                         "environment",
#                         "service"
#                     ]
#                 ],
#                 "Metrics": [
#                     {
#                         "Name": "TurbineReads",
#                         "Unit": "Count"
#                     }
#                 ]
#             }
#         ]
#     },
#     "environment": "dev",
#     "service": "booking",
#     "TurbineReads": [
#         1.0,
#         8.0
#     ]
# }
```

#### Metric provider proposal

For this new utility, we propose a new metrics class for observability providers. With an optional parameter `provider` to allow developers to pass in observability provider pre-defined or user custom provider. And the output will be in observability provider friendly format. The below code snippet is a rudimentary look at how this utility can be used and how it will function.

The default use case for metrics before is `metrics=Metrics()`
After we have this provider feature, Customers can still use original CloudWatch Metrics by `metrics=Metrics()` or `metrics=CloudWatchEMF()`. They can also use provider for third party provider by e.g.:`Metrics=DataDogMetrics()`

```python
from aws_lambda_powertools.metrics.provider import DatadogMetrics
from aws_lambda_powertools.metrics.provider import DatadogMetricsProvider

# Use datadog-defined metrics provider
provider = DatadogMetricsProvider()
metrics = DatadogMetrics(provider=provider)

@metrics.log_metrics
    metrics.add_metric(name="SuccessfulBooking", value=1, tags={"product":"ticket","ordier":"online"})
    
# JSON output to log
# {
#     "m": "SuccessfulBooking", 
#     "v": 1,
#     "e": 1678509106, 
#     "t": "['product:ticket', 'order:online']"
# }
```

#### self-defined metrics provider usage

If the customer would like to use another observability provider, or define their own metrics functions, we will define an interface that the customer can implement and pass in to the Metrics class provider parameter

```python
from aws_lambda_powertools.metrics.provider import MetricsBase, MetricsProviderBase


class DataDogProvider(MetricsProviderBase):
    def __init__(self, namespace):

    def add_metric(self, name: str, value: float, timestamp: Optional[int] = None, tags: Optional[List] = None):
        self.metrics.append({"m": name, "v": int(value), "e": timestamp, "t": tags})

    def serialize(self) -> Dict:
        # logic here is to add dimension and metadata to each metric's tag with "key:value" format
        extra_tags: List = []
        output_list: List = []

        for single_metric in self.metrics:
            output_list.append(
                {
                    "m": f"{self.namespace}.{single_metric['m']}",
                    "v": single_metric["v"],
                    "e": single_metric["e"],
                    "t": single_metric["t"] + extra_tags,
                }
            )
        return {"List": output_list}

    def flush(self, metrics):
        # flush

    def clear(self):


class DataDogMetrics(MetricsBase):
    """Class for datadog metrics standalone class.

    Example
    -------
    dd_provider = DataDogProvider(namespace="default")
    metrics = DataDogMetrics(provider=dd_provider)

    @metrics.log_metrics(capture_cold_start_metric: bool = True, raise_on_empty_metrics: bool = False)
    def lambda_handler(event, context)
        metrics.add_metric(name="item_sold",value=1,tags)
    """

    # `log_metrics` and `_add_cold_start_metric` are directly inherited from `MetricsBase`
    def __init__(self, provider):
        self.provider = provider
        super().__init__()

    def add_metric(self, name: str, value: float, timestamp: Optional[int] = None, tags: Optional[List] = None):
        self.provider.add_metric(name=name, value=value, timestamp=timestamp, tags=tags)

    def flush_metrics(self, raise_on_empty_metrics: bool = False) -> None:
        metrics = self.provider.serialize()
        if not metrics and raise_on_empty_metrics:
            warnings.warn(
                "No application metrics to publish. The cold-start metric may be published if enabled. "
                "If application metrics should never be empty, consider using 'raise_on_empty_metrics'",
                stacklevel=2,
            )
        self.provider.flush(metrics)
        self.provider.clear()
         
    
metrics = DataDogMetrics(provider=DataDogProvider())

@metrics.log_metrics
    metrics.add_metric(name="SuccessfulBooking", value=1, tags={"product":"ticket","ordier":"online"})
    # call functions in provider
    metrics.provider.add_tag({"test":True})       

# JSON output to log
# {
#     "m": "SuccessfulBooking", 
#     "v": 1,
#     "e": 1678509106, 
#     "t": "['product:ticket', 'order:online']"
# }
```

### Out of scope

Introduction of third party observability provider dependency to submit metrics to API is out of scope. The configuration included in this project should only support modifying the metrics outputted to log. But on their end, customer can still submit to API if they implement observability provider SDK in metrics' flush function 

#### How to provide more value to customer with current Metrics provider design
- Currently we just provided `log_metrics` decorator and `_add_cold_start_metric` function in the base metrics provider function. 
- Other than these two are just basic framework of metrics class.
- What else should we include in the metrics class to provide more value to our customer?

#### Dependencies

No additional dependencies required

#### Changes and Additions to Public Interfaces

* Modification in `Metrics`, `MetricsManager` Class
    * Add an optional `provider` parameter in metrics class.
    * Refactor the Metrics Class to enable use of provider. But all currently exist function will have same output.
    * Provide a AWS CloudWatch EMF Metrics provider as default metrics provider.

#### Performance Impact

Little or no performance impact expected

#### Backwards Compatibility and Upgrade Path

Full backward compatibility, old metrics codes without `provider` parameter will get exact same output as before.

### Acknowledgment

roger-zhangg:
**Issue number:** #2015

### Changes

This PR refactors Metrics class to use Metrics Provider. Enabling customer to bring their own provider for outputting custom metrics format.The default CloudWatch EMF logic is refactored into a EMF provider. 

### User experience

- For Customer using the default EMF metric format, this PR won't change any UX
- For Customer willing to implement custom metrics format in Powertools Metrics
  - Before:
    - The Metrics class only supports EMF format.
  - After:
    - User can bring their own metrics provider to support custom metrics format 

#### Metric provider UX

The default use case for metrics before is `metrics=Metrics()`
After we have this provider feature, Customers can still use original CloudWatch Metrics by `metrics=Metrics()` or `metrics=CloudWatchEMF()`. They can also use provider for third party provider by e.g.:`Metrics=DataDogMetrics()`

```python
from aws_lambda_powertools.metrics.provider import DatadogMetrics
from aws_lambda_powertools.metrics.provider import DatadogMetricsProvider

# Use datadog-defined metrics provider
provider = DatadogMetricsProvider()
metrics = DatadogMetrics(provider=provider)

@metrics.log_metrics(capture_cold_start_metric=True, raise_on_empty_metrics=False)
    metrics.add_metric(name="SuccessfulBooking", value=1, tags={"product":"ticket","ordier":"online"})
    
# JSON output to log
# {
#     "m": "SuccessfulBooking", 
#     "v": 1,
#     "e": 1678509106, 
#     "t": "['product:ticket', 'order:online']"
# }
```

#### self-defined metrics provider UX

If the customer would like to use another observability provider, or define their own metrics functions, the provider acts as an interface that the customer can implement and pass to the Metrics class using the `provider` parameter

```python
from aws_lambda_powertools.metrics.provider import MetricsBase, MetricsProviderBase


class DataDogProvider(MetricsProviderBase):
    def __init__(self, namespace):

    def add_metric(self, name: str, value: float, timestamp: Optional[int] = None, tags: Optional[List] = None):
        self.metrics.append({"m": name, "v": int(value), "e": timestamp, "t": tags})

    def serialize(self) -> Dict:
        # logic here is to add dimension and metadata to each metric's tag with "key:value" format
        extra_tags: List = []
        output_list: List = []

        for single_metric in self.metrics:
            output_list.append(
                {
                    "m": f"{self.namespace}.{single_metric['m']}",
                    "v": single_metric["v"],
                    "e": single_metric["e"],
                    "t": single_metric["t"] + extra_tags,
                }
            )
        return {"List": output_list}

    def flush(self, metrics):
        # flush

    def clear(self):


class DataDogMetrics(MetricsBase):
    """Class for datadog metrics standalone class.

    Example
    -------
    dd_provider = DataDogProvider(namespace="default")
    metrics = DataDogMetrics(provider=dd_provider)

    @metrics.log_metrics(capture_cold_start_metric: bool = True, raise_on_empty_metrics: bool = False)
    def lambda_handler(event, context)
        metrics.add_metric(name="item_sold",value=1,tags)
    """

    # `log_metrics` and `_add_cold_start_metric` are directly inherited from `MetricsBase`
    def __init__(self, provider):
        self.provider = provider
        super().__init__()

    def add_metric(self, name: str, value: float, timestamp: Optional[int] = None, tags: Optional[List] = None):
        self.provider.add_metric(name=name, value=value, timestamp=timestamp, tags=tags)

    def flush_metrics(self, raise_on_empty_metrics: bool = False) -> None:
        metrics = self.provider.serialize()
        if not metrics and raise_on_empty_metrics:
            warnings.warn(
                "No application metrics to publish. The cold-start metric may be published if enabled. "
                "If application metrics should never be empty, consider using 'raise_on_empty_metrics'",
                stacklevel=2,
            )
        self.provider.flush(metrics)
        self.provider.clear()
         
    
metrics = DataDogMetrics(provider=DataDogProvider())

@metrics.log_metrics(capture_cold_start_metric: bool = True, raise_on_empty_metrics: bool = False)
    metrics.add_metric(name="SuccessfulBooking", value=1, tags={"product":"ticket","ordier":"online"})
    # call functions in provider
    metrics.provider.add_tag({"test":True})       

# JSON output to log
# {
#     "m": "SuccessfulBooking", 
#     "v": 1,
#     "e": 1678509106, 
#     "t": "['product:ticket', 'order:online']"
# }
```

### Tasks
- [ ] Create documentation
- [x] Check the change that can generate breaking change
- [x] Use the [simplified version](https://peps.python.org/pep-0604/#simplified-syntax) of typing from Python 3.10
- [x] Create an external provider like DataDog to validate our logic

### Not covered in this PR (Added - 12/07/2023)
- [x] Method to create standard operational metrics

## Checklist

* [x] Changes are documented -> Document after changes are reviewed and finalized

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/metrics/__init__.py
```diff
@@ -20,4 +21,6 @@
     "MetricResolutionError",
     "SchemaValidationError",
     "MetricValueError",
+    "EMFProvider",
```
leandrodamascena: I'm wondering if `CloudWatchProvider` could be a better name for this. In the end we will always use the name of the provider and not the output/parser format. Examples:
`DataDogProvider`
`OtherProvider`

it makes sense?

aws_lambda_powertools/metrics/base.py
```diff
@@ -1,77 +1,34 @@
...
     """Base class for metric functionality (namespace, metric, dimension, serialization)
 
-    MetricManager creates metrics asynchronously thanks to CloudWatch Embedded Metric Format (EMF).
-    CloudWatch EMF can create up to 100 metrics per EMF object
-    and metrics, dimensions, and namespace created via MetricManager
-    will adhere to the schema, will be serialized and validated against EMF Schema.
+    provider refactor: MetricManager implements general Metrics logic, AWS EMF logic are moved to provider.EMFProvider
+
+    MetricManager is now refactored to use provider. Default provider is AWS EMF provider
+    Bring/Implement your own metrics provider to support metrics format other than EMF
```
leandrodamascena: We might need to improve this docstring. 

```suggestion
    MetricManager implements general metrics logic, with providers extending from this class to add specific functionality. Developers can easily implement their own metrics provider by implementing the provider interface. 
    The default provider for the MetricManager is AWS CloudWatch. However, developers can easily bring their own metrics provider to support metrics formats other than EMF.
```

aws_lambda_powertools/metrics/base.py
```diff
@@ -412,73 +331,19 @@ def decorate(event, context):
...
             finally:
-                self.flush_metrics(raise_on_empty_metrics=raise_on_empty_metrics)
+                if not raise_on_empty_metrics and not self.metric_set:
+                    warnings.warn(
+                        "No application metrics to publish. The cold-start metric may be published if enabled. "
+                        "If application metrics should never be empty, consider using 'raise_on_empty_metrics'",
+                        stacklevel=2,
+                    )
+                else:
+                    self.flush_metrics(raise_on_empty_metrics=raise_on_empty_metrics)
```
leandrodamascena:
```suggestion
                self.flush_metrics(raise_on_empty_metrics=raise_on_empty_metrics)
```

This `if block` is already implemented in `flush_metrics` method.

aws_lambda_powertools/metrics/provider.py
leandrodamascena, line 1: When we merge this PR we will have the possibility of having several other providers extending the Powertools metrics utility, so I think it's better to create a folder called "providers" and use the following structure:

`providers/base.py` - MetricsProvider class
`providers/cloudwatch.py` - CloudWatchProvider class
`providers/__init__.py` - exports classes

With this structure, we simplify the experience of those who are going to create a new provider, as it make easy the identification of what that file does.

aws_lambda_powertools/metrics/types.py
```diff
@@ -1,7 +1,51 @@
...
+    KilobytesPerSecond = "Kilobytes/Second"
+    MegabytesPerSecond = "Megabytes/Second"
+    GigabytesPerSecond = "Gigabytes/Second"
+    TerabytesPerSecond = "Terabytes/Second"
+    BitsPerSecond = "Bits/Second"
+    KilobitsPerSecond = "Kilobits/Second"
+    MegabitsPerSecond = "Megabits/Second"
+    GigabitsPerSecond = "Gigabits/Second"
+    TerabitsPerSecond = "Terabits/Second"
+    CountPerSecond = "Count/Second"
```
leandrodamascena: Consider adding [Literal](https://peps.python.org/pep-0586/) to annotate this.

Example:
Seconds = Literal["Seconds"]

tests/functional/test_metrics.py
```diff
@@ -441,6 +441,8 @@ def test_schema_validation_no_namespace(metric, dimension):
     with pytest.raises(SchemaValidationError, match="Must contain a metric namespace."):
         with single_metric(**metric) as my_metric:
             my_metric.add_dimension(**dimension)
+            # Need to discuss if this behavior change is a breaking change
+            my_metric.provider.Validate_Metrics = True
```
leandrodamascena: I didn't test it yet, but yes, I think this is a breaking change. I'll check it in more detail tomorrow.

leandrodamascena (changes requested):
Hi @roger-zhangg, excellent work this PR! This is a feature that we really want to have because it is a need for some customers. I made some initial considerations that can help us to make the code simpler and optimize some things.
I still have more to suggest, but I need to look at the code in more detail.

I'll leave some notes to remember what we still have to do to get this ready:

- [ ] Create documentation
- [ ] Check the change that can generate breaking change
- [ ] Use the [simplified version](https://peps.python.org/pep-0604/#simplified-syntax) of typing from Python 3.10

Thank you very much for this PR, let's work together to merge as soon as possible.

aws_lambda_powertools/metrics/provider.py
roger-zhangg: Yeah that make sense

aws_lambda_powertools/metrics/types.py
roger-zhangg: Not sure if I'm getting this right, but Literal will raise error on json.dump -> `TypeError: Object of type _LiteralGenericAlias is not JSON serializable`

aws_lambda_powertools/metrics/__init__.py
leandrodamascena: Thanks for changing the name of the class @roger-zhangg! I realized that `CloudWatchEMFProvider` could be an even better name, as we can add a provider called `CloudWatchProvider` that sends metrics through the SDK, for example.

aws_lambda_powertools/metrics/base.py
leandrodamascena: Thank you for changing this.

aws_lambda_powertools/metrics/provider.py
leandrodamascena: Thanks for refactoring this.

aws_lambda_powertools/metrics/provider/base.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    # MAX_METRICS controls how many metrics can be collected before auto-flushing
+    MAX_METRICS = 100
+    # MAX_DIMENSIONS controls how many dimension could be added
+    MAX_DIMENSIONS = 29
+    # use single metric when recording coldstart
+    Enable_Single_Metrics = True
+    # enable validation before sending out metrics. When enabled, validation will raise error if failed.
+    Validate_Metrics = False
```
leandrodamascena: Roger, according to this implementation we have to replicate these constants within each provider class we create, correct? I think this implementation makes sense, but I was wondering if there is any case where these values could be dynamic and passed by the developer in the instance of the provider class. 

I say this because when we talk about CloudWatch EMF we have control over it, we know that we can flush every 100 metrics, for example and there is no overhead, but I'm wondering if there is any provider where the developer will want to flush metrics dynamically due to tcp packet size, some batch process that needs to have real-time metrics or other situations.

I'm just guessing about it and would love to hear from you what you think or if you've found something like this in your searches for other providers.

aws_lambda_powertools/metrics/provider/cloudwatchemf_provider.py
```diff
@@ -0,0 +1,198 @@
...
+
+    """
+
+    def __init__(self):
+        self._metric_unit_valid_options = list(MetricUnit.__members__)
+        self._metric_units = [unit.value for unit in MetricUnit]
+        return super().__init__()
+
+    # generic add metrics function
+    def add_metric(self, metrics, name, unit, value, resolution) -> Dict:
```
leandrodamascena: I'm checking the signature of this method and I don't see the `name` variable being used. Is this for compatibility/non-breaking change reasons?

leandrodamascena (changes requested):

aws_lambda_powertools/metrics/provider/base.py
roger-zhangg: Yeah these should be dynamic. For current design I can imaging a custom provider can change these value in following ways
- Use different `MAX_METRICS` value in __init__ of their custom provider
- Add a `provider.set_max_metric` in their custom provider
- Use something like `provider.MAX_METRICS=50` before passing the provider into metrics class
Besides these approach, we can also add a provider config parameter. So user can pass in custom config to changes these values.
What do you think we should add here to make it more dynamic?

aws_lambda_powertools/metrics/provider/cloudwatchemf_provider.py
roger-zhangg: For this function. The name is set in the Dict: metrics. But I kept the name variable here in case other provider might need name for some internal logic.

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -119,6 +119,10 @@ def clear_metrics(self) -> None:
         self.set_default_dimensions(**self.default_dimensions)
 
 
+# add alias for original EMF format to make the provider more explicit
+CloudWatchEMF = Metrics
```
leandrodamascena: Great catch, Roger! With this class, we make the class for this specific provider more explicit.

aws_lambda_powertools/metrics/provider/__init__.py
```diff
@@ -0,0 +1,12 @@
+from aws_lambda_powertools.metrics.provider.base import MetricsBase, MetricsProviderBase
+from aws_lambda_powertools.metrics.provider.datadog_provider_draft import (
+    DataDogMetrics,
+    DataDogProvider,
+)
```
leandrodamascena: By doing this, I think we can fill in @heitorlessa's feedback to have the provider as a parameter and make it easier to switch between providers.

https://github.com/awslabs/aws-lambda-powertools-python/issues/2015#issuecomment-1550435695
>Keep the provider argument even for these standalone classes to make it easier for customers to test by swapping them with an InMemory fake provider (we could even provide that in the future).

leandrodamascena (dismissed):
@roger-zhangg great job refactoring Metrics to use the new standalone provider approach. What do you think about moving the CloudWatchEMF/Metrics to the provider folder, so in this case we have all providers using the same structure. I think you did that before changing that PR to use this new approach, right?

roger-zhangg:
> @roger-zhangg great job refactoring Metrics to use the new standalone provider approach. What do you think about moving the CloudWatchEMF/Metrics to the provider folder, so in this case we have all providers using the same structure. I think you did that before changing that PR to use this new approach, right?

Yeah I'll move it

pyproject.toml
```diff
@@ -33,6 +33,8 @@ fastjsonschema = { version = "^2.14.5", optional = true }
 pydantic = { version = "^1.8.2", optional = true }
 boto3 = { version = "^1.20.32", optional = true }
 typing-extensions = "^4.6.2"
+opentelemetry-sdk = "^1.18.0"
+opentelemetry-exporter-otlp-proto-grpc = "^1.18.0"
```
leandrodamascena: Note: after the final tests with these providers (Datadog + OTEL) we must remove the providers and these dependencies.

Please don't resolve this conversation.

leandrodamascena (commented):
@roger-zhangg We made a lot of changes in the develop branch and I fix some small problems here.

I'm finishing up the last few exploratory and DX tests, but I think we're on track to get this done.

pyproject.toml
roger-zhangg: Yeah that's true. I was thinking about the same thing

heitorlessa:
Folks - what's the latest here? 

I'm worried if we're gonna get Observability providers ready for reinvent

roger-zhangg:
> Folks - what's the latest here?
> 
> I'm worried if we're gonna get Observability providers ready for reinvent

We did multiple provider tests including Datadog and OTEL with the new metrics provider. I believe we need a final reivew before proceeding to merge?

heitorlessa:
Great! Self-assign to review -- when you can, could you update the PR body with snippets of the proposed UX please?

Same as the other PR, we should be able to know what's changed and how customers will perceive this value add change at a quick glance.

Thank you so much (again)

roger-zhangg:
@heitorlessa Just updated the UX in PR body. Thanks!

leandrodamascena:
> @heitorlessa Just updated the UX in PR body. Thanks!

Hi @heitorlessa! We need your help to make a final review of this PR and provide feedback if necessary. To make your job easier, let me explain a few points.

1 - The default provider remains `Metrics` to flush metrics for CloudWatch using EMF.

2 - `MetricManager` has been kept as the default class for the default `Metrics` provider (CloudWatch EMF). I don't believe there is a need to refactor as CloudWatch EMF is very specific and remains the default provider, we keep it as is.

3 - We added a new class `AmazonCloudWatchEMF`, which is basically an alias for Metrics. So the client can import like this:

```python
from aws_lambda_powertools import AmazonCloudWatchEMF
metrics = AmazonCloudWatchEMF()
```
Does it make sense to keep this? Is this too much? Should we create an alias for `EphemeralMetrics` also? We would like to introduce the experience of using an instance with the name of the provider.


4 - In this PR two new classes were added: `MetricsProviderBase` and `MetricsBase` to match the experience proposed here: https://github.com/aws-powertools/powertools-lambda-python/issues/2015#issuecomment-1492242654. This way we make it easy to bring any metrics provider that needs different configurations.

5 - @roger-zhangg created two experimental providers (DataDog and OTEL) to test it and I could use it with no problem. Do you want to try? Let me know and we can arrange a meeting to help on this test.

6 - We've added additional dependencies to test external providers, but don't worry, we'll remove them before merging.

7 - I remember you mentioned adding a generic method to create operational metrics that can be used by other utilities just passing an instance of Metrics to them. I think we need to discuss this in another issue or even an RFC before implementing this. This will not be implemented in this PR, does it make sense?

8- We really need your help with the documentation. Should we create a section with a real external provider? with an example? I think this is a great feature and we can take this opportunity to spread the word.

Thanks.

leandrodamascena:
Hello @heitorlessa! I think we are very close to finishing this and it will be a great achievement.

Can we book a time/day next week to focus on that and finalize it? On Friday I have a sync with @roger-zhangg and we're going to review the PR one more time, just so we don't miss anything.

Thanks

heitorlessa:
Three quick notes:

* Try switching between Metrics EMF and Datadog provider to see if the UX is on the Powertools bar -- this will improve the PR body and even the RFC (update later), including docs
    * Update PR body accordingly 
* For later, I noticed the Datadog provider has `tags={...}`. You can make the experience nicer by relying on `**kwargs` - it then becomes `add_metric(name="b", value=1, environment="prod", product="payment")`
* I noticed the snippets don't make use of the capture_cold_start metric (does it work for Datadog?)

heitorlessa:
Quick notes as we go through peer review @leandrodamascena -- Last ones before we start docs.

* [x] Metrics is our premium. It should default to instantiate `AmazonCloudWatchMetricsEMFProvider()` if no `provider` is passed. Otherwise we'll become too Java-ish in having too many options to setup something we already know upfront what good looks like.
* [x] Same with Datadog: DataDogMetrics -> instantiate DataDogMetricsProvider()
* [x] Move from high-level import to FQDN import to avoid additional lazy import logic
* [x] Move DataDog to a separate PR unless they already merged the fix
* [x] DataDog branding: DataDog -> Datadog AFAIK
* [x] Remove `MetricManager` subclassin from `Metrics, `EphemeralMetrics`, and `SingleMetric`
    * We keep `MetricManager` to not break anyone, but consider it deadzone
    * In V3, we add a deprecation warning to use CW EMF Provider if one is using it (but only remove in V4, so we give everyone ample time to react to it)
* [x] Implement all methods from `MetricManager` in `Metrics` but use `self.provider` to shim that instead
* [x] Extend `Metrics` to accept `provider` optional param

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -69,76 +73,82 @@ def lambda_handler():
...
-    def __init__(self, service: Optional[str] = None, namespace: Optional[str] = None):
-        self.default_dimensions = self._default_dimensions
-        self.dimension_set = self._dimensions
+    def _extract_metric_unit_value(self, unit: Union[str, MetricUnit]) -> str:
+        return self.provider._extract_metric_unit_value(unit=unit)
 
-        self.dimension_set.update(**self._default_dimensions)
-        super().__init__(namespace=namespace, service=service)
+    def _add_cold_start_metric(self, context: Any) -> None:
+        self.provider._add_cold_start_metric(context=context)
```
leandrodamascena, line 148: @heitorlessa We need to persist default dimension values in AmazonCloudWatchEMFProvider and Metrics.

leandrodamascena:
Hello @heitorlessa! About the documentation, do you think it makes sense to add it now OR when we add the Datadog provider? In my opinion, we should add Datadog and create the documentation.

Can you make a final review, please?

leandrodamascena:
@roger-zhangg I would like your review here also.

roger-zhangg (commented):
LGTM

aws_lambda_powertools/metrics/metrics.py
```diff
@@ -168,6 +178,129 @@ def clear_default_dimensions(self) -> None:
         self.default_dimensions.clear()
 
     def clear_metrics(self) -> None:
-        super().clear_metrics()
-        # re-add default dimensions
-        self.set_default_dimensions(**self.default_dimensions)
+        self.provider.clear_metrics()
+
```
heitorlessa, line 183: check whether tests are accurate since we're not resetting local default dimensions.

aws_lambda_powertools/metrics/provider/base.py
```diff
@@ -0,0 +1,244 @@
...
+from typing_extensions import Protocol
+
+logger = logging.getLogger(__name__)
+
+is_cold_start = True
+
+
+class MetricsProviderBase(Protocol):
+    """
+    Class for metric provider template.
```
heitorlessa:
```suggestion
    Class for metric provider interface.
```

aws_lambda_powertools/metrics/provider/base.py
```diff
@@ -0,0 +1,244 @@
...
+
+    This class serves as a template for creating your own metric provider. Inherit from this class
```
heitorlessa:
```suggestion
    This class serves as an interface for creating your own metric provider. Inherit from this class
```

heitorlessa (changes requested):
**Metrics Provider PR notes**

- Remove docs until we cleanup implementations and have our first provider (Datadog)
- Check whether we have a gap in tests for default dimensions not being updated locally but within a provider

**Separate PR for cleanup**

- Export CloudWatch EMF stuff within cloudwatch_emf
- MetricsBase and MetricsProviderBase confusion 
- Refactor `_extract_*` into standalone functions to reduce call graph
- Split up functional and unit tests with new providers

heitorlessa:
peer review last comments so we can merge this today. We can cleanup the slight confusion in OOO / Protocol, docs, etc. in a separate PR (easier to review later).

CAN'T WAIT FOR THIS!

aws_lambda_powertools/metrics/metrics.py
leandrodamascena: Hi @heitorlessa! I think his comment should be on line 177 instead of here.

I added a line to clear the default dimensions in the provider as well. My understanding of this is because we are passing `default_dimensions` to the provider and we need to clear it there and locally. Our tests cover these lines, and we have tests to back it up.

I created an exploratory test and it works as expected.


```python
import json

from aws_lambda_powertools import Logger, Metrics

metrics = Metrics(namespace="test")
metrics.set_default_dimensions(app="powertools")

    metrics.add_metric(name="metrics1", unit=MetricUnit.Count, value=1)
    metrics.clear_default_dimensions()
    metrics.flush_metrics()

    metrics.add_metric(name="metrics2", unit=MetricUnit.Count, value=1)
    metrics.flush_metrics()
    return {"sucesso": "ok"}
```

OUTPUT

```sh
{"_aws":{"Timestamp":1690907256372,"CloudWatchMetrics":[{"Namespace":"test","Dimensions":[["app","service"]],"Metrics":[{"Name":"metrics1","Unit":"Count"}]}]},"app":"powertools","service":"PowertoolsHelloWorld","metrics1":[1.0]}
{"_aws":{"Timestamp":1690907256372,"CloudWatchMetrics":[{"Namespace":"test","Dimensions":[["service"]],"Metrics":[{"Name":"metrics2","Unit":"Count"}]}]},"service":"PowertoolsHelloWorld","metrics2":[1.0]}
```


Let me know if you have any other consideration.

leandrodamascena:
> peer review last comments so we can merge this today. We can cleanup the slight confusion in OOO / Protocol, docs, etc. in a separate PR (easier to review later).
> 
> CAN'T WAIT FOR THIS!

I removed the docs for now and I'll create an additional issue to pay this tech debt.

ALMOST THERE! @roger-zhangg!

aws_lambda_powertools/metrics/metrics.py
heitorlessa: It's missing the test for it only ;) Like this but with a provider being explicitly passed: https://github.com/aws-powertools/powertools-lambda-python/blob/develop/tests/functional/test_metrics.py#L1068

heitorlessa: ah done ;) https://github.com/aws-powertools/powertools-lambda-python/pull/2194/commits/92eda07d478f5a9c283d9d787739321cb49bdd4c 

<3

heitorlessa (approved):

heitorlessa:
pending CI checks, feel free to merge ;) Gotta go cook dinner. Thank you both @roger-zhangg and @leandrodamascena - this marks the initial steps of supporting any observability provider 🍺 🎉

PR #2197, opened 2023-05-02
feat(data_masking): add new sensitive data masking utility

### LINKED ISSUE #1173: Sensitive data masking utility
**Runtime e.g. Python, Java, all of them.** Python

As a customer, I'd like to obfuscate incoming data for known fields that contain PII, so that they're not passed downstream or accidentally logged.

It could be any of these ideas or something better if anyone wants to chime in:

* Similar to [DAZN Obsfucater Middleware for NodeJS](https://github.com/getndazn/dazn-lambda-powertools/tree/master/packages/lambda-powertools-middleware-obfuscater)
* Standalone function that takes a JSON Path and does an in-place editing to mask data using masking of my choice (e.g. `***`, `###`)

A more complex operation would be for data depersonalization where I'd want to encrypt and store the correct data somewhere, in which a separate actor would have permission to decrypt it.

1. Mask data on my own with a function that walks through a graph and recursively change it.
2. Use Amazon Comprehend Medical to detect PHI data, convert it to a `str`, and run a string replacement to mask any sensitive data encountered

**Is this something you'd like to contribute if you had guidance?**

**Additional context**

### LINKED ISSUE #1858: RFC: Sensitive data masking utility

https://github.com/awslabs/aws-lambda-powertools-python/issues/1173

Other

### Summary

Customers would like to obfuscate incoming data for known fields that contain PII, so that they're not passed downstream or accidentally logged. With the increase of batch processing utilities and GDPR, this is one of the hardest tasks for customers today, specially when considering multi-account users.

AWS Encryption SDK is a good starting point but it can be too complex for the average developer, data engineer, or DevOps persona to use. As such, it is highly requested that the Powertools library have a utility to easily mask and/or encrypt sensitive data.

### Use case

The use case for this utility would be for developers who want to mask or encrypt sensitive data such as names, addresses, SSNs, etc. in order for them to not be logged in CloudWatch so such data is not compromised, and so that downstream systems like S3, DynamoDB, RDBMS, etc. will not need any additional work on handling PII data.

Additionally, developers should be able to recover encrypted sensitive data to its original form so that they can handle sensitive requests around that data on a as-needed basis.

### Proposal

The data masking utility should allow users to mask data, or encrypt and decrypt it. If they would like to encrypt their data, customers should be able to decide for themselves which encryption provider they want to use, though we will provide an out-of-the-box integration with the [AWS Encryption SDK](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/introduction.html). The below code snippet is a rudimentary look at how this utility can be used and how it will function.

**Usage**
```python
from aws_lambda_powertools.utilities.data_masking.constants import KMS_KEY_ARN
from aws_lambda_powertools.utilities.data_masking.provider.kms.aws_encryption_sdk import AwsEncryptionSdkProvider


def lambda_handler():
    
   data_masker = DataMasking()
    data = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "johndoe@example.com",
               "address": {
                    "street": "123 Main St", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }

    masked = data_masker.mask(data=data, fields=["email", "address.street"])
    """
    masked = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "*****",
               "address": {
                    "street": "*****", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """

    encryption_provider = AwsEncryptionSdkProvider(keys=[KMS_KEY_ARN])
    data_masker = DataMasking(provider=encryption_provider)

    encrypted = data_masker.encrypt(data=data, fields=["email", "address.street"])
    """
    encrypted = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "InRoaXMgaXMgYSBzdHJpbmciHsLZGx2na-XzP_TB5Bf2LNU1bLc",
               "address": {
                    "street": "XMgYSB_KDddaDJYMb-JpbmGnagTklwQ-msdaDLP", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """
 
    decrypted = data_masker.decrypt(data=encrypted, fields=["email", "address.street"])
    """
    data = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "johndoe@example.com",
               "address": {
                    "street": "123 Main St", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """
```

**AWS Encryption SDK**

The AWS Encryption SDK is a client-side encryption library that makes it easier to encrypt and decrypt data of any type in your application. The Encryption SDK is available in all the languages that Powertools supports. You can use it with customer master keys in AWS Key Management Service (AWS KMS), though the library does not require any AWS service. When you encrypt data, the SDK returns a single, portable encrypted message that includes the encrypted data and encrypted data keys. This object is designed to work in many different types of applications. You can specify many of the encryption options, including selecting an encryption and signing algorithm.

**Latencies**
Latencies for using this utility with the AWS Encryption SDK in Lambda functions configured with 128MB, 1024MB, and 1769MB, respectively.
```
plugins.metrics-by-endpoint.response_time./Prod/function128:
  min: ......................................................................... 0
  max: ......................................................................... 5373
  median: ...................................................................... 561.2
  p95: ......................................................................... 713.5
  p99: ......................................................................... 1620
plugins.metrics-by-endpoint.response_time./Prod/function1024:
  min: ......................................................................... 0
  max: ......................................................................... 5039
  median: ...................................................................... 89.1
  p95: ......................................................................... 144
  p99: ......................................................................... 232.8
plugins.metrics-by-endpoint.response_time./Prod/function1769:
  min: ......................................................................... 0
  max: ......................................................................... 1726
  median: ...................................................................... 82.3
  p95: ......................................................................... 133
  p99: ......................................................................... 183.1
```

**Custom encryption**
If customer would like to use another encryption provider, or define their own encrypt and decrypt functions, we will define an interface that the customer can implement and pass in to the DataMaskingUtility class.

```python
from aws_lambda_powertools.utilities.data_masking.provider import BaseProvider
from itsdangerous.url_safe import URLSafeSerializer

class MyCustomEncryption(BaseProvider):
    def __init__(self, secret):
        self.secret = URLSafeSerializer(secret)

    def encrypt(self, value: str) -> str:
        if value is None:
            return value
        return self.secret.dumps(value)

    def decrypt(self, value: str) -> str:
        if value is None:
            return value
        return self.secret.loads(value)


def lambda_handler():
    data = {
        "id": 1,
        "name": "John Doe",
        "age": 30,
        "email": "johndoe@example.com",
        "address": {
            "street": "123 Main St", 
            "city": "Anytown", 
            "state": "CA", 
            "zip": "12345"
        },
    }

    masking_provider = MyCustomEncryption(secret="secret-key")
    data_masker = DataMasking(provider=masking_provider)

    encrypted = data_masker.encrypt(data, fields=["email", "address.street"])
    """
    encrypted = {
        "id": 1,
        "name": "John Doe",
        "age": 30,
        "email": "InRoaXMgaXMgYSBzdHJpbmciHsLZGx2na-XzP_TB5Bf2LNU1bLc",
        "address": {
            "street": "XMgYSB_KDddaDJYMb-JpbmGnagTklwQ-msdaDLP", 
            "city": "Anytown", 
            "state": "CA", 
            "zip": "12345"
        },
    }
   """

   decrypted = data_masker.decrypt(data=encrypted, fields=["email", "address.street"])
    """
    decrypted = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "johndoe@example.com",
               "address": {
                    "street": "123 Main St", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """
```

### Out of scope

Traversing an arbitrary dictionary will be out of scope for the initial launch of this tool. This feature will be to receive instructions as to where in the given dictionary it should mask/unmask the data.

We still need to determine the most efficient method of taking input JSON path masking or encrypting the value at that path. JMESPath or JSON path can be considered for simple use cases but we need to find the fastest method.

### Potential challenges

* We will need to discuss the design and implementation with a security expert to ensure safety.
* Also need to determine if we should support envelope encryption for customers using the AWS Encryption SDK
    * Envelope encryption is the practice of encrypting plaintext data with a data key, and then encrypting the data key under another key. But, eventually, one key must remain in plaintext so you can decrypt the keys and your data. This top-level plaintext key encryption key is known as the root key. You can store root keys in AWS KMS, known as [AWS KMS keys](https://docs.aws.amazon.com/kms/latest/developerguide/concepts.html#kms_keys).
* Need to decide what the best name for the methods are. Should they still be called `encrypt` and `decrypt` even in the case where the customer only masks and won't be able to decrypt later? Should we have a `mask` method in the interface that is always accessible so that users can irretrievably mask some data and also encrypt some data?

Integration with the AWS Encryption SDK.

### Alternative solutions

_No response_

### Acknowledgment

seshubaws:
**Issue number:**
https://github.com/awslabs/aws-lambda-powertools-python/issues/1858
### Changes

As per the above RFC and this issue: https://github.com/awslabs/aws-lambda-powertools-python/issues/1173, this is the initial PR for adding a data masker for sensitive information like PII into Powertools.

This feature is meant to mask, encrypt, and decrypt any and all data types entirely, or if provided a list of keys for a given dictionary (even nested keys), it has the ability to mask, encrypt, and decrypt the corresponding values in the dictionary.

In this PR, there are additions for:

1. A new encryption Base Provider class that can mask, encrypt, and decrypt data, giving users the ability to install any encryption library they want and use it with our Data Masker.
2. A Data Masker class that takes in an encryption provider, the abstraction through which the customer would mask, encrypt, and/or decrypt their data.
3. An implementation of how the AWS Encryption SDK would implement the Base Provider class, which is an out of the box encryption provider for users to mask/encrypt their data.

NOTE: Measured latency for encrypting and decrypting nested fields in a 10KB json blob, came out to 5ms.

### User experience

This would be a new feature, so before a user would have to implement this logic themselves.

**AFTER**
```python
from aws_lambda_powertools.utilities.data_masking.constants import KMS_KEY_ARN
from aws_lambda_powertools.utilities.data_masking.provider.kms.aws_encryption_sdk import AwsEncryptionSdkProvider

    data_masker = DataMasking()
    data = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "johndoe@example.com",
               "address": {
                    "street": "123 Main St", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }

    masked = data_masker.mask(data=data, fields=["email", "address.street"])
    """
    masked = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "*****",
               "address": {
                    "street": "*****", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """

    encryption_provider = AwsEncryptionSdkProvider(keys=[KMS_KEY_ARN])
    data_masker = DataMasking(provider=encryption_provider)

    encrypted = data_masker.encrypt(data=data, fields=["email", "address.street"])
    """
    encrypted = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "InRoaXMgaXMgYSBzdHJpbmciHsLZGx2na-XzP_TB5Bf2LNU1bLc",
               "address": {
                    "street": "XMgYSB_KDddaDJYMb-JpbmGnagTklwQ-msdaDLP", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """
 
    decrypted = data_masker.decrypt(data=encrypted, fields=["email", "address.street"])
    """
    decrypted = {
              "id": 1,
               "name": "John Doe",
               "age": 30,
               "email": "johndoe@example.com",
               "address": {
                    "street": "123 Main St", 
                    "city": "Anytown", 
                    "state": "CA", 
                    "zip": "12345"
               },
      }
   """
```

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena:
Hi @seshubaws! Thank you so much for submitting this PR and for all the work put into it, this feature is a real improvement on the project as we certainly have a lot of customers who want it. The CI is failing for some reason that I haven't looked into yet, but at this point I see room for us to improve on the following and make the developer/user experience even better:

- Organization of folders/files: perhaps it is better to reorganize folders and files in order to have smaller files and with a kind of single responsibility, something like this:
	- base.py - for the base class
	- provider.py - for the provider class
	- one file per provider

- Write some performance tests for us to make this public and hear some feedback from the community.

- Make some code changes to maintain an architecture pattern similar to the one we use.

- Write documentation

- Check the dependencies and what we are missing to include

- Tests in general

I'm really excited to move forward with this PR. Would you rather I convert the draft to "Ready" and push some commits, or do I review and make some suggestions for you to work on? 
Let's collaborate to make it a success! :rocket:

seshubaws:
Hey @leandrodamascena, thanks for the comments! I just got back from PTO, I'll start working on revising this to be in smaller files like base.py, provider.py, etc. but you can convert this draft to Ready and push some commits as well!

seshubaws:
Hey all, I'm running into some confusing when trying to mock tests for the AWS Encryption SDK. I tried using Stubber to stub a KMS key, but the encryption SDK doesn’t accept that mock key to actually encrypt anything, and Stubber doesn't accept the SDK as a client since it’s not an AWS service, just an SDK. Are there are any examples on how to mock this SDK? 

Outside sources say try using the moto library, but I don't want to bring in more dependencies unless strictly necessary. @heitorlessa @rubenfonseca any thoughts on this?

tests/unit/data_masking/test_data_masking.py
```diff
@@ -0,0 +1,216 @@
...
+from aws_lambda_powertools.utilities.data_masking.base import DataMasking
+from aws_lambda_powertools.utilities.data_masking.provider import Provider
+from aws_lambda_powertools.utilities.data_masking.providers.aws_encryption_sdk import (
+    AwsEncryptionSdkProvider,
+)
+from aws_lambda_powertools.utilities.data_masking.providers.itsdangerous import (
+    ItsDangerousProvider,
+)
+
+AWS_SDK_KEY = "arn:aws:kms:us-west-2:683517028648:key/269301eb-81eb-4067-ac72-98e8e49bf2b3"
```
seshubaws: Still need to find a way to mock this key

leandrodamascena:
Hello @seshubaws! Starting to work on it to have this important feature ready in the coming weeks. The first thing I did was a small change to the new dependencies: `itsdangerous` and `aws_encryption_sdk`. 
Giving you a context of this change, in Powertools we try to keep the package with the core features as small as possible and make the other dependencies optional, so the customer can install it only if he wants to, with that we can help the customer to reduce cold start, improve the deployment package size, make deployment stages faster and so on.

So what I did in this change was create a new extra called `datamasking` that brings `itsdangerous` and `aws_encryption_sdk` dependencies and instead of installing them in any `pip install aws_lambda_powertools`, they will only be installed when the client types `pip install aws_lambda_powertools[datamasking]`.

The next steps will be to review the feature design and what improvements we might be able to make to improve the user experience. I'm working on it now.

aws_lambda_powertools/utilities/data_masking/providers/itsdangerous.py
leandrodamascena, line 1: Hello @seshubaws! I was playing around with `itsdangerous` library these days to go deeper into this implementation and I have a bunch of questions that you might have already figured out.

1 - From what I understand the `itsdangerous` library is a library to detect and avoid tampered data, right? I mean, basically what it does is add a signature to the original data to transmit it over an insecure channel and prevent any tampering. Reference: https://itsdangerous.palletsprojects.com/en/2.1.x/serializer/#api

2 - What we are using is the `URLSafeSerializer`, which is an option inside the  `itsdangerous` library to encrypt and decrypt data. There is another option which is `URLSafeTimedSerializer`, did you tried it? Just curious if we can take any advantage of this.

3 - Assuming we are going to use the `URLSafeSerializer` API, I see in every function signature that you add the `**kwargs` argument. Do we need other arguments besides salt, keys, and data? We may need to inspect the `URLSafeSerializer` API further to see what fields we really need and add this instead of `**kwargs`. Using `**kwargs` we can't add proper type hints to developers and we try to avoid this as much as possible.

4 - `serialized = URLSafeSerializer(self.keys, salt=self.salt, **kwargs)` and `URLSafeSerializer(self.keys, salt=self.salt, **kwargs)` are adding the salt again, so will we consider that a client can instantiate the class with one salt and use a different salt when encrypting/decrypting? We may not need this field here.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -0,0 +1,58 @@
...
+    def mask(self, data, fields=None, **kwargs):
+        return self._apply_action(data, fields, self.provider.mask, **kwargs)
+
+    def _apply_action(self, data, fields, action, *args, **kwargs):
+        if fields is not None:
+            return self._use_ast(data, fields, action, *args, **kwargs)
+        else:
+            return action(data, *args, **kwargs)
+
+    def _use_ast(self, data: Union[dict, str], fields, action, *args, **kwargs) -> str:
```
leandrodamascena: I think we need to do some more specific tests on this method and test its performance. I mean, at first glance, it looks like every time a client tries to encrypt and decrypt data, it goes through this function by running `json.loads/dumps` and that costs a lot of CPU. Do you have any ideas for this? I'll think of a way to validate.

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
```diff
@@ -0,0 +1,55 @@
...
+        if cls not in cls._instances:
+            instance = super().__call__(*args, **kwargs)
+            cls._instances[cls] = instance
+        return cls._instances[cls]
+
+
+class AwsEncryptionSdkProvider(Provider, metaclass=SingletonMeta):
+    CACHE_CAPACITY: int = 100
+    MAX_ENTRY_AGE_SECONDS: float = 300.0
+    MAX_MESSAGES: int = 200
```
leandrodamascena: You can move it outside this class.

aws_lambda_powertools/utilities/data_masking/provider.py
```diff
@@ -0,0 +1,26 @@
+from abc import abstractmethod
+from collections.abc import Iterable
+
+from aws_lambda_powertools.shared.constants import DATA_MASKING_STRING as MASK
```
leandrodamascena: You don't need to rename this constant, use it as defined: `DATA_MASKING_STRING`.

Using just MASK increases the cognitive load of understanding what that variable is.

leandrodamascena (changes requested):
Hello, @seshubaws! I made a few comments to address before we move on.

aws_lambda_powertools/utilities/data_masking/providers/itsdangerous.py
seshubaws: 1. Yes that’s right, it signs data upon receiving it to make sure it hasn’t been tampered with.

2. I haven’t played around with `URLSafeTimedSerializer` but from what I can tell, it’s an extension of `URLSafeSerializer` but also includes a `max_age` parameter in its `loads` method so that customers can set how long they want the signature to be valid. If the signature is outdated, it will raise a `SignatureExpired` exception. If you think this capability will be useful to cxs, we can look into replacing `URLSafeSerializer` with this one.

3. The only parameter that is required for `URLSafeSerializer` is the key, but there are several others that are optional. I’ll just put all the parameters in the init function so we don’t have to pass kwargs around, thank you for this callout!

4. I’m not quite sure what you mean by this question, I’m only sending in the `self.salt` that was set in the constructor into the `encrypt/decrypt` methods, and those methods don’t allow the user to pass in another salt parameter. Could you clarify this comment?

leandrodamascena: > I haven’t played around with URLSafeTimedSerializer but from what I can tell, it’s an extension of URLSafeSerializer but also includes a max_age parameter in its loads method so that customers can set how long they want the signature to be valid. If the signature is outdated, it will raise a SignatureExpired exception. If you think this capability will be useful to cxs, we can look into replacing URLSafeSerializer with this one.

I think makes sense we keep only the `URLSafeSerializer` for now. Thank you for the explanation.

> I’m not quite sure what you mean by this question, I’m only sending in the self.salt that was set in the constructor into the encrypt/decrypt methods, and those methods don’t allow the user to pass in another salt parameter. Could you clarify this comment?

I wasn't really clear in my question. But my question doesn't make sense anymore because I saw you update the file with the arguments :). sorry about that

aws_lambda_powertools/utilities/data_masking/base.py
seshubaws: Definitely agree! This was mostly a placeholder solution to get eyes on the design as a whole. I will test the performance of this and brainstorm alternative solutions in the next couple days! I'll work on removing `**kwargs` from the other files as well :)

seshubaws: I added performance tests to this method [here](https://github.com/aws-powertools/powertools-lambda-python/pull/2197/files#diff-7eaaf1b8a36da544776eb586fc5324672ee2fa3de790e900a48751cf2ad331a2). Here are the results:

```
--------------------------------------------------------------------------------------------- benchmark 'core': 2 tests ---------------------------------------------------------------------------------------------
Name (time in us)                                 Min                 Max                Mean            StdDev              Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_data_masking_init                        32.6320 (1.0)       32.6320 (1.0)       32.6320 (1.0)      0.0000 (1.0)       32.6320 (1.0)      0.0000 (1.0)           0;0       30.6447 (1.0)           1           1
test_data_masking_encrypt_with_json_blob     530.6170 (16.26)    530.6170 (16.26)    530.6170 (16.26)    0.0000 (1.0)      530.6170 (16.26)    0.0000 (1.0)           0;0        1.8846 (0.06)          1           1
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

Legend:
  Outliers: 1 Standard Deviation from Mean; 1.5 IQR (InterQuartile Range) from 1st Quartile and 3rd Quartile.
  OPS: Operations Per Second, computed as 1 / Mean
```

Let me know if you want me to modify the SLA times, and if we still want to explore caching around this portion.

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
```diff
@@ -0,0 +1,56 @@
...
+import botocore
+from aws_encryption_sdk import (
+    CachingCryptoMaterialsManager,
+    EncryptionSDKClient,
+    LocalCryptoMaterialsCache,
+    StrictAwsKmsMasterKeyProvider,
+)
+
+from aws_lambda_powertools.utilities.data_masking.provider import Provider
+
```
roger-zhangg:
```suggestion
from aws_lambda_powertools.shared.user_agent import register_feature_to_botocore_session
```

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
```diff
@@ -0,0 +1,56 @@
...
+CACHE_CAPACITY: int = 100
+MAX_ENTRY_AGE_SECONDS: float = 300.0
+MAX_MESSAGES: int = 200
+# NOTE: You can also set max messages/bytes per data key
+
+
+class AwsEncryptionSdkProvider(Provider, metaclass=SingletonMeta):
+    cache = LocalCryptoMaterialsCache(CACHE_CAPACITY)
+    session = botocore.session.Session()
+
```
roger-zhangg:
```suggestion
register_feature_to_botocore_session(session, "data-masking")
```

roger-zhangg (commented):
Hi Seshu, The user-agent patch botocore function is merged. You can try to pull the latest develop branch and apply the suggested change to patch the botocore session before init KMS provider. Thanks! [User-agent PR](https://github.com/aws-powertools/powertools-lambda-python/pull/2614/files)

seshubaws:
I removed unit tests for AWS Encryption SDK because of the mocking KMS key issue, I will add functional tests for that instead!

tests/unit/data_masking/test_data_masking.py
```diff
@@ -0,0 +1,247 @@
...
+
+import pytest
+from itsdangerous.url_safe import URLSafeSerializer
+
+from aws_lambda_powertools.shared.constants import DATA_MASKING_STRING
+from aws_lambda_powertools.utilities.data_masking.base import DataMasking
+from aws_lambda_powertools.utilities.data_masking.provider import Provider
+
+
+class MyEncryptionProvider(Provider):
```
seshubaws: I can pull out all the input data necessary for this into a utilities file if that would be cleaner

heitorlessa:
It looks like we had an issue with the encryption SDK dependency, looking into it before we chat.

https://github.com/aws-powertools/powertools-lambda-python/actions/runs/5814909893/job/15785493238?pr=2197

heitorlessa:
This patch should fix the CI check for CVE since our lockfile is not using the latest `cryptography` transitive dependency that AWS Encryption SDK depends on.

[0001-chore-update-lock-to-latest-range-deps.patch](https://github.com/aws-powertools/powertools-lambda-python/files/12313904/0001-chore-update-lock-to-latest-range-deps.patch)

heitorlessa:
Updating after our 1:1 - I'll be taking the functional tests and we regroup early next week.

---

* [ ] Split DataMasking with and without KMS tests
* [ ] Make tests more verbose by building test data as part of tests
* [ ] Stub or mock KMS Encryption SDK calls

seshubaws:
> This patch should fix the CI check for CVE since our lockfile is not using the latest `cryptography` transitive dependency that AWS Encryption SDK depends on.
> 
> [0001-chore-update-lock-to-latest-range-deps.patch](https://github.com/aws-powertools/powertools-lambda-python/files/12313904/0001-chore-update-lock-to-latest-range-deps.patch)

Applied the patch, thank you!

heitorlessa:
hey Seshu, I've fixed the conflict and leftovers from `itsdangerous` but didn't want to force push to `develop` due to the merge conflict -- I created another branch instead: https://github.com/seshubaws/aws-lambda-powertools-python/tree/feat/data-masking

Gonna resume working on Encryption SDK stubber now

---

Changes

* [x] Remove leftover from ItsDangerous
  * [x] `tests/performance/test_data_masking.py` (use `.mask` over `.encrypt/decrypt` for now)
  * [x] `pyproject.toml`
  * [x] `poetry.lock`
* [x] Remove `datamasking-all`
  * we need customers to validate when they'd want to use multiple providers to warrant it

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
```diff
@@ -0,0 +1,60 @@
...
+        return cls._instances[cls]
+
+
+CACHE_CAPACITY: int = 100
+MAX_ENTRY_AGE_SECONDS: float = 300.0
+MAX_MESSAGES: int = 200
+# NOTE: You can also set max messages/bytes per data key
+
+
+class AwsEncryptionSdkProvider(Provider, metaclass=SingletonMeta):
```
seshubaws: Renato and I found an edge case while writing e2e tests regarding making this a singleton class as it will only create one instance of the class. If you instantiate an instance of the AwsEncryptionSdkProvider class with a certain key, encrypt some data with it, instantiate an instance of the AwsEncryptionSdkProvider class with a *different* key, and then try to decrypt the data encrypted with the first key, it will work and decrypt the data because the second class (and so key) never actually got created. So it uses the first key that was created for everything. Seems like maybe a security issue?

Probably the general use case is that users pass in a list of keys they have in the very beginning and use that provider the whole way through, in which case there would not be a problem. But the above is a case we should discuss. Should we not have it be a singleton class? Or perhaps we could modify the current SingletonMeta class to make it so a new instance of AwsEncryptionSdkProvider gets instantiated only if there are new keys passed in?

heitorlessa: Absolutely, I was hoping you'd catch that in E2E since the code was more of a contrived POC - we're also missing config to expose those defaults that might not be good for some people.

For context, the Singleton was only added as a quick solution because Encryption provider was requesting an envelope key per `encrypt()` which adds seconds to the overall operation -- there must be a better way, I just sadly ran out of time so did a dirty Singleton.

If Singleton is the only way then absolutely, we should take keys into account and maybe an option to get a clean non-singleton for testing or cases like EC2 Nitro (one encryption operation per envelope key)

heitorlessa:
Adding the set of tasks we discussed so it's more visible on the main areas first, then we can iterate on what's left to make it prod ready.

---

* [ ] Rename `Provider` to `BaseProvider` - same convention we use in other parts
* [ ] Turn `Provider` an abstract class by subclassing `ABC`
* [ ] Add types for `Provider`'s methods
* [ ] Docstring `_apply_action` DataMasking` using Numpy style - [VSCode Extension with numpy style support](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring)
* [ ] Revisit `Turn into json string so everything has quotes around it`
* [ ] Simplify `DataMasking` self.provider logic with: `self.provider = provider or Provider()`
* [ ] Improve UX by accepting any JSON valid data in `encrypt` similar to ItsDangerous - it became apparent when writing tests that the UX would be too difficult to handle negative integers etc.
    * [ ] Optionally allow a custom serializer/deserializer so customers can handle more complex data types
* [ ] Type `provider` constructor in `DataMasking`
* [ ] Revisit error handling in `_apply_action_to_fields` for string data that isn't a JSON string
* [ ] Loop `for field in fields` is rewriting the loop target (`field`), consider another variable
* [ ] Document `for field in fields` logic
* [ ] Revisit variable names in `for field in fields` to reduce cognitive load
* [ ] Break down parametrize tests into more individual tests with related data types to ease detecting regressions
* [ ] Build simple data within a test instead of setups to ease learning about the test intent (screenshot before/after)
* [ ] E2E for AWS Encryption SDK 
* [ ] Bring `DATA_MASKING_STRING` to a local `constants.py` within Data Masking utility - we're starting to break that so we can more easily modularize it in the future
* [ ] Refactor `AwsEncryptionSdkProvider` to make production ready
  * [ ] Allow customers to set their cache capacity
  * [ ] Allow customers to set how long a cache entry will be valid
  * [ ] Allow customers to set maximum number of messages under a cache entry
  * [ ] Docstring to add more details in `encrypt`, `decrypt`, and `__init__`
  * [ ] Docstring to give an example how to use it (API docs)
  * [ ] Revisit use of `SingletonMeta` to address the underlying cause of a single KMS GenerateDataKey being called once (or more based on settings), for perf reasons. If it can't, we need to take into account more variables to make an instance unique (e.g., keys, cache settings, etc.)
* [ ] Document what 80% of customer needs to get going
  * [ ] Terminology
  * [ ] Diagrams
  * [ ] Types of data they can use
  * [ ] Minimal infra and IAM permissions (SAM) to be able to use this feature (separate encryption from decryption Lambda)
  * [ ] Advanced scenarios like custom caching, max messages encrypted etc, reusing Encryption Provider instance elsewhere in the code (along with side effects explanation)
  * [ ] Provider section so we can add others later
  * [ ] Refer to any best practices shared by the KMS team that customers should know
* [ ] Remove dependencies and references to `ItsDangerous` as it'll come in a separate PR.

seshubaws:
- [x] Rename Provider to BaseProvider - same convention we use in other parts
- [x]  Simplify DataMasking self.provider logic with: self.provider = provider or Provider()
- [x]  Bring DATA_MASKING_STRING to a local constants.py within Data Masking utility - we're starting to break that so we can more easily modularize it in the future
- [x] Remove dependencies and references to ItsDangerous as it'll come in a separate PR.

Opened a ticket with AWS Encryption SDK to discuss how their decrypt API doesn’t take in an `encryption_context` even if you pass one in to the encrypt method.

ETA: They replied in the ticket saying 

> the way I've described the API, the context passed on decrypt isn't _required_, but it allows the caller to assert that a set of pairs exist in the encryption context. So a customer not passing in an encryption context on decrypt would mean that they are not asserting anything about the context. IMO a step better than requiring the customer to do additional work themself.

Based on the above ticket, I added a check to say if the customer passes in an `encryption_context` to the decrypt method, then we check if it is the same as the one they passed into the encrypt method. If they don't pass in an encryption_context to the decrypt method, then we don't do a check as we are assuming the customer does not want to assert anything about the context as per the API.

Also, I asked @roger-zhangg to take a look at some of the Action Items until I'm back on 8/31!

tests/e2e/data_masking/test_data_masking.py
```diff
@@ -0,0 +1,129 @@
...
+
+@pytest.mark.xdist_group(name="data_masking")
+def test_encryption_context(data_masker):
+    # GIVEN an instantiation of DataMasking with the AWS encryption provider
+
+    value = bytes(str([1, 2, "string", 4.5]), "utf-8")
+
+    # WHEN encrypting and then decrypting the encrypted data with an encryption_context
+    encrypted_data = data_masker.encrypt(value, encryption_context={"this": "is_secure"})
+    decrypted_data = data_masker.decrypt(encrypted_data, encryption_context={"this": "is_secure"})
```
heitorlessa:
```suggestion
    value = bytes(str([1, 2, "string", 4.5]), "utf-8")
    context = {"this": "is_secure"}

    # WHEN encrypting and then decrypting the encrypted data with an encryption_context
    encrypted_data = data_masker.encrypt(value, encryption_context=context)
    decrypted_data = data_masker.decrypt(encrypted_data, encryption_context=context)
```

nitpick to ease reading it

tests/e2e/data_masking/test_data_masking.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # THEN the result is the original input data
+    assert decrypted_data == value
+
+
+@pytest.mark.xdist_group(name="data_masking")
+def test_encryption_diff_context_fail(data_masker):
```
heitorlessa:
```suggestion
def test_encryption_context_mismatch(data_masker):
```

nt: slightly clearer in the intent

tests/e2e/data_masking/test_data_masking.py
```diff
@@ -0,0 +1,129 @@
...
+    encrypted_data = data_masker.encrypt(value)
+
+    # THEN decrypting with an encryption_context should raise a ValueError
+    with pytest.raises(ValueError):
+        data_masker.decrypt(encrypted_data, encryption_context={"this": "is_secure"})
+
+
+# TODO: metaclass?
+@pytest.mark.xdist_group(name="data_masking")
+def test_encryption_key_fail(data_masker, kms_key2_arn):
```
heitorlessa:
```suggestion
def test_encryption_decryption_key_mismatch(data_masker, kms_key2_arn):
```

nt: slightly clearer intent

tests/e2e/data_masking/test_data_masking.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # THEN when decrypting with a different key it should fail
+    data_masker_key2 = DataMasking(provider=AwsEncryptionSdkProvider(keys=[kms_key2_arn]))
+
+    with pytest.raises(DecryptKeyError):
+        data_masker_key2.decrypt(encrypted_data)
+
+
+@pytest.mark.xdist_group(name="data_masking")
+def test_encrypted_in_logs(data_masker, basic_handler_fn, basic_handler_fn_arn):
```
heitorlessa: Suggestion to help us catch other types of issues here

* Use DataMasking within the Lambda Handler to encrypt whatever event we send
* Lambda to return the event encrypted
* Within the test, try to decrypt and compare with the original value

This will also help us debug potential issues that would only occur within the Lambda runtime (e.g., lack of CPU cycles to handle it, insufficient memory, Lambda's AWS SDK provided continues to work, etc.).

It will simplify your assertions too since traversing logs require a loop here, whereas the value is ensuring DataMasking works within any Lambda function with default settings.

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
```diff
@@ -0,0 +1,68 @@
...
+
+        expected_context = provider_options.pop("encryption_context", {})
+
+        ciphertext, decryptor_header = self.client.decrypt(
+            source=ciphertext_decoded, key_provider=self.key_provider, **provider_options
+        )
+
+        for key, value in expected_context.items():
+            if decryptor_header.encryption_context.get(key) != value:
+                raise ValueError(f"Encryption Context does not match expected value for key: {key}")
```
heitorlessa: NICE!!! One idea:

I think it'd be helpful to raise a custom exception about context mismatch (+explicitness, +easier to alert on). 

I suspect `ValueError` could be confused with the type of data provided to be decrypted if one isn't carefully reading the error message.

heitorlessa (changes requested):
Loved the encryption_context workaround until Encryption SDK provides a better experience - way to go!

Made one comment to make E2E more effective, and another one about the exception raised when there is a encryption context mismatch.

seshubaws:
- [x]  Turn Provider an abstract class by subclassing ABC
- [x]  Add types for Provider's methods
- [x] Docstring _apply_action DataMasking` using Numpy style
- [x]  Type provider constructor in DataMasking
- [x] Refactor AwsEncryptionSdkProvider to make production ready
  - [x] Allow customers to set their cache capacity
  - [x]  Allow customers to set how long a cache entry will be valid
  - [x]  Allow customers to set maximum number of messages under a cache entry
  - [x]  Docstring to add more details in encrypt, decrypt, and __init__
  - [x]  Docstring to give an example how to use it (API docs)
  
 Thanks to @roger-zhangg for helping refactor the AwsEncryptionSdkProvider!

leandrodamascena:
> * [ ] E2E for AWS Encryption SDK

Hello @heitorlessa! I had a conversation with @seshubaws and we are having issues running the e2e tests. The Powertools framework for e2e builds the layer with the wrong version for the `cryptography` library. `aws-encryption-sdk` [requires](https://github.com/aws/aws-encryption-sdk-python/blob/master/requirements.txt) `cryptography >= 3.4.0` and Powertools is building the layer with `cryptography-3.3.2` and because of that the Lambda execution fails with this error:

`........{"errorMessage": "Unsupported signing algorithm info", "errorType": "NotSupportedError", "requestId": "d7fd1f32-7230-4f9d-971e-852049cd83fa"........`

I'll see how to fix this, but in the meantime, @seshubaws will continue working on other items on this list.

tests/e2e/data_masking/test_data_masking.py
seshubaws: Implemented this test and then ran into the issue described in Leandro's comment below haha. I pushed a commit with the test anyway with a note that it doesn't work yet due to the dependency issue. Should I delete the `test_encrypted_in_logs` test since I'm just testing in the lambda handler now?

seshubaws:
- [x]  Break down parametrize tests into more individual tests with related data types to ease detecting regressions
- [x] Build simple data within a test instead of setups to ease learning about the test intent (screenshot before/after)
- [x] Remove dependencies and references to ItsDangerous as it'll come in a separate PR.
- [x]  Improve UX by accepting any JSON valid data in encrypt similar to ItsDangerous - it became apparent when writing tests that the UX would be too difficult to handle negative integers etc.
  - [x] Optionally allow a custom serializer/deserializer so customers can handle more complex data types

heitorlessa:
Updating the task list to it's quicker to review. As we discussed, I'll continue to dig deeper into the cryptography dependency conflict, but ETA wise we should be able to get it merged by end of next week - we can create another PR for docs alone as it'll be long.

---


* [x] Split DataMasking with and without KMS tests
* [x] Make tests more verbose by building test data as part of tests
* [x] Stub or mock KMS Encryption SDK calls
* [x] Remove leftover from ItsDangerous
  * [x] `tests/performance/test_data_masking.py` (use `.mask` over `.encrypt/decrypt` for now)
  * [x] `pyproject.toml`
  * [x] `poetry.lock`
* [x] Remove `datamasking-all`
  * we need customers to validate when they'd want to use multiple providers to warrant it
* [x] Rename `Provider` to `BaseProvider` - same convention we use in other parts
* [x] Turn `Provider` an abstract class by subclassing `ABC`
* [x] Add types for `Provider`'s methods
* [x] Docstring `_apply_action` DataMasking` using Numpy style - [VSCode Extension with numpy style support](https://marketplace.visualstudio.com/items?itemName=njpwerner.autodocstring)
* [x] Revisit `Turn into json string so everything has quotes around it`
* [x] Simplify `DataMasking` self.provider logic with: `self.provider = provider or Provider()`
* [x] Type `provider` constructor in `DataMasking`
* [x] Revisit error handling in `_apply_action_to_fields` for string data that isn't a JSON string
* [x] Loop `for field in fields` is rewriting the loop target (`field`), consider another variable
* [x] Document `for field in fields` logic
* [ ] Revisit variable names in `for field in fields` to reduce cognitive load
* [x] Break down parametrize tests into more individual tests with related data types to ease detecting regressions
* [x] Build simple data within a test instead of setups to ease learning about the test intent (screenshot before/after)
* [x] E2E for AWS Encryption SDK
* [x] Bring `DATA_MASKING_STRING` to a local `constants.py` within Data Masking utility - we're starting to break that so we can more easily modularize it in the future
* [x] Remove dependencies and references to `ItsDangerous` as it'll come in a separate PR.
* [x] Refactor `AwsEncryptionSdkProvider` to make production ready
  * [x] Allow customers to set their cache capacity
  * [x] Allow customers to set how long a cache entry will be valid
  * [x] Allow customers to set maximum number of messages under a cache entry
  * [x] Docstring to add more details in `encrypt`, `decrypt`, and `__init__`
  * [x] Docstring to give an example how to use it (API docs)
  * [ ] Revisit use of `SingletonMeta` to address the underlying cause of a single KMS GenerateDataKey being called once (or more based on settings), for perf reasons. If it can't, we need to take into account more variables to make an instance unique (e.g., keys, cache settings, etc.)
* [x] Improve UX by accepting any JSON valid data in `encrypt` similar to ItsDangerous - it became apparent when writing tests that the UX would be too difficult to handle negative integers etc.
    * [ ] Optionally allow a custom serializer/deserializer so customers can handle more complex data types
* [ ] Fix dependency resolution conflict for `cryptography` in x86_64 systems

---

**Separate PR**


* [ ] Document what 80% of customer needs to get going
  * [ ] Terminology
  * [ ] Diagrams
  * [ ] Types of data they can use
  * [ ] Minimal infra and IAM permissions (SAM) to be able to use this feature (separate encryption from decryption Lambda)
  * [ ] Advanced scenarios like custom caching, max messages encrypted etc, reusing Encryption Provider instance elsewhere in the code (along with side effects explanation)
  * [ ] Provider section so we can add others later
  * [ ] Refer to any best practices shared by the KMS team that customers should know

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
seshubaws: So the current implementation uses the key list the user passes in as the wrapping keys (aka envelope keys), and then we have a cache for the data keys (which are auto generated by the Encryption SDK and that actually do the encrypting of data). I don't believe the Encryption SDK has a cache mechanism for the wrapping keys.

Was the Encryption SDK requesting a *new* wrapping key per `encrypt()` call? At least one wrapping key is supposed to be provided for the `encrypt` method but it doesn't have to be new each time (wrapping keys don't time out ever either). 

So I don't think we need a Singleton class unless its for performance reasons (if in the unlikely event that users instantiate a bunch of EncryptionProvider objects with the same key list)

heitorlessa: > Was the Encryption SDK requesting a new wrapping key per encrypt() call?

Yes!! That's my recollection at least. If it doesn't, then there's no need to have a Singleton. For the initial POC, I used Pydantic validator, which I suspect now it created a new instance for every field it needed to encrypt... which gave me the impression `encrypt()` method would always get a new data key.

You can use Tracer to quickly check if a new KMS API call is being made for each field you try to encrypt/decrypt.

We should be able to find that out during a load test - a Lambda function memory 128M and another with 1,769 MB (1 CPU core).

leandrodamascena:
Reopening.. GH action is crazy.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -0,0 +1,138 @@
...
+        ```
+        """
+
+        if fields is None:
+            raise ValueError("No fields specified.")
+
+        if isinstance(data, str):
+            # Parse JSON string as dictionary
+            my_dict_parsed = json.loads(data)
+        elif isinstance(data, dict):
```
seshubaws, line 134: Here we're converting any non-string keys in the data given to us into strings. I don't know if we even want to support this. If we require that all keys in the data given to us are strings, then this elif is not needed. i.e if 
```
data = {
        3: {
            "1": {"None": "hello", "four": "world"},
            4: {"33": {"5": "goodbye", "e": "world"}},
        },
    }
```
it would need to be turned into a json string to get keys `3` and `4` to be `'3'` and `'4'`

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -0,0 +1,138 @@
...
+                f"Unsupported data type for 'data' parameter. Expected a traversable type, but got {type(data)}.",
+            )
+
+        # ['a.b.c'] in ['a.b.c', 'a.x.y']
+        for nested_key in fields:
+            # Prevent overriding loop variable
+            curr_nested_key = nested_key
+
+            # If the nested_key is not a string, convert it to a string representation
+            if not isinstance(curr_nested_key, str):
```
seshubaws, line 150: This line is also not necessary if we require that the list of fields given to us will be strings, i.e `fields=['3.4']` and not `fields=[3.4]` but don't know if we want to keep this use case for some customers

justinhauer:
Just curious - is this feature close? I have a team that would love to rip out some custom data masking we've implemented, and would favor a solution to be built into power tools

seshubaws:
Artillery load test summary

```
apdex.frustrated: .............................................................. 1975
apdex.satisfied: ............................................................... 525133
apdex.tolerated: ............................................................... 210892
http.codes.200: ................................................................ 653122
http.codes.500: ................................................................ 84873
http.codes.502: ................................................................ 5
http.downloaded_bytes: ......................................................... 904799083
http.request_rate: ............................................................. 946/sec
http.requests: ................................................................. 738000
http.response_time:
  min: ......................................................................... 0
  max: ......................................................................... 5585
  median: ...................................................................... 92.8
  p95: ......................................................................... 658.6
  p99: ......................................................................... 742.6
http.responses: ................................................................ 738000
plugins.metrics-by-endpoint./Prod/nonsingleton1024.codes.200: .................. 108951
plugins.metrics-by-endpoint./Prod/nonsingleton1024.codes.500: .................. 14046
plugins.metrics-by-endpoint./Prod/nonsingleton1024.codes.502: .................. 3
plugins.metrics-by-endpoint./Prod/nonsingleton128.codes.200: ................... 108708
plugins.metrics-by-endpoint./Prod/nonsingleton128.codes.500: ................... 14292
plugins.metrics-by-endpoint./Prod/nonsingleton1769.codes.200: .................. 108786
plugins.metrics-by-endpoint./Prod/nonsingleton1769.codes.500: .................. 14214
plugins.metrics-by-endpoint./Prod/singleton1024.codes.200: ..................... 108962
plugins.metrics-by-endpoint./Prod/singleton1024.codes.500: ..................... 14038
plugins.metrics-by-endpoint./Prod/singleton128.codes.200: ...................... 108835
plugins.metrics-by-endpoint./Prod/singleton128.codes.500: ...................... 14164
plugins.metrics-by-endpoint./Prod/singleton128.codes.502: ...................... 1
plugins.metrics-by-endpoint./Prod/singleton1769.codes.200: ..................... 108880
plugins.metrics-by-endpoint./Prod/singleton1769.codes.500: ..................... 14119
plugins.metrics-by-endpoint./Prod/singleton1769.codes.502: ..................... 1
plugins.metrics-by-endpoint.response_time./Prod/nonsingleton1024:
  min: ......................................................................... 0
  max: ......................................................................... 5039
  median: ...................................................................... 89.1
  p95: ......................................................................... 144
  p99: ......................................................................... 232.8
plugins.metrics-by-endpoint.response_time./Prod/nonsingleton128:
  min: ......................................................................... 0
  max: ......................................................................... 5373
  median: ...................................................................... 561.2
  p95: ......................................................................... 713.5
  p99: ......................................................................... 1620
plugins.metrics-by-endpoint.response_time./Prod/nonsingleton1769:
  min: ......................................................................... 0
  max: ......................................................................... 1726
  median: ...................................................................... 82.3
  p95: ......................................................................... 133
  p99: ......................................................................... 183.1
plugins.metrics-by-endpoint.response_time./Prod/singleton1024:
  min: ......................................................................... 0
  max: ......................................................................... 1826
  median: ...................................................................... 89.1
  p95: ......................................................................... 144
  p99: ......................................................................... 278.7
plugins.metrics-by-endpoint.response_time./Prod/singleton128:
  min: ......................................................................... 0
  max: ......................................................................... 5585
  median: ...................................................................... 561.2
  p95: ......................................................................... 685.5
  p99: ......................................................................... 2018.7
plugins.metrics-by-endpoint.response_time./Prod/singleton1769:
  min: ......................................................................... 0
  max: ......................................................................... 5033
  median: ...................................................................... 82.3
  p95: ......................................................................... 133
  p99: ......................................................................... 194.4
vusers.completed: .............................................................. 1230
vusers.created: ................................................................ 1230
vusers.created_by_name.0: ...................................................... 1230
vusers.failed: ................................................................. 0
vusers.session_length:
  min: ......................................................................... 128163.9
  max: ......................................................................... 190631.5
  median: ...................................................................... 148798.3
  p95: ......................................................................... 164448.1
  p99: ......................................................................... 178145

Apdex score: 0.8544430894308943 (good)
```

seshubaws:
> Just curious - is this feature close? I have a team that would love to rip out some custom data masking we've implemented, and would favor a solution to be built into power tools

@justinhauer Hi Justin! Yes, this feature is coming in early October as we work through the final touches and documentation. Thanks for checking in!

heitorlessa:
> Artillery load test summary
> 
> 
> 
> ```
> 
> Summary report @ 17:45:03(-0700)
> 
> --------------------------------
> 
> 
> 
> apdex.frustrated: .............................................................. 126542
> 
> apdex.satisfied: ............................................................... 192270
> 
> apdex.tolerated: ............................................................... 49589
> 
> errors.ETIMEDOUT: .............................................................. 2
> 
> http.codes.200: ................................................................ 368399
> 
> http.codes.502: ................................................................ 2
> 
> http.downloaded_bytes: ......................................................... 508881887
> 
> http.request_rate: ............................................................. 964/sec
> 
> http.requests: ................................................................. 368403
> 
> http.response_time:
> 
>   min: ......................................................................... 0
> 
>   max: ......................................................................... 5663
> 
>   median: ...................................................................... 92.8
> 
>   p95: ......................................................................... 685.5
> 
>   p99: ......................................................................... 1107.9
> 
> http.responses: ................................................................ 368401
> 
> plugins.metrics-by-endpoint./Prod/nonsingleton128.codes.200: ................... 122798
> 
> plugins.metrics-by-endpoint./Prod/nonsingleton128.codes.502: ................... 2
> 
> plugins.metrics-by-endpoint./Prod/nonsingleton1769.codes.200: .................. 122800
> 
> plugins.metrics-by-endpoint./Prod/singleton1769.codes.200: ..................... 122801
> 
> plugins.metrics-by-endpoint.response_time./Prod/nonsingleton128:
> 
>   min: ......................................................................... 1
> 
>   max: ......................................................................... 5663
> 
>   median: ...................................................................... 572.6
> 
>   p95: ......................................................................... 804.5
> 
>   p99: ......................................................................... 1720.2
> 
> plugins.metrics-by-endpoint.response_time./Prod/nonsingleton1769:
> 
>   min: ......................................................................... 0
> 
>   max: ......................................................................... 2305
> 
>   median: ...................................................................... 85.6
> 
>   p95: ......................................................................... 198.4
> 
>   p99: ......................................................................... 432.7
> 
> plugins.metrics-by-endpoint.response_time./Prod/singleton1769:
> 
>   min: ......................................................................... 0
> 
>   max: ......................................................................... 1932
> 
>   median: ...................................................................... 82.3
> 
>   p95: ......................................................................... 183.1
> 
>   p99: ......................................................................... 340.4
> 
> vusers.completed: .............................................................. 1228
> 
> vusers.created: ................................................................ 1230
> 
> vusers.created_by_name.0: ...................................................... 1230
> 
> vusers.failed: ................................................................. 2
> 
> vusers.session_length:
> 
>   min: ......................................................................... 75811.8
> 
>   max: ......................................................................... 105458
> 
>   median: ...................................................................... 86710.4
> 
>   p95: ......................................................................... 93932.5
> 
>   p99: ......................................................................... 97766.1
> 
> 
> 
> Apdex score: 0.5892071411315387 (poor)
> 
> ```

Is that for the 128M function right?

seshubaws:
> Is that for the 128M function right?

The first one is for singleton128. The rest are for singleton1769, nonsingleton128, and nonsingleton1769, you can see the endpoints for each. I used singleton128 as the target so I guess it doesn't show that endpoint?

heitorlessa:
Ahhh i missed these lines that separate those results - it was wrapped on
GitHub mobile app ;)

plugins.metrics-by-endpoint.response_time./Prod/singleton1769

seshubaws:
Traces:

Singleton 128MB
Screenshot 2023-09-22 at 2 36 26 PM


Singleton 1024MB
Screenshot 2023-09-22 at 2 29 45 PM


Singleton 1769MB
Screenshot 2023-09-22 at 2 37 37 PM


Non singleton 128MB
Screenshot 2023-09-22 at 2 34 38 PM


Non singleton 1024MB
Screenshot 2023-09-22 at 2 38 31 PM


Non singleton 1769MB
Screenshot 2023-09-22 at 2 40 12 PM

aws_lambda_powertools/utilities/data_masking/base.py
seshubaws: Added thorough load tests for this [here](https://github.com/aws-powertools/powertools-lambda-python/pull/2197#issuecomment-1730571597)

aws_lambda_powertools/utilities/data_masking/providers/aws_encryption_sdk.py
seshubaws: Looking at the load testing numbers, here are the differences in latencies using a singleton vs nonsingleton:
```
128MB
- nonsingleton max is 200ms faster
- singleton p95 is 20ms faster
- nonsingleton p99 is 400ms faster

1024MB
- singleton max is 3000ms faster
- nonsingleton p99 is 40ms faster

1769MB
- nonsingleton max is 3000ms faster
- nonsingleton p99 is 10ms faster
```
It seems on the whole not using a singleton is faster, so I will just get rid of the Singleton class altogether since the KMS API calls are the same for with and without singleton.

leandrodamascena:
> Just curious - is this feature close? I have a team that would love to rip out some custom data masking we've implemented, and would favor a solution to be built into power tools

Hello @justinhauer! We are merging this PR today and hope to release this feature in our next release - next week. We still have some work to do, mainly on the documentation side, but this is one of the main Powertools news this year and I'm happy that you already have some cases to use it! As soon as we officially release this utility, I will ping you.

Thanks

leandrodamascena (approved):
Hi @seshubaws, I have no words to describe the amount of great work you've done here. Implementing a utility that handles encryption, keys, key material, and a complex SDK is not simple at all!

I'm approving this PR and let start writing the documentation.  We will probably detect some user experience improvements, but they are more cosmetic changes than logical ones.

image

heitorlessa:
This is AWESOME - echoing what Leandro said about the effort. I know we'll make more internal changes to get to an unwritten standard yet but that shouldn't be a blocking factor.

A few major pieces we should handle in a separate PR before we release this feature:

* **Lambda Layer**. We need to get customers feedback on whether we include Encryption SDK as part of the Layer. It might increase ~6M in the final Layer, after removing AWS SDK and such. We need a comment in the RFC to drive customers and the community to discuss whether we include it or not. I favour including it as long as we stay within 10M boundary given how hard it is to bring these type of mixed languages dependencies in customers builds (default goes ~100M uncompressed).
* **Dataclasses/Classes and Pydantic**. We need to add support for dataclasses and pydantic models while keeping the return type accurate (generics!). If a customer uses a dataclass/pydantic model along with fields, we should return the same instance with the values encrypted/masked - tremendous value here in UX while helping customers avoid perf pitfall of deserialization+serialization overhead.
* **Update RFC**. We should update the linked RFC with the latest UX and once again after the docs PR is complete. This will serve as a future document for other languages and any discussions around it too.

heitorlessa:
Forgot two things..

@leandrodamascena we're gonna need a new label for this new feature to track future changes (changelog, issues, effort, etc.), and update our issue template as well as workflow that labels PRs based on title ;)

@seshubaws you took one of the hardest roadmap items and I can't begin to appreciate your resiliency throughout this **long** review cycle. I hope you learned a thing or two along the way too <3

Merging..

heitorlessa (approved):

justinhauer:
> > Just curious - is this feature close? I have a team that would love to rip out some custom data masking we've implemented, and would favor a solution to be built into power tools
> 
> Hello @justinhauer! We are merging this PR today and hope to release this feature in our next release - next week. We still have some work to do, mainly on the documentation side, but this is one of the main Powertools news this year and I'm happy that you already have some cases to use it! As soon as we officially release this utility, I will ping you.
> 
> Thanks

This is great, thank you and others for all the great work in this project.

Issue #2185, opened 2023-05-01
Maintenance: Migrate to new PyPi Trusted Publishers 


heitorlessa:
### Why is this needed?

PyPi just recently added support for "[Trusted Publishers](https://docs.pypi.org/trusted-publishers/)" to improve security posture by accepting OIDC tokens emitted from CIs also acting as IdP instead of PyPi tokens.

GitHub Actions natively supports it: https://docs.pypi.org/trusted-publishers/

### Which area does this relate to?

Automation

### Solution

Add a new trusted publisher in PyPi associating this org+repo+gh env+gh workflow, and updating our release process do ditch PyPi token and use the new process.

<img width="445" alt="image" src="https://user-images.githubusercontent.com/3340292/235471275-4042520c-3681-49dc-aa69-b860049eb340.png">

### Acknowledgment

- [X] This request meets [Lambda Powertools Tenets](https://awslabs.github.io/aws-lambda-powertools-python/latest/#tenets)
- [ ] Should this be considered in other Lambda Powertools languages? i.e. [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

heitorlessa:
Tasks:

* [x] Add new Trusted Publisher in PyPi
* [ ] Create new Makefile target: `release-build`
* [ ] Replace `release-prod` with `release-build`
* [ ] Use new GH Action with a SHA pinned version
* [ ] Use it for new release
* [ ] Remove PyPi Token from GH Secrets if release was successful

PR #2198, opened 2023-05-03
chore(ci): use new pypi trusted publisher for increase security

heitorlessa:
**Issue number:** #2185 

## Summary

Updates our release to use new PyPi Trusted Publisher to exchange GHA OIDC token for a valid PyPi token for publishing.

### Changes

* [x] Update release steps to use new GH Action for PyPi publishing with OIDC
* [x] Remove PyPi token
* [x] Add this repository as a trusted publisher in PyPi prod and PyPi test
* [x] Split build and release jobs for least privilege
* [x] Update release process in-workflow docs

### User experience

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -52,6 +52,7 @@ jobs:
     runs-on: aws-lambda-powertools_ubuntu-latest_4-core
     permissions:
       contents: write
+      id-token: write # OIDC for PyPi Trusted Publisher feature
     outputs:
       RELEASE_VERSION: ${{ steps.release_version.outputs.RELEASE_VERSION }}
     env:
@@ -95,16 +96,13 @@ jobs:
       # March 1st: PyPi test is under maintenance....
       # - name: Upload to PyPi test
       #   if: ${{ !inputs.skip_pypi }}
-      #   run: make release-test
-      #   env:
-      #     PYPI_USERNAME: __token__
-      #     PYPI_TEST_TOKEN: ${{ secrets.PYPI_TEST_TOKEN }}
+      #   uses: pypa/gh-action-pypi-publish@0bf742be3ebe032c25dd15117957dc15d0cfc38d # v1.8.5
+      #   with:
+      #     repository-url: https://test.pypi.org/legacy/
+
       - name: Upload to PyPi prod
         if: ${{ !inputs.skip_pypi }}
-        run: make release-prod
-        env:
-          PYPI_USERNAME: __token__
-          PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
+        uses: pypa/gh-action-pypi-publish@0bf742be3ebe032c25dd15117957dc15d0cfc38d # v1.8.5
       - name: Update version in trunk
         if: steps.versioning.outcome == 'success'
         run: |
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -98,6 +98,12 @@ jobs:
         uses: actions/cache/save@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
         with:
           path: dist/
+          # NOTE: cache key uses a hash of (Runner OS + Version to be released + Deps)
+          # since a new release might not change a dependency but version
+          # otherwise we might accidentally reuse a previously cached artifact for a newer release
+          # the reason we don't add pyproject.toml here is to avoid racing conditions
+          # where git checkout might happen too fast and doesn't pick up the latest
+          # and also future-proof for when we switch to protected branch and update via PR
           key: ${{ runner.os}}-${{ env.RELEASE_VERSION }}-${{ hashFiles('**/poetry.lock')}}
 
       - name: Update version in trunk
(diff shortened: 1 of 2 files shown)


heitorlessa:
cc @rubenfonseca @leandrodamascena - just pending docs but please let me know if you've got any questions.

rubenfonseca:
Looking at it now.

.github/workflows/release.yml
```diff
@@ -90,21 +91,21 @@ jobs:
...
+      - name: Cache release artifact
+        id: cache-release-build
+        uses: actions/cache/save@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+        with:
+          path: dist/
+          # NOTE: cache key uses a hash of (Runner OS + Version to be released + Deps)
+          # since a new release might not change a dependency but version
+          # otherwise we might accidentally reuse a previously cached artifact for a newer release
+          # the reason we don't add pyproject.toml here is to avoid racing conditions
+          # where git checkout might happen too fast and doesn't pick up the latest
```
rubenfonseca:
```suggestion
          # otherwise we might accidentally reuse a previously cached artifact for a newer release.
          # The reason we don't add pyproject.toml here is to avoid racing conditions
          # where git checkout might happen too fast and doesn't pick up the latest version
```

.github/workflows/release.yml
```diff
@@ -90,21 +91,21 @@ jobs:
...
+          # and also future-proof for when we switch to protected branch and update via PR
+          key: ${{ runner.os}}-${{ env.RELEASE_VERSION }}-${{ hashFiles('**/poetry.lock')}}
```
rubenfonseca:
```suggestion
          key: ${{ runner.os }}-${{ env.RELEASE_VERSION }}-${{ hashFiles('**/poetry.lock') }}
```

.github/workflows/release.yml
```diff
@@ -115,6 +116,36 @@ jobs:
...
+    env:
+      RELEASE_VERSION: ${{ needs.build.outputs.RELEASE_VERSION }}
+    steps:
+      - uses: actions/checkout@v3
+      - name: Restore release artifact from cache
+        id: restore-release-build
+        uses: actions/cache/restore@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+        with:
+          path: dist/
+          key: ${{ runner.os}}-${{ env.RELEASE_VERSION }}-${{ hashFiles('**/poetry.lock')}}
```
rubenfonseca:
```suggestion
          key: ${{ runner.os }}-${{ env.RELEASE_VERSION }}-${{ hashFiles('**/poetry.lock') }}
```

.github/workflows/release.yml
```diff
@@ -115,6 +116,36 @@ jobs:
...
+        uses: pypa/gh-action-pypi-publish@0bf742be3ebe032c25dd15117957dc15d0cfc38d # v1.8.5
+        with:
+          repository-url: https://test.pypi.org/legacy/
+
+      # March 1st: PyPi test is under maintenance....
+      # - name: Upload to PyPi test
+      #   if: ${{ !inputs.skip_pypi }}
+      #   uses: pypa/gh-action-pypi-publish@0bf742be3ebe032c25dd15117957dc15d0cfc38d # v1.8.5
+      #   with:
+      #     repository-url: https://test.pypi.org/legacy/
```
rubenfonseca, line 147: Should we uncoment this now and put it in the correct place?

.github/workflows/release.yml
```diff
@@ -124,27 +155,27 @@ jobs:
   # NOTE: Watch out for the depth limit of 4 nested workflow_calls.
   # publish_layer -> publish_v2_layer -> reusable_deploy_v2_layer_stack -> reusable_update_v2_layer_arn_docs
   publish_layer:
-    needs: release
+    needs: [build, release]
```
rubenfonseca, line 158: Maybe it's more correct to have it only depending on release (which already depends on build

```suggestion
    needs: [release]
```)

rubenfonseca (changes requested):
Some small things, but amazing job!

.github/workflows/release.yml
heitorlessa: I'm in two minds about this given how often it fails -- should we perhaps not fail the entire release IF we can't push it to Test?

Put another way, what safety do we gain from publishing to the Test repository?

rubenfonseca: Agree, I didn't get that it was failing constantly. Then I agree to leave it as it is!

.github/workflows/release.yml
heitorlessa: nooo, it fails because of the `{{ needs.build.outputs.RELEASE_VERSION }}`, since the `build` object will be empty. Unless you want me to declare RELEASE_VERSION in the `release` job so it cascades down?

rubenfonseca: ahahah I knew there was a reason :P It's good as it is then!

heitorlessa: you might ask how I know hehehe... our pre-commit hook caught this bug earlier :) 

image

rubenfonseca (approved):
LGTM!

PR #2200, opened 2023-05-03
feat(event_source): support custom json_deserializer; add json_body in SQSEvent

### LINKED ISSUE #2199: Feature request: support custom JSON deserializer in event source classes
### Use case

In some event classes (such as SQS, Kafka, and others), certain string fields (such as `body`) can be converted to JSON using the `json.loads` function. However, it is possible that the customer may have created the message with a JSON serialization method that differs from the default `json` class. To cover these scenarios, we need to provide customers with the flexibility to use a custom deserialization function of their choice.

This is important because by allowing customers to use their preferred deserialization function, we can ensure that their messages are properly deserialized into the intended format.

The default if nothing is informed will continue to be `json.loads`

### Solution/User Experience

To address the issue, we can modify the DictWrapper class to introduce a new parameter. This parameter would allow customers to provide their preferred deserialization function for converting string fields to JSON.

```python
class DictWrapper(Mapping):
    """Provides a single read only access to a wrapper dict"""

    def __init__(self, data: Dict[str, Any], json_deserializer: Optional[Callable] = None):
        self._data = data
        self._json_data: Optional[Any] = None
        self._json_deserializer = json_deserializer or json.loads
```

### Alternative solutions

_No response_

### Acknowledgment

- [ ] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)

leandrodamascena:
**Issue number:** #2199 

### Changes

This PR aims to add support for a custom `json` deserialization method in the event source classes.

### User experience

**BEFORE**

```python
from aws_lambda_powertools.utilities.data_classes import event_source, SQSEvent
import json 

def lambda_handler(event: SQSEvent, context):
    parsed = SQSEvent(data=event)
    # Multiple records can be delivered in a single event
    for record in parsed.records:
        print(json.loads(record.body))
```

**AFTER**

```python
from aws_lambda_powertools.utilities.data_classes import event_source, SQSEvent
import ujson

def lambda_handler(event: SQSEvent, context):
    
    parsed = SQSEvent(data = event, json_deserializer=ujson.loads)
    # Multiple records can be delivered in a single event
    for record in parsed.records:
        print(record.json_body)
```

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -1,7 +1,7 @@
 import base64
 import json
 from collections.abc import Mapping
-from typing import Any, Dict, Iterator, List, Optional
+from typing import Any, Callable, Dict, Iterator, List, Optional
 
 from aws_lambda_powertools.shared.headers_serializer import BaseHeadersSerializer
 
@@ -9,9 +9,10 @@ from aws_lambda_powertools.shared.headers_serializer import BaseHeadersSerialize
 class DictWrapper(Mapping):
     """Provides a single read only access to a wrapper dict"""
 
-    def __init__(self, data: Dict[str, Any]):
+    def __init__(self, data: Dict[str, Any], json_deserializer: Optional[Callable] = None):
         self._data = data
         self._json_data: Optional[Any] = None
+        self._json_deserializer = json_deserializer or json.loads
 
     def __getitem__(self, key: str) -> Any:
         return self._data[key]
@@ -122,7 +123,7 @@ class BaseProxyEvent(DictWrapper):
     def json_body(self) -> Any:
         """Parses the submitted body as json"""
         if self._json_data is None:
-            self._json_data = json.loads(self.decoded_body)
+            self._json_data = self._json_deserializer(self.decoded_body)
         return self._json_data
 
     @property
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -10,6 +10,15 @@ class DictWrapper(Mapping):
     """Provides a single read only access to a wrapper dict"""
 
     def __init__(self, data: Dict[str, Any], json_deserializer: Optional[Callable] = None):
+        """
+        Parameters
+        ----------
+        data : Dict[str, Any]
+            Lambda Event Source Event payload
+        json_deserializer : Callable, optional
+            function to deserialize `str`, `bytes`, bytearray` containing a JSON document to a Python `obj`,
+            by default json.loads
+        """
         self._data = data
         self._json_data: Optional[Any] = None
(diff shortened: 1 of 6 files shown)


aws_lambda_powertools/utilities/data_classes/sqs_event.py
```diff
@@ -103,6 +103,16 @@ def body(self) -> str:
...
 
+    @property
+    def json_body(self) -> Dict:
+        """Parses the submitted body as json"""
+        try:
+            if self._json_data is None:
+                self._json_data = self._json_deserializer(self["body"])
+            return self._json_data
+        except Exception:
+            return self["body"]
```
leandrodamascena: The reason I add a `try/except` block here is because it could be a simple string that cannot be deserialized as `json`.
Should we consider this for the other `json_body` fields in Kinesis/Kafka/APIGW? Should we return the plain body if the deserialization fails? 
I would like to hear from you.

heitorlessa: We should fail hard here to give the caller a chance to handle this error correctly. 

Reasoning in a mental model fashion:

* Do we offer a non-JSON body property e.g., `body`?
    * If so, `json_body` should fail if it can't deserialize it
* If the customer received a poison pill record, a non-JSON string, will they be given the chance to handle a JSON deserialization error?
    * If we return `self["body"]`, it can result to a totally different error we can't even anticipate, e.g., `json_body["blah"]`

As a rule of thumb, you'd want to propagate the error as close to the actual problem so they instinctively know how to handle it. If we catch all, we're masking the actual error and leading them to an unknown error later on.

> tiny note: until Python 3.11, try/catch has a performance overhead even if it's tiny, so we can also improve here by letting them do it (not accidentally twice).

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -1,17 +1,18 @@
...
+from typing import Any, Callable, Dict, Iterator, List, Optional
 
 from aws_lambda_powertools.shared.headers_serializer import BaseHeadersSerializer
 
 
 class DictWrapper(Mapping):
     """Provides a single read only access to a wrapper dict"""
 
-    def __init__(self, data: Dict[str, Any]):
+    def __init__(self, data: Dict[str, Any], json_deserializer: Optional[Callable] = None):
```
heitorlessa, line 12: docstring ;) we apparently missed back in the days for DictWrapper, but given the json_deserializer will be added to all data classes, we can propagate the docstring to everyone benefits from it.

pushing that change

image

heitorlessa (approved):

PR #2239, opened 2023-05-11
chore(ci): bump package version after release via pull request

heitorlessa:
**Issue number:** #2210

**DEPENDENCY**: https://github.com/awslabs/aws-lambda-powertools-python/pull/2238

## Summary

Use new create-pr action introduced in https://github.com/awslabs/aws-lambda-powertools-python/pull/2238 to bump package version via PR instead of pushing directly to `develop` as part of the release.

### Changes

* [x] Create new job `bump_version` with scoped permissions to do exactly that
* [x] Lower permission in `build` job as it no longer needs to push to `develop`
* [x] Remove `changelog` job as it's now handled perfectly in every push via PR - introduced in https://github.com/awslabs/aws-lambda-powertools-python/pull/2238

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


.github/workflows/release.yml
```diff
@@ -163,6 +140,28 @@ jobs:
       latest_published_version: ${{ needs.build.outputs.RELEASE_VERSION }}
       pre_release: ${{ inputs.pre_release }}
 
+  bump_version:
+    needs: [build, release]
+    permissions:
+      contents: write
+      pull-requests: write
```
leandrodamascena: Are these permissions not conflicting with each other? I think the `pull-requests: write` permission is enough for that, no? The `contents: write` permission seems to be much broader and allows things we don't need.

References:
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#pull-requests
https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#contents

leandrodamascena (changes requested):
I just left a comment about permissions, but for everything else, this looks good!

.github/workflows/release.yml
heitorlessa: Not in this case :) 

**`create-pr`** action needs to create a temporary branch to push these staged files (`pyproject.toml`) - hence `contents: write`. Then, creates a PR against `develop` using our temporary branch as the HEAD - hence `pull-requests: write`

.github/workflows/release.yml
```diff
@@ -163,6 +140,28 @@ jobs:
       latest_published_version: ${{ needs.build.outputs.RELEASE_VERSION }}
       pre_release: ${{ inputs.pre_release }}
 
+  bump_version:
+    needs: [build, release]
+    permissions:
+      contents: write
```
heitorlessa:
```suggestion
      contents: write  # create-pr action creates a temporary branch
```

heitorlessa:
```suggestion
      contents: write  # create-pr action creates a temporary branch
```

.github/workflows/release.yml
heitorlessa:
```suggestion
      pull-requests: write # create-pr action creates a PR using the temporary branch
```

leandrodamascena (approved):
LGTM!

PR #2277, opened 2023-05-16
feat(docs): Move docs to S3

sthulb:
**Issue number:**

## Summary
As the project changes, we want to host our documentation on our own domain, this frees us of GitHub Pages and related outages. After this change our docs will be hosted in two locations GH Pages and on docs.powertools.aws.dev, this is a middle step for further changes later.

### Changes

This pull request adds some extra steps to the release docs process, it copies the generated docs to an S3 bucket.

### User experience

Docs will be available in S3 and GH Pages

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/.github/workflows/rebuild_latest_docs.yml
+++ b/.github/workflows/rebuild_latest_docs.yml
@@ -14,11 +14,15 @@ on:
         default: "2.0.0"
         required: true
 
+permissions:
+  id-token: write
+
 jobs:
   release-docs:
     permissions:
       contents: write
       pages: write
+      id-token: write
     uses: ./.github/workflows/reusable_publish_docs.yml
     with:
       version: ${{ inputs.latest_published_version }}
--- a/.github/workflows/on_push_docs.yml
+++ b/.github/workflows/on_push_docs.yml
@@ -10,11 +10,15 @@ on:
       - "examples/**"
       - "CHANGELOG.md"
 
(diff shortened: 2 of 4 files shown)


.github/workflows/rebuild_latest_docs.yml
```diff
@@ -14,11 +14,15 @@ on:
         default: "2.0.0"
         required: true
 
+permissions:
+  id-token: write
+
```
leandrodamascena, line 19: Simon, we are trying to keep the permissions very granular in our worflows and I don't think this is necessary as you are adding this permission on the `release-docs` job. Have you tried without it?

leandrodamascena (changes requested):

.github/workflows/rebuild_latest_docs.yml
sthulb: It seems that if you don't have id-token: write on the caller, it won't pass it down – I kept getting a `id-token` should be none error

leandrodamascena: Thank you for the explanation, Simon! In the future we'll refactor this workflows and we can review this.

leandrodamascena (approved):
Great work, Simon! APPROVED!

PR #2279, opened 2023-05-16
feat(event_handler): allow multiple CORS origins

### LINKED ISSUE #1006: Support multiple Allowed Origins in CORS

rubenfonseca:
**Issue number:** #1006

### Changes

This PR adds support for additional origins for CORS. It also changes the behavior to never return Origin information outside of a CORS request.

### User experience

carbon (8)

**NOTE:** this PR changes the default behavior when a non-CORS request is sent. Previously, we would return the configured Origin as part of the response. Now we only do it if there's an Origin in the request and it matches one of the allowed origins. Should we do something about this?

## Checklist

- [x] Changes are documented
- [x] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


docs/core/event_handler/api_gateway.md
```diff
@@ -331,7 +335,7 @@ You can use the `Response` class to have full control over the response. For exa
 
 === "fine_grained_responses.py"
 
-    ```python hl_lines="9 28-32"
+    ```python hl_lines="9 29-35"
```
rubenfonseca, line 351: The highlight was wrong.

docs/core/event_handler/api_gateway.md
```diff
@@ -310,6 +311,9 @@ For convenience, these are the default values when using `CORSConfig` to enable
 ???+ warning
     Always configure `allow_origin` when using in production.
 
+???+ tip "Multiple allowed origins?"
+    If you require multiple allowed origins, pass the additional origins using the `extra_origins` key.
+
```
leandrodamascena: Ruben, please add the `extra_origins` field in the table below.

How about adding a new tab with an example to show how to use this new field? What do you think?

docs/core/event_handler/api_gateway.md
leandrodamascena: Good catch!

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -128,16 +132,29 @@ def __init__(
...
         """Builds the configured Access-Control http headers"""
+
+        # If there's no Origin, don't add any CORS headers
+        if not origin:
+            return {}
+
+        # If the origin doesn't match any of the allowed origins, and we don't allow all origins ("*"),
+        # don't add any CORS headers
+        if origin not in self.allowed_origins and "*" not in self.allowed_origins:
+            return {}
```
leandrodamascena: Woww! :heart: :clap: :clap:

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -128,16 +132,29 @@ def __init__(
         allow_credentials: bool
             A boolean value that sets the value of `Access-Control-Allow-Credentials`
         """
-        self.allow_origin = allow_origin
```
leandrodamascena, line 131: I saw that you needed to change the test and I'm Just wondering if it doesn't make sense to keep this field for compatibility issues. I'm not sure if anyone has used this in their local tests and we can break something or it's better to ask for forgiveness than permission.

For me we can remove that but that's just a question.

leandrodamascena (changes requested):
I tested this with several scenarios, resolvers and origins and in all of them it worked perfectly! You are truly amazing, Ruben!

I made some comments and after you answer them I think we are ready to merge!!

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -84,6 +84,7 @@ def with_cors():
 
     cors_config = CORSConfig(
         allow_origin="https://wwww.example.com/",
+        extra_origins=["https://www1.example.com/"],
```
heitorlessa: Try another distinct origin like maybe dev.example.com so it's easier to parse the differences

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: Yup, since it was public attr we dunno how many people took a dependency on that to workaround the fact we didn't support multiple Origins.

Add a note so we can remove in v3 (We need that breaking change doc)

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -128,16 +132,29 @@ def __init__(
...
+        self.allowed_origins = [allow_origin]
+        if extra_origins:
+            self.allowed_origins.extend(extra_origins)
         self.allow_headers = set(self._REQUIRED_HEADERS + (allow_headers or []))
         self.expose_headers = expose_headers or []
         self.max_age = max_age
         self.allow_credentials = allow_credentials
 
-    def to_dict(self) -> Dict[str, str]:
+    def to_dict(self, origin: Optional[str]) -> Dict[str, str]:
```
heitorlessa, line 143: origin: str = ""?

Annotation says it's optional but we are not setting a default value. 

You can drop the Optional (None), and simply set to an empty str

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -207,9 +224,9 @@ def __init__(self, response: Response, route: Optional[Route] = None):
         self.response = response
         self.route = route
 
-    def _add_cors(self, cors: CORSConfig):
+    def _add_cors(self, event: BaseProxyEvent, cors: CORSConfig):
         """Update headers to include the configured Access-Control headers"""
-        self.response.headers.update(cors.to_dict())
+        self.response.headers.update(cors.to_dict(event.get_header_value("Origin")))
```
heitorlessa, line 229: Hard to see in mobile but perhaps at this point the self.current_event is already populated so you wouldn't need to expand the signature to get the event.

Unless you want to make it more explicit or future proof to work in containers next year which is sensible too

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -644,7 +661,7 @@ def _not_found(self, method: str) -> ResponseBuilder:
         headers: Dict[str, Union[str, List[str]]] = {}
         if self._cors:
             logger.debug("CORS is enabled, updating headers.")
-            headers.update(self._cors.to_dict())
+            headers.update(self._cors.to_dict(self.current_event.get_header_value("Origin")))
```
heitorlessa, line 664: Same as before

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -113,7 +113,7 @@ def get_header_value(
 class BaseProxyEvent(DictWrapper):
     @property
     def headers(self) -> Dict[str, str]:
-        return self["headers"]
+        return self.get("headers") or {}
```
heitorlessa, line 116: Nice!!!

docs/core/event_handler/api_gateway.md
```diff
@@ -280,7 +280,8 @@ To address this API Gateway behavior, we use `strip_prefixes` parameter to accou
 
 You can configure CORS at the `APIGatewayRestResolver` constructor via `cors` parameter using the `CORSConfig` class.
 
-This will ensure that CORS headers are always returned as part of the response when your functions match the path invoked.
+This will ensure that CORS headers are returned as part of the response when your functions match the path invoked and the `Origin`
```
heitorlessa, line 283: 🫶

docs/core/event_handler/api_gateway.md
heitorlessa: Maybe I'm dizzy in the taxi but this reads odd, perhaps what you meant was:

Multiple origins?

If you need to allow multiple origins ...

As in, Allow-Origins is an explicit CORS terminology, but here we can be more flexible in wording

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -533,6 +533,36 @@ def rest_func() -> Response:
...
+    @app.get("/cors")
+    def get_with_cors():
+        return {}
+
+    # WHEN calling the event handler with the correct Origin
+    event = {"path": "/cors", "httpMethod": "GET", "headers": {"Origin": "https://origin3"}}
+    result = app(event, None)
+
+    # THEN routes by default return the custom cors headers
+    assert "multiValueHeaders" in result
```
heitorlessa: Do we need to assert it? It'd fail the next assertions either way

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -533,6 +533,36 @@ def rest_func() -> Response:
...
+    headers = result["multiValueHeaders"]
+    assert headers["Content-Type"] == [content_types.APPLICATION_JSON]
+    assert headers["Access-Control-Allow-Origin"] == ["https://origin3"]
+
+    # WHEN calling the event handler with the wrong origin
+    event = {"path": "/cors", "httpMethod": "GET", "headers": {"Origin": "https://wrong.origin"}}
+    result = app(event, None)
+
+    # THEN routes by default return the custom cors headers
+    assert "multiValueHeaders" in result
```
heitorlessa: Do we need to assert it? It'd fail the next assertions either way

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -660,7 +690,7 @@ def custom_method():
         ...
 
     # WHEN calling the handler
-    result = app({"path": "/some-call", "httpMethod": "OPTIONS"}, None)
+    result = app({"path": "/some-call", "httpMethod": "OPTIONS", "headers": {"Origin": "https://example.org"}}, None)
```
heitorlessa: nitpick: this is quite long now, perhaps a variable in the given section so it's easier to understand the intent of the test

heitorlessa (commented):
Some tiny suggestions but looks great!

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: As we've agreed on a call, we're going to:

- keep the removal of the undocumented property, and do a quick rollback if anyone is impacted 
- prefix the new variables with "_" so we can start signaling private properties

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: Does this mean I would have to compare `if origin == ""` later on instead of `not origin`? Doesn't it sound wrong?

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: I prefer to make it explicit, and since this a private method, I think it's ok to change the signature. Is that ok?

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: TIL `not` will eventually translate to `__bool__`, so it works for empty strings too

rubenfonseca: I've changed my mind again and I believe it's better to keep the Optional. Reason is, the caller will try to fetch an Origin from the headers. The Origin is not always present, so `mypy` would complain that I can't pass an Optional to the `to_dict`. So I think in this case the Optional makes sense.

rubenfonseca:
As promised, added E2E tests.

leandrodamascena (approved):
I ran the end2end test on this branch and everything is working as expected, Ruben! There are some random bugs at specific runtimes, but they are known bugs and nothing related to this change. - https://github.com/awslabs/aws-lambda-powertools-python/actions/runs/5006899451/jobs/8972788300

One more amazing work, dude!!!

PR #2301, opened 2023-05-22
chore(ci): source code tampering protection for release

heitorlessa:
**Issue number:** #2206

## Summary

Implements mechanism to seal source code, further break `build` job, further isolate `release` job with binaries only, and fixes the Git Tag creation process.

### Changes

- [x] Split quality check and build jobs
- [x] Create integrity check
- [x] Create a job to seal source code after bumping version
- [x] Bump version and seal to prevent further boilerplate in code and workarounds
- [x] Cache-artifact
- [x] Restore-artifact and ensure it overwrites any checked out code
- [x] Pin down Poetry to prevent any code tampering on build + publishing
- [x] Update `bump_version` and `create_tag` to use sealed source code with version bumped
- [x] Replace subsequent checkouts to use RELEASE_COMMIT and restore artifact
- [x] Document every job with its details
- [x] Scale down Runners to use `ubuntu-latest` as we no longer need those 4 cores (savings!!)
- [x] Lower permissions further as jobs are isolated
- [x] Add simple debug to display `pyproject.toml` in case sealed source code is incorrect
- [x] Forcefully fail release if sealed source code hasn't been found or uploaded correctly
- [x] Test new release
- [x] Test whether tag created contains latest source code with version bumped
- [x] Document overall process at the top
- [x] Verify integrity hash (source and build) in key steps

**Additional fixes**

- [x] bump_version needs to install poetry
- [x] Fix `create_tag` as it needs to add staged changes before pushing... otherwise it pushes the current commit
- [x] No longer need custom build machines for build/release.. but layer only

**Threats to address**

- [x] **Checking out newer code during the release process**
  - Forcefully use `RELEASE_COMMIT` during checkout
  - Verifies hash throughout jobs (source and build)
- [x] **Source code tampering from dependencies or workflows**
  - `seal` job to hash and reuse artifact throughout release
  - `pipx install git+<url+SHA>` for poetry to prevent rogue package manager
- [x] **Tamper build artifact during release**
  - `release-<build-hash>` for traceability and immutability
  - `pipx install git+<url+SHA>` for poetry to prevent rogue package manager
  - `release` job doesn't install any dependencies but a pinned `poetry`
- [x] **Tamper git tag during release**
  - `seal` job to hash and reuse artifact
  - No dependencies installed
  - Forcefully add `pyproject.toml` that changed in `seal` to our local git tree, then push that with the new tag only. This also solves the data race condition where PR is only merged later and our tag uses an older version.
- [x] **Push modified code to repository via 3rd-party actions / dependencies**
  - Protected branch enabled in trunk
  - All source code modifications generate PR and require maintainer's review
  - Least privilege permissions are enforced per job
  - For Git tags, source code tampering check is also enforced
- [x] **Secret ex-filtration 3rd-party action / dependencies**
  - GitHub Environment Secrets are used over standard Secrets
  - Encrypted secrets are only available for `release` environment
  - `release` environment requires manual approval before proceeding
- [x] **Release credentials overtake**
  - Setup PyPi Trusted Publisher
  - Release credentials are temporary and only issued to what's configured in PyPi Trusted Publisher (org+repo+workflow+environment name)

### User experience

## Checklist

* [X] Changes are documented
* [X] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/.github/actions/download-artifact/action.yml
+++ b/.github/actions/download-artifact/action.yml
@@ -38,7 +38,7 @@ runs:
   using: composite
   steps:
     - name: Download artifacts
-      uses: actions/download-artifact@cbed621e49e4c01b044d60f6c80ea4ed6328b281 # v2.1.1
+      uses: actions/download-artifact@9bc31d5ccc31df68ecc42ccf4149144866c47d8a # v3.0.2
       with:
         name: ${{ inputs.name }}
         path: ${{ inputs.path }}
--- a/.github/actions/download-artifact/action.yml
+++ b/.github/actions/download-artifact/action.yml
@@ -1,0 +1,58 @@
+name: Download artifact
+description: Wrapper around GitHub's official action, with additional extraction before download
+
+# PROCESS
+#
+# 1. Downloads artifact using actions/download-artifact action
+# 2. Extracts and overwrites tarball previously uploaded
+# 3. Remove archive after extraction
+
+# NOTES
+#
+# Upload-artifact and download-artifact takes ~2m40s to upload 8MB
+# so this is custom action cuts down the entire operation to 1s
+# by uploading/extracting a tarball while relying on the official upload-artifact/download-artifact actions
+#
+
+# USAGE
+#
+# NOTE: Meant to be used with ./.github/actions/upload-artifact
+#
+# - name: Restore sealed source code
+# uses: ./.github/actions/download-artifact
+# with:
+#   name: ${{ needs.seal.outputs.INTEGRITY_HASH }}
+#   path: .
+
+# https://github.com/actions/download-artifact/blob/main/action.yml
+inputs:
+  name:
+    description: Artifact name
+    required: true
+  path:
+    description: Destination path. By default, it will download to the current working directory.
+    required: false
+    default: .
+
+runs:
+  using: composite
+  steps:
+    - name: Download artifacts
+      uses: actions/download-artifact@cbed621e49e4c01b044d60f6c80ea4ed6328b281 # v2.1.1
+      with:
+        name: ${{ inputs.name }}
+        path: ${{ inputs.path }}
+
+    - name: Extract artifacts
+      run: tar -xvf "${ARCHIVE}"
+      env:
+        ARCHIVE: ${{ inputs.name }}.tar
+      shell: bash
+      working-directory: ${{ inputs.path }}
+
+    - name: Remove archive
+      run: rm -f "${ARCHIVE}"
+      env:
+        ARCHIVE: ${{ inputs.name }}.tar
+      shell: bash
+      working-directory: ${{ inputs.path }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -205,7 +205,7 @@ jobs:
       - name: Upload build artifact
         uses: ./.github/actions/upload-artifact
         with:
-          name: build-${{ steps.integrity.outputs.INTEGRITY_HASH}}
+          name: build-${{ steps.integrity.outputs.BUILD_INTEGRITY_HASH}}
           path: dist/
 
   # This job uses release artifact to publish to PyPi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,15 +4,17 @@ name: Release
 #
 # === Automated activities ===
 #
-# 1. Run tests, linting, security and complexity base line
-# 2. Bump package version and build release artifact
-# 3. Publish package to PyPi prod repository using cached artifact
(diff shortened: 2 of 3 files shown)


rubenfonseca:
Reviewing now

.github/actions/download-artifact/action.yml
```diff
@@ -0,0 +1,58 @@
...
+# PROCESS
+#
+# 1. Downloads artifact using actions/download-artifact action
+# 2. Extracts and overwrites tarball previously uploaded
+# 3. Remove archive after extraction
+
+# NOTES
+#
+# Upload-artifact and download-artifact takes ~2m40s to upload 8MB
+# so this is custom action cuts down the entire operation to 1s
```
rubenfonseca, line 13: ❤️

.github/workflows/release.yml
```diff
@@ -50,72 +50,161 @@ on:
...
+          name: ${{ steps.integrity.outputs.INTEGRITY_HASH }}
+          path: .
+
+
+  # This job runs our automated test suite, complexity and security baselines
+  # it ensures previously merged have been tested as part of the pull request process
+  #
+  # NOTE
+  #
+  # we don't upload the artifact after tested to prevent no dependencies tamper our source code
```
rubenfonseca:
```suggestion
  # we don't upload the artifact after testing to prevent any tampering of our source code dependencies
```

.github/workflows/release.yml
```diff
@@ -50,72 +50,161 @@ on:
...
+          ref: ${{ env.RELEASE_COMMIT }}
+
+      - name: Restore sealed source code
+        uses: ./.github/actions/download-artifact
+        with:
+          name: ${{ needs.seal.outputs.INTEGRITY_HASH }}
+          path: .
+
+      - name: Debug cache restore
+        run: cat pyproject.toml
```
rubenfonseca, line 137: Is this still needed?

.github/workflows/release.yml
```diff
@@ -50,72 +50,161 @@ on:
...
         run: poetry build
 
-      - name: Cache release artifact
-        id: cache-release-build
-        uses: actions/cache/save@88522ab9f39a2ea568f7027eddc7d8d8bc9d59c8 # v3.3.1
+      - name: Upload build artifact
+        uses: ./.github/actions/upload-artifact
         with:
+          name: release-${{ needs.seal.outputs.INTEGRITY_HASH }}
           path: dist/
```
rubenfonseca: So if I'm understanding this correctly, whatever happens when we run `poetry build` we ignore, and just replace the original sealed artifacts. I believe it's ok because we're just packaging the module, but just wanted to double check with you first.

rubenfonseca (changes requested):
Super good job, TIL that you can have your own custom actions on the same repo. Just left some questions and a tiny comment.

.github/workflows/release.yml
heitorlessa: I left this one thinking it'd be useful in case something ever goes wrong in any other stage, we'd know what's the bumped version and project metadata looked like in the pipeline (no secret leak here).

Happy to remove if you think otherwise. My gut tells me to leave as debugging GH Actions is always a pain after it fails.

.github/workflows/release.yml
heitorlessa: A tiny correction: We create a **new** artifact named **release-<sealed_source_code_hash>** that only contains the package distribution (wheel, tar.gz) that PyPi understands.

Optionally, we could also hash the distributed folder and use that as the hash, but felt a bit unnecessary, and it was great to trace a release to a given sealed source code (same hash)

rubenfonseca: That makes 100% sense

.github/workflows/release.yml
rubenfonseca: Leave it!

rubenfonseca (approved):
Amazing!

heitorlessa:
@rubenfonseca gonna need your review one more time as I've added a hash verifier as per @leandrodamascena ask. Leandro reviewed

heitorlessa:
Successful run publishing to Test PyPi repo (minus layers for good reasons and no change there): https://github.com/heitorlessa/aws-lambda-powertools-test/actions/runs/5053853611

leandrodamascena (approved):
APPROVED!! Amazing work @heitorlessa!

heitorlessa:
Aleluia! <3

PR #2320, opened 2023-05-24
fix(logger): add setLevel function to set level programmatically

### LINKED ISSUE #2303: Bug: programmatically changing the log level fails
### Expected Behaviour

When changing the log level of a Logger via code, the log level should be changed. 
```
powertools_logger = Logger()
powertools_logger.setLevel("DEBUG")    # the log level should become DEBUG
powertools_logger.debug("Hello, world!")    # this should produce a log message
```

### Current Behaviour

When changing the log level of a Logger via code, the log level is not changed. 
```
powertools_logger = Logger()
powertools_logger.setLevel("DEBUG")    # the log level remains at the default, INFO
powertools_logger.debug("Hello, world!")    # no log message comes out
```

### Code snippet

```python
See above, also the linked repo below
```

### Possible Solution

_No response_

### Steps to Reproduce

See https://github.com/paulsuh/aws-powertools-logging-test-sam for detailed test cases and output. 

latest

3.9

### Packaging format used

Serverless Application Repository (SAR) App

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #2303 

### Changes

When changing the logging level of a Logger via code, the logging level must be changed. With the current implementation, this was not happening because it was modifying the superclass and not the Logger instance. This PR aims to fix that.

### User experience

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -455,6 +455,7 @@ class Logger(logging.Logger):  # lgtm [py/missing-call-to-init]
             The level to set. Can be a string representing the level name: 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'
             or an integer representing the level value: 10 for 'DEBUG', 20 for 'INFO', 30 for 'WARNING', 40 for 'ERROR', 50 for 'CRITICAL'. # noqa: E501
         """
+        self.log_level = level
         self._logger.setLevel(level)
 
     def info(
--- a/aws_lambda_powertools/logging/logger.py
+++ b/aws_lambda_powertools/logging/logger.py
@@ -91,7 +91,9 @@ class Logger(logging.Logger):  # lgtm [py/missing-call-to-init]
     service : str, optional
         service name to be appended in logs, by default "service_undefined"
     level : str, int optional
-        logging.level, by default "INFO"
+        The level to set. Can be a string representing the level name: 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'
+        or an integer representing the level value: 10 for 'DEBUG', 20 for 'INFO', 30 for 'WARNING', 40 for 'ERROR', 50 for 'CRITICAL'. # noqa: E501
+        by default "INFO"
     child: bool, optional
         create a child Logger named <service>.<caller_file_name>, False by default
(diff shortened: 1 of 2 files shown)


aws_lambda_powertools/logging/logger.py
```diff
@@ -443,6 +445,18 @@ def decorate(event, context, *args, **kwargs):
...
+        """
+        Set the logging level for the logger.
+
+        Parameters:
+        -----------
+        level str | int
+            The level to set. Can be a string representing the level name: 'DEBUG', 'INFO', 'WARNING', 'ERROR', 'CRITICAL'
+            or an integer representing the level value: 10 for 'DEBUG', 20 for 'INFO', 30 for 'WARNING', 40 for 'ERROR', 50 for 'CRITICAL'. # noqa: E501
+        """
+        self._logger.setLevel(level)
```
heitorlessa, line 459: we also need to update our custom attribute: `self.log_level` here.

heitorlessa (changes requested):
one tiny change and LGTM. Also a reminder for us that this is a workaround until we have a way to fix proxy object (`__getattribute__` and `__getattr__`) while keeping mypy detecting non-existent attrs.

aws_lambda_powertools/logging/logger.py
leandrodamascena: Yes, I missed it. Fixed.

heitorlessa (approved):

heitorlessa:
NOTE for future when we do this properly...

* Remove subclassing from `logging.Logger`) and subclass from `object` instead - already commented on why
* Override `__getattribute__` to proxy to self._logger -- had to remove it before, and add `__getattr__` under `TYPE_CHECKING` due to Mypy not detecting non-existent attributes

Then we can remove these ugly workarounds, including things like `.log_level` and anything else that goes straight to the underlying logger

```python
    def __getattribute__(self, item):
        try:
            # Return actual item from our class first (e.g., append_keys, etc.)
            return object.__getattribute__(self, item)
        except AttributeError:
            # Return actual item from the underlying logger (e.g., setLevel, addHandler, etc.)
            return self._logger.__getattribute__(item)
```

PR #2334, opened 2023-05-26
fix(logger): remove subclassing and move unnecessary APIs

### LINKED ISSUE #2356: Bug: Handler registered with addHandler not logging extra keys added by append_keys
### Discussed in https://github.com/awslabs/aws-lambda-powertools-python/discussions/2261

<sup>Originally posted by **nimish-salve** May 15, 2023</sup>
`aws-lambda-powertools version: 2.15.0`

I am trying logger to use 2 handlers.
1. FileHandler
2. Streamhandler

When I added extra keys via append_keys, newly registered handlers are not logging extra keys.
Sample code below:
```
import logging

file_handler = logging.FileHandler('logs.log')
_LOGGER = Logger(service="test", level=logging.INFO, logger_handler=logging.NullHandler())
#_LOGGER = Logger(service="test", level=logging.INFO, logger_handler=file_handler) # this works

def main():
    file_handler.setFormatter(LambdaPowertoolsFormatter())
    _LOGGER.addHandler(file_handler)
    _LOGGER.append_keys(test="test")
    _LOGGER.info("in main 1")

    stream_handler = logging.StreamHandler()
    stream_handler.setFormatter(LambdaPowertoolsFormatter())
    _LOGGER.addHandler(stream_handler)
    _LOGGER.info("in main 2")

if __name__ == "__main__":
    main()
```

Whats could be the issue? 

This is reproducible even with just 1 handler. If we register the handler while creating the logger itself, it works without any issues. When we explicitly register the handler, it does not work.</div>

heitorlessa:
**Issue number:**  #2356 

## Summary

Long standing tech debt to remove `logging.Logger` sub-classing, and move backwards compatible private functions to `compat.py` to keep Logger public API cleaner.

**Why?**

If we were to add `__getattr__`, `mypy` would then fail to detect non-existent objects like `logger.foo`, because it can't infer from that.

### Changes

* [x] Remove subclassing to [prevent erroneous behaviour](https://github.com/awslabs/aws-lambda-powertools-python/discussions/2261)
* [x] Move backwards compatibility to `compat.py`.
  * [x] Double check caller filename patch
* [x] Correct docstring for `registered_handler` and `registered_formatter` to say it picks up the original handler only for convenience.
  * This is due to niche customers wanting multiple handlers; it shall prevent confusion.
* [x] Proxy `log_level` to `_logger.level` to make it less error-prone

**Historical events that made this possible now (older to recent)**

* [Introduced sub-classing to address autocomplete in PyCharm](https://github.com/awslabs/aws-lambda-powertools-python/pull/110)
* [Introduced log statements method (info/error/exception/critical/warning/debug) and addressed backward compatibility with 3.8+ stacklevel issue](https://github.com/awslabs/aws-lambda-powertools-python/pull/1658) (e.g., location line)
  * This made it possible to get rid of the subclassing as that was the main issue
* [Introduced `setLevel` method to correctly apply level in private logger](https://github.com/awslabs/aws-lambda-powertools-python/issues/2303)
  * This was only needed due to the subclassing we had, where originally was to fix autocompletion and location key, eventually leading to this confusion this PR finally addresses.

### User experience

Escape hatching to standard python logging methods like `addHandler`, `setLevel` will work as expected and effect the underlying logger. However, `mypy` customers will receive a error at type checking - if necessary, we can port these into our Logger API on a per case basis. 

image

**Output**

```json
{
    "level": "DEBUG",
    "location": "<module>:7",
    "message": "hello from debug",
    "timestamp": "2023-05-26 16:50:51,239+0200",
    "service": "payment"
}
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/logging/logger.py
```diff
@@ -445,19 +439,6 @@ def decorate(event, context, *args, **kwargs):
 
         return decorate
 
-    def setLevel(self, level: Union[str, int]):
```
rubenfonseca, line 448: So we dropped this method from the API?

heitorlessa: no, it's proxied to ._logger via descriptors. Using it will work as before. The only difference is if a customer uses Mypy it'll complain it doesn't exist, as it can't infer -- in these cases we can add them as needed. 

This removes the main issue of these previous methods already being exposed, not documented, and affecting the superclass Logger, not our underlying Logger thus the confusion.

rubenfonseca (approved):
Awesome!

aws_lambda_powertools/logging/logger.py
leandrodamascena: This isn't in production yet, so it's okay to remove it.

This PR supersedes this other one [PR](https://github.com/awslabs/aws-lambda-powertools-python/pull/2320)

PR #2342, opened 2023-05-30
feat(tracer): Support for external observability providers - Tracer

roger-zhangg:
**Issue number:**
[2030](https://github.com/awslabs/aws-lambda-powertools-python/issues/2030)

## Summary
Support provider in tracer
### Changes
Modify the provider with more genreral tracer feature, provide a new x-ray provider.

### User experience
This change doesn't affect users who just use default x-ray provider.
This change allows users to bring their own tracer provider to use other solution provider other than X-ray

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


roger-zhangg:
Drafted the first PR for tracer provider.
- Provided a more general Tracer Provider interface.
- Drafted a x-ray provider
I wasn't included in the initial discussions in this projects so I might have missed something.

Questions:
- Decide how to refactor `BaseSegment` Class
- Check on current x-ray provider's design.(Doesn't support `add_span`, `end_span`)
- In the RFC there's a capture_method function included in the provider. But currently this function is in Tracer class. Do we need to move it into our provider?

leandrodamascena:
Hey @rubenfonseca! Could you please review this PR? As we discussed offline, here are some key points to consider during the review:

1 - DD and OTEL are just placeholders/examples in the code and will be removed before merging this PR.
2 - We've updated naming conventions to use more conventional names in the methods.
3 - Regarding the example, would it be better to create a new one instead of modifying the existing example?

Thanks for your attention to these points.

aws_lambda_powertools/tracing/base.py
```diff
@@ -5,6 +5,7 @@
 from typing import Any, Generator, List, Optional, Sequence, Union
 
 
+## this class is deprecated, keep this class for backwards compatibility
```
rubenfonseca:
```suggestion
## MAINTENANCE: deprecated, kept for backwards compatibility until v3 is released
```

And can you please create a new issue to track the removal of this class on V3?

aws_lambda_powertools/tracing/base.py
```diff
@@ -16,6 +17,7 @@ def close(self, end_time: Optional[int] = None):
         Parameters
         ----------
         end_time: int
+        # TO-discuss: Providers typically use ns(time_ns -> nanosecond) as start or close time
             Time in epoch seconds, by default current time will be used.
         """
```
rubenfonseca: This needs to be addressed on the RFC. I have a hunch that instrumentation needs nanoseconds, but we need to make sure this will be the common denominator amongst all providers.

aws_lambda_powertools/tracing/provider/__init__.py
```diff
@@ -0,0 +1,5 @@
+from .datadog.dd_tracer import DDTraceProvider
+from .otel.otel_tracer import OtelProvider
+from .xray.xray_tracer import XrayProvider
+
+__all__ = ["DDTraceProvider", "OtelProvider", "XrayProvider"]
```
rubenfonseca: Note: in the future, maybe we should avoid this, as loading one provider will probably load all of them:

`from aws_lambda_powertools.tracing.provider import XrayProvider`

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+import numbers
+from contextlib import asynccontextmanager, contextmanager
+from typing import AsyncGenerator, Generator, Sequence, Union
+
+
+class BaseSpan(abc.ABC):
+    """Holds common properties and methods on span."""
+
+    @abc.abstractmethod
+    def set_attribute(self, key: str, value: Union[str, numbers.Number, bool], **kwargs) -> None:
```
rubenfonseca: I can't find any reference in the RFC why a `number.Number` is used here. Can you give me more context?

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+        """set attribute for span with a key-value pair.
```
rubenfonseca:
```suggestion
        """Set an attribute for a span with a key-value pair.
```

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+        ----------
+        key: str
+            Attribute key
+        value: Union[str, numbers.Number, bool]
+            Attribute value
+        """
+
+    @abc.abstractmethod
+    def record_exception(self, exception: BaseException, **kwargs):
+        """Add an exception to trace entities.
```
rubenfonseca:
```suggestion
        """Records an exception.
```

I'm sorry but "to trace entities" means nothing in my mind :( Can it maybe be rephrased?

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Parameters
+        ----------
+        exception: Exception
+            Caught exception
+            Output from `traceback.extract_stack()`.
```
rubenfonseca: I don't understand this line.

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
+import abc
+import numbers
+from contextlib import asynccontextmanager, contextmanager
+from typing import AsyncGenerator, Generator, Sequence, Union
+
+
+class BaseSpan(abc.ABC):
+    """Holds common properties and methods on span."""
```
rubenfonseca: Can we get a bigger description about what a "Span" is?

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Parameters
+        ----------
+        exception: Exception
+            Caught exception
+            Output from `traceback.extract_stack()`.
+        """
+
+
+class BaseProvider(abc.ABC):
```
rubenfonseca, line 37: Can we have string docs for this class please?

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+    @abc.abstractmethod
+    @contextmanager
+    def trace(self, name: str, **kwargs) -> Generator[BaseSpan, None, None]:
+        """Return a span context manger.
```
rubenfonseca: I need much more than this line of documentation. I know already what it returns based on the signature. I'm more interested about what it does and when it should be used.

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Parameters
+        ----------
+        name: str
+            Span name
+        kwargs: Optional[dict]
+            Optional parameters to be propagated to span
```
rubenfonseca:
```suggestion
            Optional parameters to be propagated to the span
```

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+
+    @abc.abstractmethod
+    @asynccontextmanager
+    def trace_async(self, name: str, **kwargs) -> AsyncGenerator[BaseSpan, None]:
+        """Return a async span context manger.
```
rubenfonseca: I need much more than this line of documentation. I know already what it returns based on the signature. I'm more interested about what it does and when it should be used.

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+
+        Parameters
+        ----------
+        name: str
+            Span name
+        kwargs: Optional[dict]
+            Optional parameters to be propagated to span
```
rubenfonseca:
```suggestion
            Optional parameters to be propagated to the span
```

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+
+    @abc.abstractmethod
+    def set_attribute(self, key: str, value: Union[str, numbers.Number, bool], **kwargs) -> None:
+        """set attribute on current active trace entity with a key-value pair.
```
rubenfonseca: Is "trace entity" the "current span" here?

aws_lambda_powertools/tracing/provider/base.py
```diff
@@ -0,0 +1,85 @@
...
+        ----------
+        key: str
+            attribute key
+        value: Union[str, numbers.Number, bool]
+            attribute value
+        """
+
+    @abc.abstractmethod
+    def patch(self, modules: Sequence[str]) -> None:
+        """Instrument a set of supported libraries
```
rubenfonseca: I need more explanation here, why a string? How can libraries be passed on to this method? Can we have an example on the string doc?

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,69 @@
...
+        self.put_annotation = self.subsegment.put_annotation
+        self.put_metadata = self.subsegment.put_metadata
+        self.add_exception = self.subsegment.add_exception
+        self.close = self.subsegment.close
+
+    def set_attribute(self, key: str, value: str | Number | bool, **kwargs) -> None:
+        if kwargs.get("namespace", "") != "":
+            self.put_metadata(key=key, value=value, namespace=kwargs["namespace"])
+        else:
+            self.put_annotation(key=key, value=value)
```
rubenfonseca: I'm trusting you did your research here, as I don't have enough knowledge to help here :D

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -15,7 +15,8 @@
 )
 from aws_lambda_powertools.shared.lazy_import import LazyLoader
 from aws_lambda_powertools.shared.types import AnyCallableT
-from aws_lambda_powertools.tracing.base import BaseProvider, BaseSegment
+from aws_lambda_powertools.tracing.provider import XrayProvider
```
rubenfonseca: Again, I understand that you are removing the other providers; however, the manner in which you have written the code results in importing all providers despite utilizing only one of them.

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -177,6 +178,31 @@ def __init__(
...
+            tracer = Tracer(service="payment")
+            tracer.set_attribute("PaymentStatus", "CONFIRMED")
+        """
+        if self.disabled:
+            logger.debug("Tracing has been disabled, aborting set_attribute")
+            return
+
+        logger.debug(f"setting attribute on key '{key}' with '{value}'")
+
+        self.provider.set_attribute(key=key, value=value)
```
rubenfonseca: I believe since no `kwargs` are being passed here, all  "attributes" will be treated as metadata, and never as annotation. Is this the correct behaviour?

```python
    def set_attribute(self, key: str, value: str | Number | bool, **kwargs) -> None:
        if kwargs.get("namespace", "") != "":
            self.put_metadata(key=key, value=value, namespace=kwargs["namespace"])
        else:
            self.put_annotation(key=key, value=value)
```

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -227,7 +257,10 @@ def put_metadata(self, key: str, value: Any, namespace: Optional[str] = None):
 
         namespace = namespace or self.service
         logger.debug(f"Adding metadata on key '{key}' with '{value}' at namespace '{namespace}'")
-        self.provider.put_metadata(key=key, value=value, namespace=namespace)
+        if self._is_custom_provider():
+            self.provider.set_attribute(key=f"{namespace}.{key}", value=value)
+        else:
+            self.provider.put_metadata(key=key, value=value, namespace=namespace)  # type: ignore
```
rubenfonseca: What if you just use `set_attribute` and make sure it works with both annotations and metadata on the Xray provider?

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -199,7 +225,11 @@ def put_annotation(self, key: str, value: Union[str, numbers.Number, bool]):
             return
 
         logger.debug(f"Annotating on key '{key}' with '{value}'")
-        self.provider.put_annotation(key=key, value=value)
+
+        if self._is_custom_provider():
+            self.provider.set_attribute(key=key, value=value)
+        else:
+            self.provider.put_annotation(key=key, value=value)  # type: ignore
```
rubenfonseca: What if you just use `set_attribute` and make sure it works with both annotations and metadata on the Xray provider?

examples/tracer/src/sdk_escape_hatch.py
```diff
@@ -11,9 +11,9 @@ def collect_payment(charge_id: str) -> str:
 @tracer.capture_lambda_handler
 def lambda_handler(event: dict, context: LambdaContext) -> str:
     charge_id = event.get("charge_id", "")
-    with tracer.provider.in_subsegment("## collect_payment") as subsegment:
-        subsegment.put_annotation(key="PaymentId", value=charge_id)
+    with tracer.provider.trace("## collect_payment") as span:
+        span.set_attribute(key="PaymentId", value=charge_id)
         ret = collect_payment(charge_id=charge_id)
-        subsegment.put_metadata(key="payment_response", value=ret)
+        span.set_attribute(key="payment_response", value=ret)
```
rubenfonseca: Please keep the old example as it was to make sure it still works and maybe create a new one?

poetry.toml
```diff
@@ -0,0 +1,2 @@
+[virtualenvs]
+in-project = true
```
rubenfonseca: Are these changes intended on this PR?

pyproject.toml
```diff
@@ -47,7 +47,11 @@ redis = { version = ">=4.4,<6.0", optional = true }
 typing-extensions = "^4.11.0"
 datadog-lambda = { version = ">=4.77,<6.0", optional = true }
 aws-encryption-sdk = { version = "^3.1.1", optional = true }
+ddtrace = "^2.5.2"
+opentelemetry-api = "^1.22.0"
+opentelemetry-sdk = "^1.22.0"
 jsonpath-ng = { version = "^1.6.0", optional = true }
+types-aws-xray-sdk = "^2.13.0.20240308"
```
rubenfonseca: Probably needs to be moved to a dev dependency.

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -831,7 +876,11 @@ def _disable_xray_trace_batching(self):
         aws_xray_sdk.core.xray_recorder.configure(streaming_threshold=0)
 
     def _is_xray_provider(self):
-        return "aws_xray_sdk" in self.provider.__module__
+        return isinstance(self.provider, XrayProvider)
+
+    def _is_custom_provider(self):
+        # check if provider is not default xray_provider. Avoid test conflits
+        return not self._is_xray_provider() and isinstance(self.provider, BaseProvider)
```
rubenfonseca: I don't understand this code. If I'm passing another provider that is not an instance of the BaseProvider, it will use the XRayProvider, is that correct? Is this intendend?

Overall I would prefer to remove this method, and fix this in the concrete implementation. This would remove a lot of the `# type:ignore` errors that leaves me scared.

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -703,7 +742,7 @@ def _add_response_as_metadata(
         self,
         method_name: Optional[str] = None,
         data: Optional[Any] = None,
-        subsegment: Optional[BaseSegment] = None,
+        subsegment: Optional[BaseSpan] = None,
```
rubenfonseca, line 739: BaseSpan is not compatible with BaseSegment, so I would say this is a breaking change. BUT since it's a "private" method, I think it's ok. Just wanted you to double check it's fine.

rubenfonseca (changes requested):
Thank you for the PR, there's a lot of great work here! 

I've left some suggestions. Can I also asked to update the PR body with a better description of the changes? It took me a while going back and forth to understand what was the scope of the PR.

aws_lambda_powertools/tracing/base.py
roger-zhangg: Created #4105

aws_lambda_powertools/tracing/base.py
roger-zhangg: @leandrodamascena and I have a discussion on this and we come to the conclusion that using float in epoch seconds this case is more appropriate. I also created a [issue](https://github.com/aws/aws-xray-sdk-python/issues/424)  for this in X-Ray python SDK as I believe the docstring here is referencing the X-Ray Python SDK. Will make sure to mention this in RFC

aws_lambda_powertools/tracing/provider/__init__.py
roger-zhangg: removed

aws_lambda_powertools/tracing/provider/base.py
roger-zhangg: Changed to Any to remove limitation.

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
roger-zhangg: need further discussion, see [here](https://github.com/aws-powertools/powertools-lambda-python/pull/2342#discussion_r1560159107)

aws_lambda_powertools/tracing/tracer.py
roger-zhangg: removed

aws_lambda_powertools/tracing/tracer.py
roger-zhangg: changed the behavior of x_ray provider's `set_attribute`. Now namespace will be passed in.

aws_lambda_powertools/tracing/tracer.py
roger-zhangg: This is to make sure the Tracer works the same way as it works before on X-Ray provider. We have considered unifying the behavior using `set_attribute`. 

To do so, firstly we need to make sure all the existing code calling `Tracer.put_annotation` is still calling `put_annotation` in SDK. But we didn't figure out a good way to do that. 

Secondly, current unit test is testing extensively if `provider.put_annotation`/`provider.put_metadata` is called. We need to change a lot of test cases to change `put_annotation` call `provider.set_attribute` instead. I think we will need an additional discussion to signoff the final design on this.

aws_lambda_powertools/tracing/tracer.py
roger-zhangg: same as above

aws_lambda_powertools/tracing/tracer.py
roger-zhangg: Yes this is tested.

poetry.toml
roger-zhangg: remove before merge

pyproject.toml
roger-zhangg: remove before merge

aws_lambda_powertools/tracing/tracer.py
roger-zhangg: The reason for this method to exist is to keep the current x-ray tracer and test work as intended. this `_is_custom_provider` is used only to decide if we should call `provider.set_attribute` or the original `provider.put_annotation`/`provider.put_metadata` in Tracer `put_annotation`, `put_metadata` methods. 

The idea is when we introduce custom provider in the future, these two method: `put_annotation`, `put_metadata` should call `provider.set_attribute` instead. But for X-Ray provider and the test dummy provider will get `false` from `_is_custom_provider` and uses current `provider.put_annotation`/`provider.put_metadata`. 

As for your suggestion: " I would prefer to remove this method, and fix this in the concrete implementation" I do agree. To do so, we need to make `provider.set_attribute` to work for both `put_annotation`, `put_metadata` in X-ray sdk and also fix all the impacted test cases (for example [here](https://github.com/aws-powertools/powertools-lambda-python/blob/0ac38003c860749cb9d0baef8630de7eba1c8733/tests/unit/test_tracing.py#L141-L160)). My thinking before is we should avoid modifying test cases for this feature.

roger-zhangg:
@rubenfonseca Thanks for the super detailed feedback. I've resolved all the comments now. Please help with a final review, thanks.

examples/tracer/src/sdk_escape_hatch.py
```diff
@@ -11,7 +11,7 @@ def collect_payment(charge_id: str) -> str:
 @tracer.capture_lambda_handler
 def lambda_handler(event: dict, context: LambdaContext) -> str:
     charge_id = event.get("charge_id", "")
-    with tracer.provider.in_subsegment("## collect_payment") as subsegment:
+    with tracer.provider.in_subsegment("## collect_payment") as subsegment:  # type: ignore
```
rubenfonseca, line 14: We should review this to make sure we don't have a typing breaking change

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -177,6 +178,36 @@ def __init__(
         if self._is_xray_provider():
             self._disable_xray_trace_batching()
 
+    def set_attribute(self, key: str, value: Any, **kwargs):
+        """Set attribute on current active span with a key-value pair.
```
rubenfonseca:
```suggestion
        """Set an attribute on current active span with a key-value pair.
```

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -177,6 +178,36 @@ def __init__(
...
+
+        Parameters
+        ----------
+        key : str
```
rubenfonseca:
```suggestion
        key: str
```

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -177,6 +178,36 @@ def __init__(
...
+            attribute key
+        value : Any
```
rubenfonseca:
```suggestion
        value: Any
```

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -177,6 +178,36 @@ def __init__(
...
+            Value for attribute
+        kwargs: Optional[dict]
+            Optional parameters to be passed to provider.set_attributes
+
+        Example
+        -------
+        Set attribute for a pseudo service named payment
```
rubenfonseca:
```suggestion
        Set an attribute for a pseudo service named payment
```

aws_lambda_powertools/tracing/tracer.py
```diff
@@ -812,11 +849,8 @@ def _patch_xray_provider(self):
         # Due to Lazy Import, we need to activate `core` attrib via import
         # we also need to include `patch`, `patch_all` methods
         # to ensure patch calls are done via the provider
```
rubenfonseca: Due to the new code, this comment can probably be deleted

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
+from __future__ import annotations
+
+from contextlib import asynccontextmanager, contextmanager
+from numbers import Number
+from typing import Any, AsyncGenerator, Generator, Literal, Sequence, Union
+
+from ....shared import constants
+from ....shared.lazy_import import LazyLoader
+from ..base import BaseProvider, BaseSpan
```
rubenfonseca: I think we're trying to use full imports now

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def set_attribute(
+        self,
+        key: str,
+        value: Any,
+        category: Literal["Annotation", "Metadata", "Auto"] = "Auto",
+        **kwargs,
+    ) -> None:
+        """
+        Set attribute on this span with a key-value pair.
```
rubenfonseca:
```suggestion
        Set an attribute on this span with a key-value pair.
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Parameters
+        ----------
+        key : str
```
rubenfonseca:
```suggestion
        key: str
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+            attribute key
+        value : Any
```
rubenfonseca:
```suggestion
        value: Any
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+            Value for attribute
+        category : Literal["Annotation","Metadata","Auto"] = "Auto"
```
rubenfonseca:
```suggestion
        category: Literal["Annotation","Metadata","Auto"] = "Auto"
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+
+    def set_attribute(
+        self,
+        key: str,
+        value: Any,
+        category: Literal["Annotation", "Metadata", "Auto"] = "Auto",
+        **kwargs,
+    ) -> None:
+        """
+        Set attribute on the current active span with a key-value pair.
```
rubenfonseca:
```suggestion
        Set an attribute on the current active span with a key-value pair.
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+
+        Parameters
+        ----------
+        key : str
```
rubenfonseca:
```suggestion
        key: str
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+            attribute key
+        value : Any
```
rubenfonseca:
```suggestion
        value: Any
```

aws_lambda_powertools/tracing/provider/xray/xray_tracer.py
```diff
@@ -0,0 +1,141 @@
...
+            Value for attribute
+        category : Literal["Annotation","Metadata","Auto"] = "Auto"
```
rubenfonseca:
```suggestion
        category: Literal["Annotation","Metadata","Auto"] = "Auto"
```

aws_lambda_powertools/tracing/base.py
```diff
@@ -5,17 +5,18 @@
 from typing import Any, Generator, List, Optional, Sequence, Union
 
 
+## MAINTENANCE: deprecated, kept for backwards compatibility until v3 is released
 class BaseSegment(abc.ABC):
     """Holds common properties and methods on segment and subsegment."""
 
     @abc.abstractmethod
-    def close(self, end_time: Optional[int] = None):
+    def close(self, end_time: Optional[float] = None):
```
rubenfonseca: Is this a breaking change? Or are we assuming that no one is inheriting from `BaseSegment` today?

rubenfonseca (changes requested):
Super good progress! Just left some tiny picks and some questions on breaking changes. @leandrodamascena can you help implement the changes so we commit quickly?

aws_lambda_powertools/tracing/base.py
leandrodamascena: We will deprecate this in Powertools v3. We're not using this class today, but we never know if customers are, so it's best to revert this change.

Reverted.

leandrodamascena:
We can consider this PR completed, but we will not merge it now. We'll start conversations about Powertools V3 and add this refactor when we're ready for it:

1 - We will deprecate the [BaseSegment](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/tracing/base.py#L8) in Powertools v3.
2 - This is a [potential breaking change](https://github.com/aws-powertools/powertools-lambda-python/pull/2342/files#diff-ac2e7c32fce52e9e17243aafe7f69a62615e44d9d3a15dabd164d16b476dc5daR18) for customers who are [accessing the XRay](https://docs.powertools.aws.dev/lambda/python/latest/core/tracer/#escape-hatch-mechanism) methods directly.
3 - We can take this opportunity to standardize method names like `trace` instead of `subsegment` in our documentation and ease for customers adopt new Observability providers.

I'm approving this PR and changing it to Draft state just so I don't mess up our stats.

leandrodamascena (approved):
Thanks for another stellar job @roger-zhangg! This refactor will open the opportunity to expand the Tracer utility and add more observability providers.

I left a comment explaining why we are approving but not merging this PR right now: 
https://github.com/aws-powertools/powertools-lambda-python/pull/2342#issuecomment-2061734362

leandrodamascena:
Hi @roger-zhangg! I'm closing this PR in favor of #4902! It was really hard to fix the entire merge conflict here, so I've created a new one.

Thank you for all the effort you've put in here, and I'll give you credit for this work when I release v3.

PR #2419, opened 2023-06-07
feat(logger): type log record in LambdaPowertoolsFormatter with TypedDict

### LINKED ISSUE #2401: Feature request: Add typing for log parameter in LambdaPowertoolsFormatter
### Use case

Change discussed in #2014
Improve the existing mechanism of the LambdaPowertoolsFormatter by typing the log parameter with TypedDict

### Solution/User Experience

Create a new types.py for the logger that will include the new TypedLog class, similar to:
```
from typing_extensions import TypedDict


class TypedLog(TypedDict, total=False):
    level: str
    location: str
    timestamp: str
    service: str
    event: str
```
When overriding the serialize() function of the LambdaPowertoolsFormatter, customers can have type checking on our log fields using the new TypedLog class, similar to:
```
from aws_lambda_powertools.logging.types import TypedLog


class CustomFormatter(LambdaPowertoolsFormatter):
    def serialize(self, log: TypedLog) -> str:
        """Serialize final structured log dict to JSON str"""
        log["event"] = log.pop("message")  # rename message key to event
        return self.json_serializer(log)  # use configured json serializer


logger = Logger(service="payment", logger_formatter=CustomFormatter())
logger.info("hello")
```
### Alternative solutions

_No response_

### Acknowledgment

erikayao93:
**Issue number:**
#2401 
### Changes

Added types.py file with TypedLog class

### User experience

New typing has not been implemented into any other files, so user experience is not impacted.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
Looking...

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any, Dict, Iterable, Union
+
+from typing_extensions import TypedDict
+
+class TypedLog(TypedDict, total=False):
```
heitorlessa: Two suggestions here:

* Rename to a more explicit name like `PowertoolsLogRecord` - this makes it easier to quickly understand what this is about
* Remove `total=False` as that removes our guarantee that these keys will be present. There's another way you can be more assertive about what's required vs what's not required - sending a snippet for you

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,10 @@
+from typing import Any, Dict, Iterable, Union
+
```
heitorlessa, line 2:
```suggestion
from __future__ import annotations
```

This will gives us a nicer syntax for Unions among other things. For example:

* `timestamp: Union[str, int]` becomes `timestamp: str | int`

If you'd like to learn more about this topic, check this out on how this enables Postponed Evaluation of Annotations (or type hint)

https://peps.python.org/pep-0563/

:)

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,10 @@
...
+from typing_extensions import TypedDict
+
+class TypedLog(TypedDict, total=False):
+    level: Union[str, int]
```
heitorlessa:
```suggestion
    level: str
```

we're controlling the logging level so in this case it'll always be string. The `Logger(level=20)` will be transformed into `INFO` as a string later.

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,10 @@
...
+    location: str
+    timestamp: Union[str, int]
```
heitorlessa:
```suggestion
    timestamp: str | int
```

see how much nicer it becomes after you bring in `from __future__ import annotations` I suggested earlier?

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,10 @@
...
+    service: str
+    message: Union[Dict[str, Any], str, bool, Iterable]
```
heitorlessa:
```suggestion
    message: Dict[str, Any] | str | bool | Iterable
```

see how much nicer it becomes after you bring in `from __future__ import annotations` I suggested earlier?

I've replaced `Iterable` for `List` as Iterable encompass a wider type of data structures that JSON doesn't, for example a `Set` is Iterable but would be serialized as a string.


```python

logger = Logger(service="sample")

item_set = {"banana", "apple"}  # this is set()
logger.info(item_set)
```

The output would be more or less like this

```json
{"level":"INFO","location":"<module>:19","message":"{'banana', 'apple'}","timestamp":"2023-06-08 10:50:34,451+0200","service":"sample"}
```

aws_lambda_powertools/logging/types.py
heitorlessa: Extend your TypedDict class to include optional fields when customers choose to inject Lambda context, correlation ID, etc. - our documentation has examples of those fields.

Here's a complete example to make it easier and when they're added for context:

```python

from typing_extensions import NotRequired

class PowertoolsLogRecord(TypedDict):
    level: str
    location: str
    timestamp: str | int
    service: str
    message: str | bool | Dict[str, Any] | List[Any]
    cold_start: NotRequired[bool]  # logger.inject_lambda_context
    function_name: NotRequired[str]  # logger.inject_lambda_context
    function_memory_size: NotRequired[int]  # logger.inject_lambda_context
    function_arn: NotRequired[str]  # logger.inject_lambda_context
    function_request_id: NotRequired[str]  # logger.inject_lambda_context
    xray_trace_id: NotRequired[str]  # logger.inject_lambda_context if AWS X-Ray is enabled
    correlation_id: NotRequired[str]  # logger.set_correlation_id
    exception_name: NotRequired[str]  # logger.exception
    exception: NotRequired[str]  # logger.exception
```


**Caveat**

You need to adjust your import to make it conditional for both `NotRequired` and `TypedDict`. The former (NotRequired) is only supported in 3.11 so import from `typing_extensions` would fail in 3.11, and `TypedDict` is available in 3.8+ and it'd also fail later.

> Docs: https://docs.python.org/3/library/typing.html

I'll let you figure out how to make an conditional import depending on the Python version, but please shout if you get stuck!

examples/logger/src/bring_your_own_formatter.py
```diff
@@ -1,12 +1,13 @@
...
+from aws_lambda_powertools.logging.types import TypedLog
 
 
 class CustomFormatter(LambdaPowertoolsFormatter):
-    def serialize(self, log: dict) -> str:
+    def serialize(self, log: TypedLog) -> str:  # type: ignore
         """Serialize final structured log dict to JSON str"""
-        log["event"] = log.pop("message")  # rename message key to event
-        return self.json_serializer(log)  # use configured json serializer
+        log["event"] = log.pop("message")  # type: ignore
```
heitorlessa: This shouldn't be needed with the right class now.

examples/logger/src/bring_your_own_formatter.py
```diff
@@ -1,12 +1,13 @@
...
+        return self.json_serializer(log)  # type: ignore
```
heitorlessa: Check if you still need this after modifying the class.. `type: ignore` is a last resort, we should avoid it. If truly necessary, you can make it more specific like based on Mypy error, e.g., `type: ignore[override]`

examples/logger/src/bring_your_own_formatter.py
```diff
@@ -1,12 +1,13 @@
 from aws_lambda_powertools import Logger
 from aws_lambda_powertools.logging.formatter import LambdaPowertoolsFormatter
+from aws_lambda_powertools.logging.types import TypedLog
 
 
 class CustomFormatter(LambdaPowertoolsFormatter):
-    def serialize(self, log: dict) -> str:
+    def serialize(self, log: TypedLog) -> str:  # type: ignore
```
heitorlessa:
```suggestion
    def serialize(self, log: PowertoolsLogRecord | Mapping) -> str:  # type: ignore
```

`type: ignore` is a last resort, we should avoid it. What you'll have to do here is to first allow a Typed Dictionary (PowertoolsLogRecord you defined earlier), and any Dictionary to allow inheritance and override (liskov substitution principle).

From here, you'll have to address other parts in the Formatter code which expected a Dict to serialize/deserialize and use Mapping.

Mapping is a generic hashable value, a protocol in Python words, which would satisfy our typing constraints - more info here: https://docs.python.org/3/library/stdtypes.html#mapping-types-dict

Please shout if you get stuck. We should be changing our actual implementation in `formatter.py` too in addition to this example.

heitorlessa (changes requested):
Great start Erika!! Made comments on how to make this change production ready, and links to help you navigate intricacies of structured typing.

Please do shout if you get stuck

aws_lambda_powertools/logging/types.py
erikayao93: Out of curiosity, is there a reason why this isn't used for some of the other typing hints, like for return values on functions?

examples/logger/src/bring_your_own_formatter.py
erikayao93: The issue mypy doesn't like here is two-fold:
- The addition of a key not defined in the TypedDict (in this case "event") raises a "typeddict-unknown-key" error
- Attempting to delete the "message" key from the TypedDict raises a "misc" error because it is a required field (since total=False was removed)

aws_lambda_powertools/logging/types.py
heitorlessa: You can definitely use for return values. There's generally two gotchas with this feature:

1. [Reflections will see the annotated type as string until runtime](https://peps.python.org/pep-0649/). This means an extra step to get a forward reference (-> Response is now -> "Response").
2. IDEs will show '|' as a regular `Union` as `|` it's merely syntactic sugar. Eventually, it's easy to go overboard with potentially multiple function returns and it can degrade the caller experience, as it now has to make multiple checks before proceeding.

Hopefully that makes sense, let me know if I misunderstood your question otherwise.

examples/logger/src/bring_your_own_formatter.py
heitorlessa: Thanks a lot @erikayao93 for explaining it. Apparently we're not the only ones - known issue here: https://github.com/python/mypy/issues/4976

I'd guess a `Dict[str, Any]` would address the first issue, while the former **might** need a specific `type: ignore[typeddict-unknown-key]`. 

**UPDATE**: Read the thread. Dict would indeed address the first issue and the latter needs two ignore until Mypy agrees TypedDict is mutable.

```python
log["event"] = log.pop("message")  # type: ignore[typeddict-unknown-key, misc]
```

For the former, you can use another feature named `TypeAlias` to make it easier to use this union then replace what we've got earlier with `PowertoolsLogRecord | Mapping`:

```python
LogRecord: TypeAlias = Union[Dict[str, Any], PowertoolsLogRecord]
```
> Sadly, TypeAlias doesn't support yet `|` syntax :/

Let me know if that doesn't work and what alternatives you can find, and I can jump in early next week.

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,38 @@
+from __future__ import annotations
+
+import sys
+
+if sys.version_info < (3, 11):
+    from typing_extensions import NotRequired, TypedDict
+else:
+    from typing import NotRequired, TypedDict
```
heitorlessa: You'd need an additional check for 3.8 - when TypedDict was introduced, and removed from typing_extensions.

aws_lambda_powertools/logging/types.py
```diff
@@ -0,0 +1,38 @@
...
+    cold_start: NotRequired[bool]
+    function_name: NotRequired[str]
+    function_memory_size: NotRequired[int]
+    function_arn: NotRequired[str]
+    function_request_id: NotRequired[str]
+    # From logger.inject_lambda_context if AWS X-Ray is enabled
+    xray_trace_id: NotRequired[str]
+
+    # If sample_rate is defined
+    sampling_rate: NotRequired[float]
```
heitorlessa, line 43: Great catch, Erika!! Superb!

examples/logger/src/bring_your_own_formatter.py
erikayao93: I did some local testing, away from the repo, and it looks like even with the `TypeAlias`, mypy will still give a `[typeddict-unknown-key]` error when trying to add a key. 

I tried out defining some kind of `Dict[str, Any]` with a `TypedDict` attribute, which actually does work. 

**Pros:**
- elements can be added AND deleted without raising errors

**Cons:**
- the type checking is gone after the initial `TypedDict` definition

I pretty much casted the `TypedDict` to a regular `Dict[str, Any]`, but I don't know if that's what we really want? It mght not be particularly feasible for our use case, especially since any typing is gone after the initial definition.

I attached my small test file here, which compiles without complaint from `mypy` and ran on Python 3.10.6.
[typedict_test.zip](https://github.com/awslabs/aws-lambda-powertools-python/files/11716892/typedict_test.zip)

Let me know your thoughts!

heitorlessa: Feel free to add that specific `type: ignore[typeddict-unknown-key]` for that line - it describes the intent well and much improved compared to the previous ignores we had to make for every line. Casting is similar to `type: ignore` and can accidentally lead to customers having multiple casts when they bring their own TypedDict etc later.

Given this is an example only, we're more than fine with that type ignore as we're explicitly remapping it. Most customers will end up creating a new dictionary object, and copying the values of our `log` TypedDict object, since their format might be radically different based on their company policies.

Thank you so much for digging this through! Let's add that type ignore specific error and we can do a final pass before merging ;)

erikayao93: I missed this earlier, since I had ended up using a single `type: ignore`, but that ignore is actually muting four errors now -- two new errors came up because of using Mapping
Since the mapping is read-only (which is what makes it compatible with TypedDict), there is an `index` error and a `union-attr` error on that line.
So overall on that line we have:
- error: Unsupported target for indexed assignment ("Union[PowertoolsLogRecord, Mapping[Any, Any]]")  [index]
- error: TypedDict "PowertoolsLogRecord" has no key "event"  [typeddict-unknown-key]
  - Could be addressed by adding the field as `NotRequired` as suggested [here](https://github.com/python/mypy/issues/4976#issuecomment-1584955773), but but this is super inflexible for any fields customers might want to actually add, as you basically said in the reply to that suggestion.
- error: Item "Mapping[Any, Any]" of "Union[PowertoolsLogRecord, Mapping[Any, Any]]" has no attribute "pop"  [union-attr]
- error: Key "message" of TypedDict "PowertoolsLogRecord" cannot be deleted  [misc]
  - Could be addressed by making all of the fields of the `TypedDict` `NotRequired`? This is what is being suggested in that thread [here](https://github.com/python/mypy/issues/4976#issuecomment-1585694263).

Let me know how you want me to proceed on this!

heitorlessa:
Looking into this today and sending the fixes along w/ comments before merging

examples/logger/src/bring_your_own_formatter.py
heitorlessa: just pushed the changes to fix the first one (`unsupported target for indexed assignment`) and ignored the last two checks with an explanation as to why we did it.

We wouldn't want to add `message` as NotRequired because it wouldn't be accurate (it's always there) - I replied to that thread with context. It's the same as saying `logger.info()` is a valid API call instead of `logger.info("any message here")`.

heitorlessa:
I've pushed the changes to fix Mypy and bring `LogRecord` as a TypeAlias for `Dict | PowertoolsLogRecord` - waiting for CI before merging.

The only unrelated change I noticed here was the `pre-commit` hook for the `markdownlint-cli` which I suspected it came from an offline chat with @seshubaws due to NodeJS setup -- always use `git sha` whenever changing a pre-commit or GitHub Action, otherwise it opens the door for abuse since `v0.34.0` is a mutable git tag.

**THANK YOU** so much for going through this change @erikayao93. Much appreciated.

---

**Changes**

* [x] Simplify conditional imports to ease reading and future updates
* [x] Bring LogRecord `TypeAlias`
* [x] Update `LambdaPowertoolsFormatter` and `DatadogLogFormatter` to use `LogRecord`
  * [x] constructor: `json_serializer/deserializer`
  * [x] method: `serialize`
* [x] Fix `bring_your_own_formatter.py` docs highlighting
* [x] Confirm conditional imports work across all versions in CI
* [x] Fix pre-commit hook to use immutable image
  * [x] markdownlint-cli
  * [x] actionlint (tech debt)
  * [x] pre-commit hooks (tech debt)

**Unrelated changes**

* [x] Pay old tech debt
  * [x] use `|` shortcut for Union types in `formatter.py`
  * [x] use FQDN imports in `formatter.py`
  * [x] Update Union type in formatter to `|`
  * [x] immutable pre-commit actionlint
  * [x] immutable pre-commit pre-commit hooks

heitorlessa (approved):

PR #2458, opened 2023-06-14
fix(event_handler): prioritize static over dynamic route to prevent order of route registration mismatch

### LINKED ISSUE #2439: APIGatewayRestResolver route matching order
### Expected Behaviour

When there are two routes:
1. GET /studies/fetch
2. GET /studies/{studyid}

And a request is sent to /studies/fetch, the request should be routed to the first endpoint.

### Current Behaviour

Request if routed to the second endpoint.

### Code snippet

```python

@app.get("/studies/<studyid>")
def get_study(studyid):
    return "get_study"


@app.get("/studies/fetch")
def fetch_studies():
    return "fetch_studies"


def handler(event, context):
```

### Possible Solution

_No response_

### Steps to Reproduce

Create a SAM project with the snippet as a lambda. 
Invoke with /studies/fetch

latest

3.10

### Packaging format used

Serverless Application Repository (SAR) App

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #2439

### Changes

This PR changes the priority of route matching, so that the most specific route is matched first.

### User experience

Take into consideration the following code:

```python

@app.get("/studies/<studyid>")
def get_study(studyid):
    return "get_study"


@app.get("/studies/fetch")
def fetch_studies():
    return "fetch_studies"


def handler(event, context):
```

Before this change, `/studies/fetch` would match the first route, because it's a prefix of `/studies/<studyid>`. After this change, the most specific route is matched first, so `/studies/fetch` will match the second route.

## Checklist

- [ ] Changes are documented
- [ ] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -483,7 +483,8 @@ class ApiGatewayResolver(BaseRouter):
             with api gateways with multiple custom mappings.
         """
         self._proxy_type = proxy_type
-        self._routes: List[Route] = []
+        self._dynamic_routes: List[Route] = []
+        self._static_routes: List[Route] = []
         self._route_keys: List[str] = []
         self._exception_handlers: Dict[Type, Callable] = {}
         self._cors = cors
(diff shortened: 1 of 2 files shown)


rubenfonseca:
As the failing test is showing, it seems we might be breaking some existing behavior. Let me look into this deeper.

rubenfonseca:
Ok, tried a new approach which changes some private properties, which I think it's ok.

heitorlessa:
looking 👀

heitorlessa (approved):

heitorlessa:
all good here provided CI pass now. Leaving to @leandrodamascena to merge as he expressed he wanted to learn about the fix.

heitorlessa:
Leandro just shared he's learned it 2 hours ago. GH Actions seems to be experiencing operational issues... :/, so I'll merge it on the basis that `make pr` worked locally.

We'll know either way once they resolve capacity management issues.

GREAT work @rubenfonseca investigating and fixing it.

PR #2506, opened 2023-06-20
refactor(event_source): convert functional tests to unit tests

### LINKED ISSUE #509: Maintenance: Break down event_source_data_classes tests
Issue: #504

Functional tests for event source data classes utility are now 1.3K LOC in a single file. This issue tracks resolving them into a test per event source to ease maintenance and contrib

leandrodamascena:
**Issue number:** #509 

### Changes

This pull request aims to enhance the test structure and organization within the `event_source_data_classes` module. By breaking down the existing tests into smaller, more focused units, this refactor improves the maintainability and readability of the codebase. The `test_data_class.py file` had more than 2k LOC and it was becoming impossible to maintain it, so I divided it into small tests to make it easier to maintain.

I've also changed function tests to unit tests, which makes more sense for the type of testing we're doing. Just to explain a little more, these tests specifically test whether given an event source, Powertools can turn it into an object.

### User experience

There is no change in the user experience.

## Checklist

* [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda-python/#tenets)
* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
looking 👀

heitorlessa:
@leandrodamascena PR body can be improved - add a note about functional -> unit, before/after (functional -> unit change for a single test), etc.

tests/unit/data_classes/test_api_gateway_proxy_event.py
```diff
@@ -0,0 +1,219 @@
+from aws_lambda_powertools.utilities.data_classes import (
+    APIGatewayProxyEvent,
+    APIGatewayProxyEventV2,
+)
+from tests.functional.utils import load_event
+
+
+def test_default_api_gateway_proxy_event():
+    event = APIGatewayProxyEvent(load_event("apiGatewayProxyEvent_noVersionAuth.json"))
```
heitorlessa: separate `raw_event` from `parsed_event`, otherwise you're testing Python getters of `APIGatewayProxyEvent` and not whether the data has been parsed correctly.

example

```python
raw_event = load_event("apiGatewayProxyEvent_noVersionAuth.json")
parsed_event = APIGatewayProxyEvent(**raw_event)

assert parsed_event.resource == raw_event["resource"]
```

heitorlessa (changes requested):
I'm so so happy we're finally getting this tech debt done - THANK YOU.

Two major changes before merging:

* We need a functional test for `@event_source` decorator
* Split `raw_event` from `parsed_event` in every example, and assert whether `parsed_event.<property>` has the same data as the `raw_event[<property>]`

leandrodamascena:
> @leandrodamascena PR body can be improved - add a note about functional -> unit, before/after (functional -> unit change for a single test), etc.

Can you check again please?

tests/unit/data_classes/test_api_gateway_proxy_event.py
leandrodamascena: Hi @heitorlessa! This was lot of work, dude! hahaha

leandrodamascena:
> * We need a functional test for `@event_source` decorator

We already have this test - https://github.com/aws-powertools/powertools-lambda-python/pull/2506/files#diff-1a9fabb294fa232197e0c50cb7b3ee1b8994448f3e798c6a5ff16afc24c52bf6R401

> * Split `raw_event` from `parsed_event` in every example, and assert whether `parsed_event.<property>` has the same data as the `raw_event[<property>]`

Done

tests/unit/data_classes/test_active_mq_event.py
```diff
@@ -6,33 +6,34 @@
...
     assert message.delivery_mode is None
     assert message.correlation_id is None
     assert message.reply_to is None
     assert message.get_type is None
     assert message.expiration is None
     assert message.priority is None
 
-    messages = list(event.messages)
+    messages = list(parsed_event.messages)
     message = messages[1]
```
heitorlessa: you can use `message` previously defined instead (same effect).

`message` property takes one message at a time from the list using `next()`

This means:

```python
parsed_event.message == list(parsed_event.messages)[0]
```

https://github.com/aws-powertools/powertools-lambda-python/blob/a3d7bd54685a94af5a0cb02fcfa17c96f6e2e790/aws_lambda_powertools/utilities/data_classes/active_mq_event.py#L130

heitorlessa: Actually there's a bug in ActiveMQ implementation for `message` which will **always** return the same message, since the generator is being redefined every time when accessing `.messages`.

---

More concise test to explain the bug

```python
class Messages:
    def __init__(self):
        self._messages = ["one", "two", "three"]
        ## NOTE: This defines a generator at instantiation
        # self.messages = (msg for msg in self._messages)

    ## NOTE: This defines a generator every time we access it
    ## which is fine for expected use in a loop (for msg in messages)
    ## but this doesn't work in the way `message` property is doing
    @property
    def messages(self):
        for msg in self._messages:
            yield msg

    @property
    def message(self):
        return next(self.messages)

data = Messages()
first_message = data.message
second_message = data.message

assert first_message != second_message
```

heitorlessa (changes requested):
found a hidden bug in ActiveMQ implementation of `message` property when reviewing an older assertion it was doing.

We need to search for other uses of generators with `next()` in properties, as it'll lead to the same problem.

tests/unit/data_classes/test_active_mq_event.py
leandrodamascena: Amazing catch @heitorlessa!

The same happens in the following data class:

https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/data_classes/kafka_event.py#L120

https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/data_classes/s3_event.py#L314

https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/data_classes/s3_event.py#L312

https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/data_classes/sns_event.py#L116

I'll try to change these files to see if it is a potential breaking change or not.

leandrodamascena: Hi @heitorlessa!! Check out this commit - [`d8b2739` (#2506)](https://github.com/aws-powertools/powertools-lambda-python/pull/2506/commits/d8b273994543f75dbb6f52ab51383b263be9663d)

I addressed this feedback and fixed this property in the `ActiveMQ` and `Kafka` event source classes. For S3, SNS and SES don't need to change because it is always just one record per Lambda trigger.

Thanks for giving this feedback, it was very helpful to learn new things.

aws_lambda_powertools/utilities/data_classes/active_mq_event.py
```diff
@@ -128,4 +132,4 @@ def messages(self) -> Iterator[ActiveMQMessage]:
 
     @property
     def message(self) -> ActiveMQMessage:
-        return next(self.messages)
+        return next(self._messages)
```
heitorlessa, line 151:
```suggestion
        if self._messages is None:
            self._messages = self.messages
        return next(self._messages)
```

aws_lambda_powertools/utilities/data_classes/active_mq_event.py
```diff
@@ -128,4 +132,4 @@ def messages(self) -> Iterator[ActiveMQMessage]:
 
     @property
     def message(self) -> ActiveMQMessage:
```
heitorlessa, line 134: docstring ;)

aws_lambda_powertools/utilities/data_classes/kafka_event.py
```diff
@@ -87,6 +87,11 @@ class KafkaEvent(DictWrapper):
     - https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html
     """
 
+    def __init__(self, data: Dict[str, Any]):
+        super().__init__(data)
+        for chunk in self["records"].values():
+            self._records = (KafkaEventRecord(record) for record in chunk)
```
heitorlessa: same here: https://github.com/aws-powertools/powertools-lambda-python/pull/2506/files/f92761763342fe2f82ead4a618e94978ae6ec4aa..5f4425065c861d039f347d095b5fdf81c405ae54#r1239927667

heitorlessa (dismissed):
🚀 ⛴️

heitorlessa:
Just need to fix the typing for self._messages and records because they're being inferred as None, then assigned a generator later

heitorlessa (approved):

PR #2521, opened 2023-06-22
feat(event_handler): allow stripping route prefixes using regexes

### LINKED ISSUE #2494: Feature request: Better strip-prefixes for APIGatewayRestResolver
### Use case

I have multiple API Gateway Rest apis backed by AWS Lambda.
I want to these APIs to the same **Custom domain name** with different **path**s. 

For example I have APIs with these urls, each backed by a different lambda:
1. https://xxx.execute-api.us-east-1.amazonaws.com/v1
2. https://yyy.execute-api.us-east-1.amazonaws.com/v1
3. https://xxx_test.execute-api.us-east-1.amazonaws.com/v1
4. https://yyy_test.execute-api.us-east-1.amazonaws.com/v1

I want both of them to be available from:
1. https://mydomain.com/api/v1/xxx
2. https://mydomain.com/api/v1/yyy
3. https://mydomain.com/api/test/xxx
4. https://mydomain.com/api/test/yyy
respectively.

When Init-ing the APIGatewayRestResolver class I would have to pass in the **strip_prefixes**. 
```
strip_prefixes=["/api/v1/xxx", "/api/test/xxx"]
```
We could potentials go to a high level of options due to combinatorics.

### Solution/User Experience

1. Use regex:
```
strip_prefixes=["^.*?xxx"]
```

2. The path will be stripped of the prefixes one by one from left to right.
```
strip_prefixes=["api", "v1" ,"test","xxx"]
```
So if out initial path is `/api/v1/xxx` it will strip "api" first:
 `/api/v1/xxx` ->  `/v1/xxx`
Then v1 
`/v1/xxx` -> `/xxx`
Then it will try to strip "test" but nothing would change:
`/xxx` - > `/xxx`
Then it will strip "xxx"
`/xxx` - > `/`
 

### Alternative solutions

_No response_

### Acknowledgment

royassis:
**Issue number:** #2494

### Changes

This PR adds the ability to set a [regex](https://xkcd.com/208/) when removing prefixes from routes. This is in addition to the existing mechanism where we strip static strings as prefixes. 

### User experience

carbon (16)

## Checklist

* [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda-python/#tenets)
* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -921,7 +921,7 @@ class APIGatewayRestResolver(ApiGatewayResolver):
         cors: Optional[CORSConfig] = None,
         debug: Optional[bool] = None,
         serializer: Optional[Callable[[Dict], str]] = None,
-        strip_prefixes: Optional[List[str]] = None,
+        strip_prefixes: Optional[List[Union[str, Pattern]]] = None,
     ):
         """Amazon API Gateway REST and HTTP API v1 payload resolver"""
         super().__init__(ProxyEventType.APIGatewayProxyEvent, cors, debug, serializer, strip_prefixes)
@@ -952,7 +952,7 @@ class APIGatewayHttpResolver(ApiGatewayResolver):
         cors: Optional[CORSConfig] = None,
         debug: Optional[bool] = None,
         serializer: Optional[Callable[[Dict], str]] = None,
-        strip_prefixes: Optional[List[str]] = None,
+        strip_prefixes: Optional[List[Union[str, Pattern]]] = None,
     ):
         """Amazon API Gateway HTTP API v2 payload resolver"""
         super().__init__(ProxyEventType.APIGatewayProxyEventV2, cors, debug, serializer, strip_prefixes)
@@ -966,7 +966,7 @@ class ALBResolver(ApiGatewayResolver):
         cors: Optional[CORSConfig] = None,
         debug: Optional[bool] = None,
         serializer: Optional[Callable[[Dict], str]] = None,
-        strip_prefixes: Optional[List[str]] = None,
+        strip_prefixes: Optional[List[Union[str, Pattern]]] = None,
     ):
         """Amazon Application Load Balancer (ALB) resolver"""
         super().__init__(ProxyEventType.ALBEvent, cors, debug, serializer, strip_prefixes)
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -724,10 +724,11 @@ class ApiGatewayResolver(BaseRouter):
             if isinstance(prefix, Pattern):
                 path = re.sub(prefix, "", path)
 
-        # When using regexes, we might get into a point where everything is removed
-        # from the string, so we check if it's empty and change it accordingly.
-        if not path:
-            path = "/"
+                # When using regexes, we might get into a point where everything is removed
+                # from the string, so we check if it's empty and return /, since there's nothing
+                # else to strip anymore.
+                if not path:
+                    return "/"
 
         return path
 
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -715,8 +715,10 @@ class ApiGatewayResolver(BaseRouter):
         for prefix in self._strip_prefixes:
             if path == prefix:
                 return "/"
-            if self._path_starts_with(path, prefix):
-                return path[len(prefix) :]
+            path = re.sub(rf"^/?({prefix})/", r"/", path)
+
+        if not path:
+            path = "/"
 
         return path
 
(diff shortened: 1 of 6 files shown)


rubenfonseca:
Looking into this now

rubenfonseca (changes requested):
Thank you for your contribution! 🎉  Two asks:

- can you make this non-iterative? Meaning, when there is a match in the regex, we stop processing more items from the `strip_prefixes` array? Reason: This would probably be a breaking change and using the regex solves the initial problem in the first place
- do you mind adding a couple of functional test cases to cover the new code?

Please let me know if you need assistance with those, and we would be happy to help!

leandrodamascena:
> Thank you for your contribution! 🎉 Two asks:
> 
> * can you make this non-iterative? Meaning, when there is a match in the regex, we stop processing more items from the `strip_prefixes` array? Reason: This would probably be a breaking change and using the regex solves the initial problem in the first place
> * do you mind adding a couple of functional test cases to cover the new code?
> 
> Please let me know if you need assistance with those, and we would be happy to help!

I asked for your review @rubenfonseca again! Since we haven't heard from other clients, we need to decide whether to go ahead with this PR and close it.

Thanks

heitorlessa:
Hey! We'll be making the necessary adjustments to the PR to guarantee it solves your problem and of similar customers.

rubenfonseca:
Looking into this now

rubenfonseca:
@heitorlessa @leandrodamascena I would like your opinion on this new code, before I write documentation. Interpreting bare strings into regexes could lead us to problem. So to circumvent that, I've changed the typing to allow either `str` or `Pattern`, and threat the differently in the stripping code.

Example usage:

```python
app = ApiGatewayResolver(strip_prefixes=[re.compile(r"/(dev|stg)")])

@app.get("/foo") # will match /dev/foo and /stg/foo
...
```

Do you like that direction?

heitorlessa:
Yes, happy with that. Because it's regex, I'd suggest giving a detailed and common example in the docs.

leandrodamascena:
Hei @heitorlessa, can you review this, please?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -534,9 +534,9 @@ def __init__(
             environment variable
         serializer : Callable, optional
             function to serialize `obj` to a JSON formatted `str`, by default json.dumps
-        strip_prefixes: List[str], optional
+        strip_prefixes: List[Union[str, Pattern]], optional
             optional list of prefixes to be removed from the request path before doing the routing. This is often used
-            with api gateways with multiple custom mappings.
+            with api gateways with multiple custom mappings. Each prefix can be a static string or a compiled regex
```
heitorlessa:
```suggestion
            with api gateways with multiple custom mappings. Each prefix can be a static string or a compiled regex pattern
```

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -713,10 +713,20 @@ def _remove_prefix(self, path: str) -> str:
...
+                if self._path_starts_with(path, prefix):
+                    return path[len(prefix) :]
+
+            if isinstance(prefix, Pattern):
+                path = re.sub(prefix, "", path)
+
+        # When using regexes, we might get into a point where everything is removed
+        # from the string, so we check if it's empty and change it accordingly.
+        if not path:
+            path = "/"
```
heitorlessa: nice catch!! should we warn? thinking of unintentional regex accidents or maybe it's intentional

docs/core/event_handler/api_gateway.md
```diff
@@ -293,6 +293,14 @@ To address this API Gateway behavior, we use `strip_prefixes` parameter to accou
 
 	For example, when using `strip_prefixes` value of `/pay`, there is no difference between a request path of `/pay` and `/pay/`; and the path argument would be defined as `/`.
 
+For added flexibility, you can use regexes to strip a prefix. This is helpful when you have many options due to different combinations of prefixes (e.g: multiple environments, multiple versions).
```
heitorlessa, line 296: Niiicely written

tests/functional/event_handler/test_api_gateway.py
heitorlessa, line 1: can we have an additional test to cover when regex might lead to: https://github.com/aws-powertools/powertools-lambda-python/pull/2521/files#diff-bdb4b43087f89fdf398381f3d88048f6182562ca67e8907259d5c56d658465c1R726

heitorlessa (changes requested):
Looks great, specially docs!

Asked for an additional test (if I read it correctly), and one question on whether we should warn

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: I've added a test and also moved the code so it's more consistent with the previous path.

heitorlessa (approved):
thank you!! this looks great! 🚀

leandrodamascena (approved):

leandrodamascena (approved):

rubenfonseca (approved):

PR #1914, opened 2023-02-08
feat(idempotency): adding redis as idempotency backend

Vandita2020:
**Issue number:** #1181

## Summary
PR adds support for Redis as Idempotecny backend

The following features for Redis are currently supported by this PR

- Add, update, get and delete idempotency record.
- Standalone and Cluster architecture for Redis. The third type Sentinel will be added to this soon.
- The database index, which is kind of namespace for information to be saved/stored between them.
- User ACL
- TTL support

### User experience

> Please share what the user experience looks like before and after this change

User can now use Redis as backend. You can connect it using Redis Standalone or Redis Cluster as shown in the example.

```python
from aws_lambda_powertools.utilities.connections import RedisStandalone, RedisCluster
from aws_lambda_powertools.utilities.idempotency import (
    idempotent,
    RedisCachePersistenceLayer,
    IdempotencyConfig
)    

# For connection using Redis Standalone architecture
redis_connection = RedisStandalone(host="192.168.68.112", port=6379, password="pass", db_index=0) 

# For connection using Redis Cluster architecture
#redis_connection = RedisCluster(url="redis://:bitnami@192.168.68.112:6379/0")

persistence_layer = RedisCachePersistenceLayer(connection=redis_connection)
config =  IdempotencyConfig(
    expires_after_seconds=1*60,  # 1 minutes
)

@idempotent(config=config, persistence_store=persistence_layer)
def lambda_handler(event, context):
    return {"message":"Hello"}
```

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [x] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented
* [x] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #2567, opened 2023-06-23
feat(idempotency): adding redis as idempotency backend

### LINKED ISSUE #1181: Add support for MemoryDB/ElasticCache/Redis as Idempotency backend
Updated topic to mention ElasticCache and general Redis as well.

https://aws.amazon.com/about-aws/whats-new/2021/08/amazon-memorydb-redis/

roger-zhangg:
**Issue number:** #1181
**previous pr:** [#1914](https://github.com/aws-powertools/powertools-lambda-python/pull/1914)

## Summary
PR adds support for Redis as Idempotecny backend

The following features for Redis are currently supported by this PR

* Add, update, get and delete idempotency record.
* Standalone and Cluster architecture for Redis. The third type Sentinel will be added to this soon.
* The database index, which is kind of namespace for information to be saved/stored between them.
* User ACL
* TTL support

User can now use Redis as backend. You can connect it using Redis Standalone or Redis Cluster as shown in the example.

```python
    idempotent,
    RedisCachePersistenceLayer,
    IdempotencyConfig
)    

persistence_layer = RedisCachePersistenceLayer(host="192.168.68.112", port=6379, db_index=0)
config =  IdempotencyConfig(
    expires_after_seconds=1*60,  # 1 minutes
)

@idempotent(config=config, persistence_store=persistence_layer)
    return {"message":"Hello"}
```

## Checklist

* [x]  [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [x]  I have performed a self-review of this change
* [ ]  Changes have been tested
* [x]  Changes are documented
* [x]  PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
Hey @roger-zhangg is this ready for review?

roger-zhangg:
Hi @heitorlessa Thanks for the reply! This one is still under active discussion with @leandrodamascena . Can review it once we finalize confusions and documents.

tomaszdudek7:
@roger-zhangg Is there any ETA for this feature? I desperately need a Redis alternative, as DDB is quite costly, especially with large volumes of traffic and on-demand mode. 

Also - thank you for working on this! <3

leandrodamascena:
> @roger-zhangg Is there any ETA for this feature? I desperately need a Redis alternative, as DDB is quite costly, especially with large volumes of traffic and on-demand mode.
> 
> Also - thank you for working on this! <3

Hi @tomaszdudek7! Thanks for adding this comment. Knowing that our customers want this Redis support as an Idempotency backend is really important to us because helps to organize our board and prioritize internally. We are currently working on merging PR's #2194, #1998 and #2197 in the coming weeks.

We haven't nailed down an exact ETA for this PR, but we can ballpark a reasonable deadline for the end of August. We'll do our best to get this done within the given timeframe, but it's just an estimate. If anything changes, we'll let you know and adjust accordingly.

leandrodamascena:
Hello @roger-zhangg! I'm going to go over the code today and create some bullet points on items that we can work together to get this PR ready to merge. I see a possible problem in the execution logic of the Idempotency, but I need to review in more detail.

aws_lambda_powertools/utilities/connections/redis.py
```diff
@@ -0,0 +1,108 @@
...
+class RedisConnection(BaseConnectionSync):
+    def __init__(
+        self,
+        host: Optional[str] = None,
+        port: Optional[int] = None,
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        db_index: Optional[int] = None,
+        url: Optional[str] = None,
+        **extra_options,
```
leandrodamascena: Hello @roger-zhangg! What do you think instead of adding the type of connection like `kwargs` we create a specific parameter for it, we assume "Standalone" as default (in case of not informing) and we return the established connection?

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -37,7 +37,7 @@ class DataRecord:
 
     def __init__(
         self,
-        idempotency_key: str,
+        idempotency_key: str = "",
```
leandrodamascena: We need to double-check if this can be a potential problem.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -32,7 +32,7 @@ class DynamoDBPersistenceLayer(BasePersistenceLayer):
     def __init__(
         self,
         table_name: str,
-        key_attr: str = "id",
+        key_attr: str = "id",  # remove optional for type checking.
```
leandrodamascena: Should we remove this comment?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,149 @@
...
+
+            logger.debug(f"Putting record on Redis for idempotency key: {data_record.idempotency_key}")
+            self._connection.hset(**item)
+            # hset type must set expiration after adding the record
+            # Need to review this to get ttl in seconds
+            # Q: should we replace self.expires_after_seconds with _get_expiry_timestamp? more consistent
+            self._connection.expire(name=data_record.idempotency_key, time=self.expires_after_seconds)
+        except Exception:
+            logger.debug(f"Failed to put record for already existing idempotency key: {data_record.idempotency_key}")
+            raise IdempotencyItemAlreadyExistsError
```
leandrodamascena: I'm creating a local project to validate this logic. Were you able to validate this logic by running a local Lambda (with SAM for example) and see how Idempotency works? I feel this logic is missing something, but I need to inspect more.

Makefile
```diff
@@ -30,12 +31,16 @@ lint-docs-fix:
...
-	poetry run pytest -m "not perf" --ignore tests/e2e --cov=aws_lambda_powertools --cov-report=xml
+	poetry run pytest -m "not perf" --ignore tests/e2e --ignore tests/integration --cov=aws_lambda_powertools --cov-report=xml
 	poetry run pytest --cache-clear tests/performance
 
 unit-test:
 	poetry run pytest tests/unit
 
+int-test-idem:
+	docker run --name int-test-idem -d -p 63005:6379 redis
+	poetry run pytest tests/integration/idempotency;docker stop int-test-idem;docker rm int-test-idem
```
leandrodamascena: I liked it, great catch! I think we should rename it to `test-idempotency-redis`.

leandrodamascena (changes requested):
Hello @roger-zhangg! I made some comments.

I'm still testing the idempotency logic in the `_put_record` method to validate that we're not missing something.

In our current idempotency documentation you can find [request flow](https://docs.powertools.aws.dev/lambda/python/2.22.0/utilities/idempotency/#idempotency-request-flow). It might be interesting if we can validate this logic using these diagrams as a base.

Thanks

leandrodamascena:
Can you fix the conflicts before we move on? Thanks!

roger-zhangg:
Looking into it, thanks for the review&comment!

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
roger-zhangg: I tried it locally for basic functionality(a random generator that return same number after idem record hits), it works as I imaged, do you have a specific case I should focus on?

aws_lambda_powertools/utilities/connections/redis.py
roger-zhangg: Yeah we can change to this more straight forward way

roger-zhangg:
## A bullet points for what's remaining to discuss in this PR
- Review on Mock Redis implementation
    - I implemented a mock Redis client for tests but haven't been reviewed yet.
- integration test renaming
    - now that we have fake Redis in function test, should we just abandon int test?
- check out what other tests we want to cover, current tests
    - `test_idempotent_lambda_redis_no_decode`
      - -> check `decode_responses=True`, throw IdempotencyRedisClientConfigError if this is False (we rely on this option or all Redis response will be in bytes)
    - `test_idempotent_function_and_lambda_handler_redis_basic`
      - -> test if the basic function can run
    - `test_idempotent_function_and_lambda_handler_redis_cache`
      - -> call function twice and gets the same result, call with different idem key and gets diff result
    - `test_idempotent_lambda_redis_in_progress`
      - -> call function using the idem key that match in-progress in redis store, return IdempotencyItemAlreadyExistsError
    - `test_idempotent_lambda_redis_delete`
      - -> call function and delete idem record, call function again with same idem key, gets diff result
    - `test redis acl`
      - -> I removed this because this seems like just testing redis, instead of testing our idem functionality.
- finalize Connection setup
    - There's a bunch of pros for allowing customer to pass in the redis connections they have already established.(we don't need to take care of stdalone/cluster/sentinel setup, don't need to take care of credential). If we still want to provide a connection class for our user to pass in their connections. I would suggest we follow the DynamoDB approach, which is we allow our customer to bring their own connection as well as support pass in credential and we create the connection for them.
    - https://github.com/aws-powertools/powertools-lambda-python/blob/b90891d3b3a5d350a0b90424f5323a9fdf1e68c8/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py#L90-L95
- Docs update: coming later today
- Code polishing: polish the code with Heitor's suggestions, I intend to do this after we have a final decision on Connection setup.

roger-zhangg:
Hey @leandrodamascena This is now ready for review.

leandrodamascena:
Hey @roger-zhangg! I pushed a commit with some changes to simplify the developer experience:

1 - We don't need the Config Class, we can pass the parameters during the initialization and choose between the client that the customer passes or create a connection if needed.
2 - We need to fix tests after this adjustment.
3 - Mypy is complaining about some errors, but it's easy to fix.
4 - Can you create an [Amazon ElastiCache](https://aws.amazon.com/elasticache/) in your account and test its use? You need to place your Lambda inside the VPC. I'm testing here and I was successful, but it's good for someone else to do an exploratory test with multiple invocations, concurrency, and situations where Lambda fails.
5 - After we finish these tests we can review the documentation.

Thanks.

roger-zhangg:
Redis operation failed/Function time out just after we created idem record, and just before we set a timeout on the idem record
(Ophan Record: where a record failed to auto delete in Redis)
- during _put_record, if failed to execute hsetnx(means key already exist)
- do a fetch on redis, vaildate with status, time. to check if record is an Ophan
    - if in_progress, check if in_progress_expiry_attr> now
    - if complete, check if expiry_timestamp < now
    - if above check failed, or keys not found means this record is Ophan.
-  if not Ophan, raise IdempotencyItemAlreadyExistsError
- if Ophan, overwrite

There are multiple benefits with this design. 
- Normal request will rarely hits idempotency record. Use Hsetnx only need one request to redis instead of two with prev solution (fetch/set)
- Hsetnx solve the racing condition when a record is set in another lambda after fetch
- We do checking(more than 1 redis request) only if Hsetnx failed. Which should be a rare case. Thus imporved performance.

leandrodamascena:
> 
> There are multiple benefits with this design.
> 
> * Normal request will rarely hits idempotency record. Use Hsetnx only need one request to redis instead of two with prev solution (fetch/set)
> * Hsetnx solve the racing condition when a record is set in another lambda after fetch
> * We do checking(more than 1 redis request) only if Hsetnx failed. Which should be a rare case. Thus imporved performance.

Hey @roger-zhangg! I really liked this improvement to improve the performance and make access to Redis faster than the previous implementation. We just need to fix the problem of refreshing the TTL in every invocation.

Next steps:
- [ ] Fix TTL
- [ ] mprove comments
- [ ] Make variable names more explicit and easier to understand
- [ ] Create a stress/load test

roger-zhangg:
In the previous commit I introduced a lock for safely overwrite Orphan records

Orphan records could be generated due to Previous runtime timeout, Redis expiry not working properly or a misbehavior in idemopotency code. In either case, we need to overwrite the Orphan record with new valid idempotency record.

However, due to the nature of write after read. This introduce a possibility for race condition, illustrated in the flowchart below.
image

- To prevent this race condition, I introduced a lock before overwriting the Orphan records. illustrated in the flowchart below.

image

leandrodamascena:
Hello @heitorlessa! We've almost finished the work here and we need a review before merging this. @roger-zhangg created a workflow on how an Idempotent record works in Redis, which is a little different from DynamoDB, and we validated several scenarios. Test coverage is ok and we are performing functional tests without the need to run a docker container. 
It might make sense to add the e2e test, but this is complex due to Elasticache Redis creation time, Lambda configuration with VPC and others. So, I think we can rely on functional tests.

This current review would be focused on:

1 - DX
2 - The code we created in Redis persistent code.
3 – Documentation
4 - Tests

Can we schedule a meeting this or the next week to validate this?

Thanks

aws_lambda_powertools/utilities/idempotency/persistence/base.py
roger-zhangg: Using empty idem key ("" or {}) will result in:
```UserWarning: No idempotency key value found. Skipping persistence layer and validation operations. jmespath: 
    idempotency_key = self._get_hashed_idempotency_key(data=data)
```
And idempotency will not work(handler executing) So this shouldn't be a problem anymore

roger-zhangg:
Added 2 test cases for simulating the race condition this Redis layer can get. Please check @leandrodamascena . Thanks!

leandrodamascena:
I'm working on tests with Redis in providers like Upstash and Redis Cloud and it is working as expected. Some details about my findings:

### Upstash

1 - Upstash supports the official library [redis-py](https://github.com/redis/redis-py) and we are using the same in our implementation.
2 - Upstash works by bringing the redis connection or using our `RedisCachePersistenceLayer` to create a new one.
3 - Upstash supports TLS just by informing the `ssl=True` parameter. @roger-zhangg we need to add this in our constructor, ok?
4 - When using Upstash with its redis-py implementation ([upstash-redis](https://github.com/upstash/redis-python)) we have some problems.

- 	4.1 - The method `set` defined in upstash-redis uses the name `key` for the key name, while redis-py uses `name`. We must need our implementation if want to support this.
-  4.2 - The upstash-redis library is a wrap over HTTP, which means we lose some controls we have when using the REDIS protocol and we need to change our implementation.
-  4.3 - In my tests, the upstash-redis library is a little slower than redis-py.
-  4.4 - I'm not sure if we need to support this library since Upstash works with the REDIS protocol/library by default.

### Redis Cloud

1 - Works as expected when using Redis Cloud, nothing to highlight here.

Thanks

roger-zhangg:
Create a bullet to track current progress

- [x] add a helper `abs_lambda_path` function for getting lambda filepath
- [x] using context manager to optimize logic on race lock
- [x] Protocol for Redis client (But need discussion on typing)
- [x] set default string value to ""
- [x] add an `ssl` parameter to connection
- [x] Doc: Modify Doc Order to show persistence layer before `Idempotency request flow`
- [x] Doc: change naming of "idempotent-decorator-for-redis" 
- [x] Doc: add elasticache setup template in `RedisPersistenceLayer`
- [x] Doc: create a example for using secret manager
- [x] Doc: create a example for using local cert file to establish ssl redis connection
- [x] e2e test: create e2e test using Amazon ElastiCache Serverless for Redis

leandrodamascena:
We are blocked from creating an e2e test due to this bug in CDK - https://github.com/aws/aws-cdk/issues/28246

I'm following the issue and as soon as it is fixed we can create one.

pyproject.toml
```diff
@@ -94,7 +95,8 @@ datadog-lambda = "^5.85.0"
 parser = ["pydantic"]
 validation = ["fastjsonschema"]
 tracer = ["aws-xray-sdk"]
-all = ["pydantic", "aws-xray-sdk", "fastjsonschema"]
+redis = ["redis"]
+all = ["pydantic", "aws-xray-sdk", "fastjsonschema","redis"]
```
leandrodamascena: Review before merge

rubenfonseca:
Starting reviewing now

.github/workflows/quality_check_pydanticv2.yml
```diff
@@ -58,7 +58,9 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: "poetry"
       - name: Replacing Pydantic v1 with v2 > 2.0.3
-        run: poetry add "pydantic=^2.0.3"
+        run: |
+          rm -rf poetry.lock
+          poetry add "pydantic=^2.0.3"
```
rubenfonseca, line 63: Do we still need to remove poetry.lock? Just making sure this is really necesary.

aws_lambda_powertools/utilities/idempotency/__init__.py
```diff
@@ -9,6 +9,18 @@
     DynamoDBPersistenceLayer,
 )
 
+# import RedisCachePersistenceLayer here mean we will need redis as a required lib? Do we want to make it optional?
```
rubenfonseca: Can we please address this comment before the next review?

aws_lambda_powertools/utilities/idempotency/exceptions.py
```diff
@@ -83,3 +83,21 @@ class IdempotencyNoSerializationModelError(BaseError):
...
+    """
+
+
+class IdempotencyRedisConnectionError(BaseError):
+    """
+    Redis connection error
+    """
+
+
+class IdempotencyOrphanRecordError(BaseError):
```
rubenfonseca: Since this is specific to Redis, can we rename it?

```suggestion
class IdempotencyRedisOrphanRecordError(BaseError):
```

rubenfonseca: Note: in general I'm not a fan of having exceptions that are specific to a persistence layer. I would be happier with generic names as:

- IdempotencyPersistenceConfigError
- IdempotencyPersistenceConnectionError
- IdempotencyPersistenceConsistencyError

One of the reasons is that we might add additional persistence layers in the future, and this file will only continue to grow bigger (+ cognitive load). WDYT?

aws_lambda_powertools/utilities/idempotency/persistence/base.py
rubenfonseca: Can you please explain me this change? I had the ideia that the idempotency_key is something we need explicitly.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -66,6 +66,8 @@ def __init__(
             DynamoDB attribute name for status, by default "status"
         data_attr: str, optional
             DynamoDB attribute name for response data, by default "data"
+        validation_key_attr: str, optional
+            DynamoDB attribute name for hashed representation of the parts of the event used for validation
```
rubenfonseca, line 70: ❤️

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+
+import datetime
+import json
+import logging
+from contextlib import contextmanager
+from datetime import timedelta
+from typing import Any, Dict
+
+import redis
+from typing_extensions import Literal, Protocol
```
rubenfonseca: Should we import from `shared/types` instead?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+from aws_lambda_powertools.utilities.idempotency.persistence.base import (
+    STATUS_CONSTANTS,
+    DataRecord,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class RedisClientProtocol(Protocol):
+    def get(self, name: bytes | str | memoryview) -> bytes | str | None:
```
rubenfonseca, line 65: Awesome! I thought we couldn't use the "|" syntax because of python 3.7, but tests pass!

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+)
+from aws_lambda_powertools.utilities.idempotency.persistence.base import (
+    STATUS_CONSTANTS,
+    DataRecord,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class RedisClientProtocol(Protocol):
```
rubenfonseca, line 29: Can we add documentation to this protocol? Including each method.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+    def __init__(
+        self,
+        url: str = "",
+        host: str = "",
+        port: int = 6379,
+        username: str = "",
+        password: str = "",  # nosec - password for Redis connection
+        db_index: int = 0,
+        mode: Literal["standalone", "cluster"] = "standalone",
+        ssl: bool = False,
```
rubenfonseca: Can we have this `true` by default? /cc @sthulb

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            redis host
+        port: int, optional: default 6379
+            redis port
+        username: str, optional
+            redis username
+        password: str, optional
+            redis password
+        url: str, optional
+            redis connection string, using url will override the host/port in the previous parameters
+        db_index: str, optional: default 0
```
rubenfonseca:
```suggestion
        db_index: int, optional: default 0
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        self,
+        url: str = "",
+        host: str = "",
+        port: int = 6379,
+        username: str = "",
+        password: str = "",  # nosec - password for Redis connection
+        db_index: int = 0,
+        mode: Literal["standalone", "cluster"] = "standalone",
+        ssl: bool = False,
+        ssl_cert_reqs: Literal["required", "optional", "none"] = "none",
```
rubenfonseca: Same here, should the default be `required`?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            client = redis.cluster.RedisCluster
+        else:
+            raise IdempotencyRedisClientConfigError(f"Mode {self.mode} not supported")
+
+        try:
+            if self.url:
+                logger.debug(f"Using URL format to connect to Redis: {self.host}")
+                return client.from_url(url=self.url)
+            else:
+                logger.debug(f"Using other parameters to connect to Redis: {self.host}")
```
rubenfonseca:
```suggestion
                logger.debug(f"Using arguments to connect to Redis: {self.host}")
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        self,
+        url: str = "",
+        host: str = "",
+        port: int = 6379,
+        username: str = "",
+        password: str = "",  # nosec - password for Redis connection
+        db_index: int = 0,
+        mode: Literal["standalone", "cluster"] = "standalone",
+        ssl: bool = False,
+        ssl_cert_reqs: Literal["required", "optional", "none"] = "none",
```
rubenfonseca: See note above about making SSL the default

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        client: RedisClientProtocol | None = None,
+        in_progress_expiry_attr: str = "in_progress_expiration",
+        expiry_attr: str = "expiration",
+        status_attr: str = "status",
+        data_attr: str = "data",
+        validation_key_attr: str = "validation",
+    ):
+        """
+        Initialize the Redis Persistence Layer
```
rubenfonseca, line 227:
```suggestion
        Initialize the Redis Persistence Layer
        
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            redis host
+        port: int, optional: default 6379
+            redis port
+        username: str, optional
+            redis username
+        password: str, optional
+            redis password
+        url: str, optional
+            redis connection string, using url will override the host/port in the previous parameters
+        db_index: str, optional: default 0
```
rubenfonseca:
```suggestion
        db_index: int, optional: default 0
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            redis db index
+        mode: str, Literal["standalone","cluster"]
+            set redis client mode, choose from standalone/cluster
+        ssl: bool, optional: default False
+            set whether to use ssl for Redis connection
+        ssl_cert_reqs: str, optional: default "none"
+            set whether to use ssl cert for Redis connection, choose from required/optional/none
+        client: RedisClientProtocol, optional
+            You can bring your established Redis client that follows RedisClientProtocol.
+            If client is provided, all connection config above will be ignored
```
rubenfonseca:
```suggestion
            Bring your own Redis client that follows RedisClientProtocol.
            If provided, all other connection configuration options will be ignored
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        self.data_attr = data_attr
+        self.validation_key_attr = validation_key_attr
+        self._json_serializer = json.dumps
+        self._json_deserializer = json.loads
+        super(RedisCachePersistenceLayer, self).__init__()
+        self._orphan_lock_timeout = min(10, self.expires_after_seconds)
+
+    def _get_expiry_second(self, expiry_timestamp: int | None = None) -> int:
+        """
+        return seconds of timedelta from now to the given unix timestamp
```
rubenfonseca:
```suggestion
	Calculates the number of seconds remaining until a specified expiry time
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        """
+        if expiry_timestamp:
+            return expiry_timestamp - int(datetime.datetime.now().timestamp())
+        return self.expires_after_seconds
+
+    def _item_to_data_record(self, idempotency_key: str, item: Dict[str, Any]) -> DataRecord:
+        in_progress_expiry_timestamp = item.get(self.in_progress_expiry_attr)
+        if isinstance(in_progress_expiry_timestamp, str):
+            in_progress_expiry_timestamp = int(in_progress_expiry_timestamp)
```
rubenfonseca: Any reason why we need to convert it to a integer all the time? Can't we store it as integer in Redis? Or is this a problem with the redis drivers?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            status=item[self.status_attr],
+            in_progress_expiry_timestamp=in_progress_expiry_timestamp,
+            response_data=str(item.get(self.data_attr)),
+            payload_hash=str(item.get(self.validation_key_attr)),
+            expiry_timestamp=item.get("expiration", None),
+        )
+
+    def _get_record(self, idempotency_key) -> DataRecord:
+        # See: https://redis.io/commands/get/
+        response = self.client.get(idempotency_key)
```
rubenfonseca, line 339:
```suggestion
        response = self.client.get(idempotency_key)
        
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        # key not found
+        if not response:
+            raise IdempotencyItemNotFoundError
```
rubenfonseca, line 343:
```suggestion
            raise IdempotencyItemNotFoundError
            
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        try:
+            item = self._json_deserializer(response)
+        except json.JSONDecodeError:
+            raise IdempotencyOrphanRecordError
```
rubenfonseca: Can we add a comment why failing JSON decoding is considered an Ophaned Record? For future reference it should be easy for the reader to understand why.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        return self._item_to_data_record(idempotency_key, item)
+
+    def _put_in_progress_record(self, data_record: DataRecord) -> None:
+        item: Dict[str, Any] = {}
+        item = {
```
rubenfonseca: I believe this two lines can be merged.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+
+            # The idempotency key does not exist:
+            #    - first time that this invocation key is used
+            #    - previous invocation with the same key was deleted due to TTL
+            #    - SET see https://redis.io/commands/set/
+
+            logger.debug(f"Putting record on Redis for idempotency key: {item['name']}")
+            encoded_item = self._json_serializer(item["mapping"])
+            ttl = self._get_expiry_second(expiry_timestamp=item["mapping"][self.expiry_attr])
+
```
rubenfonseca, line 393: Suggestion: can we fetch the arguments from data_record instead of doing all this dictionary lookups? Seems we are wasting a lot of cycles here.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            redis_response = self.client.set(name=item["name"], value=encoded_item, ex=ttl, nx=True)
+
+            # redis_response:True -> Redis set succeed, idempotency key does not exist before
+            # return to idempotency and proceed to handler execution phase. Most cases should return here
```
rubenfonseca:
```suggestion
	    # If redis_response is True, the Redis SET operation was successful and the idempotency key was not
	    # previously set. This indicates that we can safely proceed to the handler execution phase.
	    # Most invocations should successfully proceed past this point.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            if redis_response:
+                return
+
+            # redis_response:None -> Existing record on Redis, continue to checking phase
+            # The idempotency key exist:
+            #    - previous invocation with the same key and not expired(active idempotency)
+            #    - previous invocation timed out (Orphan Record)
+            #    - previous invocation record expired but not deleted by Redis (Orphan Record)
```
rubenfonseca:
```suggestion
	    # If redis_response is None, it indicates an existing record in Redis for the given idempotency key.
	    # This could be due to:
	    # - An active idempotency record from a previous invocation that has not yet expired.
	    # - An orphan record where a previous invocation has timed out.
            # - An expired idempotency record that has not been deleted by Redis.
	    # In any case, we proceed to retrieve the record for further inspection.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+
+            idempotency_record = self._get_record(item["name"])
+
+            # status is completed and expiry_attr timestamp still larger than current timestamp
+            # found a valid completed record
```
rubenfonseca:
```suggestion
            # If the status of the idempotency record is 'COMPLETED' and the record has not expired
            # (i.e., the expiry timestamp is greater than the current timestamp), then a valid completed
            # record exists. We raise an error to prevent duplicate processing of a request that has already
            # been completed successfully.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            if idempotency_record.status == STATUS_CONSTANTS["COMPLETED"] and not idempotency_record.is_expired:
+                raise IdempotencyItemAlreadyExistsError
+
+            # in_progress_expiry_attr exist means status is in_progress, and still larger than current timestamp,
+            # found a vaild in_progress record
```
rubenfonseca:
```suggestion
            # If the idempotency record has a status of 'INPROGRESS' and has a valid in_progress_expiry_timestamp
            # (meaning the timestamp is greater than the current timestamp in milliseconds), then we have encountered
            # a valid in-progress record. This indicates that another process is currently handling the request, and
            # to maintain idempotency, we raise an error to prevent concurrent processing of the same request.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            if (
+                idempotency_record.status == STATUS_CONSTANTS["INPROGRESS"]
+                and idempotency_record.in_progress_expiry_timestamp
+                and idempotency_record.in_progress_expiry_timestamp > int(now.timestamp() * 1000)
+            ):
+                raise IdempotencyItemAlreadyExistsError
+
+            # If the code reaches here means we found an Orphan record.
```
rubenfonseca:
```suggestion
            # Reaching this point indicates that the idempotency record found is an orphan record. An orphan record is
            # one that is neither completed nor in-progress within its expected time frame. It may result from a
            # previous invocation that has timed out or an expired record that has yet to be cleaned up by Redis.
            # We raise an error to handle this exceptional scenario appropriately.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            raise IdempotencyOrphanRecordError
+
+        except IdempotencyOrphanRecordError:
```
rubenfonseca: Instead of raising and immediately except, why not just handle it directly?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            # deal with orphan record here
+            # aquire a lock for default 10 seconds
```
rubenfonseca:
```suggestion
            # Handle an orphan record by attempting to acquire a lock, which by default lasts for 10 seconds.
            # The purpose of acquiring the lock is to prevent race conditions with other processes that might
            # also be trying to handle the same orphan record. Once the lock is acquired, we set a new value
            # for the idempotency record in Redis with the appropriate time-to-live (TTL).
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            with self._acquire_lock(name=item["name"]):
+                self.client.set(name=item["name"], value=encoded_item, ex=ttl)
+
+            # lock was not removed here intentionally. Prevent another orphan operation in race condition.
```
rubenfonseca: Sorry this was a little bit hard for me to understand. What about this?

```suggestion
            # Not removing the lock here serves as a safeguard against race conditions,
            # preventing another operation from mistakenly treating this record as an orphan while the
            # current operation is still in progress.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        except (redis.exceptions.RedisError, redis.exceptions.RedisClusterException) as e:
+            raise e
+        except Exception as e:
+            logger.debug(f"encountered non-redis exception:{e}")
```
rubenfonseca:
```suggestion
            logger.debug(f"encountered non-redis exception: {e}")
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+            raise e
+
+    @contextmanager
+    def _acquire_lock(self, name: str):
+        """
+        aquire a lock for default 10 seconds
```
rubenfonseca:
```suggestion
        Attempt to acquire a lock for a specified resource name, with a default timeout.

        This context manager attempts to set a lock using Redis to prevent concurrent
        access to a resource identified by 'name'. It uses the 'nx' flag to ensure that
        the lock is only set if it does not already exist, thereby enforcing mutual exclusion.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        """
+        try:
+            acquired = self.client.set(name=f"{name}:lock", value="True", ex=self._orphan_lock_timeout, nx=True)
+            logger.debug("acquiring lock to overwrite orphan record")
+            if acquired:
+                logger.debug("lock acquired")
+                yield
+            else:
+                # lock failed to aquire, means encountered a race condition. Just return
```
rubenfonseca:
```suggestion
                # If the lock acquisition fails, it suggests a race condition has occurred. In this case, instead of
                # proceeding, we log the event and raise an error to indicate that the current operation should be
                # retried after the lock is released by the process that currently holds it.
```

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+
+        finally:
+            ...
+
+    def _put_record(self, data_record: DataRecord) -> None:
+        if data_record.status == STATUS_CONSTANTS["INPROGRESS"]:
+            self._put_in_progress_record(data_record=data_record)
+        else:
+            # current this function only support set in_progress. set complete should use update_record
+            raise NotImplementedError
```
rubenfonseca, line 481: This looks like a leaky abstraction. What about calling `update_record` instead of raising?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+
+    def _update_record(self, data_record: DataRecord) -> None:
+        item: Dict[str, Any] = {}
+
+        item = {
```
rubenfonseca: Can we merge this lines?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
```diff
@@ -0,0 +1,440 @@
...
+        }
+        logger.debug(f"Updating record for idempotency key: {data_record.idempotency_key}")
+        encoded_item = self._json_serializer(item["mapping"])
+        ttl = self._get_expiry_second(data_record.expiry_timestamp)
+        # need to set ttl again, if we don't set ex here the record will not have a ttl
+        self.client.set(name=item["name"], value=encoded_item, ex=ttl)
+
+    def _delete_record(self, data_record: DataRecord) -> None:
+        # This function only works when Lambda handler has already been invoked
+        # Or you'll get empty idempotency_key
```
rubenfonseca:
```suggestion
        """
        Deletes the idempotency record associated with a given DataRecord from Redis.

        This function is designed to be called after a Lambda handler invocation has completed processing.
        It ensures that the idempotency key associated with the DataRecord is removed from Redis to
        prevent future conflicts and to maintain the idempotency integrity.

        Note: it is essential that the idempotency key is not empty, as that would indicate the Lambda
        handler has not been invoked or the key was not properly set.
        """
```

docs/utilities/idempotency.md
```diff
@@ -14,6 +14,7 @@ The idempotency utility provides a simple solution to convert your Lambda functi
 * Select a subset of the event as the idempotency key using JMESPath expressions
 * Set a time window in which records with the same payload should be considered duplicates
 * Expires in-progress executions if the Lambda function times out halfway through
+* Support Amazon DynamoDB and Redis as persistence layer
```
rubenfonseca:
```suggestion
* Support Amazon DynamoDB and Redis as persistence layers
```

docs/utilities/idempotency.md
```diff
@@ -51,6 +52,9 @@ classDiagram
 
 ## Getting started
 
+???+ note
+    This section uses DynamoDB as default idempotent persistence storage layer. If you are interested in using Redis as persistence storage layer, Check out the [Redis as persistence storage layer](#redis-as-persistent-storage-layer-provider) Section.
```
rubenfonseca:
```suggestion
    This section uses DynamoDB as the default idempotent persistence storage layer. If you are interested in using Redis as the persistence storage layer, check out the [Redis as persistence storage layer](#redis-as-persistent-storage-layer-provider) section.
```

docs/utilities/idempotency.md
```diff
@@ -62,7 +66,7 @@ Your Lambda function IAM Role must have `dynamodb:GetItem`, `dynamodb:PutItem`,
 
 Before getting started, you need to create a persistent storage layer where the idempotency utility can store its state - your lambda functions will need read and write access to it.
 
-As of now, Amazon DynamoDB is the only supported persistent storage layer, so you'll need to create a table first.
+We currently support Amazon DynamoDB and Redis as a storage layer. This example demonstrates how to create a table in DynamoDB. If you want to use Redis, please go to the section [RedisPersistenceLayer](#redispersistencelayer)
```
rubenfonseca:
```suggestion
We currently support Amazon DynamoDB and Redis as storage layers. The following example demonstrates how to create a table in DynamoDB. If you prefer to use Redis, please refer to the [RedisPersistenceLayer](#redispersistencelayer) section.
```

docs/utilities/idempotency.md
```diff
@@ -333,6 +337,51 @@ If an Exception is raised _outside_ the scope of the decorated function and afte
 
     As this happens outside the scope of your decorated function, you are not able to catch it if you're using the `idempotent` decorator on your Lambda handler.
 
+### Persistence layers
+
+#### DynamoDBPersistenceLayer
+
+This persistence layer is built-in, and you can either use an existing DynamoDB table or create a new one dedicated for idempotency state (recommended).
```
rubenfonseca:
```suggestion
This persistence layer is built-in, allowing you to use an existing DynamoDB table or create a new one dedicated to idempotency state (recommended).
```

docs/utilities/idempotency.md
```diff
@@ -333,6 +337,51 @@ If an Exception is raised _outside_ the scope of the decorated function and afte
...
+
+=== "Customizing DynamoDBPersistenceLayer to suit your table structure"
+
+    ```python hl_lines="7-15"
+    --8<-- "examples/idempotency/src/customize_persistence_layer.py"
+    ```
+
+When using DynamoDB as a persistence layer, you can alter the attribute names by passing these parameters when initializing the persistence layer:
```
rubenfonseca:
```suggestion
When using DynamoDB as the persistence layer, you can customize the attribute names by passing the following parameters during the initialization of the persistence layer:
```

docs/utilities/idempotency.md
```diff
@@ -333,6 +337,51 @@ If an Exception is raised _outside_ the scope of the decorated function and afte
...
+| **data_attr**               |                    | `data`                               | Stores results of successfully executed Lambda handlers                                                  |
+| **validation_key_attr**     |                    | `validation`                         | Hashed representation of the parts of the event used for validation                                      |
+| **sort_key_attr**           |                    |                                      | Sort key of the table (if table is configured with a sort key).                                          |
+| **static_pk_value**         |                    | `idempotency#{LAMBDA_FUNCTION_NAME}` | Static value to use as the partition key. Only used when **sort_key_attr** is set.                       |
+
+#### RedisPersistenceLayer
+
+This persistence layer is built-in, and you can use an existing Redis service. For optimal performance and compatibility, we strongly advise using a Redis service version 7 or higher.
```
rubenfonseca:
```suggestion
This persistence layer is built-in, allowing you to use an existing Redis service. For optimal performance and compatibility, it is strongly recommended to use a Redis service version 7 or higher.
```

docs/utilities/idempotency.md
```diff
@@ -333,6 +337,51 @@ If an Exception is raised _outside_ the scope of the decorated function and afte
...
+
+=== "Customizing RedisPersistenceLayer to suit your data structure"
+
+    ```python hl_lines="14-20"
+    --8<-- "examples/idempotency/src/customize_persistence_layer_redis.py"
+    ```
+
+When using Redis as a persistence layer, you can alter the attribute names by passing these parameters when initializing the persistence layer:
```
rubenfonseca:
```suggestion
When using Redis as the persistence layer, you can customize the attribute names by providing the following parameters upon initialization of the persistence layer:
```

docs/utilities/idempotency.md
```diff
@@ -333,6 +337,51 @@ If an Exception is raised _outside_ the scope of the decorated function and afte
...
+
+| Parameter                   | Required           | Default                              | Description                                                                                              |
+| --------------------------- | ------------------ | ------------------------------------ | -------------------------------------------------------------------------------------------------------- |
+| **in_progress_expiry_attr** |                    | `in_progress_expiration`             | Unix timestamp of when record expires while in progress (in case of the invocation times out)            |
+| **status_attr**             |                    | `status`                             | Stores status of the lambda execution during and after invocation                                        |
```
rubenfonseca:
```suggestion
| **status_attr**             |                    | `status`                             | Stores status of the Lambda execution during and after invocation                                        |
```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
 </center>
 
-## Advanced
+## Redis as persistent storage layer provider
 
-### Persistence layers
+### Redis resources
 
-#### DynamoDBPersistenceLayer
+You need an existing Redis service before setting up Redis as persistent storage layer provider. You can also use Redis compatible services like [Amazon ElastiCache for Redis](https://aws.amazon.com/elasticache/redis/){target="_blank"} or [Amazon MemoryDB for Redis](https://aws.amazon.com/memorydb/){target="_blank"} as persistent storage layer provider.
```
rubenfonseca:
```suggestion
Before setting up Redis as the persistent storage layer provider, you must have an existing Redis service. Alternatively, you can use Redis compatible services such as [Amazon ElastiCache for Redis](https://aws.amazon.com/elasticache/redis/){target="_blank"} or [Amazon MemoryDB for Redis](https://aws.amazon.com/memorydb/){target="_blank"} as your persistent storage layer provider.

```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
+???+ tip "No existing Redis service?"
+    If you don't have an existing Redis service, we recommend using [DynamoDB](#dynamodbpersistencelayer) as persistent storage layer provider.
```
rubenfonseca:
```suggestion
    If you don't have an existing Redis service, we recommend using [DynamoDB](#dynamodbpersistencelayer) as the persistent storage layer provider.
```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
+    --8<-- "examples/idempotency/templates/cfn_redis_serverless.yaml"
+    ```
 
-    ```python hl_lines="7-15"
-    --8<-- "examples/idempotency/src/customize_persistence_layer.py"
+    1. Replace the Security Group ID and Subnet ID to match your VPC settings.
+
+### VPC Access
+
+Your Lambda Function must be able to reach the Redis endpoint before using it for idempotency persistent storage layer. In most cases you will need to [configure VPC access](https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html){target="_blank"} for your Lambda Function.
```
rubenfonseca:
```suggestion
Your Lambda Function must have network access to the Redis endpoint before using it as the idempotency persistent storage layer. In most cases, you will need to [configure VPC access](https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html){target="_blank"} for your Lambda Function.
```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
+
+???+ tip "Amazon ElastiCache/MemoryDB for Redis as persistent storage layer provider"
+    If you intend to use Amazon ElastiCache for Redis for idempotency persistent storage layer, you can also reference [This AWS Tutorial](https://docs.aws.amazon.com/lambda/latest/dg/services-elasticache-tutorial.html){target="_blank"}.
+    If you are using Amazon MemoryDB for Redis, reference [This AWS Tutorial](https://aws.amazon.com/blogs/database/access-amazon-memorydb-for-redis-from-aws-lambda/){target="_blank"} for only VPC setup part.
```
rubenfonseca:
```suggestion
???+ tip "Amazon ElastiCache/MemoryDB for Redis as persistent storage layer provider"
    If you plan to use Amazon ElastiCache for Redis as the idempotency persistent storage layer, you may find [this AWS tutorial](https://docs.aws.amazon.com/lambda/latest/dg/services-elasticache-tutorial.html){target="_blank"} helpful.
    For those using Amazon MemoryDB for Redis, refer to [this AWS tutorial](https://aws.amazon.com/blogs/database/access-amazon-memorydb-for-redis-from-aws-lambda/){target="_blank"} specifically for the VPC setup guidance.
```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
+
+After VPC setup, you can follow the templates down below to setup Lambda functions with VPC internal subnet access.
```
rubenfonseca:
```suggestion
After completing the VPC setup, you can use the templates provided below to set up Lambda functions with access to VPC internal subnets.
```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
-| **data_attr**               |                    | `data`                               | Stores results of successfully executed Lambda handlers                                                  |
-| **validation_key_attr**     |                    | `validation`                         | Hashed representation of the parts of the event used for validation                                      |
-| **sort_key_attr**           |                    |                                      | Sort key of the table (if table is configured with a sort key).                                          |
-| **static_pk_value**         |                    | `idempotency#{LAMBDA_FUNCTION_NAME}` | Static value to use as the partition key. Only used when **sort_key_attr** is set.                       |
+### Configuring Redis persistence layer
+
+You can quickly start by initializing the `RedisCachePersistenceLayer` class and using it with the `idempotent` decorator on your lambda handler. Check out detailed example of `RedisCachePersistenceLayer` in [Persistence layers section](#redispersistencelayer).
```
rubenfonseca:
```suggestion
You can quickly get started by initializing the `RedisCachePersistenceLayer` class and applying the `idempotent` decorator to your Lambda handler. For a detailed example of using the `RedisCachePersistenceLayer`, refer to the [Persistence layers section](#redispersistencelayer).
```

docs/utilities/idempotency.md
```diff
@@ -527,33 +576,93 @@ sequenceDiagram
...
+
+=== "Sample event"
+
+    ```json
+    --8<-- "examples/idempotency/src/getting_started_with_idempotency_payload.json"
+    ```
+
+### Custom advanced settings
+
+For advanced settings, including SSL certificates and the ability to customize parameters such as a custom timeout, you can use the Redis client to accommodate these specific settings.
```
rubenfonseca:
```suggestion
For advanced configurations, such as setting up SSL certificates or customizing parameters like a custom timeout, you can utilize the Redis client to tailor these specific settings to your needs.
```

docs/utilities/idempotency.md
```diff
@@ -854,6 +963,40 @@ This means it is possible to pass a mocked Table resource, or stub various metho
     --8<-- "examples/idempotency/tests/app_test_io_operations.py"
     ```
 
+### Testing with Redis
+
+To test locally, You can either utilize [fakeredis-py](https://github.com/cunla/fakeredis-py) or check out the [MockRedis](https://github.com/aws-powertools/powertools-lambda-python/blob/ba6532a1c73e20fdaee88c5795fd40e978553e14/tests/functional/idempotency/persistence/test_redis_layer.py#L34-L66) Class we used in our test.
```
rubenfonseca:
```suggestion
To test locally, you can either utilize [fakeredis-py](https://github.com/cunla/fakeredis-py) for a simulated Redis environment or refer to the [MockRedis](https://github.com/aws-powertools/powertools-lambda-python/blob/ba6532a1c73e20fdaee88c5795fd40e978553e14/tests/functional/idempotency/persistence/test_redis_layer.py#L34-L66) class used in our tests to mock Redis operations.
```

docs/utilities/idempotency.md
```diff
@@ -854,6 +963,40 @@ This means it is possible to pass a mocked Table resource, or stub various metho
...
+    --8<-- "examples/idempotency/tests/test_with_mock_redis.py"
+    ```
+
+=== "mock_redis.py"
+
+    ```python
+    --8<-- "examples/idempotency/tests/mock_redis.py"
+    ```
+
+If you want to actually setup a Real Redis client for integration test, reference the code below
```
rubenfonseca:
```suggestion
If you want to set up a real Redis client for integration testing, you can reference the code provided below.
```

tests/e2e/utils/infrastructure.py
```diff
@@ -177,9 +177,9 @@ def deploy(self) -> Dict[str, str]:
             CloudFormation Stack Outputs with output key and value
         """
         stack_file = self._create_temp_cdk_app()
-        synth_command = f"npx cdk synth --app 'python {stack_file}' -o {self._cdk_out_dir}"
+        synth_command = f"npx cdk@latest synth --app 'python {stack_file}' -o {self._cdk_out_dir}"
```
rubenfonseca: This is not a good idea as it will create mismatches with the pinned version of CDK in our deps. We rely on dependabot to update the CDK version everywhere (JS and Python).

tests/e2e/utils/infrastructure.py
```diff
@@ -177,9 +177,9 @@ def deploy(self) -> Dict[str, str]:
...
         deploy_command = (
-            f"npx cdk deploy --app '{self._cdk_out_dir}' -O {self._stack_outputs_file} "
+            f"npx cdk@latest deploy --app '{self._cdk_out_dir}' -O {self._stack_outputs_file} "
```
rubenfonseca: This is not a good idea as it will create mismatches with the pinned version of CDK in our deps. We rely on dependabot to update the CDK version everywhere (JS and Python).

rubenfonseca (changes requested):
Thanks for the hard work!!

Most of my comments are some nitpicks. The majority of them where about adding or rewording comments because it took me a long time to understand the code, and I want to make the life easier for the next reader :)

Besides the comments, I want to highlight two things:

- **inconsistency of "redis" vs "Redis"**: in the documentation, we freely move between the two versions. Can we decide on a single one and use it everywhere?
- **JSON SerDe**: waiting from @leandrodamascena to confirm that GET/SET + JSON SerDe is not slower than using HSETNX/HGET

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
roger-zhangg: Multiple case can raise `IdempotencyOrphanRecordError`, like json decode error above so it's better to have a centailized place like this to handle the error

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
roger-zhangg: yes we can use data_record instead

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
roger-zhangg: Yes, I'll add it. Json decoding error is considered an Orphan Record cause this scenario will also introduce possible race condition just like Orphan record. As two lambda handlers is possible to reach this line of code almost simultaneously. If we simply regard this record as non-valid record. The two lambda handlers will both start the overwrite process without knowing each other. Causing this value being overwritten twice (ultimately two Lambda Handlers will both be executed, which is against idempotency). Behaving exactly the same as Orphan Record does. So this case should also be handled by the error handling in `except IdempotencyOrphanRecordError` part to avoid the possible race condition.

.github/workflows/quality_check_pydanticv2.yml
leandrodamascena: Sometimes Poetry crashes because it can't resolve dependencies in our pipeline and it takes a long time. By removing `poetry.lock` we can be sure that Poetry will install Pydantic correctly and without problems.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: We are importing `from __future__ import annotations` in the first line and it is safe to add this kind of type annotation. We use this in other parts of our codebase, like metrics and batch processing, for example.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: This is a good question! By default, Elasticache Redis 7.0 (serverless, standalone, and cluster) comes with in-transit encryption enabled by default, but the customer can disable it. Other Redis caches may come without TLS enabled. But I agree that Elasticache is our primer resource and we must enable this.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: No. We only need this parameter if the customer uses a self-signed certificate, which is not common in most cases. Maybe this field isn't even necessary here and if the cusomer wants to bring self-signed certifications, they can bring their instance of the Redis cli.

References: 
https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/connect-tls.html

https://redis-py.readthedocs.io/en/stable/examples/ssl_connection_examples.html#Connecting-to-a-Redis-instance-via-SSL,-while-specifying-a-self-signed-SSL-certificate.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: This is a good question! By default, Elasticache Redis 7.0 (serverless, standalone, and cluster) comes with in-transit encryption enabled by default, but the customer can disable it. Other Redis caches may come without TLS enabled. But I agree that Elasticache is our primer resource and we must enable this.

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: I did not get it! Are you using the same code in the review?

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: I did not get it! Are you using the same code in the review?

tests/e2e/utils/infrastructure.py
leandrodamascena: You're 100% right, Ruben! It was a leftover from my local tests.
Removed

tests/e2e/utils/infrastructure.py
leandrodamascena: You're 100% right, Ruben! It was a leftover from my local tests.
Removed

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
leandrodamascena: Done

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
roger-zhangg: double checked and we can remove this conversion. Thanks!

aws_lambda_powertools/utilities/idempotency/persistence/redis.py
roger-zhangg: Thanks for the suggestion and refine to all my comments. You are amazing @rubenfonseca !

leandrodamascena:
> Thanks for the hard work!!
> 
> Most of my comments are some nitpicks. The majority of them where about adding or rewording comments because it took me a long time to understand the code, and I want to make the life easier for the next reader :)

Thanks for this detailed review, @rubenfonseca! I'm 100% sure that we will release a better version after this review.

> 
> Besides the comments, I want to highlight two things:
> 
> * **inconsistency of "redis" vs "Redis"**: in the documentation, we freely move between the two versions. Can we decide on a single one and use it everywhere?

It's better to use Redis and I changed the documentation to use Redis.

> * **JSON SerDe**: waiting from @leandrodamascena to confirm that GET/SET + JSON SerDe is not slower than using HSETNX/HGET

This is an excellent point, Ruben! This PR had changes and optimizations throughout its development and one of the modifications was the migration of `HSETNX/HGETALL` to simplify with `GET/SET` and use JSON SerDe.
To use `HSETNX/HGETALL` we have a change of concept in this implementation, as the traditional way is to make an `HSETNX` call for each key (in the dict) that we want to store in Redis as an object, after which we have to set ttl for all of them. But at this point, we made an optimization using the concept of `pipelines` in Redis, which optimizes latency, as it uses batch to write to Redis and reduces RTT, one of the main reasons for latency in Redis. But this makes the code more verbose and the business logic more complicated, in addition to the fact that the batch writing has to be transactional (this is by default), and in tests, this took even longer than using set/get with JSON SerDe, in some cases.

I recovered the [old code](https://github.com/aws-powertools/powertools-lambda-python/pull/2567/commits/7ceb45e203412941e2c87e17d14152d8cc7626ad), ran some tests, and saw no consistent difference in execution time between the two. My opinion is that we can continue with this implementation, and perhaps wait for feedback from some customers.

What do you think about?

leandrodamascena:
Hello @rubenfonseca! We are ready for you to review this PR a second time.

cc @roger-zhangg!

aws_lambda_powertools/utilities/idempotency/exceptions.py
```diff
@@ -83,3 +83,21 @@ class IdempotencyNoSerializationModelError(BaseError):
     """
     No model was supplied to the serializer
     """
+
+
+class IdempotencyPersistenceConfigError(BaseError):
+    """
+    The Redis connection passed in has unsupported config
```
rubenfonseca:
```suggestion
    The idempotency persistency configuration was unsupported
```

aws_lambda_powertools/utilities/idempotency/exceptions.py
```diff
@@ -83,3 +83,21 @@ class IdempotencyNoSerializationModelError(BaseError):
...
+    """
+
+
+class IdempotencyPersistenceConnectionError(BaseError):
+    """
+    Redis connection error
```
rubenfonseca:
```suggestion
    Idempotency persistence connection error
```

aws_lambda_powertools/utilities/idempotency/exceptions.py
```diff
@@ -83,3 +83,21 @@ class IdempotencyNoSerializationModelError(BaseError):
...
+    """
+
+
+class IdempotencyPersistenceConsistencyError(BaseError):
+    """
+    Redis Orphan Record found, need to be removed
```
rubenfonseca:
```suggestion
    Idempotency persistency consistency error, needs to be removed
```

rubenfonseca (changes requested):
Just tiny things, super close now :D

rubenfonseca:
> > * **JSON SerDe**: waiting from @leandrodamascena to confirm that GET/SET + JSON SerDe is not slower than using HSETNX/HGET
> 
> This is an excellent point, Ruben! This PR had changes and optimizations throughout its development and one of the modifications was the migration of `HSETNX/HGETALL` to simplify with `GET/SET` and use JSON SerDe. To use `HSETNX/HGETALL` we have a change of concept in this implementation, as the traditional way is to make an `HSETNX` call for each key (in the dict) that we want to store in Redis as an object, after which we have to set ttl for all of them. But at this point, we made an optimization using the concept of `pipelines` in Redis, which optimizes latency, as it uses batch to write to Redis and reduces RTT, one of the main reasons for latency in Redis. But this makes the code more verbose and the business logic more complicated, in addition to the fact that the batch writing has to be transactional (this is by default), and in tests, this took even longer than using set/get with JSON SerDe, in some cases.
> 
> I recovered the [old code](https://github.com/aws-powertools/powertools-lambda-python/pull/2567/commits/7ceb45e203412941e2c87e17d14152d8cc7626ad), ran some tests, and saw no consistent difference in execution time between the two. My opinion is that we can continue with this implementation, and perhaps wait for feedback from some customers.

Cool! I'm 100% convinced this is the best approach. Thanks for the explanation!

roger-zhangg:
- Orphan record Race condition without lock
```mermaid
graph TD;
    A(Existing orphan record in redis)--without lock-->A1;
    A1[Two Lambda invoke at same time]-->B1[Lambda handler1];
    B1-->B2[Fetch from Redis];
    B2-->B3[Handler1 got orphan record];
    B3-->B4[Handler1 overwrite orphan record];
    B4-->B5[Handler1 continue to execution];
    A1-->C1[Lambda handler2];
    C1-->C2[Fetch from Redis];
    C2-->C3[Handler2 got orphan record];
    C3-->C4[Handler2 overwrite orphan record];
    C4-->C5[Handler2 continue to execution];
    B5-->D(Lambda handler executed twice!);
    C5-->D;
```

- Orphan record Race condition with lock
```mermaid
graph TD;
    A(Existing orphan record in redis)--with lock-->A1;
    A1[Two Lambda invoke at same time]-->B1[Lambda handler1];
    B1-->B2[Fetch from Redis];
    B2-->B3[Handler1 got orphan record];
    B3-->B4[Handler1 acquired lock];
    B4-->B5[Handler1 overwrite orphan record]
    B5-->B6[Handler1 continue to execution];
    A1-->C1[Lambda handler2];
    C1-->C2[Fetch from Redis];
    C2-->C3[Handler2 got orphan record];
    C3-->C4[Handler2 failed to acquire lock];
    C4-->C5[Handler2 wait and fetch from Redis];
    C5-->C6[Handler2 return without executing];
    B6-->D(Lambda handler executed only once!);
    C6-->D;
```

tests/functional/idempotency/persistence/test_redis_layer.py
```diff
@@ -0,0 +1,631 @@
...
+
+
+def test_idempotent_function_and_lambda_handler_redis_validation(
+    persistence_store_standalone_redis: RedisCachePersistenceLayer,
+    lambda_context,
+):
+    mock_event = {"user_id": "xyz", "time": "1234"}
+    persistence_layer = persistence_store_standalone_redis
+    result = {"message": "Foo"}
+    expected_result = copy.deepcopy(result)
```
rubenfonseca: Sonarcube is telling me this variable is not used. Can you check?

tests/functional/idempotency/persistence/test_redis_layer.py
```diff
@@ -0,0 +1,631 @@
...
+    config = IdempotencyConfig(event_key_jmespath="user_id", payload_validation_jmespath="time")
+
+    @idempotent(persistence_store=persistence_layer, config=config)
+    def lambda_handler(event, context):
+        return result
+
+    # WHEN calling the function and handler with idempotency and event_key_jmespath,payload_validation_jmespath
+    handler_result = lambda_handler(mock_event, lambda_context)
```
rubenfonseca: Sonarcube is telling me this variable is not used. Can you check?

tests/functional/idempotency/persistence/test_redis_layer.py
```diff
@@ -0,0 +1,631 @@
...
+    # THEN we expect the function and lambda handler to execute successfully
+
+    result = {"message": "Bar"}
+    mock_event = {"user_id": "xyz", "time": "2345"}
+    # Given idempotency record already in Redis
+    # When we modified the payload where validation is on and invoke again.
+    # Then should raise IdempotencyValidationError
+    with pytest.raises(IdempotencyValidationError):
+        handler_result2 = lambda_handler(mock_event, lambda_context)
```
rubenfonseca: Sonarcube is telling me this variable is not used. Can you please check?

tests/integration/idempotency/test_idempotency_redis.py
```diff
@@ -0,0 +1,198 @@
...
+        persistence_layer = RedisCachePersistenceLayer(client=redis_client)
+        result = {"message": "Foo"}
+
+        @idempotent(persistence_store=persistence_layer)
+        def lambda_handler(event, context):
+            return result
+
+        # first run is just to populate function infos for deletion.
+        # delete_record won't work if the function was not run yet. bug maybe?
+        handler_result = lambda_handler(mock_event, lambda_context)
```
rubenfonseca: Sonarcube is telling me this variable is not used. Can you please check?

rubenfonseca (changes requested):
Just some small things left

rubenfonseca (dismissed):
Awesome work everyone!

rubenfonseca (approved):

leandrodamascena (dismissed):
APPROVED!

PR #2733, opened 2023-07-10
feat(parser): add support for Pydantic v2

### LINKED ISSUE #2427: Feature request: Start Evaluating Pydantic v2
### Use case

Pydantic is already included in aws-lamdba-powertools as the provider for deep data validation and parsing as an optional dependency. Pydantic V2 is now in [beta evaluation](https://docs.pydantic.dev/dev-v2/), when it becomes a stable version, it should be introduced as an option into powertools without major breaking changes. We need a way to test the integration with the rest of the library and benchmark performance within Lambda.

### Solution/User Experience

Today the parser is included as an extra using Pydantic. With the launch of PydanticV2 we could include a second extra, add max version, or identify if we can do this transparently to the user with an optional switch.

To keep powertools lean, having both dependencies on a single extra seems less than ideal.

What is the best way to evaluate an impact like this without falling out of sync with the enhancements to powertools?

### Alternative solutions

```markdown
- Have two extras tagged to Pydantic versions
- Freeze the parser to Pydantic v1
- Identify if there are any breaking implementations within parser for breaking API changes of the Pydantic upgrade (not personally recommended)
- Identify how to handle the version difference in the full install layer as users will choose one version, not both in their function.
```

### Acknowledgment

- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/), and [.NET](https://github.com/awslabs/aws-lambda-powertools-dotnet/)
```[tasklist]
### Tasks
```

### LINKED ISSUE #2672: RFC: Add support for Pydantic v2

https://github.com/aws-powertools/powertools-lambda-python/issues/2427

Parser

### Summary

This RFC proposes changes to Parser’s Pydantic Models to support both [Pydantic V2](https://docs.pydantic.dev/2.0/) and V1 without breaking changes.

In Powertools v3 (~EOL, date not settled yet), we will then remove support for Pydantic V1, and update our Parser’s dependency to use Pydantic v2.

```mermaid
stateDiagram-v2
    Models: Update Parser Pydantic models
    DocsV2: Document how to bring Pydantic v2
    Compatibility: Ensure Pydantic v1 and v2 can coexist
    POC: Make POC available for beta testers
    Compatibility --> Models
    Models --> DocsV2
    DocsV2 --> POC
    POC --> PR

```

#### Compatibility table

| Pydantic v1     | Pydantic v2           | V2 deprecation     | V2 removed | In use?            | Code change required |
| --------------- | --------------------- | ------------------ | ---------- | ------------------ | -------------------- |
| @validator      | @field_validator      | :heavy_check_mark: |            | :heavy_check_mark: | :heavy_check_mark:   |
| @root_validator | @model_validator      | :heavy_check_mark: |            | :heavy_check_mark: | :heavy_check_mark:   |
| .parse_obj()    | model_validate()      | :heavy_check_mark: |            | :heavy_check_mark: |                      |
| .json()         | model_dump_json()     | :heavy_check_mark: |            | :heavy_check_mark: |                      |
| .dict()         | model_dump()          | :heavy_check_mark: |            | :heavy_check_mark: |                      |
| .parse_raw()    | model_validate_json() | :heavy_check_mark: |            | :heavy_check_mark: |                      |

-----------

### Use case

Pydantic V2, the latest version of Pydantic, has been launched with enhanced features and performances. Customers using  Powertools for AWS Lambda (Python) in their AWS Lambda functions express interest to update it to Pydantic V2. To meet the customer's needs, the integration of Pydantic V2 in Powertools is required. 

This integration enables customers to update their workloads to use Pydantic V2 while ensuring a seamless transition for existing customers using Pydantic V1.

-----------

### Proposal

To accommodate both Pydantic v1 and Pydantic v2, the proposed design is to refactor the parser models with minimal changes. These changes should work in both versions of Pydantic and be transparent to users.

**TL;DR**: Proposed actions summary
* [x] Set default value of `None` for optional fields
* [x] Keep `@validator` and `@root_validator` deprecated features with a note to remove in V3
* [x] Keep `@parse_obj` deprecated features with a note to remove in V3
* [x] Investigate empty Dicts/List fail validation
* [x] Investigate .dict() and .json() removal
* [x] Handle TypeError validation
* [x] Investigate `datetime` coercion
* [x] Investigate development dependencies conflict
* [x] Document how to bring Pydantic v2 with Powertools
* [x] Document how to disable deprecation warnings for Pydantic V2
* [x] Create PR with POC

<h4>Optional fields must have default value</h4>

Pydantic v1
```python
class SqsAttributesModel(BaseModel):
    ApproximateReceiveCount: str
    ApproximateFirstReceiveTimestamp: datetime
    MessageDeduplicationId: Optional[str]
    MessageGroupId: Optional[str]
    SenderId: str
    SentTimestamp: datetime
    SequenceNumber: Optional[str]
    AWSTraceHeader: Optional[str]
```

Pydantic v2
```python
class SqsAttributesModel(BaseModel):
    ApproximateReceiveCount: str
    ApproximateFirstReceiveTimestamp: datetime
    MessageDeduplicationId: Optional[str] = None
    MessageGroupId: Optional[str] = None
    SenderId: str
    SentTimestamp: datetime
    SequenceNumber: Optional[str] = None
    AWSTraceHeader: Optional[str] = None
```

<h4>Validators are deprecated</h4>

Both `@root_validator` and `@validator` validators are deprecated in Pydantic V2 and will be removed in Pydantic V3. Pydantic recommends using the new `@model_validator` and `@field_validator` validators. However, we can continue using the deprecated validators in Powertools to avoid breaking changes and plan their removal for Powertools v3.

Pydantic v1
```python 
@root_validator(allow_reuse=True)
def check_message_id(cls, values):
    message_id, event_type = values.get("messageId"), values.get("eventType")
    if message_id is not None and event_type != "MESSAGE":
        raise TypeError("messageId is available only when the `eventType` is `MESSAGE`")
    return values
```

Pydantic v2
```python
@root_validator(allow_reuse=True, skip_on_failure=True)
def check_message_id(cls, values):
    message_id, event_type = values.get("messageId"), values.get("eventType")
    if message_id is not None and event_type != "MESSAGE":
        raise ValueError("messageId is available only when the `eventType` is `MESSAGE`")
    return values
```

Another alternative is to check the Pydantic version and add the conditional import. We will also need to create a function to wrap the validator decorators and check the Pydantic version. This workaround may make the models harder to read and understand for maintenance purposes.

<h4>Powertools Layer</h4>

The Powertools Layer is built with Pydanticv1 and this can be a potential problem if the customer uses our layer and brings Pydantic v2 as an external dependency. 

In tests with Lambda Powertools Layer + Pydanticv2 installed as an external dependency, Lambda first includes the `/var/task` path, that is, the external dependency will have preference over the one used in the Layer and it allows the customer brings their preferred Pydantic version.

image

Path
```sh
{"level":"INFO","location":"<module>:8","message":["/var/task","/opt/python/lib/python3.10/site-packages","/opt/python","/var/runtime","/var/lang/lib/python310.zip","/var/lang/lib/python3.10","/var/lang/lib/python3.10/lib-dynload","/var/lang/lib/python3.10/site-packages","/opt/python/lib/python3.10/site-packages"],"timestamp":"2023-07-05 09:04:52,691+0000","service":"service_undefined"}
```

Pydantic Version
```sh
{"level":"INFO","location":"lambda_handler:11","message":"Pydantic version -> 2.0.1","timestamp":"2023-07-05 09:04:52,694+0000","service":"service_undefined","xray_trace_id":"1-64a53234-0ca06edc18c523524775237c"}
```

<h4>Warnings</h4>

- To ensure a smooth transition and minimize disruptions for our users, we have temporarily suppressed the `PydanticDeprecatedSince20` and `PydanticDeprecationWarning` warnings (related to these functions). This allows existing applications to continue functioning as expected without outputting warnings.

- If needed, you can enable the warnings yourself with something like the code below. Reference: https://docs.python.org/3/library/warnings.html

```python
from aws_lambda_powertools.utilities.parser import event_parser, BaseModel, envelopes
from aws_lambda_powertools.utilities.parser.models import (
    SqsModel,
)

import pydantic

import warnings
warnings.simplefilter('default')

```

-----------
### Out of scope

Refactorings involving breaking change for customers who want to use v1. If there is something that involves breaking, it will be left out of this change.

We could this opportunity to evaluate the performance of Pydantic V2 and potentially enhance the performance of Parser utility - not required tho.

-----------
### Potential challenges

Most of the challenges were addressed and I was able to use the Powertools for AWS Lambda (Python) Parser utility with Pydantic v2 with several models. But some challenges still need to be understood whether this is a breaking change or not.

<h4> Working with datetime fields</h4>

In Pydantic v1, when using `datetime` fields, the UTC offset is included and the tests work fine. However, in Pydantic v2, the UTC offset is not included, causing our tests to fail.

Codebase

```python
from datetime import datetime

import pydantic


class Model(BaseModel):
    datefield: datetime = None

epoch_time = 1659687279885

m = Model(
    datefield=epoch_time,
)

print(f"Pydantic version -> {pydantic.__version__}")
print(f"Raw epoch time -> {epoch_time}")
print(f"Raw pydantic field -> {m.datefield}")
print(f"Pydantic converted epoch time -> {int(round(m.datefield.timestamp() * 1000))}")

assert epoch_time == int(round(m.datefield.timestamp() * 1000))
```

Pydantic v1
```sh
/tmp/pydantic2 via 🐍 v3.10.6 (.env) on ☁️  (us-east-1) 
❯ python v1.py
Pydantic version -> 1.10.11
Raw epoch time -> 1659687279885
Raw pydantic field -> 2022-08-05 08:14:39.885000+00:00
Pydantic converted epoch time -> 1659687279885
```

Pydantic v2
```sh
/tmp/pydantic2 via 🐍 v3.10.6 (.env) on ☁️  (us-east-1) 
❯ python v2.py
Pydantic version -> 2.0.1
Raw epoch time -> 1659687279885
Raw pydantic field -> 2022-08-05 08:14:39.885000
Pydantic converted epoch time -> 1659683679885
  File "/tmp/pydantic2/v2.py", line 21, in <module>
    assert epoch_time == int(round(m.datefield.timestamp() * 1000))
AssertionError
```

<h4>Batch processing</h4>

Some Batch processing tests are failing and I need to investigate why.

```sh
FAILED tests/functional/test_utilities_batch.py::test_batch_processor_model_with_partial_validation_error - AttributeError: 'NoneType' object has no attribute 'message_id'
FAILED tests/functional/test_utilities_batch.py::test_batch_processor_dynamodb_context_model_with_partial_validation_error - AttributeError: 'NoneType' object has no attribute 'dynamodb'
FAILED tests/functional/test_utilities_batch.py::test_batch_processor_kinesis_context_parser_model_with_partial_validation_error - AttributeError: 'NoneType' object has no attribute 'kinesis'
FAILED tests/functional/test_utilities_batch.py::test_async_batch_processor_model_with_partial_validation_error - AttributeError: 'NoneType' object has no attribute 'message_id'
FAILED tests/functional/test_utilities_batch.py::test_async_batch_processor_dynamodb_context_model_with_partial_validation_error - AttributeError: 'NoneType' object has no attribute 'dynamodb'
FAILED tests/functional/test_utilities_batch.py::test_async_batch_processor_kinesis_context_parser_model_with_partial_validation_error - AttributeError: 'NoneType' object has no attribute 'kinesis'
```

<h4>Developer environment</h4>

Since some of our dependencies (`cfn-lint `/ `aws-sam-translator`) have a requirement of Pydantic v1, we'll need to remove them from our development environment in order to accommodate Pydantic v2.

```sh
❯ poetry add "pydantic>=2.0"                       

Updating dependencies
Resolving dependencies... (0.6s)

Because no versions of aws-sam-translator match >1.68.0,<1.69.0 || >1.69.0,<1.70.0 || >1.70.0
 and aws-sam-translator (1.68.0) depends on pydantic (>=1.8,<2.0), aws-sam-translator (>=1.68.0,<1.69.0 || >1.69.0,<1.70.0 || >1.70.0) requires pydantic (>=1.8,<2.0).
And because aws-sam-translator (1.69.0) depends on pydantic (>=1.8,<2.0), aws-sam-translator (>=1.68.0,<1.70.0 || >1.70.0) requires pydantic (>=1.8,<2.0).
And because aws-sam-translator (1.70.0) depends on pydantic (>=1.8,<2.0)
 and cfn-lint (0.77.10) depends on aws-sam-translator (>=1.68.0), cfn-lint (0.77.10) requires pydantic (>=1.8,<2.0).
So, because aws-lambda-powertools depends on both pydantic (>=2.0) and cfn-lint (0.77.10), version solving failed.
```
---------
### Dependency resolution for Pydantic v1 and v2

Customers will run into a conflict when having either `requirements.txt`:

**All Powertools features and Pydantic v2**

```txt
aws-lambda-powertools[all]
pydantic # or pydantic>=2
```

**Powertools parser feature and Pydantic v2**

```txt
aws-lambda-powertools[parser]
pydantic # or pydantic>=2
```

#### Recommendation

We should include a new section in the documentation to explain how to use Pydantic v2 with Parser.

For example, customers should refrain from using `[all]` or `[parser]` when bringing Pydantic v2 as part of their dependencies.

* `aws-lambda-powertools[all]` becomes `aws-lambda-powertools[validation,tracer,aws-sdk]`

Because of the `Optional[str] = None` breaking change in v2, we should keep our pydantic pinning to v1 until we launch v3 and move away. We cannot guarantee a customer is using additional Pydantic v1 features through Powertools - or [followed our docs to the letter](https://docs.powertools.aws.dev/lambda/python/latest/utilities/parser/#defining-models).

This also gives us room to recommend disabling warnings for deprecated features we're keeping for backwards compatibility (e.g., validators). This prevents Pydantic littering customers' logs ($$) when bringing Pydantic v2.

---------
### Ad-hoc test for pydantic v2 dep

Until Powertools V3 is a thing, we should guard against backwards incompatibility for newer Pydantic models, or changes to existing models that might be contributed externally from a Pydantic v2 customer.

#### Recommendation

Setup a new temporary GitHub Action workflow to trigger on changes to Parser's models. We can use [Nox](https://nox.thea.codes/en/stable/) to streamline dependencies and easily trigger Parser's unit tests.

Alternatively, within this temporary workflow, we could call `make dev`, remove Pydantic, install Pydantic v2`, and run Parser's unit tests.

Nox's benefit is that it's more stable, easier to reason, and it will lead us to address an unrelated area we aren't testing today -- tests w/o optional dependencies (e.g., Batch without parser code).

---------

_No response_

### Alternative solutions

_No response_

### Acknowledgment

leandrodamascena:
**Issue number:** #2672 
Additional issue: #2427 

### Changes

After an extensive week of investigation, experimentation, and implementation, I'm excited to open this pull request, which introduces support for Pydantic v2 in the parser feature. I wanted to provide you with an overview of the changes and considerations made during this process:

**Support for Pydantic v1 and v2:**

- The parser utility has been enhanced to ensure compatibility with both Pydantic v1 and v2.
- Users now have the flexibility to choose their preferred version of Pydantic for parse utility.

**Powertools and the support for Pydantic:**

- Historically, Powertools has considered Pydantic v1 as the official library for parse utility, given its long-standing compatibility and robustness within our code.
- However, with this update, we are embracing the advancements and enhancements introduced in Pydantic v2, allowing users to leverage the latest features and improvements.
- It's important to highlight that while we have introduced support for Pydantic v2, our official version for Powertools remains Pydantic v1. This decision has been made to prioritize compatibility and provide a seamless experience for our existing users. By maintaining Pydantic v1 as our official library, we ensure that our ecosystem and the applications built with it continue to function as expected without any disruptions.

**No-breaking changes were made**

**NOTE**: We haven't detected any breaking changes, but please let us know as soon as possible if you find any.

- Importantly, this update does not introduce any breaking changes to the existing functions within the Powertools framework. The functions that were already in use remain intact and fully functional, ensuring compatibility with the existing codebase. 
- However, it's worth noting that with the support for Pydantic v2, some Pydantic functions used in Powertools are now considered deprecated. This deprecation serves as a signal to users that these functions will be removed in a future major release, specifically Powertools v3.

#### Compatibility table

| Pydantic v1     | Pydantic v2           | V2 deprecation     | In use?            | Code change required |
| --------------- | --------------------- | ------------------ | ------------------ | -------------------- |
| @validator      | @field_validator      | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:   |
| @root_validator | @model_validator      | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark:   |
| .parse_obj()    | model_validate()      | :heavy_check_mark: | :heavy_check_mark: |                      |
| .json()         | model_dump_json()     | :heavy_check_mark: | :heavy_check_mark: |                      |
| .dict()         | model_dump()          | :heavy_check_mark: | :heavy_check_mark: |                      |
| .parse_raw()    | model_validate_json() | :heavy_check_mark: | :heavy_check_mark: |                      |

- To ensure a smooth transition and minimize disruptions for our users, we have temporarily suppressed the `PydanticDeprecatedSince20` and `PydanticDeprecationWarning` warnings (related to these functions). This allows existing applications to continue functioning as expected without outputting warnings.
- If needed, you can enable the warnings yourself with something like the code below. Reference: https://docs.python.org/3/library/warnings.html
```python
from aws_lambda_powertools.utilities.parser import event_parser, BaseModel, envelopes
from aws_lambda_powertools.utilities.parser.models import (
    SqsModel,
)

import pydantic

import warnings
warnings.simplefilter('default')

```

**How to test before we merge this PR**

To facilitate the testing of Pydantic v2 before merging this pull request, we have created a separate branch called `poc/pydanticv2` where you can find the relevant code changes. Here are the steps to follow in order to test it effectively:

If using sourcecode:
	- Clone Powertools repository
	- Fetch the `poc/pydanticv2` branch
	- Check out the branch locally by running the command: `git checkout poc/pydanticv2`.

If using pip:

Create or modify the `requirements.txt` file and add the following content:
```
git+https://github.com/aws-powertools/powertools-lambda-python@poc/pydanticv2#egg=aws_lambda_powertools
pydantic>2
```
Then install by running: `pip install -r requirements.txt`.

If using Poetry:

```sh
poetry add git+https://github.com/aws-powertools/powertools-lambda-python#poc/pydanticv2
poetry add pydantic>2
```

### Note for documentation

Since the Powertools currently relies on Pydantic v1 as its official dependency, you need to change the `aws-lambda-powertools[tracer]` or `aws-lambda-powertools[all]` installation to `aws-lambda-powertools` or `aws-lambda-powertools[validation,tracer,aws-sdk]` and add Pydantic v2 as an extra dependency.

### Missing in this PR
- [x] Documentation
- [x] Workflow in develop branch to ensure that any changes will work in Pydanticv2
- [x] Community feedback

### User experience

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
hooorraay

.github/workflows/quality_check_temp_pydanticv2.yml
```diff
@@ -0,0 +1,82 @@
...
+    steps:
+      - uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9  # v3.5.3
+      - name: Install poetry
+        run: pipx install poetry
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@bd6b4b6205c4dbad673328db7b31b7fab9e241c0 # v4.6.1
+        with:
+          python-version: ${{ matrix.python-version }}
+          cache: "poetry"
+      - name: Removing cfn-lint
```
heitorlessa:
```suggestion
      - name: Removing dev dependencies locked to Pydantic v1
```

aws_lambda_powertools/shared/functions.py
```diff
@@ -173,3 +173,33 @@ def extract_event_from_common_models(data: Any) -> Dict | Any:
...
+        if int(version[0]) == 2:
+            import warnings
+
+            from pydantic import PydanticDeprecatedSince20, PydanticDeprecationWarning
+
+            warnings.filterwarnings("ignore", category=PydanticDeprecationWarning)
+            warnings.filterwarnings("ignore", category=PydanticDeprecatedSince20)
+
+    except ImportError:
+        pass
```
heitorlessa: move to `parser`

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -348,6 +348,11 @@ def _to_batch_type(self, record: dict, event_type: EventType) -> EventSourceData
 
     def _to_batch_type(self, record: dict, event_type: EventType, model: Optional["BatchTypeModels"] = None):
         if model is not None:
+            # If a model is provided, we assume Pydantic is installed and we need to disable v2 warnings
+            from aws_lambda_powertools.shared.functions import disable_pydantic_v2_warning
```
heitorlessa:
```suggestion
            from aws_lambda_powertools.utilities.parser.compat import disable_pydantic_v2_warning
```

aws_lambda_powertools/utilities/parser/models/__init__.py
```diff
@@ -1,11 +1,15 @@
-from .alb import AlbModel, AlbRequestContext, AlbRequestContextData
-from .apigw import (
+from aws_lambda_powertools.shared.functions import disable_pydantic_v2_warning
+
+disable_pydantic_v2_warning()
+
+from .alb import AlbModel, AlbRequestContext, AlbRequestContextData  # noqa: E402
```
heitorlessa: add the ignore directly in Ruff (ignore the whole file)

aws_lambda_powertools/utilities/parser/models/cloudwatch.py
```diff
@@ -15,6 +15,14 @@ class CloudWatchLogsLogEvent(BaseModel):
     timestamp: datetime
     message: Union[str, Type[BaseModel]]
 
+    # Validator to normalize the timestamp field
+    # Converts the provided timestamp value to a UTC datetime object
+    # See: https://github.com/pydantic/pydantic/issues/6518
+    @validator("timestamp", pre=True)
+    def coerc_timestamp(cls, value):
```
heitorlessa:
```suggestion
    def coerce_timestamp(cls, value):
```

tests/unit/parser/test_apigw.py
```diff
@@ -138,7 +138,7 @@ def test_apigw_event_with_invalid_websocket_request():
     errors = err.value.errors()
     assert len(errors) == 1
     expected_msg = "messageId is available only when the `eventType` is `MESSAGE`"
-    assert errors[0]["msg"] == expected_msg
+    assert expected_msg in errors[0]["msg"]
```
heitorlessa, line 143: add a comment on what happened here

tests/unit/parser/test_cloudwatch.py
```diff
@@ -86,7 +86,7 @@ def test_handle_invalid_cloudwatch_trigger_event_no_envelope():
     with pytest.raises(ValidationError) as context:
         CloudWatchLogsModel(**raw_event)
 
-    assert context.value.errors()[0]["msg"] == "unable to decompress data"
+    assert "unable to decompress data" in context.value.errors()[0]["msg"]
```
heitorlessa, line 91: same here

heitorlessa (changes requested):
Also add a requirements.txt

```txt
git+....
pydantic>=2
```

and poetry command to add this dep for customers to test

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -500,8 +505,13 @@ def _process_record(self, record: dict) -> Union[SuccessResponse, FailureRespons
             # we need to handle that exception differently.
             # We check for a public attr in validation errors coming from Pydantic exceptions (subclass or not)
             # and we compare if it's coming from the same model that trigger the exception in the first place
-            model = getattr(exc, "model", None)
-            if model == self.model:
+
+            # Pydantic v1 raises a ValidationError with ErrorWrappers and store the model instance in a class variable.
+            # Pydantic v2 simplifies this by adding a title variable to store the model name directly.
```
leandrodamascena, line 510: @heitorlessa pls suggest a better comment, if needed.

aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -644,8 +654,13 @@ async def _async_process_record(self, record: dict) -> Union[SuccessResponse, Fa
...
```
leandrodamascena, line 659: Same here

.github/workflows/quality_check_temp_pydanticv2.yml
leandrodamascena: done

aws_lambda_powertools/shared/functions.py
leandrodamascena: Moved it to parser.

aws_lambda_powertools/utilities/parser/models/__init__.py
leandrodamascena: Great catch, Heitor! Done

leandrodamascena:
Hi Powertools customers and community at large! We have the PR for Pydanticv2 ready for testing and would love to hear your feedback. Our goal is to merge this PR on July 21st and it would be nice to hear something before then.

Tagging people we know are directly interested in this support: @ran-isenberg @djfurman @half2me @Wurstnase @taharah @wickeat @waveFrontSet

ran-isenberg:
@leandrodamascena  is there a pypi PR version I can consume in my project and run the tests?
BTW, i'd like to still have the deprecated warning so it'd easier for me to fix it  . Did you disable it?

leandrodamascena:
> @leandrodamascena is there a pypi PR version I can consume in my project and run the tests?

So far there isn't @ran-isenberg! But you have an alternative to install from your local folder or from GitHub. Do these alternatives not work for you? Does it really need to be a package on Pypi?

I ask because we considered the idea of having a package on Pypi in alpha release, but we chose to leave the installation directly from the GitHub branch/local folder, as this facilitates any changes we have to make.

leandrodamascena:
> BTW, i'd like to still have the deprecated warning so it'd easier for me to fix it . Did you disable it?

Yeah, we disabled it but you can enable again:

```python
from aws_lambda_powertools.utilities.parser import event_parser, BaseModel, envelopes
from aws_lambda_powertools.utilities.parser.models import (
    SqsModel,
)

import pydantic

import warnings
warnings.simplefilter('default')

logger = Logger()


@event_parser(model=SqsModel)
def lambda_handler(event: SqsModel, context):
    logger.info(f"Pydantic version -> {pydantic.__version__}")
    for record in event.Records:
        logger.info(f"Event body -> {record.body}")
```

ran-isenberg:
> > @leandrodamascena is there a pypi PR version I can consume in my project and run the tests?
> 
> So far there isn't @ran-isenberg! But you have an alternative to install from your local folder or from GitHub. Do these alternatives not work for you? Does it really need to be a package on Pypi?
> 
> I ask because we considered the idea of having a package on Pypi in alpha release, but we chose to leave the installation directly from the GitHub branch/local folder, as this facilitates any changes we have to make.

If you could share the proper toml definition for a github branch that would be amazing and save me some chatgpt time :)

ran-isenberg:
So, it's working and it didnt break me. Well done @leandrodamascena  and the entire team!
 
Here's a PR on my AWS cookbook repo with pydantic 2 upgrade (i removed all deprecated issues).
https://github.com/ran-isenberg/aws-lambda-handler-cookbook/pull/682

It worked without me having to change anything, but it was hard digging through the warnings. Funny enough, pydantic's own legacy support code raises these warnings too. It's going to be a pain fixing it in more complex projects.

aws_lambda_powertools/utilities/parser/models/s3.py
```diff
@@ -77,12 +77,12 @@ class S3EventNotificationEventBridgeDetailModel(BaseModel):
...
-    restore_expiry_time: Optional[str] = Field(None, alias="restore-expiry-time")
-    source_storage_class: Optional[str] = Field(None, alias="source-storage-class")
-    destination_storage_class: Optional[str] = Field(None, alias="destination-storage-class")
-    destination_access_tier: Optional[str] = Field(None, alias="destination-access-tier")
+    reason: Optional[str] = None
+    deletion_type: Union[str, None] = Field(None, alias="deletion-type")
+    restore_expiry_time: Union[str, None] = Field(None, alias="restore-expiry-time")
+    source_storage_class: Union[str, None] = Field(None, alias="source-storage-class")
+    destination_storage_class: Union[str, None] = Field(None, alias="destination-storage-class")
+    destination_access_tier: Union[str, None] = Field(None, alias="destination-access-tier")
```
tonnico:
```suggestion
    reason: Optional[str] = None
    deletion_type: Optional[str] = Field(None, alias="deletion-type")
    restore_expiry_time: Optional[str] = Field(None, alias="restore-expiry-time")
    source_storage_class: Optional[str] = Field(None, alias="source-storage-class")
    destination_storage_class: Optional[str] = Field(None, alias="destination-storage-class")
    destination_access_tier: Optional[str] = Field(None, alias="destination-access-tier")
```

aws_lambda_powertools/utilities/parser/models/apigwv2.py
```diff
@@ -29,9 +29,9 @@ class RequestContextV2AuthorizerJwt(BaseModel):
 
 
 class RequestContextV2Authorizer(BaseModel):
-    jwt: Optional[RequestContextV2AuthorizerJwt]
-    iam: Optional[RequestContextV2AuthorizerIam]
-    lambda_value: Optional[Dict[str, Any]] = Field(None, alias="lambda")
+    jwt: Optional[RequestContextV2AuthorizerJwt] = None
+    iam: Optional[RequestContextV2AuthorizerIam] = None
+    lambda_value: Union[Dict[str, Any], None] = Field(None, alias="lambda")
```
tonnico:
```suggestion
    lambda_value: Optional[Dict[str, Any]] = Field(None, alias="lambda")
```

aws_lambda_powertools/utilities/parser/models/ses.py
```diff
@@ -36,10 +36,10 @@ class SesMailHeaders(BaseModel):
...
     header_from: List[str] = Field(None, alias="from")
     to: List[str]
-    cc: Optional[List[str]]
-    bcc: Optional[List[str]]
-    sender: Optional[List[str]]
-    reply_to: Optional[List[str]] = Field(None, alias="reply-to")
+    cc: Optional[List[str]] = None
+    bcc: Optional[List[str]] = None
+    sender: Optional[List[str]] = None
+    reply_to: Union[List[str], None] = Field(None, alias="reply-to")
```
tonnico:
```suggestion
    reply_to: Optional[List[str]] = Field(None, alias="reply-to")
```

tests/functional/parser/test_parser.py
```diff
@@ -53,6 +54,33 @@ def handle_no_envelope(event: Dict, _: LambdaContext):
     handle_no_envelope(dummy_event["payload"], LambdaContext())
 
 
+def test_pydanticv2_validation():
+    from pydantic import __version__
+
+    version = __version__.split(".")
+
+    # GIVEN pydantic v2 version
+    if int(version[0]) == 2:
```
tonnico: I would prefer a pytest.skip and detent the other code.

```suggestion
    if version[0] != "2":
        pytest.skip("test only for pydantic v2.")
```

or maybe better as a fixture?

```
@pytest.fixture
def pydanticv2_only():
    from pydantic import __version__

    version = __version__.split(".")
    if version[0] != "2":
        pytest.skip("pydanticv2 test only")

@pytest.mark.usefixture("pydanticv2_only")
def test_pydanticv2_validation():
    ...

Comment (leandrodamascena) [reply tests/functional/parser/test_parser.py]: Hi @Wurstnase! Thanks a lot for suggesting this. It makes perfect sense to use a fixture for this because we can introduce new tests for v2 without repeating too much code.

File: aws_lambda_powertools/utilities/parser/models/s3.py (hunk 10 above)
Comment (leandrodamascena) [reply aws_lambda_powertools/utilities/parser/models/s3.py]: I'm curious about this suggestion.
Both options are widely used and accepted in Python to represent an `Optional` field and I wonder if you suggest this because `Optional` is more explicit than `Union`.

File: aws_lambda_powertools/utilities/parser/models/apigw.py [hunk 14]
```diff
@@ -21,74 +21,82 @@ class ApiGatewayUserCert(BaseModel):
...
+            raise ValueError("messageId is available only when the `eventType` is `MESSAGE`")
         return values
 
+    # Validator to normalize the requestTimeEpoch field
+    # Converts the provided timestamp value to a UTC datetime object
+    # See: https://github.com/pydantic/pydantic/issues/6518
+    @validator("requestTimeEpoch", pre=True)
+    def coerce_timestamp(cls, value):
+        date_utc = datetime.fromtimestamp(int(value) / 1000, tz=timezone.utc)
+        return date_utc
```
leandrodamascena: The Pydantic project will revert the behavior back to what we had in v1, so we can remove this validator after that.

Issue: https://github.com/pydantic/pydantic/issues/6518

aws_lambda_powertools/utilities/parser/models/apigwv2.py
```diff
@@ -55,17 +55,25 @@ class RequestContextV2(BaseModel):
     timeEpoch: datetime
     http: RequestContextV2Http
 
+    # Validator to normalize the timeEpoch field
+    # Converts the provided timestamp value to a UTC datetime object
+    # See: https://github.com/pydantic/pydantic/issues/6518
+    @validator("timeEpoch", pre=True)
+    def coerce_timestamp(cls, value):
+        date_utc = datetime.fromtimestamp(int(value) / 1000, tz=timezone.utc)
+        return date_utc
```
leandrodamascena: The Pydantic project will revert the behavior back to what we had in v1, so we can remove this validator after that.

Issue: https://github.com/pydantic/pydantic/issues/6518

aws_lambda_powertools/utilities/parser/models/cloudwatch.py
```diff
@@ -15,6 +15,14 @@ class CloudWatchLogsLogEvent(BaseModel):
     timestamp: datetime
     message: Union[str, Type[BaseModel]]
 
+    # Validator to normalize the timestamp field
+    # Converts the provided timestamp value to a UTC datetime object
+    # See: https://github.com/pydantic/pydantic/issues/6518
+    @validator("timestamp", pre=True)
+    def coerce_timestamp(cls, value):
+        date_utc = datetime.fromtimestamp(value / 1000, tz=timezone.utc)
+        return date_utc
```
leandrodamascena: The Pydantic project will revert the behavior back to what we had in v1, so we can remove this validator after that.

Issue: https://github.com/pydantic/pydantic/issues/6518

aws_lambda_powertools/utilities/parser/models/kafka.py
```diff
@@ -34,6 +34,14 @@ def decode_headers_list(cls, value):
                 header[key] = bytes(values)
         return value
 
+    # Validator to normalize the timestamp field
+    # Converts the provided timestamp value to a UTC datetime object
+    # See: https://github.com/pydantic/pydantic/issues/6518
+    @validator("timestamp", pre=True)
+    def coerce_timestamp(cls, value):
+        date_utc = datetime.fromtimestamp(int(value) / 1000, tz=timezone.utc)
+        return date_utc
```
leandrodamascena: The Pydantic project will revert the behavior back to what we had in v1, so we can remove this validator after that.

Issue: https://github.com/pydantic/pydantic/issues/6518

aws_lambda_powertools/utilities/parser/models/sqs.py
```diff
@@ -1,25 +1,33 @@
...
+    SequenceNumber: Optional[str] = None
+    AWSTraceHeader: Optional[str] = None
+
+    # Validator to normalize the ApproximateFirstReceiveTimestamp and SentTimestamp fields
+    # Converts the provided timestamp value to a UTC datetime object
+    # See: https://github.com/pydantic/pydantic/issues/6518
+    @validator("ApproximateFirstReceiveTimestamp", "SentTimestamp", pre=True)
+    def coerce_timestamp(cls, value):
+        date_utc = datetime.fromtimestamp(int(value) / 1000, tz=timezone.utc)
+        return date_utc
```
leandrodamascena: The Pydantic project will revert the behavior back to what we had in v1, so we can remove this validator after that.

Issue: https://github.com/pydantic/pydantic/issues/6518

leandrodamascena (commented):
Adding notes.

aws_lambda_powertools/utilities/parser/models/s3.py
tonnico: Just be consistent. The other code uses `Optional`.

leandrodamascena: @Wurstnase I agree with that and reverted all those changes.

waveFrontSet:
> So, it's working and it didnt break me. Well done @leandrodamascena and the entire team!

Same here!

I've got two comments that might be helpful to others testing out the poc (or generally migrating to pydantic v2).

1. There's a [code transformation tool called bump-pydantic](https://docs.pydantic.dev/dev-v2/migration/#code-transformation-tool) that tremendously helped me doing the upgrade. It could transform some things automatically and left `TODO` markers wherever it couldn't.
2. After fixing the deprecation warnings in our library, I was still left with deprecation warnings from aws-lambda-powertools when running `pytest` (even though [@leandrodamascena mentioned that these deprecation warnings are filtered out](https://github.com/aws-powertools/powertools-lambda-python/pull/2733#issuecomment-1629465204)). This appears to be [intended behaviour of pytest always showing deprecation warnings when running tests](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#deprecationwarning-and-pendingdeprecationwarning).

leandrodamascena:
> > So, it's working and it didnt break me. Well done @leandrodamascena and the entire team!
> 
> Same here!

That's amazing to hear @waveFrontSet! Thanks for spending time validating this.

> 
> I've got two comments that might be helpful to others testing out the poc (or generally migrating to pydantic v2).
> 
> 1. There's a [code transformation tool called bump-pydantic](https://docs.pydantic.dev/dev-v2/migration/#code-transformation-tool) that tremendously helped me doing the upgrade. It could transform some things automatically and left `TODO` markers wherever it couldn't.

Yes, I've tried that, but in our case, that's not very useful because we don't want to change the code to only support v2. We are maintaining support for v1 while adding support for v2.

> 2. After fixing the deprecation warnings in our library, I was still left with deprecation warnings from aws-lambda-powertools when running `pytest` (even though [@leandrodamascena mentioned that these deprecation warnings are filtered out](https://github.com/aws-powertools/powertools-lambda-python/pull/2733#issuecomment-1629465204)). This appears to be [intended behaviour of pytest always showing deprecation warnings when running tests](https://docs.pytest.org/en/stable/how-to/capture-warnings.html#deprecationwarning-and-pendingdeprecationwarning).

Good point about `pytest` deprecation messages. I understand that we can suppress messages in `pytest`, but I'm not sure if we're missing to suppress these messages in other pieces of our codebase or if it's just `pytest` behavior. Can you please share one of that messages here? Please remove any confidential data.

waveFrontSet:
> Good point about `pytest` deprecation messages. I understand that we can suppress messages in `pytest`, but I'm not sure if we're missing to suppress these messages in other pieces of our codebase or if it's just `pytest` behavior. Can you please share one of that messages here? Please remove any confidential data.

Sure, here you go:

```
 .../lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/parser.py:163: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.2/migration/
    return model.parse_obj(event)
 .../lib/python3.10/site-packages/pydantic/main.py:941: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.2/migration/
    warnings.warn('The `parse_obj` method is deprecated; use `model_validate` instead.', DeprecationWarning)
```

I'm pretty sure you're already catching this warning, though: If I run the same code as a script the warning won't show up.

leandrodamascena:
> Sure, here you go:
> 
> 
> 
> ```
> 
>  .../lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/parser.py:163: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.2/migration/
> 
>     return model.parse_obj(event)
> 
>  .../lib/python3.10/site-packages/pydantic/main.py:941: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.2/migration/
> 
>     warnings.warn('The `parse_obj` method is deprecated; use `model_validate` instead.', DeprecationWarning)
> 
> ```
> 
> 
> 
> I'm pretty sure you're already catching this warning, though: If I run the same code as a script the warning won't show up.

Thanks for sharing this, it confirmed what I was suspecting. I'm away today, but I'll sort it out by Monday and let you know.

aws_lambda_powertools/utilities/parser/models/apigw.py
leandrodamascena: They fixed this bug in v2.0.3 - https://github.com/pydantic/pydantic/releases/tag/v2.0.3.

We have removed this extra code now.

aws_lambda_powertools/utilities/parser/models/apigwv2.py
leandrodamascena: They fixed this bug in v2.0.3 - https://github.com/pydantic/pydantic/releases/tag/v2.0.3.

We have removed this extra code now.

aws_lambda_powertools/utilities/parser/models/cloudwatch.py
leandrodamascena: They fixed this bug in v2.0.3 - https://github.com/pydantic/pydantic/releases/tag/v2.0.3.

We have removed this extra code now.

aws_lambda_powertools/utilities/parser/models/sqs.py
leandrodamascena: They fixed this bug in v2.0.3 - https://github.com/pydantic/pydantic/releases/tag/v2.0.3.

We have removed this extra code now.

leandrodamascena:
> I'm pretty sure you're already catching this warning, though: If I run the same code as a script the warning won't show up.

Hi @waveFrontSet, can you test again?

waveFrontSet:
Hi @leandrodamascena, updated to commit `1774a1c`, still the same behaviour. I'm seeing the same warnings while running pytest, but not while executing it as a script.
```
.../lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/parser.py:163: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.3/migration/
    return model.parse_obj(event)
.../lib/python3.10/site-packages/pydantic/main.py:944: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.3/migration/
    warnings.warn('The `parse_obj` method is deprecated; use `model_validate` instead.', DeprecationWarning)
```

leandrodamascena:
> Hi @leandrodamascena, updated to commit `1774a1c`, still the same behaviour. I'm seeing the same warnings while running pytest, but not while executing it as a script.
> 
> ```
> .../lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/parser.py:163: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.3/migration/
>     return model.parse_obj(event)
> .../lib/python3.10/site-packages/pydantic/main.py:944: PydanticDeprecatedSince20: The `parse_obj` method is deprecated; use `model_validate` instead. Deprecated in Pydantic V2.0 to be removed in V3.0. See Pydantic V2 Migration Guide at https://errors.pydantic.dev/2.0.3/migration/
>     warnings.warn('The `parse_obj` method is deprecated; use `model_validate` instead.', DeprecationWarning)
> ```

Hi @waveFrontSet! I could reproduce the problem in my local environment and fixed it. Can you check again?

Thank you.

waveFrontSet:
Hi @leandrodamascena,
this has done the trick! No warnings when running pytest. Great job. 👍

.github/workflows/quality_check_pydanticv2.yml
```diff
@@ -0,0 +1,82 @@
...
+on:
+  pull_request:
+    paths:
+      - "aws_lambda_powertools/**"
+      - "tests/**"
+      - "pyproject.toml"
+      - "poetry.lock"
+      - "mypy.ini"
+    branches:
+      - poc/pydanticv2
```
rubenfonseca: I guess we need to change this before we merge?

.github/workflows/quality_check_pydanticv2.yml
```diff
@@ -0,0 +1,82 @@
...
+  push:
+    paths:
+      - "aws_lambda_powertools/**"
+      - "tests/**"
+      - "pyproject.toml"
+      - "poetry.lock"
+      - "mypy.ini"
+    branches:
+      - poc/pydanticv2
```
rubenfonseca: Same here

aws_lambda_powertools/utilities/parser/compat.py
```diff
@@ -0,0 +1,28 @@
+def disable_pydantic_v2_warning():
+    """
+    Disables the Pydantic version 2 warning by filtering out the related warnings.
+
+    This function checks the version of Pydantic currently installed and if it is version 2,
+    it filters out the PydanticDeprecationWarning and PydanticDeprecatedSince20 warnings
+    to suppress them.
+
+    Note: This function assumes that Pydantic is already imported.
```
rubenfonseca: Is this still valid? I can see pydantic being loaded later on anyways "from pydantic"

rubenfonseca (changes requested):
Superb work Leandro! The only concern I have is that we are calling "disable_pydantic_v2_warning" in so many different places, and I wonder if we could improve performance by memoizing it so it only runs once, no matter how many times it gets called. What do you think?

.github/workflows/quality_check_pydanticv2.yml
leandrodamascena: Done

rubenfonseca (approved):

heitorlessa (approved):

Issue #2826, opened 2023-07-22
Feature request: Setting Parameters and Secrets


stephenbawks:
### Use case

The parameters module currently only allows users to read secrets and parameters from the Secret and Parameter Store. I would like to extend the parameters module to allow users to write secrets and parameters back to the stores. This would make it easier for users to update values that are stored in the Secret Store, without having to create their own boto3 client.

### Solution/User Experience

The parameters module provides a convenient and easy-to-use way to read values from the Secret and Parameter Store. It eliminates the need to set up your own session, making it a more streamlined and efficient process. I would like to replicate these features by adding the ability to write values back to the stores.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [X] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

heitorlessa:
hey @stephenbawks thanks for another nice idea! Got any ideas on what the experience could look like?

Like this?

```python
provider = SSMProvider()
provider.set(param_or_secret_name, value, **sdk_options)
```

SSM has quite a few options so I'd start by defining what are the most common options so we can have defaults or infer to keep it simple: https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ssm/client/put_parameter.html

stephenbawks:
Yah I started on this actually this morning. Might commit some early stuff and get some initial feedback since the parameter code is quite complex like you are saying with so many options. 
@heitorlessa

leandrodamascena:
Hei @heitorlessa and @stephenbawks! Is this issue related to the PR https://github.com/aws-powertools/powertools-lambda-python/pull/2858?

I really like this idea because we can make it easy to set and update secrets and parameters. Should we include AppConfig also? 

Heitor's suggestion:
> hey @stephenbawks thanks for another nice idea! Got any ideas on what the experience could look like?
> 
> Like this?
>
> ```python
> provider = SSMProvider()
> provider.set(param_or_secret_name, value, **sdk_options)
> ```

I like this experience, but we need to add high-level functions also, right?

So, lets see if we can work to get it merged soon.

Thanks

[...]

PR #2858, opened 2023-07-27
feat(parameters): add feature for creating and updating Parameters and Secrets

### LINKED ISSUE #3040: RFC: Setting SSM Parameters and Secret Manager Secrets

https://github.com/aws-powertools/powertools-lambda-python/issues/2826

Parameters

### Summary

There are many use cases where the Parameter store is used with PowerTools everywhere.  It's an invaluable tool for getting SSM Parameters and the ability to cache them has been fantastic to save time going back to SSM repeatedly.  There is one thing that I find to be missing and that is the ability to put SSM parameters as well as the ability to write secrets back to Secrets Manager as well.  

### Parameter Store
Adding the function to `put_parameter` back to SSM with the ability to make sure it's also being cached when it is written back.  This should also include the ability to have the `overwrite` flag as optional.  When parameters are being over-written to Parameter store you need to set the overwrite flag because it's set to `false` by default.  

### Secrets Manager
Would also like to add the ability to write values to the Secrets Store as well.  There are many uses, especially in Lambda where I need to cache/write something secret, (JWT Tokens for example) so that multiple Lambda's are not all going back to get a new token every time a new one is invoked.  "Caching" that value in Secret Manager (or as a SecureString in Parameter Store) is a good use case for this.  

### Tasks

- [ ] Define the API for creating/updating secrets/parameters
- [ ] Define whether we will cache
- [ ] Create a PoC to prove the experience.

### Parameter Store
For getting values in SSM, Powertools is using `get_parameter`.  I am proposing that we add a new option for `put_paramter` for the ability to write a value back to a parameter in SSM.  

By default, you have to specify the `overwrite=true` to write a value to a parameter if there is already an existing value.  I am wondering if this should be set to `true` by default or if we should be setting that on every time you need to overwrite a value.  When it is written to Parameter store the `put_parameter` will return a response that includes the new version of the parameter.  So in theory there is already the ability to go back versions and it will hold up to 100 of the previous versions before it will start rolling over the oldest.

### Secret Manager
Adding the ability to create, set, and update secrets in Secret Manager.  

I think `create()` makes sense for creating a new one.

`set()` might be translated into boto3 as `put_secret_value`.  The newly created secret would become the latest version of the secret.

`update()` sounds very similar to `set()` but could be used for changing the version of the secret that would be retrieved.

`update(secret_name="mySuperSecret", secret_value="toomanysecrets")`

`update(secret_name="mySuperSecret", secret_value="toomanysecrets", stage="development")`

`get()` would be nice to have with the optional way to get different versions of a secret.

[`generate_random_secret()` ](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/get_random_password.html) is also a nice feature for Secrets Manager.  This might be a really nice way for people to generate random secrets that can be added to Powertools and saving the need for our fellow Powertools-ers to write custom code to generate random strings or secrets.

### Secrets Manager
I think being able to cache that value would be great but I may have to think through if they are updating a secret with a different version.

I have been thinking about how this might eventually be used for versioning around being able to do "canary" or even cutovers to new secrets. But that is probably future work....

### Secrets Manager 

When it comes to having the SecretsManager class also cache the new value, I think it would be great but I may have to think through if they are updating a secret with a different version. Do we cache the new version AND the old version if it exists?  What would be the behavior around this?

_No response_

### Alternative solutions

_No response_

### Acknowledgment

stephenbawks:
**Issue number:** #2826 

## Summary

This PR boosts parameter and secret management by allowing updates and creations. With enhanced functionality, it allow users to efficiently manage existing parameters and secrets, while also facilitating the creation of new ones as needed.

#### Setting parameter

```python


    try:
        # Set a single parameter, returns the version ID of the parameter
        parameter_version = parameters.set_parameter(name="/mySuper/Parameter", value="PowerToolsIsAwesome")

        return {"mySuperParameterVersion": parameter_version, "statusCode": 200}
    except parameters.exceptions.SetParameterError as error:
        return {"comments": None, "message": str(error), "statusCode": 400}
```

#### Setting secret

```python
from typing import Any

logger = Logger(serialize_stacktrace=True)


def access_token(client_id: str, client_secret: str, audience: str) -> str:
    # example function that returns a JWT Access Token
    # add your own logic here
    return f"{client_id}.{client_secret}.{audience}"


    try:
        client_id: Any = parameters.get_parameter("/aws-powertools/client_id")
        client_secret: Any = parameters.get_parameter("/aws-powertools/client_secret")
        audience: Any = parameters.get_parameter("/aws-powertools/audience")

        jwt_token = access_token(client_id=client_id, client_secret=client_secret, audience=audience)

        # set-secret will create a new secret if it doesn't exist and return the version id
        update_secret_version_id = parameters.set_secret(name="/aws-powertools/jwt_token", value=jwt_token)

        return {"access_token": "updated", "statusCode": 200, "update_secret_version_id": update_secret_version_id}
    except parameters.exceptions.SetSecretError as error:
        logger.exception(error)
        return {"access_token": "updated", "statusCode": 400}
```
## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


stephenbawks:
@heitorlessa @aradyaron @leandrodamascena 
Wondering if I can get a temperature check on this PR so far?

heitorlessa:
@sthulb would you be able to do a quick pass on UX?

I can look into it tomorrow afternoon (Boston time this week).

Hey @stephenbawks Leandro, Ruben and I are on events this week (delays) so we appreciate the nudge

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -116,6 +118,69 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
         """
         raise NotImplementedError()
 
+    def set_secret(
+            self,
+            *, # force keyword arguments
+            name: str,
+            secret: Optional[str],
+            secret_binary: Optional[str],
```
aradyaron: I suggest to put the * after the secret key, as

set(key,value) makes a lot of sense.

I also suggest to change secret to value

```suggestion
    def set_secret(
            self,
            name: str,
            value: Union[str, bytes],
             *, # force keyword arguments

```

stephenbawks: I can move that and change that to `value`.  However `value` or `secret_binary` would be required.  One or the other.  That was why I had left it as `Optional`. 

@aradyaron 

**EDIT**
Nevermind I see what you are suggesting.  Removing `secret_binary`.  Good idea.

aradyaron: Once 'value' is either bytes or str, you can check the type by yourself and occupy the correct parameter accordingly, having better UX, and less error prone than the boto3 sdk
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/put_secret_value.html#:~:text=of%20the%20secret.-,You%20must%20include%20SecretBinary%20or%20SecretString%2C%20but%20not%20both.,-We%20recommend%20you

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -188,6 +191,72 @@ def _get(self, name: str, decrypt: bool = False, **sdk_options) -> str:
...
+        self,
+        path: str,
+        value: str,
+        *, # force keyword arguments
+        parameter_type: SSM_PARAMETER_TYPES = "String",
+        overwrite: bool = False,
+        tier: SSM_PARAMETER_TIER = "Standard",
+        description: Optional[str] = None,
+        kms_key_id: Optional[str] = None,
+        transform: Optional[str] = None,
```
leandrodamascena: What is the idea of the `transform` parameter here? I think it is not needed here.

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -188,6 +191,72 @@ def _get(self, name: str, decrypt: bool = False, **sdk_options) -> str:
 
         return self.client.get_parameter(**sdk_options)["Parameter"]["Value"]
 
+    def set(
```
leandrodamascena, line 211: set is a built-in name in Python and we should avoid that.

leandrodamascena (changes requested):
Hi @stephenbawks! I made two comments in this PR, but I have an idea of how we can work together to achieve this.

The RFC looks good to me and we can move forward with the implementation to validate things, but I see that things are missing, so I can work to change some things in this PR and have the code ready to validate this PoC. Make sense?


I tried to run the code and I couldn't because I'm getting errors:

```python
from aws_lambda_powertools.utilities.parameters import set_parameter

    test = set_parameter("/aaaaa","xyz")
```

`{"errorMessage": "Can't instantiate abstract class SSMProvider with abstract method _set", "errorType": "TypeError", "requestId": "435ee1a9-7365-4533-b6c3-8db9237b7953", "stackTrace": ["  File \"/var/task/app.py\", line 5, in lambda_handler\n    setar = set_parameter(\"/aaaaa\",\"xyz\")\n", "  File \"/var/task/aws_lambda_powertools/utilities/parameters/ssm.py\", line 838, in set_parameter\n    DEFAULT_PROVIDERS[\"ssm\"] = SSMProvider()\n"]}%      `

aws_lambda_powertools/utilities/parameters/ssm.py
stephenbawks: Yah I was going to remove the transform but just wanted to make sure no one could think of a reason to keep it that I was missing.

aws_lambda_powertools/utilities/parameters/ssm.py
stephenbawks: Changed it over to `_set`

aws_lambda_powertools/utilities/parameters/ssm.py
stephenbawks: Updated and tested.  Working on my side.

stephenbawks:
One thing I just ran into and wondering what others think, but if the Secrets Manager Secret does NOT exist, the current code is throwing this error:

```
aws_lambda_powertools.utilities.parameters.exceptions.SetParameterError: An error occurred (ResourceNotFoundException) when calling the PutSecretValue operation: Secrets Manager can't find the specified secret.
```

So I am wondering if the secret does not exist should be create it? Or should we default to the secret existing before someone can write a value to it?

The current behavior when using `set_parameter` is that it will create the parameter if it does not exist.  Wondering if should make the behavior the same for `set_secret` to match that or not?

@leandrodamascena  @aradyaron

aradyaron:
> One thing I just ran into and wondering what others think, but if the Secrets Manager Secret does NOT exist, the current code is throwing this error:
> 
> ```
> aws_lambda_powertools.utilities.parameters.exceptions.SetParameterError: An error occurred (ResourceNotFoundException) when calling the PutSecretValue operation: Secrets Manager can't find the specified secret.
> ```
> 
> So I am wondering if the secret does not exist should be create it? Or should we default to the secret existing before someone can write a value to it?
> 
> The current behavior when using `set_parameter` is that it will create the parameter if it does not exist. Wondering if should make the behavior the same for `set_secret` to match that or not?
> 
> @leandrodamascena @aradyaron

IMHO, I think that catching this error and trying to create the secret is the wanted behavior.
That said, secrets have additional configurations upon creation, meaning that implementing your creation will NOT include all the options.
It feels like a boolean for 'create if not exist' shuold be added.
@leandrodamascena, @stephenbawks  What do you think about adding it as a config to the Secrets provider, like I suggested here https://github.com/aws-powertools/powertools-lambda-python/issues/3040#issuecomment-1729989108

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -116,6 +119,72 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+            value = self.client.put_secret_value(**sdk_options)
+            return value["VersionId"]
+        except ClientError as exc:
+            if exc.response['Error']['Code'] == 'ResourceNotFoundException' and create_not_exists:
+                sdk_options.pop("SecretId")
+                sdk_options["Name"] = name
+                value = self.client.create_secret(**sdk_options)
+                return value["VersionId"]
+        except Exception as exc:
+            raise SetParameterError(str(exc)) from exc
```
aradyaron, line 186:
```suggestion
        except ClientError as exc:
            if exc.response['Error']['Code'] == 'ResourceNotFoundException' and create_not_exists:
                sdk_options.pop("SecretId")
                sdk_options["Name"] = name
                value = self.client.create_secret(**sdk_options)
                return value["VersionId"]
          raise SetParameterError(str(exc)) from exc

        except Exception as exc:
            raise SetParameterError(str(exc)) from exc
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -116,6 +119,72 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+
+        try:
+            value = self.client.put_secret_value(**sdk_options)
+            return value["VersionId"]
+        except ClientError as exc:
+            if exc.response['Error']['Code'] == 'ResourceNotFoundException' and create_not_exists:
+                sdk_options.pop("SecretId")
+                sdk_options["Name"] = name
+                value = self.client.create_secret(**sdk_options)
+                return value["VersionId"]
```
aradyaron:
```suggestion
            if exc.response['Error']['Code'] != 'ResourceNotFoundException' :
                raise SetParameterError('Failed to set parameter') from exc
            elif not raise SetParameterError(str(exc)) from exc:
                raise SetParameterError('Parameter does not exist, create before setting') from exc

            sdk_options.pop("SecretId")
            sdk_options["Name"] = name
            try:
              value = self.client.create_secret(**sdk_options)
              return value["VersionId"]

            except Exception as exc:
                raise SetParameterError('Failed to create parameter') from exc

```
Please note indentation might be off

aradyaron (commented):
Looks really nice, suggested some error handling additions

stephenbawks:
> Looks really nice, suggested some error handling additions

One follow up I just noticed is that I did add the `create_not_exists` `bool` and wondering if that should need to be specified before it will do the actual create or should we just default to create first unless specified?

@aradyaron @leandrodamascena

aradyaron:
> > Looks really nice, suggested some error handling additions
> 
> One follow up I just noticed is that I did add the `create_not_exists` `bool` and wondering if that should need to be specified before it will do the actual create or should we just default to create first unless specified?
> 
> @aradyaron @leandrodamascena

Yeah I agree, I think it should be called ‘create’ and default to True

leandrodamascena:
Hello everyone! Giving a heads-up here, thanks for working on this. I couldn't review it due to some exceptional issues. I will work this weekend to review it and make some comments/changes.

Thank you so much for continuing to work on this PR and bringing this utility, I'm sure we will merge this very soon.

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -116,6 +119,79 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
         """
         raise NotImplementedError()
 
+    def _set(
+        self,
+        name: str,
+        value: Union[str, dict, bytes],
+        *, # force keyword arguments
+        idempotency_id: Optional[str] = None,
```
leandrodamascena: What is the idea of `idempotency_id`? Is this field an alias for [ClientRequestToken](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/create_secret.html)? If so, I think it might confuse customers because it's not explicitly what idempotency means here. I think we can think about other name here.

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -116,6 +119,79 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+            elif not create:
+                raise SetParameterError("Parameter does not exist, create before setting or set 'create' to True.")
+            else:
+                sdk_options.pop("SecretId")
+                sdk_options["Name"] = name
+                try:
+                    value = self.client.create_secret(**sdk_options)
+                    return value["VersionId"]
+                except Exception as exc:
+                    raise SetParameterError(str(exc)) from exc
```
leandrodamascena: Can we simplify and comment on this code? I had to read 3 times to understand what it is doing. 

Also, I'm still not convinced if a unique method called `set_secret` is better than 2 methods called `update_secret` and `create_secret`. I don't know, but looking from the customers' side, I think the client can lose control of what they are creating/updating. This can incur extra costs and a bad user experience.

What do you think @heitorlessa @stephenbawks @aradyaron?

leandrodamascena (changes requested):
Hi @stephenbawks! Let's focus on this PR to merge it soon!

I left two comments and there are no tests for this new methods. Can you add this? Thanks

aws_lambda_powertools/utilities/parameters/secrets.py
aradyaron: I agree to the separation.
Secrets creation is, AFAIK, done in a different stage in the lifetime of the secret

heitorlessa: We need to two distinct decisions here:

A. **Keep a standalone `set_secret` OR keep at the provider level only (`SecretProvider().set`** 

B. **Abstract branching and error handling logic from customers on secret creation/update**

I'll leave A decision to @leandrodamascena to discuss with all of you. If you do go this route, keep in mind docs must be crisp AND we should accept a provider option, to allow customers to bring their own secrets provider (extended or something else entirely). The latter is a design flaw of parameters today, all standalone functions should accept an optional provider option.

On B, we should have a single method with sane but overridable defaults.

**Reasoning**

In a lower level SDK, you should have distinct operations for Create and Update. After all, you might use CloudFormation, AWS Console, AWS CLI, and many other methods to create secrets -- what do they have in common?

They are not a managed runtime, it's a manual and immediate feedback-response scenario. If something goes wrong, you receive the error upfront and have the opportunity to do something about it. 

**However**... when you factor an application creating/updating a secret, you now have to create error handling logic, IF/ELSE conditions to know when to use create vs update, and handle idempotency (update secret again but different value should fail), etc.

If we were to offer a `create_secret`, `update_secret`, we're not providing any value to the end customer compared to the SDK other than caching.

Instead, they would have a Lambda function that would look like this pseudo code.

```python
from powertools import create_secret, update_secret

def handler(event, context):
    if event.get("options").get("create"):
        try:
            create_secret(....)
        except SecretAlreadyExists as e:
            # more logic to handle what to do here
    elif event.get("options").get("update"):
        try:
            update_secret(...)
        ...
```

VS a `set` function that implies it'll create and/or update if one already exists

```python
from powertools import set_secret

def handler(event, context):
    set_secret(...)
```

If you now want to be intentional and not create a secret, you can now see it in the code clearly, and any error will now be an indicative that you have an anomaly in your runtime/code/metrics.

```python
from powertools import set_secret

def handler(event, context):
    set_secret(..., create=False)
```

The only thing left here would be a good citizen to explain to customers that they should have a secrets strategy, otherwise they will end up with a big bill if they create unique secrets for every Lambda invocation. [There's a warning in the SDK too on creating one secret value every 10 minutes](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/put_secret_value.html).

stephenbawks: I do tend to lean towards the fact that we should set the `create` option to False by default.  

We want someone to specify when the secret should be created.  I think to @heitorlessa  point we would not want someone to have some process that is failing and being re-invoked and they happen to be using some random name for the secret and then they then have 193493 (exaggerating here a bit.... but possibly... ) secrets created for them.

I think they could certainly default `create` to be `False` and or catch the `SetParameterError` and THEN turn the `create` to `True`.

```python
def handler(event, context):
    try:
        set_secret(....)
    except SetParameterErroras e:
        set_secret(...., create=True)
        ...
```

leandrodamascena:
Hi @stephenbawks! After some time, I'll work on this PR again and will merge it soon!

@roger-zhangg worked on making some changes on his branch https://github.com/roger-zhangg/aws-lambda-powertools-python/tree/setparams. Tomorrow I will create a patch for this branch and apply it here to discuss the changes with you @stephenbawks!

Thanks for your patience.

heitorlessa:
Rebasing to confirm latest CI checks before review

heitorlessa:
hey @stephenbawks when you get a chance, could you look into why tests are failing - feature flags use Parameters Providers under the hood hence why they're failing too :) 

I can look into it next week if you're unable (day-to-day work etc.)

image

stephenbawks:
> hey @stephenbawks when you get a chance, could you look into why tests are failing - feature flags use Parameters Providers under the hood hence why they're failing too :)
> 
> I can look into it next week if you're unable (day-to-day work etc.)
> 

Going to see if I can knock this out this weekend. @heitorlessa

aws_lambda_powertools/utilities/parameters/secrets.py
stephenbawks: For the time being I removed the create option.  Have to revisit that one.

aws_lambda_powertools/utilities/parameters/secrets.py
stephenbawks: Changed this to `client_request_token`  so it lines up with the boto argument name. 
@leandrodamascena

heitorlessa:
Updating the fork to force CI to kick in to understand why tests are failing ;)

heitorlessa:
removed `_set` as an abstract method since that would be a breaking change hence why AppConfig and faker tests were failing :) 

I'll start doing a deep review tomorrow and hopefully get this to a great shape by EOW!

Looking promising!

heitorlessa:
Did a quick look through from tests back to implementation. I'm gonna through pending conversations to re-learn context I missed here, before I make progress.

I'll edit tasks below as I learn new or break existing ones. This should give a better sense of progress for anyone watching.

**Tasks**

* [x] `SSMProvider`
  * [x] Use `@overload` to handle `value` type for `String` vs `StringList`
  * [x] Implement `set` with simplest logic
  * [x] Handle SDK options
  * [x] Enforce keyword argument (`*` syntax)
  * [x] Tests
    * [x] Revisit `test_ssm_provider_set`
    * [x] Revisit `test_ssm_provider_set_default_config`
    * [x] Create new test with custom options like Description, Tier, others
* [x] `set_parameter`
  * [x] Rename `path` argument to `name` to allow future growth _(imagine setting multiple params at once)_
  * [x] Doc string update (it's partly reusing `get_parameter`)
  * [x] Revisit whether boto put_parameter accepts sane defaults `""` over `None` to the same effect _(reduce CPU cycles)_
  * [x] Double check defaults
    * [x] Parameter type
    * [x] Parameter tier
* [x] `set_secret`
  * [x] Learn about [current implementation](https://github.com/aws-powertools/powertools-lambda-python/pull/2858/files#diff-10991aa2442d160cf5ecb9808fce7d5063455ed52cb159d005d8b8f61a238279R123)
  * [x] Think whether we can have a better name for `client_request_token` _(it'll be important if we support any provider)_
  * [x] Think about creating a better return contract
    * Example: what would people do with the returned Secret ID if it's the same as the secret name?
  * [x] Raise a specific Secret exception, not Parameter
* [x] Documentation
  * [x] Revisit IAM permissions
  * [x] Revisit examples
* [x] Exceptions
  * [x] `SetParameterError`
  * [x] One for secret
* [x] Add debugging
* [x] `BaseProvider`
  * [x] Implement `set` with common parameters to allow extension
    * [x] `name`
    * [x] `value`
    * [x] `overwrite`
  * [x] Delete `_set` as setting a parameter and a secret are wildly different, we won't be able to get static typing correctly
* [x] Re-learn requirements/UX
  * [x] Param/Secret overwrite
  * [x] Re-read pending conversations

**Random notes (lowest priority)**

* [ ] Consider whether Enum provides a more natural ergonomics _(those w/o mypy, using Console)_
* [ ] Consider whether we want to provide serialization, e.g., `set_parameter/secret(value={})`
  * Challenge here is that they must be equally aware when fetching; easy to make a mistake if list[str] vs list[dict] on auto-serialization
* [x] Double check `@overload` in SSMProvider `set` when `SecureString` is used, since KMS is mandatory

>  We made a mistake in naming `SecretsProvider` instead of `SecretsManagerProvider`, as that would allow general purpose secret solutions

**Unrelated**

* Maybe introduce the concept of `provider` argument to higher level fn to ease testing

aws_lambda_powertools/utilities/parameters/exceptions.py
```diff
@@ -9,3 +9,7 @@ class GetParameterError(Exception):
 
 class TransformParameterError(Exception):
     """When a provider fails to transform a parameter value"""
+
+
+class SetParameterError(Exception):
```
heitorlessa, line 14: we need another one for secrets.

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -115,6 +120,65 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+            sdk_options["VersionStages"] = version_stages
+        if client_request_token:
+            sdk_options["ClientRequestToken"] = client_request_token
+
+        try:
+            value = self.client.put_secret_value(**sdk_options)
+            return value["VersionId"]
+        except ClientError as exc:
+            if exc.response["Error"]["Code"] != "ResourceNotFoundException":
+                raise SetParameterError(str(exc)) from exc
```
heitorlessa: we need a more specific secret exception here

aws_lambda_powertools/utilities/parameters/exceptions.py
leandrodamascena: Done

aws_lambda_powertools/utilities/parameters/secrets.py
leandrodamascena: Done

leandrodamascena:
Hey everyone! I've made some changes based on @heitorlessa's list/feedback, and now we're nearly done—just a few final touches left to complete the secrets stuff.

* [x] `set_secret`
  * [x] Learn about [current implementation](https://github.com/aws-powertools/powertools-lambda-python/pull/2858/files#diff-10991aa2442d160cf5ecb9808fce7d5063455ed52cb159d005d8b8f61a238279R123)
  * [x] Think whether we can have a better name for `client_request_token` _(it'll be important if we support any provider)_
  * [x] Think about creating a better return contract
    * Example: what would people do with the returned Secret ID if it's the same as the secret name?
  * [x] Raise a specific Secret exception, not Parameter
  * [x] Refactor the method to update or create
  * [x] Review the logic

leandrodamascena:
This PR is now ready for the final reviews before merging! Huge thanks to @stephenbawks and @heitorlessa for all their hard work and contributions!

Following some additional information about itens I've addressed:

> Think whether we can have a better name for client_request_token (it'll be important if we support any provider)

We are working exclusively with AWS on this provider, if we have to extend to another provider from `BaseProvider`, we can think of another name for this provider, but I think `client_request_token` makes a lot of sense to be 100% following the SDK.

> Think about creating a better return contract

Yes, it makes sense to return the dict that comes from the SDK so that the customer can choose what information he wants to use. For this, I created `SetSecretResponse` to help with the IDE's typing hit

Please let me know if you have any questions.

rubenfonseca:
Starting reviewing now

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -154,6 +154,9 @@ def _get(self, name: str, **sdk_options) -> Union[str, bytes, Dict[str, Any]]:
         """
         raise NotImplementedError()
 
+    def set(self, name: str, value: Any, *, overwrite: bool = False, **kwargs):
```
rubenfonseca, line 157: Needs docstring

aws_lambda_powertools/utilities/parameters/exceptions.py
```diff
@@ -9,3 +9,11 @@ class GetParameterError(Exception):
...
+    """When a provider raises an exception on setting a SSM parameter"""
```
rubenfonseca: To keep it consistent with the Get exceptions on the same file, I would recommend:
```suggestion
    """When a provider raises an exception on writing a parameter"""
```

aws_lambda_powertools/utilities/parameters/exceptions.py
```diff
@@ -9,3 +9,11 @@ class GetParameterError(Exception):
...
+
+
+class SetSecretError(Exception):
+    """When a provider raises an exception on setting an AWS Secret"""
```
rubenfonseca: To keep it consistent with the Get exceptions on the same file, I would recommend:

```suggestion
    """When a provider raises an exception on writing a secret"""
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -2,21 +2,27 @@
...
-from aws_lambda_powertools.utilities.parameters.types import TransformOptions
+from aws_lambda_powertools.utilities.parameters.types import SetSecretResponse, TransformOptions
 
 if TYPE_CHECKING:
     from mypy_boto3_secretsmanager import SecretsManagerClient
 
 from aws_lambda_powertools.shared import constants
 from aws_lambda_powertools.shared.functions import resolve_max_age
+from aws_lambda_powertools.utilities.parameters.base import DEFAULT_MAX_AGE_SECS, DEFAULT_PROVIDERS, BaseProvider
+from aws_lambda_powertools.utilities.parameters.exceptions import SetSecretError
```
rubenfonseca: Can we merge all the powertools imports instead of splitting them with TYPE_CHECKING?

```suggestion
if TYPE_CHECKING:
    from mypy_boto3_secretsmanager import SecretsManagerClient

from aws_lambda_powertools.shared import constants
from aws_lambda_powertools.shared.functions import resolve_max_age
from aws_lambda_powertools.utilities.parameters.base import DEFAULT_MAX_AGE_SECS, DEFAULT_PROVIDERS, BaseProvider
from aws_lambda_powertools.utilities.parameters.exceptions import SetSecretError
from aws_lambda_powertools.utilities.parameters.types import SetSecretResponse, TransformOptions
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -117,6 +123,135 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+        """
+        Modify the details of a secret or create a new secret if it doesn't already exist.
+        It includes metadata and the secret value.
+
+        We aim to minimize API calls by assuming that the secret already exists and needs updating.
+        If it doesn't exist, we attempt to create a new one. Refer to the following workflow for a better understanding:
+
+
+                          ┌────────────────────────┐      ┌─────────────────┐
+                ┌───────▶│Resource NotFound error?│────▶│Create Secret API│─────┐
```
rubenfonseca, line 176: Can we add a space or two to make the box align? Great explanation btw :D

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -117,6 +123,135 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+        self,
+        name: str,
+        value: Union[str, dict, bytes],
+        *,  # force keyword arguments
+        client_request_token: Optional[str] = None,
+        **sdk_options,
+    ) -> SetSecretResponse:
+        """
+        Modify the details of a secret or create a new secret if it doesn't already exist.
+        It includes metadata and the secret value.
```
rubenfonseca: I don't understand this sentence :( Either remove it or rewrite it please. What is "metadata"?

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -117,6 +123,135 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+        └─────────────────┘                                              └─────────────────────┘
+
+        Parameters
+        ----------
+        name: str
+            The ARN or name of the secret to add a new version to or create a new one.
+        value: str, dict or bytes
+            Specifies text data that you want to encrypt and store in this new version of the secret.
+        client_request_token: str, optional
+            This value helps ensure idempotency. Recommended that you generate
```
rubenfonseca:
```suggestion
            This value helps ensure idempotency. It's recommended that you generate
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -117,6 +123,135 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+            a UUID-type value to ensure uniqueness within the specified secret.
+            This value becomes the VersionId of the new version. This field is
+            auto-populated if not provided.
```
rubenfonseca:
```suggestion
            auto-populated if not provided, but no idempotency will be enforced this way.
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -117,6 +123,135 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+                )
+
+        URLs:
+        -------
+            https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/put_secret_value.html
+            https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/create_secret.html
+        """
+
+        if isinstance(value, dict):
+            value = json.dumps(value)
```
rubenfonseca:
```suggestion
	    from aws_lambda_powertools.shared.json_encoder import Encoder

            value = json.dumps(value, cls=Encoder)
```

Just to make sure we support more than the standard types

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -117,6 +123,135 @@ def _get_multiple(self, path: str, **sdk_options) -> Dict[str, str]:
...
+
+        if isinstance(value, bytes):
+            sdk_options["SecretBinary"] = value
+        else:
+            sdk_options["SecretString"] = value
+
+        if client_request_token:
+            sdk_options["ClientRequestToken"] = client_request_token
```
rubenfonseca, line 243: I'm not seeing a default ClientRequestToken being generated as described on the docstring. Is it a bug? Or does the SDK generate one automatically?

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -224,3 +359,88 @@ def get_secret(
...
+def set_secret(
+    name: str,
+    value: Union[str, bytes],
+    *,  # force keyword arguments
+    client_request_token: Optional[str] = None,
+    **sdk_options,
+) -> str:
+    """
+    Modify the details of a secret or create a new secret if it doesn't already exist.
+    It includes metadata and the secret value.
```
rubenfonseca: I don't understand this sentence :( Either remove it or rewrite it please. What is "metadata"?

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -224,3 +359,88 @@ def get_secret(
...
+
+    We aim to minimize API calls by assuming that the secret already exists and needs updating.
+    If it doesn't exist, we attempt to create a new one. Refer to the following workflow for a better understanding:
+
+
+                      ┌────────────────────────┐      ┌─────────────────┐
+            ┌───────▶│Resource NotFound error?│────▶│Create Secret API│─────┐
```
rubenfonseca, line 378: Same formatting, just a couple more spaces or dashes? :D It's REALLY awesome

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -224,3 +359,88 @@ def get_secret(
...
+    ----------
+    name: str
+        The ARN or name of the secret to add a new version to or create a new one.
+    value: str, dict or bytes
+        Specifies text data that you want to encrypt and store in this new version of the secret.
+    client_request_token: str, optional
+        This value helps ensure idempotency. Recommended that you generate
+        a UUID-type value to ensure uniqueness within the specified secret.
+        This value becomes the VersionId of the new version. This field is
+        auto-populated if not provided.
```
rubenfonseca:
```suggestion
            auto-populated if not provided, but no idempotency will be enforced this way.
```

aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -224,3 +359,88 @@ def get_secret(
...
+
+    Raises
+    ------
+    SetSecretError
+        When attempting to update or create a secret fails.
+
+    Returns:
+    -------
+    SetSecretResponse:
+        The dict returned by boto3.
```
rubenfonseca, line 409: The function signature is returning a `str`. Which one is correct?

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -169,6 +179,126 @@ def get(  # type: ignore[override]
...
+        overwrite: bool, optional
+            If the parameter value should be overwritten, False by default
+        tier: str, optional
+            The parameter tier to use.  Allowed values are Standard, Advanced, and Intelligent-Tiering
+        description: str, optional
+            The description of the parameter
+        kms_key_id: str, optional
+            The KMS key id to use to encrypt the parameter
+        sdk_options: dict, optional
+            Dictionary of options that will be passed to the Parameter Store get_parameter API call
```
rubenfonseca: Same order as declared on the signature:

```suggestion
        name: str
            The fully qualified name includes the complete hierarchy of the parameter name and name.
        value: str
            The parameter value
        overwrite: bool, optional
            If the parameter value should be overwritten, False by default
        description: str, optional
            The description of the parameter
        parameter_type: str, optional
            Type of the parameter.  Allowed values are String, StringList, and SecureString
        tier: str, optional
            The parameter tier to use.  Allowed values are Standard, Advanced, and Intelligent-Tiering
        kms_key_id: str, optional
            The KMS key id to use to encrypt the parameter
        sdk_options: dict, optional
            Dictionary of options that will be passed to the Parameter Store get_parameter API call
```

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -811,6 +941,81 @@ def get_parameters(
...
+    overwrite: bool, optional
+        If the parameter value should be overwritten, False by default
+    tier: str, optional
+        The parameter tier to use.  Allowed values are Standard, Advanced, and Intelligent-Tiering
+    description: str, optional
+        The description of the parameter
+    kms_key_id: str, optional
+        The KMS key id to use to encrypt the parameter
+    sdk_options: dict, optional
+        Dictionary of options that will be passed to the Parameter Store get_parameter API call
```
rubenfonseca: Same order as declared on the signature:

```suggestion
        name: str
            The fully qualified name includes the complete hierarchy of the parameter name and name.
        value: str
            The parameter value
        overwrite: bool, optional
            If the parameter value should be overwritten, False by default
        description: str, optional
            The description of the parameter
        parameter_type: str, optional
            Type of the parameter.  Allowed values are String, StringList, and SecureString
        tier: str, optional
            The parameter tier to use.  Allowed values are Standard, Advanced, and Intelligent-Tiering
        kms_key_id: str, optional
            The KMS key id to use to encrypt the parameter
        sdk_options: dict, optional
            Dictionary of options that will be passed to the Parameter Store get_parameter API call
```

examples/parameters/src/getting_started_set_ssm_parameter_overwrite.py
```diff
@@ -0,0 +1,17 @@
+from aws_lambda_powertools.utilities import parameters
+from aws_lambda_powertools.utilities.typing import LambdaContext
+
+
+def lambda_handler(event: dict, context: LambdaContext) -> dict:
+    try:
+        # Set a single parameter, but overwrite if it already exists
+        # by default, overwrite is False, explicitly set it to True
```
rubenfonseca: Took me a while to understand the sentences without punctuation :D 

```suggestion
        # Set a single parameter, but overwrite if it already exists.
        # Overwrite is False by default, so we explicitly set it to True
```

rubenfonseca (changes requested):
Great job, first pass. Will look into the documentation on the next pass.

aws_lambda_powertools/utilities/parameters/secrets.py
leandrodamascena: According to the AWS Documentation, it's autopopulated by boto3 if not provided.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/put_secret_value.html

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/secretsmanager/client/create_secret.html

aws_lambda_powertools/utilities/parameters/secrets.py
leandrodamascena: Please take a screenshot of what you especially want. I only partially understand hahaha

aws_lambda_powertools/utilities/parameters/secrets.py
leandrodamascena: Leftover, fixed!

aws_lambda_powertools/utilities/parameters/secrets.py
leandrodamascena: Yep! Some leftover, I removed the sentence.

aws_lambda_powertools/utilities/parameters/secrets.py
leandrodamascena: Please take a screenshot of what you especially want. I only partially understand hahaha

aws_lambda_powertools/utilities/parameters/base.py
leandrodamascena: Done

docs/utilities/parameters.md
```diff
@@ -32,8 +32,10 @@ This utility requires additional permissions to work as expected.
...
 | SSM       | **`get_parameters`**, **`SSMProvider.get_multiple`**                   | **`ssm:GetParametersByPath`**                                                        |
 | SSM       | **`get_parameters_by_name`**, **`SSMProvider.get_parameters_by_name`** | **`ssm:GetParameter`** and **`ssm:GetParameters`**                                   |
+| SSM       | **`set_parameter`**, **`SSMProvider.set_parameter`**                   | **`ssm:PutParameter`**                                                               |
 | SSM       | If using **`decrypt=True`**                                            | You must add an additional permission **`kms:Decrypt`**                              |
 | Secrets   | **`get_secret`**, **`SecretsProvider.get`**                            | **`secretsmanager:GetSecretValue`**                                                  |
+| Secrets   | **`set_secret`**, **`SecretsProvider.set`**                            | **`secretsmanager:PutSecretValue`** and or **`secretsmanager:CreateSecret`**         |
```
rubenfonseca:
```suggestion
| Secrets   | **`set_secret`**, **`SecretsProvider.set`**                            | **`secretsmanager:PutSecretValue`** and **`secretsmanager:CreateSecret`** (if creating secrets)         |
```

docs/utilities/parameters.md
```diff
@@ -84,6 +86,22 @@ For multiple parameters, you can use either:
...
+
+You can set a parameter using the `set_parameter` high-level function. This will create a new parameter if it doesn't exist.
+
+=== "getting_started_set_single_ssm_parameter.py"
+    ```python hl_lines="8"
+    --8<-- "examples/parameters/src/getting_started_set_single_ssm_parameter.py"
+    ```
+
+=== "getting_started_set_ssm_parameter_overwrite.py"
+    There are occasions where sometimes you are setting a parameter and then you may need to update that parameter later on. In this case, you can use the `overwrite` parameter to overwrite the parameter value if it already exists. If you do not set this parameter, then the parameter will not be overwritten and an exception will be raised.
```
rubenfonseca:
```suggestion
    Sometimes you may be setting a parameter that you will have to update later on. Use the `overwrite` option to overwrite any existing value. If you do not set this option, the parameter value will not be overwritten and an exception will be raised.
```

too many parameters ahah

rubenfonseca (dismissed):
Just left two small things, otherwise approved!

rubenfonseca (approved):

leandrodamascena (approved):

PR #2961, opened 2023-08-15
fix(parser): API Gateway V2 request context scope field should be optional

### LINKED ISSUE #2953: Bug: RequestContextV2AuthorizerJwt requires scopes
### Expected Behaviour

I'm using the RequestContextV2 Model for parsing the event payload within a AWS Lambda function called by an AWS API Gateway. The API Gateway is configured to allow only authenticated requests with a valid JWT Token. The token is generated by our self-hosted GitLab instance with the help of the https://docs.gitlab.com/ee/ci/yaml/#id_tokens keyword. 

The token itself has no scopes and it's not possible to configure any. The API Gateway is configured without any scope requirements which is a valid configuration according to AWS docs.

### Current Behaviour

Right now this configuration is not supported by the RequestContextV2AuthorizerJwt Model as it requires that a scopes field is present in the object. The validation fails and an exception is thrown

[ERROR] ValidationError: 1 validation error for APIGatewayProxyEventV2Model
requestContext.authorizer.jwt.scopes
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.1/v/list_type
  File "/opt/python/lib/python3.10/site-packages/aws_lambda_powertools/middleware_factory/factory.py", line 135, in wrapper
    response = middleware()
  File "/opt/python/lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/parser.py", line 83, in event_parser
    parsed_event = parse(event=event, model=model, envelope=envelope) if envelope else parse(event=event, model=model)
  File "/opt/python/lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/parser.py", line 155, in parse
    return envelope().parse(data=event, model=model)
  File "/opt/python/lib/python3.10/site-packages/aws_lambda_powertools/utilities/parser/envelopes/apigwv2.py", line 30, in parse
    parsed_envelope: APIGatewayProxyEventV2Model = APIGatewayProxyEventV2Model.parse_obj(data)
  File "/opt/python/lib/python3.10/site-packages/typing_extensions.py", line 2562, in wrapper
    return __arg(*args, **kwargs)
  File "/opt/python/lib/python3.10/site-packages/pydantic/main.py", line 961, in parse_obj
    return cls.model_validate(obj)
  File "/opt/python/lib/python3.10/site-packages/pydantic/main.py", line 496, in model_validate
    return cls.__pydantic_validator__.validate_python(

### Code snippet

```python
@event_parser(model=Pipeline, envelope=envelopes.ApiGatewayV2Envelope)
def lambda_handler(event: Pipeline, context: LambdaContext):
  pass
```

### Possible Solution

https://github.com/aws-powertools/powertools-lambda-python/compare/develop...smittnacht:powertools-lambda-python:jwt-scopes-as-optional

### Steps to Reproduce

Create an AWS API Gateway with JWT Authorizer and with a Lambda Function integration. Don't configure any required scopes. Send an authenticated request to the endpoint without any scopes defined.

latest

3.10

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #2953 

### Changes

Include the Optional type to make the "scope" field optional.

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/parser/models/apigwv2.py
+++ b/aws_lambda_powertools/utilities/parser/models/apigwv2.py
@@ -25,7 +25,7 @@ class RequestContextV2AuthorizerIam(BaseModel):
 
 class RequestContextV2AuthorizerJwt(BaseModel):
     claims: Dict[str, Any]
-    scopes: List[str]
+    scopes: Optional[List[str]] = None
 
 
 class RequestContextV2Authorizer(BaseModel):
--- a/tests/unit/parser/test_apigwv2.py
+++ b/tests/unit/parser/test_apigwv2.py
@@ -63,6 +63,13 @@ def test_apigw_v2_event_jwt_authorizer():
     assert parsed_event.stageVariables == raw_event["stageVariables"]
 
 
+def test_apigw_v2_event_empty_jwt_scopes():
+    raw_event = load_event("apiGatewayProxyV2Event.json")
+    raw_event["requestContext"]["authorizer"]["jwt"]["scopes"] = None
+
+    APIGatewayProxyEventV2Model(**raw_event)
+
+
 def test_api_gateway_proxy_v2_event_lambda_authorizer():
     raw_event = load_event("apiGatewayProxyV2LambdaAuthorizerEvent.json")
     parsed_event: APIGatewayProxyEventV2Model = APIGatewayProxyEventV2Model(**raw_event)


aws_lambda_powertools/utilities/parser/models/apigwv2.py
```diff
@@ -25,7 +25,7 @@ class RequestContextV2AuthorizerIam(BaseModel):
 
 class RequestContextV2AuthorizerJwt(BaseModel):
     claims: Dict[str, Any]
-    scopes: List[str]
+    scopes: Optional[List[str]] = None
```
heitorlessa, line 28: can we have a quick test to avoid future regression?

We can take this event, remove the `scopes` dict key and run the model - no need to assert is `None`, if the Model works without it, it means validation worked as expected.

https://github.com/aws-powertools/powertools-lambda-python/blob/bfde1c9124910851f3de41b1db54e48e2bba7f16/tests/unit/parser/test_apigwv2.py#L41

heitorlessa (changes requested):
tiny test to avoid regressions

aws_lambda_powertools/utilities/parser/models/apigwv2.py
leandrodamascena: I would like to keep this test and this field to validate that we are parsing as a List when a value is provided.

My thoughts:

1 - I can add more lines in this test to remove this field and run the model again
```python
    .
    .
    .
    assert parsed_event.stageVariables == raw_event["stageVariables"]

    # Removing scope field
    raw_event["requestContext"]["authorizer"]["jwt"]["scopes"] = None
    parsed_event: APIGatewayProxyEventV2Model = APIGatewayProxyEventV2Model(**raw_event)
```

2 - I can create another test to test specific this, but I think this is too much in my opinion.

What do you suggest?

heitorlessa: Create another test specific to the claim being none. It'll be only a few lines, and if it fails we know what contract it broke.

I'll post a sample what I mean by that after a meeting

heitorlessa:
```python
def test_apigw_v2_event_empty_jwt_scopes():
    raw_event = load_event("apiGatewayProxyV2Event.json")
    raw_event["requestContext"]["authorizer"]["jwt"]["scopes"] = None

    APIGatewayProxyEventV2Model(**raw_event)
```

It fails (without this PR)

```python
```python
E   pydantic.error_wrappers.ValidationError: 1 validation error for APIGatewayProxyEventV2Model
E   requestContext -> authorizer -> jwt -> scopes
E     none is not an allowed value (type=type_error.none.not_allowed)

pydantic/main.py:341: ValidationError
```

```

leandrodamascena: Thanks for creating the test @heitorlessa! I created a new one before you send this haha! :heart:

heitorlessa (approved):

PR #3029, opened 2023-08-30
feat(event_source): add Kinesis Firehose Data Transformation data class

### LINKED ISSUE #2440: Feature request: Kinesis Firehose Response Record data class
### Use case

Constructing response objects for use in Kinesis Firehose transformation functions.

This is a continuation of #1059 which describes the event object as well as the response object. The implementation for that issue can be found at #1540, but that does not include the response object.

### Solution/User Experience

A data class that can be populated during the execution of a function that will be properly formed as a response to a `KinesisFirehoseEvent` invocation.

**Rough idea**

```
KinesisFirehoseResponse:
   records: list[KinssisFirehoseResponseRecord]
```

```
KinesisFirehoseResponseRecord:
   record_id: str
   result: Literal["Ok", "ProcessingFailed"]
   data: bytes
   metadata: KinesisFirehoseResponseRecordMetadata
```

```
KinesisFirehoseResponseRecordMetadata:
   partition_keys: dict
```

_Note: ☝🏼 I'm not sure if this is not an exhaustive list of options that can be returned_

### Alternative solutions

Previously, I've used basic dictionaries for this, but it would be nice to have a more structured data class to use.

The Go example in the [Dynamic Partitioning in Kinesis Data Firehose](https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html) has the concept of a `KinesisFirehoseResponse` in their events package.

I believe it would be possible to re-use the `KinesisFirehoseEvent` data class from the `utilities.data_classes` module, but this seems like it is more geared for the event invocation object as opposed to the response object.

### Acknowledgment

roger-zhangg:
**Issue number:**#2440

## Summary
Support Kinesis Firehose Response Record data class in event_sources.

This PR is still in draft status, much to discuss including
- Dataclasses are mostly used for read only Dicts, this request need to construct dict before using the data classes, which is different. Need to review on the current approach
- Discuss what functionality should be supported. E.g (data.from_str, data.from_binary)

### Changes

Added three data_classes and their constructor that supports Kinesis Firehose Response Record

### User experience

**BEFORE**

```python
import base64

print('Loading function')


    output = []

    for record in event['records']:
        print(record['recordId'])
        payload = base64.b64decode(record['data']).decode('utf-8')

        # Do custom processing on the payload here

        output_record = {
            'recordId': record['recordId'],
            'result': 'Ok',
            'data': base64.b64encode(payload.encode('utf-8')).decode('utf-8')
        }
        output.append(output_record)

    print('Successfully processed {} records.'.format(len(event['records'])))

    return {'records': output}
```

**AFTER**

```python
import base64

from aws_lambda_powertools.utilities.data_classes import (
    KinesisFirehoseDataTransformationRecord,
    KinesisFirehoseDataTransformationResponse,
)
from aws_lambda_powertools.utilities.serialization import base64_from_json


    result = KinesisFirehoseDataTransformationResponse()

    for record in event["records"]:
        print(record["recordId"])
        payload = base64.b64decode(record["data"]).decode("utf-8")
        ## do all kind of stuff with payload
        ## generate data to return
        transformed_data = {"tool_used": "powertools_dataclass", "original_payload": payload}

        processed_record = KinesisFirehoseDataTransformationRecord(
            record_id=record["recordId"],
            data=base64_from_json(transformed_data),
        )
        result.add_record(processed_record)

    # return transformed records
    return result.asdict()
```

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


roger-zhangg:
from here https://github.com/aws-powertools/powertools-lambda-python/blob/04cf87f59a49b9cea79925b2af5f5eff48f39355/aws_lambda_powertools/utilities/data_classes/common.py#L13-L14
This `DictWrapper` is designed to be read-only and used by all data-classes. Currently all data-classes in powertools are used to parse events. But for this issue, we are using data-classes to create the event. And this use case doesn't exist in powertools data-classes yet.
In this case I can't set value directly to data class expect init the class with data parameter. As you can see in my current code, 

https://github.com/aws-powertools/powertools-lambda-python/blob/6eb8530165d7ab562c3d12eaa9a43c25ef4072e0/aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py#L89-L103

I'm using a `factory` method to create each of these data classes. But I don't think this is a good solution. So it's better we can discuss to figure out a better way to create these data classes

From my perspective I can think of three possible solution of to support this
- Add a setter method to `DictWrapper` and init the class as a data class
```python
# possible UX
result = KinesisFirehoseResponce({})
for record in event.records:
        data = record.data_as_json
        processed_record = KinesisFirehoseResponceRecord(
            record_id=record.record_id,
            result=FirehoseStateOk,
        )
        processed_record.data_from_json(data)
        result.add_record(processed_record)
return result

```
- Add a new Wrapper class to support this specific case for creating events in lambda
```python
# possible UX
        processed_record = KinesisFirehoseResponceRecord(
            record_id=record.record_id,
            result=FirehoseStateOk,
            data=data,
        )
```
- Do nothing, just keep the current data approach
```python
# possible UX
data={
    "record_id":record.record_id,
    "result":FirehoseStateOk,
    "data":data_raw,
}
record = KinesisFirehoseResponceRecord(data=data)

```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
+from __future__ import annotations
```
leandrodamascena: Please remove this. In Data Class and parser, we are not using the new types. We probably will refactor that when drop support to Python 3.7

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+
+import base64
+from typing import Callable, Iterator, List, Optional, Union
+
+from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
+
+FirehoseStateOk = "Ok"
+FirehoseStateDropped = "Dropped"
+FirehoseStateFailed = "ProcessingFailed"
```
leandrodamascena: We don't need this, we can use Literal.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+    - https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
+    """
+
+    @property
+    def record_id(self) -> str:
+        """Record ID; uniquely identifies this record within the current batch"""
+        return self["recordId"]
+
+    @property
+    def result(self) -> Union[FirehoseStateOk, FirehoseStateDropped, FirehoseStateFailed]:
```
leandrodamascena:
```suggestion
    def result(self) -> Literal["Ok", "Dropped", "ProcessingFailed"]:
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+        return self["metadata"]  # could raise KeyError
+
+    @property
+    def partition_keys(self) -> Optional[dict[str, str]]:
+        """Kinesis stream partition key; present only when Kinesis Stream is source"""
+        return self._metadata["partitionKeys"]
+
+
+def KinesisFirehoseResponseRecordMetadataFactory(
+    partition_keys: dict[str, str],
```
leandrodamascena:
```suggestion
    partition_keys: Dict[str, str],
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+    - https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
+    """
+
+    @property
+    def _metadata(self) -> Optional[dict]:
+        """Optional: metadata associated with this record; present only when Kinesis Stream is source"""
+        return self["metadata"]  # could raise KeyError
+
+    @property
+    def partition_keys(self) -> Optional[dict[str, str]]:
```
leandrodamascena:
```suggestion
    def partition_keys(self) -> Optional[Dict[str, str]]:
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
+from __future__ import annotations
+
+import base64
+from typing import Callable, Iterator, List, Optional, Union
```
leandrodamascena:
```suggestion
from typing import Callable, Iterator, List, Optional, Dict
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+
+class KinesisFirehoseResponseRecordMetadata(DictWrapper):
+    """
+    Documentation:
+    --------------
+    - https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
+    """
+
+    @property
+    def _metadata(self) -> Optional[dict]:
```
leandrodamascena:
```suggestion
    def _metadata(self) -> Optional[Dict]:
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+        """Decoded base64-encoded data as bytes"""
+        return base64.b64decode(self.data)
+
+    @property
+    def data_as_text(self) -> str:
+        """Decoded base64-encoded data as text"""
+        return self.data_as_bytes.decode("utf-8")
+
+    @property
+    def data_as_json(self) -> dict:
```
leandrodamascena:
```suggestion
    def data_as_json(self) -> Dict:
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_response.py
```diff
@@ -0,0 +1,125 @@
...
+        """Decoded base64-encoded data loaded to json"""
+        if self._json_data is None:
+            self._json_data = self._json_deserializer(self.data_as_text)
+        return self._json_data
+
+
+def KinesisFirehoseResponceRecordFactory(
+    record_id: str,
+    result: Union[FirehoseStateOk, FirehoseStateDropped, FirehoseStateFailed],
```
leandrodamascena:
```suggestion
    result: Literal["Ok", "Dropped", "ProcessingFailed"],
```

leandrodamascena (changes requested):
Hi @roger-zhangg! I'm reviewing the code, your comments, and reading the AWS documentation for Kinesis Firehose Response, but in the meantime, I think you should fix these issues.

Please mark this PR as "ready to review".

leandrodamascena:
> * Add a setter method to `DictWrapper` and init the class as a data class
> 
> ```python
> # possible UX
> result = KinesisFirehoseResponce({})
> for record in event.records:
>         data = record.data_as_json
>         processed_record = KinesisFirehoseResponceRecord(
>             record_id=record.record_id,
>             result=FirehoseStateOk,
>         )
>         processed_record.data_from_json(data)
>         result.add_record(processed_record)
> return result

I like this UX, easy and clean to create records and add these in the response. I'm just wondering if we really need `DictWrapper`. As you said, the `DictWrapper` class is very much meant for parsing events and getting elements, what do you think about just keeping this class as @dataclasses?

leandrodamascena:
Please review the class name, there is a typo: `Responce` instead of `Response`.

Thanks

roger-zhangg:
Thanks for the suggestion, I'll refactor to remove the `DictWrapper` and keep this UX

heitorlessa:
You can make this UX even nicer by having a method in the Kinesis Record
that returns a standalone KinesisFirehoseResponceRecord.

That way you don’t need to ask for the record ID since you’re building it
from the record already, and can have more fluid options like JSON to get
the payload in the expected shape etc.

Play around with it later

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,8 +1,126 @@
...
+import sys
+from dataclasses import dataclass
+from typing import Any, Callable, Dict, Iterator, List, Optional
 
 from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
 
+if sys.version_info >= (3, 8):
+    from typing import Literal
+else:
+    from typing_extensions import Literal
```
roger-zhangg: just typing_extensions

leandrodamascena:
```suggestion
from typing_extensions import Literal
```

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -15,7 +15,7 @@ class DictWrapper(Mapping):
 
     def __init__(self, data: Dict[str, Any], json_deserializer: Optional[Callable] = None):
         """
-        Parameters
+        ParametersW
```
leandrodamascena:
```suggestion
        Parameters
```

examples/event_sources/src/kinesis_firehose_delivery_stream.py
```diff
@@ -1,28 +1,25 @@
...
         # if data was delivered as json; caches loaded value
         data = record.data_as_json
 
-        processed_record = {
-            "recordId": record.record_id,
-            "data": base64.b64encode(json.dumps(data).encode("utf-8")),
-            "result": "Ok",
-        }
+        ## do all kind of stuff with data
+
```
leandrodamascena: Add data transformation here. Don't need something too complex, but just to show an example.

examples/event_sources/src/kinesis_firehose_delivery_stream.py
```diff
@@ -1,28 +1,25 @@
...
+        processed_record = record.create_firehose_response_record(result="Ok")
+        processed_record.data_from_json(data=data)
 
-        result.append(processed_record)
+        result.add_record(processed_record)
 
     # return transformed records
-    return {"records": result}
+    return result.asdict
```
leandrodamascena:
```suggestion
    return result.asdict()
```

leandrodamascena (changes requested):
I made some few comments.

examples/event_sources/src/kinesis_firehose_delivery_stream.py
```diff
@@ -1,28 +1,25 @@
...
 
-        processed_record = {
-            "recordId": record.record_id,
-            "data": base64.b64encode(json.dumps(data).encode("utf-8")),
-            "result": "Ok",
-        }
+        ## do all kind of stuff with data
+
+        processed_record = record.create_firehose_response_record(result="Ok")
+        processed_record.data_from_json(data=data)
```
roger-zhangg: check data size
error

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,123 @@
...
+
+    def asdict(self) -> Optional[Dict]:
+        if self.partition_keys is not None:
+            return {"partitionKeys": self.partition_keys}
+        return None
+
+
+@dataclass
+class KinesisFirehoseResponseRecord:
+    """Record in Kinesis Data Firehose event
```
troyswanson: May want to change this to "Record in Kinesis Data Firehose response object"

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,123 @@
...
+    """
+
+    """Record ID; uniquely identifies this record within the current batch"""
+    record_id: str
+    """processing result, supported value: Ok, Dropped, ProcessingFailed"""
+    result: Literal["Ok", "Dropped", "ProcessingFailed"]
+    """data blob, base64-encoded, optional at init. Allows pass in base64-encoded data directly or
+            use either function like `data_from_text`, `data_from_json` to populate data"""
+    data: Optional[str] = None
+    """Optional: metadata associated with this record; present only when Kinesis Stream is source"""
```
troyswanson: This doc is incorrect, should refer to the response metadata object which can contain partition keys.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,123 @@
...
+        if not self.data_as_text:
+            return None
+        if self._json_data is None:
+            self._json_data = self.json_deserializer(self.data_as_text)
+        return self._json_data
+
+
+@dataclass
+class KinesisFirehoseResponse:
+    """Kinesis Data Firehose event
```
troyswanson: May want to change this to "Kinesis Data Firehose response object" so it's not confused with the event source object.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -77,6 +191,23 @@ def data_as_json(self) -> dict:
             self._json_data = self._json_deserializer(self.data_as_text)
         return self._json_data
 
+    def create_firehose_response_record(
+        self,
+        result: Literal["Ok", "Dropped", "ProcessingFailed"],
+        data: Optional[str] = None,
+    ) -> KinesisFirehoseResponseRecord:
```
troyswanson: This is a cool method, but it needs the ability to include response metadata (eg: partition keys) as well.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,123 @@
...
+@dataclass
+class KinesisFirehoseResponse:
+    """Kinesis Data Firehose event
+
+    Documentation:
+    --------------
+    - https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
+    """
+
+    records: Optional[List[KinesisFirehoseResponseRecord]] = None
```
troyswanson: I'm not sure it makes sense for this to be `Optional`. All of the records that were given should be annotated, transformed (if applicable), and returned, even if the script failed to process them or if they were intentionally dropped.

(Although I will admit, I've never tested what Firehose will do if your response completely omits some records... Either way, I don't think that should pattern be encouraged.)

leandrodamascena: Hey @troyswanson! I agree that this field should not be `Optional`. I tested the Kinesis Firehose with Lambda returning an empty response (without records) and I got this error:

```json
{
    "deliveryStreamARN": "arn:aws:firehose:us-east-1:xxxxxxx:deliverystream/PUT-S3-7AI2w",
    "destination": "arn:aws:s3:::xxxxxxxx",
    "deliveryStreamVersionId": 1,
    "message": "One or more record Ids were not returned. Ensure that the Lambda function returns all received record Ids.",
    "errorCode": "Lambda.MissingRecordId",
    "processor": "arn:aws:lambda:us-east-1:xxxxx:function:datatransform:$LATEST"
}
```

We should not encourage this pattern.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
leandrodamascena: >May want to change this to "Record in Kinesis Data Firehose response object"

+1

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
leandrodamascena: > May want to change this to "Kinesis Data Firehose response object" so it's not confused with the event source object.

+1

leandrodamascena (changes requested):

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
roger-zhangg: The Optional here enables users to init the `KinesisFirehoseResponse` without record, and use `add_record` function to add records to it. So that user doesn't need to have the full record before using `KinesisFirehoseResponse` class. If we worry about returning empty response, we can add a check in `asdict` to raise if no record is found. Does this make sense to you?

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,128 @@
...
+
+    partition_keys: Optional[Dict[str, str]]
+
+    def asdict(self) -> Optional[Dict]:
+        if self.partition_keys is not None:
+            return {"partitionKeys": self.partition_keys}
+        return None
+
+
+@dataclass
```
heitorlessa: can't remember by heart now the optimizations, but the default @dataclass will even include things like comparison and other code generation we won't need.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,128 @@
...
+
+    Documentation:
+    --------------
+    - https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
+    """
+
+    # Record ID; uniquely identifies this record within the current batch"""
+    record_id: str
+    # Processing result, supported value: Ok, Dropped, ProcessingFailed"""
+    result: Literal["Ok", "Dropped", "ProcessingFailed"]
```
heitorlessa: Shouldn't we have `Ok` as the default? I'm assuming 70%+ of the use cases are to build a successful response, unless otherwise.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,128 @@
...
+    - https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
+
+    Parameters
+    ----------
+    records : List[KinesisFirehoseResponseRecord]
+        records of Kinesis Data Firehose response object,
+        optional parameter at start. can be added later using `add_record` function.
+    """
+
+    records: Optional[List[KinesisFirehoseResponseRecord]] = None
```
heitorlessa:
```suggestion
    records: List[KinesisFirehoseResponseRecord] = field(default_factory=list)
```

You can create one by default, so you don't need yet another check and doesn't sacrifice readability ;)

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -77,6 +196,26 @@ def data_as_json(self) -> dict:
             self._json_data = self._json_deserializer(self.data_as_text)
         return self._json_data
 
+    def create_firehose_response_record(
```
heitorlessa:
```suggestion
    def build_data_transformation_response(
```

tip: think **Intent** first, then align what's the documentation is suggesting and how a customer eventually finds this information.

If we change the name now, you can easily find Kinesis Data Transformation docs as well as a few articles on how other customers handle data transformation in Firehose :)

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -77,6 +196,26 @@ def data_as_json(self) -> dict:
...
+        self,
+        result: Literal["Ok", "Dropped", "ProcessingFailed"],
```
heitorlessa:
```suggestion
        result: Literal["Ok", "Dropped", "ProcessingFailed"] = "Ok",
```

You can confirm with the customer, but I suspect this is 70+% of the use cases. If they want to drop, it'll be clearer when reading a code that something it's not the standard behaviour.

Similarly, ProcessingFailed will likely appear during a try/catch only. We could also optimize here with another method but without customer feedback is an overkill atm ;)

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,128 @@
...
+    partition_keys: Optional[Dict[str, str]]
+
+    def asdict(self) -> Optional[Dict]:
+        if self.partition_keys is not None:
+            return {"partitionKeys": self.partition_keys}
+        return None
+
+
+@dataclass
+class KinesisFirehoseResponseRecord:
```
heitorlessa:
```suggestion
class KinesisFirehoseDataTransformationRecord:
```

same as before, think of the intent, and how "searchable" it is if anyone needs to go to Google or LLM etc.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,128 @@
...
+    --------------
+    - https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
+    """
+
+    partition_keys: Optional[Dict[str, str]]
+
+    def asdict(self) -> Optional[Dict]:
+        if self.partition_keys is not None:
+            return {"partitionKeys": self.partition_keys}
+        return None
```
heitorlessa:
```suggestion
    def asdict(self) -> Dict:
        if self.partition_keys is not None:
            return {"partitionKeys": self.partition_keys}
        return {}
```

Since it's a new data class altogether, try having a good default. This improves the overall experience of the caller handling it.

For example:

**From**

```python
payload = blah.as_dict() or {}
```

**To**

```python
payload = blah.as_dict() # one less IF too!
```

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,128 @@
...
+        return self.data_as_bytes.decode("utf-8")
+
+    @property
+    def data_as_json(self) -> Optional[Dict]:
+        """Decoded base64-encoded data loaded to json"""
+        if not self.data_as_text:
+            return None
+        if self._json_data is None:
+            self._json_data = self.json_deserializer(self.data_as_text)
+        return self._json_data
```
heitorlessa: Try returning the default empty representation of the data type to improve data handling for the caller - it reduces the number of conditional checks for None, Empty data, and hacks due to Mypy thinking it's a None unless you check them.

We've made that mistake in the Event Source Data Class in multiple places which we'll correct in v3 :) 

```python
default_str = ""
default_bytes = b""
default_dict = {}
```

etc.

examples/event_sources/src/kinesis_firehose_delivery_stream.py
```diff
@@ -1,28 +1,27 @@
...
-            "recordId": record.record_id,
-            "data": base64.b64encode(json.dumps(data).encode("utf-8")),
-            "result": "Ok",
-        }
+        ## do all kind of stuff with data
+        ## generate data to return
+        new_data = {"tool_used": "powertools_dataclass", "original_payload": data}
+
+        processed_record = record.create_firehose_response_record(result="Ok")
+        processed_record.data_from_json(data=new_data)
```
heitorlessa: is this correct? did you mean to pass `new_data` as part of L21 when creating a data transformation response?

This is a normal Python statement which will be return a dictionary from a processed_record and doesn't go anywhere.

examples/event_sources/src/kinesis_firehose_response.py
```diff
@@ -0,0 +1,24 @@
...
+
+    for record in event["records"]:
+        print(record["recordId"])
+        payload = base64.b64decode(record["data"]).decode("utf-8")
+        ## do all kind of stuff with payload
+        ## generate data to return
+        new_data = {"tool_used": "powertools_dataclass", "original_payload": payload}
+
+        processed_record = KinesisFirehoseResponseRecord(record_id=record["recordId"], result="Ok")
+        processed_record.data_from_text(data=new_data)
```
heitorlessa: same here

heitorlessa (changes requested):
Some notes to improve UX, Mypy, and some issues that could mislead people reading code afterwards (Python statement vs mutation)

examples/event_sources/src/kinesis_firehose_delivery_stream.py
roger-zhangg: Hey Heitor, I think the confusion comes from `data_from_json` and `data_as_json`. The later one does what you described. But the `data_from_json` is assigning data to self.data

examples/event_sources/src/kinesis_firehose_response.py
roger-zhangg: `data_from_text` and `data_as_text`

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
roger-zhangg: This is a really good optimization. Thanks!

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
roger-zhangg: I think I just adopt the name from the go lib. In this case, should I also change `KinesisFirehoseResponse` -> `KinesisFirehoseDataTransformationResponse`

heitorlessa:
Yep, change that too for consistency.

roger-zhangg:
Hey @troyswanson there's lot of optimization went through in the last few commits, we would really appreciate if you could give a final review on this commit before we proceed to merge it. Thanks!

docs/utilities/data_classes.md
```diff
@@ -981,12 +981,23 @@ inline, and re-emit them back to the Delivery Stream.
 Similar to Kinesis Data Streams, the events contain base64 encoded data. You can use the helper
 function to access the data either as json or plain text, depending on the original payload.
 
+When constructing response to Firehose, You can utilize the `KinesisFirehoseResponse` class shown
+in the example below.
```
troyswanson: I think this should be `KinesisFirehoseDataTransformationResponse` now

troyswanson:
> Hey @troyswanson there's lot of optimization went through in the last few commits, we would really appreciate if you could give a final review on this commit before we proceed to merge it. Thanks!

I'm really liking the way this looks! Thank you for your effort ❤️

examples/event_sources/src/kinesis_firehose_delivery_stream.py
leandrodamascena: Hello @heitorlessa! I follow @roger-zhangg's solution. The customer still has the option of passing the `data` parameter on L21, but the response must be base64 encoded, and using a method (L22) avoids some boilerplate code needed by the customer.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
+
+    def asdict(self) -> Dict:
+        r: Dict[str, Any] = {
+            "recordId": self.record_id,
+            "result": self.result,
+            "data": self.data,
+        }
+        if self.metadata:
+            r["metadata"] = self.metadata.asdict()
+        return r
```
leandrodamascena:
```suggestion
        record: Dict[str, Any] = {
            "recordId": self.record_id,
            "result": self.result,
            "data": self.data,
        }
        if self.metadata:
            record["metadata"] = self.metadata.asdict()
        return record
```

It is good to avoid variable names with a single letter.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
+    records: List[KinesisFirehoseDataTransformationRecord] = field(default_factory=list)
+
+    def add_record(self, record: KinesisFirehoseDataTransformationRecord):
+        self.records.append(record)
+
+    def asdict(self) -> Dict:
+        if not self.records:
+            raise ValueError("Kinesis Firehose doesn't accept empty response")
+
+        return {"records": [r.asdict() for r in self.records]}
```
leandrodamascena:
```suggestion
        return {"records": [record.asdict() for record in self.records]}
```

leandrodamascena (changes requested):
Hei @roger-zhangg! I just made 2 comments and I'm good to approve this PR!

Any considerations @heitorlessa?

leandrodamascena (dismissed):
Approved.

heitorlessa:
reviewing today, please update PR description as UX seems wrong?

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
+
+from typing_extensions import Literal
 
 from aws_lambda_powertools.utilities.data_classes.common import DictWrapper
 
 
+@dataclass(repr=False, order=False, frozen=True)
+class KinesisFirehoseDataTransformationRecordMetadata:
+    """
+    Documentation:
```
heitorlessa, line 22: Explain what it is, then share the docs for more details. This avoids death by a thousand browser tabs

leandrodamascena:
I think we need to come to a common sense on the best way to allow the customer to add the `data` field in the response. Below I explain what we did and the expected output.

#### Definition of the data field

According to [official documentation](https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html), the `data` field must be base64 encoded.


#### Manually encoding the data field

The customer can add the `data` parameter when invoking the `build_data_transformation_response` method, but this data must be base64 encoded and the client must encode it manually.

```python
import base64
import json

from aws_lambda_powertools.utilities.data_classes import (
    KinesisFirehoseDataTransformationResponse,
    KinesisFirehoseEvent,
    event_source,
)

@event_source(data_class=KinesisFirehoseEvent)
def lambda_handler(event: KinesisFirehoseEvent, context: LambdaContext):
    
    result = KinesisFirehoseDataTransformationResponse()

    for record in event.records:
        # get original data using data_as_text property
        data = record.data_as_text

        ## do all kind of stuff with data
        ## generate data to return
        transformed_data = {"new_data": "transformed data using Powertools", "original_payload": data}
        transformed_data_encoded = base64.b64encode(json.dumps(transformed_data).encode())
        

        processed_record = record.build_data_transformation_response(result="Ok", data=transformed_data_encoded)

        result.add_record(processed_record)

    # return transformed records
    return result.asdict()
```

#### Automatically encoding the data field

We can shorten the path for the customer and provide a method to automatically base64-encode this data by using `processed_record.data_from_json()`. The `processed_record` is an instance of `KinesisFirehoseDataTransformationRecord` and implements methods that can mutate/change the `self.data` value.

I agree that `processed_record.data_from_json()` can be confusing to understand because we are not setting the value to a variable or even don't know what was changed behind the scenes, but I don't see a better alternative for this.

```python
from aws_lambda_powertools.utilities.data_classes import (
    KinesisFirehoseDataTransformationResponse,
    KinesisFirehoseEvent,
    event_source,
)

@event_source(data_class=KinesisFirehoseEvent)
def lambda_handler(event: KinesisFirehoseEvent, context: LambdaContext):
    
    result = KinesisFirehoseDataTransformationResponse()

    for record in event.records:
        # get original data using data_as_text property
        data = record.data_as_text

        ## do all kind of stuff with data
        ## generate data to return
        transformed_data = {"new_data": "transformed data using Powertools", "original_payload": data}

        processed_record = record.build_data_transformation_response(result="Ok")
        processed_record.data_from_json(transformed_data)

        result.add_record(processed_record)

    # return transformed records
    return result.asdict()
```

We just need to improve this before merging this PR.

Any ideas?

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
+    --------------
+    - https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
+    """
+
+    partition_keys: Optional[Dict[str, str]]
+
+    def asdict(self) -> Dict:
+        if self.partition_keys is not None:
+            return {"partitionKeys": self.partition_keys}
+        return {}
```
heitorlessa:
```suggestion
    partition_keys: Dict[str, str] = field(default_factory=lambda: {})
```

Tip: Any Python class by default has a `__dict__` dunder method OR `vars(class_instance)` which dumps the attributes as a dictionary. 

If you were to create a factory with a different structure it'd be fine to have `asdict` like you have for the Response class.

Here, that logic can be fully replaced with a `__dict__`, `vars()`, or even `dataclasses.asdict`. The latter is the slowest but perfect when you have nested dataclasses.

tip 2: have good defaults to avoid checking on them later (mypy, if, etc.) unless you're going to serialize them in a different way OR need them as `None` to assert intentional falsy behaviour. Here, we can use `default_factory()` to instantiate an empty dict during dataclass initialization IF a customer doesn't pass one.

heitorlessa (dismissed):
a few more logic deduplication that can be removed

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
+    result: Literal["Ok", "Dropped", "ProcessingFailed"] = "Ok"
+    # data blob, base64-encoded, optional at init. Allows pass in base64-encoded data directly or
+    # use either function like `data_from_text`, `data_from_json` to populate data"""
+    data: Optional[str] = None
+    # Optional: Metadata associated with this record; can contain partition keys
+    # See - https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
+    metadata: Optional[KinesisFirehoseDataTransformationRecordMetadata] = None
+    _json_data: Optional[Any] = None
+    json_serializer: Callable = json.dumps
+    json_deserializer: Callable = json.loads
```
heitorlessa:
```suggestion
@dataclass(repr=False, order=False)
class KinesisFirehoseDataTransformationRecord:
    """Record in Kinesis Data Firehose response object.

    Parameters
    ----------
    record_id: str
        uniquely identifies this record within the current batch
    result: Literal["Ok", "Dropped", "ProcessingFailed"]
        record data transformation status, whether it succeeded, should be dropped, or failed.
    data: str
        base64-encoded payload, by default empty string.

        Use `data_from_text` or `data_from_json` methods to convert data if needed.

    metadata: Optional[KinesisFirehoseDataTransformationRecordMetadata]
        Metadata associated with this record; can contain partition keys.

        See: https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
    json_serializer: Callable
        function to serialize `obj` to a JSON formatted `str`, by default json.dumps
    json_deserializer: Callable
        function to deserialize `str`, `bytes`, bytearray` containing a JSON document to a Python `obj`,
        by default json.loads

    Documentation:
    --------------
    - https://docs.aws.amazon.com/firehose/latest/dev/data-transformation.html
    """

    record_id: str
    result: Literal["Ok", "Dropped", "ProcessingFailed"] = "Ok"
    data: str = ""
    metadata: Optional[KinesisFirehoseDataTransformationRecordMetadata] = None
    json_serializer: Callable = json.dumps
    json_deserializer: Callable = json.loads
    _json_data: Optional[Any] = None
```

Docstring to provide actionable information for caller.

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
+        """Populate data field using a byte like data"""
+        self.data = base64.b64encode(data).decode("utf-8")
+
+    def data_from_text(self, data: str):
+        """Populate data field using a string like data"""
+        self.data_from_byte(data.encode("utf-8"))
+
+    def data_from_json(self, data: Any):
+        """Populate data field using any structure that could be converted to json"""
+        self.data_from_text(data=self.json_serializer(data))
```
heitorlessa:
```suggestion
```

In-place mutation can almost always go wrong. It looks like what you're trying to improve in UX is make it easier to transform data that will be added in the final processed_record.

Since you accept a `data=` argument, it's easier to have a bunch of standalone functions that handle X to base64.

---

Example

```python
        processed_record = KinesisFirehoseDataTransformationRecord(
            record_id=record.record_id, result="Ok", metadata=metadata_partition, data=base64_from_json(data)
        )
        response.add_record(record=processed_record)
```

tests/unit/data_classes/test_kinesis_firehose_response.py
```diff
@@ -0,0 +1,69 @@
+from aws_lambda_powertools.utilities.data_classes import (
+    KinesisFirehoseDataTransformationRecord,
+    KinesisFirehoseDataTransformationRecordMetadata,
+    KinesisFirehoseDataTransformationResponse,
+    KinesisFirehoseEvent,
+)
+from tests.functional.utils import load_event
+
+
+def test_kinesis_firehose_response():
```
heitorlessa, line 25: tip: use GIVEN/WHEN/THEN to help reviewers pick up the intent of the test.

For example, reading this code, I noticed we're deserializing b64 to str, then back again. If the intent is to test the Response record serializing correctly with the input data, we can pass `data=record.data` directly :) 

Then have a separate test where we transform the data to confirm it differs from the original record

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
```diff
@@ -1,9 +1,125 @@
...
 
+@dataclass(repr=False, order=False, frozen=True)
+class KinesisFirehoseDataTransformationRecordMetadata:
+    """
+    Documentation:
+    --------------
+    - https://docs.aws.amazon.com/firehose/latest/dev/dynamic-partitioning.html
+    """
+
+    partition_keys: Optional[Dict[str, str]]
```
heitorlessa:
```suggestion
    partition_keys: Optional[Dict[str, Any]]
```

to match your example with "year": 2023

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
roger-zhangg: For this, we still need `partitionKeys` in dict instead of `partition_keys` that would be produced by `__dict__`

aws_lambda_powertools/utilities/data_classes/kinesis_firehose_event.py
roger-zhangg: Sorry this is a mistake by me, We shouldn't use int in the Map. Although it's not clearly documented in the Firehose's doc, but every example they provide is in `Dict[str, str]` format. So reverting this back to str:str and changed 2023->"2023" in test case

roger-zhangg:
Hey @leandrodamascena @heitorlessa please help to review the test. I also added `asdict` back to metadata class for `__dict__` will remain the key as `partition_keys` instead of supposed `partitionKeys` value

heitorlessa:
coming at this later tonight... had to switch to https://github.com/aws-powertools/powertools-lambda-python/issues/3091 as it seemed a potential customer issue

heitorlessa:
Three quick things only and we can merge <3 

* [x] It’s missing “Examples” in docstring (for the ones you believe a customer will be mostly exposed to, 1 example suffice)
* [x] During result validation, you’re creating a list with valid strings for every iteration - you can use a class variable (class definition) and use a tuple (fixed array size for memory footprint)
* [ ] `result="Ok"` is the default now, we can remove it to keep it clean. Then in the docs, we can have a tabbed example with the non-ok examples so they know when to use `Dropped` and `ProcessingFailed`.

heitorlessa:
Item 1 and 2 are done.

---

Docstring experience

image

API reference experience

image

heitorlessa:
As soon as the docs contain both the default (`ok`), `Dropped`, and `ProcessingFailed`, along with key parts highlighted (what to import, where standalone function is used, result.add_record, asdict), and any comments or code annotation... it's good to merge ;)

cc @leandrodamascena 

Gonna to go to bed (10:33pm)

leandrodamascena (approved):

heitorlessa:
Docs final touches sent

* Use code annotations to provide links, rich formatting, for comments where you can provide more contextual information. Whereas comments would be limited to a single line (cognitive load)
* Always link to the official docs for authorative source
* Provide information upfront to ease customers learning curve in complex topics - in this case, it'd take a while for customers to understand the nuances between the 3 modes (OK, DROP, FAILED)
* Use realistic examples - always put yourself in the customers shoes, specially when they might not be developers. In the exception example, we want to make it true that you could drop an unwanted record :) 
* The failed example now uses a code annotation to explain how Kinesis Firehose handles failure. Adding comments would be a distraction, and even if we were to explain it'd be hard to keep it up to date if it changes. Instead, we go straight to the point and add a link for more details.
* Don't be afraid of using newlines and removing multi-line comments. Whenever there's more than one line of comment, maybe there's a better way of phrasing it, or use code annotation.

heitorlessa:
Now I can go to sleep hahaha. Thank you SO MUCH both of you for the great work here ;)

troyswanson:
Thank you all! This is a very exciting feature for Powertools

PR #3113, opened 2023-09-20
fix(logger): add explicit None return type annotations

### LINKED ISSUE #3112: Static typing: `logger.append_keys` produces a `[no-untyped-call]` mypy error
### Static type checker used

mypy (project's standard)

3.10

latest

## Output

```plaintext
error: Call to untyped function "append_keys" in typed context  [no-untyped-call]
        logger.append_keys(something="yes please")
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

## Config

```toml
# pyproject.toml
[tool.mypy]
files = [ "**/*.py" ]
ignore_missing_imports = true
strict = true
pretty = true
show_error_codes = true
warn_unreachable = true
enable_error_code = [ "ignore-without-code", "redundant-expr", "truthy-bool" ]
plugins = [ "pydantic.mypy" ]
```

## Run

I just run `mypy` on the command line from within the virtual environment

### Code snippet

```python

from typing import TYPE_CHECKING, Any

if TYPE_CHECKING:


logger = Logger(utc=True)


def lambda_handler(event: dict[str, Any], context: LambdaContext) -> dict[str, Any]:
    logger.append_keys(something="yes please") # Error here
    return {"hello": "world"}
```

### Possible Solution

It seems that in a few places in powertools where a function or method would return `None`, the return type annotation is omitted. One such example is `append_keys` https://github.com/aws-powertools/powertools-lambda-python/blob/812b63decac1da1311e544c92d423fd6fee1ac03/aws_lambda_powertools/logging/logger.py#L596

The solution would be to simply add `-> None:` as the return type annotation 🙂, preferably in other places where it's missing too.

Happy to do a PR for this, just waiting for triage 👍🏻

FollowTheProcess:
**Issue number:** #3112 

## Summary

Improves the type annotations in the `logging` module, specifically around those functions/methods that return None (with some bonus low hanging fruit elsewhere); solving the issue raised in #3112.

However, this has revealed a few new typing issues that weren't present before; some of which were simple fixes so I have just made them. Others I have questions on how we want to handle which I will post as comments below.

### Changes

Added explicit `-> None` return type annotations where appropriate in the `logging` module

### User experience

Should mean that users using mypy in strict mode no longer have to suppress type errors on certain logging functions/methods

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


FollowTheProcess:
This revealed a few issues with mypy that didn't occur when I first checked the repo out:

image

It looks as if the `registered_handler` property accesses `self._logger.parent.handlers` even though technically `self._logger.parent` returns `Logger | None`

image

Now I'm guessing but the `self.child` attribute seems to encode the info that there _is_ indeed a parent so this looks more like mypy can't understand that from context? Just wondering on the best way you'd like to address this? I did try changing `if self.child` condition to `if self._logger.parent is not None` which reads the same but it broke a lot of tests so there must be something else going on there.

The second issue is that mypy now appears not to like the `_get_caller_filename` at all as `inspect.currentframe()` again returns `FrameType | None` so each call to `f_back` could be `None` which we're not handling in that function:

image

Again this seems like a case where we know better than mypy to me and it's not getting the full context as we know by definition (and the comment) that there will always be at least 3 frames.

Are we happy with `# type: ignore` in these cases? Or do we want to explicitly handle the possibility of `None` (however unlikely)

heitorlessa:
> This revealed a few issues with mypy that didn't occur when I first checked the repo out:
> 
> image
> 
> It looks as if the `registered_handler` property accesses `self._logger.parent.handlers` even though technically `self._logger.parent` returns `Logger | None`
> 
> image
> 
> Now I'm guessing but the `self.child` attribute seems to encode the info that there _is_ indeed a parent so this looks more like mypy can't understand that from context? Just wondering on the best way you'd like to address this? I did try changing `if self.child` condition to `if self._logger.parent is not None` which reads the same but it broke a lot of tests so there must be something else going on there.
> 
> The second issue is that mypy now appears not to like the `_get_caller_filename` at all as `inspect.currentframe()` again returns `FrameType | None` so each call to `f_back` could be `None` which we're not handling in that function:
> 
> image
> 
> Again this seems like a case where we know better than mypy to me and it's not getting the full context as we know by definition (and the comment) that there will always be at least 3 frames.
> 
> Are we happy with `# type: ignore` in these cases? Or do we want to explicitly handle the possibility of `None` (however unlikely)

The former - I'd be happy with adding `type: ignore[Specific_Error_to_Ignore]`. I'm genuinely nervous with a blank `type: ignore` as it could miss on other future regressions vs ignoring the None instance only.

The main reason we didn't go with strict mode in Mypy is that it ended up causing more trouble than saving the effort. One item in the back of my mind is to take yet another look at Pyright -- it was better at these things, but I wasn't sure if it was going to be maintained long enough after all those years (it did!!).

aws_lambda_powertools/logging/filters.py
```diff
@@ -2,10 +2,10 @@
 
 
 class SuppressFilter(logging.Filter):
-    def __init__(self, logger):
+    def __init__(self, logger: str):
         self.logger = logger
 
-    def filter(self, record):  # noqa: A003
+    def filter(self, record: logging.LogRecord) -> bool:  # noqa: A003
```
heitorlessa, line 8: Thank you for the attention to detail ;-)

heitorlessa (approved):
LGTM pending CI results -- lemme know otherwise if there's any other change.

Appreciate those extra comments and explicit ignore types 🫶

Making a release tomorrow

heitorlessa:
Released in 2.25.1 in PyPi. Building and optimizing Lambda Layers now -- should be all done in ~20m or so. Next Layer version being 45.

PR #3133, opened 2023-09-25
docs(contributing): initial structure for revamped contributing guide

heitorlessa:
**Issue number:** #1916
**Additional issue:** #2181

## Summary

This begins the initial structure for the newly revamped contributing guide. It keeps `CONTRIBUTING.md` intact until we have it in prod otherwise it can lead to 404 or stale content.

### Changes

> Please provide a summary of what's being changed

* [x] Add Contributing section in Processes
* [x] Add Development Environment
* [x] Add Your first contribution section
* [x] Create initial structure for final navigation (commented until we have docs)

Unrelated:

* [x] Update Event Handler roadmap
* [x] Update Sensitive Data Masking roadmap
* [x] Update Gov Cloud roadmap 

### User experience

> Please share what the user experience looks like before and after this change

image

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented
* [x] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #3147, opened 2023-09-28
feat(logger): new stack_trace field with rich exception details

### LINKED ISSUE #1268: Feature request: Parse exceptions into structured logs
### Use case

When using the logger and logging an exception, the exception traceback appears as a single string that includes the stack trace, each frame of which has a file, a line number, a function name, and the statement on that line. Parsing that line, either with the Mk I eyeball or with code, is difficult

### Solution/User Experience

Because logger is all about structured logging, it would be nice if the stack backtrace were structured as well. Something like:

```
Traceback = [
  {
    "File" : "myfile.py",
    "Line" : 44,
    "Function" : "do_something_risky",
    "Statement": "do_division(x, 0)"
  },
  {
    "File" : "myotherfile.py",
    "Line": 12,
    "Function" : "do_division",
    "Statement" : "return numerator/denominator
  }
]
```

### Alternative solutions

```markdown
There are no good ones that I'm aware of. I might be able to overload the existing functions in the traceback module, but doing that for every lambda function in my solution is not ideal
```

### Acknowledgment

stevrobu:
**Issue number:** #1268 

### Changes

Added environment variable and formatter constructor parameter to allow users to include additional JSON formatted full stack trace to their logger.exception output. 

To test the change, the following Lambda code will enable the enhanced stacktrace and log the exception along with the stacktrace:

```python
import requests

ENDPOINT = "http://httpbin.org/status/500"
logger = Logger(serialize_stacktrace=True)


    try:
        ret = requests.get(ENDPOINT)
        ret.raise_for_status()
    except requests.HTTPError as e:
        logger.exception(e)
        raise RuntimeError("Unable to fullfil request") from e

    return "hello world"
```

Response in CloudWatch logs should be:

```json
{
   "level":"ERROR",
   "location":"lambda_handler:16",
   "message":"500 Server Error: INTERNAL SERVER ERROR for url: http://httpbin.org/status/500",
   "timestamp":"2023-10-09 17:47:50,191+0000",
   "service":"service_undefined",
   "exception":"Traceback (most recent call last):\n  File \"/var/task/app.py\", line 14, in lambda_handler\n    ret.raise_for_status()\n  File \"/var/task/requests/models.py\", line 1021, in raise_for_status\n    raise HTTPError(http_error_msg, response=self)\nrequests.exceptions.HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: http://httpbin.org/status/500",
   "exception_name":"HTTPError",
   "stack_trace":{
      "type":"HTTPError",
      "value":"500 Server Error: INTERNAL SERVER ERROR for url: http://httpbin.org/status/500",
      "module":"requests.exceptions",
      "frames":[
         {
            "file":"/var/task/app.py",
            "line":14,
            "function":"lambda_handler",
            "statement":"ret.raise_for_status()"
         },
         {
            "file":"/var/task/requests/models.py",
            "line":1021,
            "function":"raise_for_status",
            "statement":"raise HTTPError(http_error_msg, response=self)"
         }
      ]
   }
}

```

### User experience

User can now parse JSON stacktraces instead of string stacktraces. 

## Checklist

* [x] Changes are documented

<details>

Not a breaking change.

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/logging/formatter.py
```diff
@@ -148,16 +153,34 @@ def __init__(
...
         """Serialize structured log dict to JSON str"""
         return self.json_serializer(log)
+    
+    def serialize_traceback(self, e: Exception) -> list:
+        return [{"file": fs.filename, 
+                    "line": fs.lineno,
+                    "column": fs.colno,
+                    "function": fs.name,
+                    "statement": fs.line
+                    } for fs in traceback.extract_tb(e.__traceback__)]
```
leandrodamascena:
```suggestion
    def serialize_traceback(self, log_record: logging.LogRecord) -> list:
        print(log_record.exc_info[0], log_record.exc_info[1],log_record.exc_info[2])
        exception_info = {
            "type": log_record.exc_info[0].__name__,
            "value": str(log_record.exc_info[1]),
            "module": log_record.exc_info[1].__class__.__module__,
            "frames": []
        }

        for fs in traceback.extract_tb(log_record.exc_info[2]):
            frame_info = {
                "file": fs.filename,
                "line": fs.lineno,
                "function": fs.name,
                "statement": fs.line
            }
            exception_info["frames"].append(frame_info)

        return exception_info
```

If we assume that `serialize_traceback` always expects an exception, we break our current behavior where a customer can use `logger.exception("failed etc etc etc")`. We can extract the exception from the current `log_record` and parser it.

I left a [comment](https://github.com/aws-powertools/powertools-lambda-python/issues/1268#issuecomment-1743483745) in the issue talking about the data that we should include in the exception. I still don't have a strong opinion that this model I proposed is the best, so I would like to hear opinions.

aws_lambda_powertools/logging/formatter.py
```diff
@@ -148,16 +153,34 @@ def __init__(
...
+
 
     def format(self, record: logging.LogRecord) -> str:  # noqa: A003
         """Format logging record as structured JSON str"""
         formatted_log = self._extract_log_keys(log_record=record)
         formatted_log["message"] = self._extract_log_message(log_record=record)
+
+        if self.include_stacktrace:
+            # Generate the traceback from the traceback library
+            formatted_log["stack_trace"] = self.serialize_traceback(record.msg) #JSR
```
leandrodamascena:
```suggestion
            formatted_log["stack_trace"] = self.serialize_traceback(log_record=record)
```

tests/functional/test_logger.py
```diff
@@ -959,3 +959,62 @@ def test_stream_defaults_to_stdout(service_name, capsys):
...
+        logger.exception(e)
+
+    log = capture_logging_output(stdout)
+    assert "stack_trace" not in log
+
+def test_logger_logs_stack_trace_with_parameter(stdout):
+    logger = Logger(stream=stdout, logger_formatter=LambdaPowertoolsFormatter(include_stacktrace=True))
+
+    try:
+        val = 1 + "someString"
```
leandrodamascena: We have linter enabled in this project, so, if you initiate a variable and don't use that in the rest of the code then the Linter will complain about:

`tests/functional/test_logger.py:968:9: F841 Local variable `val` is assigned to but never used`

You can do something like this:

```python
try:
        raise ValueError("something went wrong")
    except Exception:
        logger.exception("Received an exception")
```

tests/functional/test_logger.py
```diff
@@ -959,3 +959,62 @@ def test_stream_defaults_to_stdout(service_name, capsys):
...
+
+    log = capture_logging_output(stdout)
+    assert "stack_trace" in log
+
+def test_logger_logs_stack_trace_with_env_var(stdout, monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv(constants.POWERTOOLS_STACKTRACE_ENV, "true")
+    logger = Logger(stream=stdout)
+
+    try:
+        val = 1 + "someString"
```
leandrodamascena: Same here

tests/functional/test_logger.py
```diff
@@ -959,3 +959,62 @@ def test_stream_defaults_to_stdout(service_name, capsys):
...
+    log = capture_logging_output(stdout)
+    assert "stack_trace" in log
+
+
+def test_logger_logs_no_stack_trace_with_env_var(stdout, monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv(constants.POWERTOOLS_STACKTRACE_ENV, "false")
+    logger = Logger(stream=stdout)
+
+    try:
+        val = 1 + "someString"
```
leandrodamascena: Same here

tests/functional/test_logger.py
```diff
@@ -959,3 +959,62 @@ def test_stream_defaults_to_stdout(service_name, capsys):
...
+
+    log = capture_logging_output(stdout)
+    assert "stack_trace" not in log
+
+def test_logger_logs_no_stack_trace_with_parameter_override(stdout, monkeypatch: pytest.MonkeyPatch):
+    monkeypatch.setenv(constants.POWERTOOLS_STACKTRACE_ENV, "true")
+    logger = Logger(stream=stdout, logger_formatter=LambdaPowertoolsFormatter(include_stacktrace=False))
+
+    try:
+        val = 1 + "someString"
```
leandrodamascena: Same here

tests/functional/test_logger.py
```diff
@@ -959,3 +959,62 @@ def test_stream_defaults_to_stdout(service_name, capsys):
     # NOTE: we can't assert on capsys.readouterr().err due to a known bug: https://github.com/pytest-dev/pytest/issues/5997
     log = json.loads(capsys.readouterr().out.strip())
     assert log["message"] == msg
+
+def test_logger_logs_no_stack_trace_without_parameter(stdout):
+    logger = Logger(stream=stdout)
+
+    try:
+        val = 1 + "someString"
```
leandrodamascena: Same here

leandrodamascena (changes requested):
Hi @stevrobu! Thanks for submitting this PR, this is an amazing first contribution! We need to change a few things before merging this.

One thing I'm missing is the `include_stacktrace` parameter in the Logger instance. I see you added it in `PowertoolsFormatter`, but I think we can improve the user experience by adding the Logger class in the same way. You can keep this parameter in `PowertoolsFormatter` because someone can use the Formatter class in some special cases, but adding this in the Logger class improves the experience.

```python

logger = Logger(include_stacktrace=True)
```

aws_lambda_powertools/logging/formatter.py
```diff
@@ -77,6 +81,7 @@ def __init__(
         log_record_order: List[str] | None = None,
         utc: bool = False,
         use_rfc3339: bool = False,
+        include_stacktrace: bool = None,
```
dreamorosi: I'm thinking that we could improve the naming here (naming is hard!).

If I understand correctly the feature, we already include the stack trace in the log entry, but now we are serializing it.

If this is the case, should we maybe make it more explicit by renaming the option to `serialize_stacktrace`? The current name makes me think that if this is `False` I won't have any stack trace in my logs, which is not the case.

Also, should the name be plural `serialize_stacktraces` or singular `serialize_stacktrace`?

stevrobu:
@leandrodamascena : I will add the parameter to the logger as you suggested. It actually already works that way because any of the kwargs from the Logger class initialization are passed to the formatter in the structure_logs method. I actually didn't realize this until I wrote a test prior to making the update and the test passed. 

I've updated the tests so that the linter does not complain and I'm checking out the other suggestions.

tests/functional/test_logger.py
stevrobu: Done

tests/functional/test_logger.py
stevrobu: Done

tests/functional/test_logger.py
stevrobu: Done

tests/functional/test_logger.py
stevrobu: Done

tests/functional/test_logger.py
stevrobu: Done

leandrodamascena:
> @leandrodamascena : I will add the parameter to the logger as you suggested. It actually already works that way because any of the kwargs from the Logger class initialization are passed to the formatter in the structure_logs method. I actually didn't realize this until I wrote a test prior to making the update and the test passed.
> 
> I've updated the tests so that the linter does not complain and I'm checking out the other suggestions.

Hi @stevrobu! I think you need to push your local commits, right? I can't see the new code.

stevrobu:
I have not finished yet. I should be able to push tomorrow.

On Oct 3, 2023 5:58 PM, Leandro Damascena ***@***.***> wrote:

@leandrodamascena<https://github.com/leandrodamascena> : I will add the parameter to the logger as you suggested. It actually already works that way because any of the kwargs from the Logger class initialization are passed to the formatter in the structure_logs method. I actually didn't realize this until I wrote a test prior to making the update and the test passed.

I've updated the tests so that the linter does not complain and I'm checking out the other suggestions.

Hi @stevrobu<https://github.com/stevrobu>! I think you need to push your local commits, right? I can't see the new code.

leandrodamascena:
> I have not finished yet. I should be able to push tomorrow. 

Aaa, sorry for my mistake. I saw you resolve conversations and assumed you made the commits. No rush, take your time.

aws_lambda_powertools/logging/formatter.py
stevrobu: I made the suggested updates.

aws_lambda_powertools/logging/formatter.py
stevrobu: I updated the code per your suggestion.

aws_lambda_powertools/logging/formatter.py
stevrobu: This was a good suggestion. I updated to serialize_stacktrace. I believe singular is correct for this scenario.

stevrobu:
@leandrodamascena - I have checked in the latest updates. Thanks for the feedback!

leandrodamascena (dismissed):
Hey @heitorlessa can you pls review this PR?

For me, it is approved! Amazing work @stevrobu! :rocket:

rubenfonseca:
Looking at this now

docs/core/logger.md
```diff
@@ -319,6 +319,22 @@ Logger can optionally log uncaught exceptions by setting `log_uncaught_exception
     --8<-- "examples/logger/src/logging_uncaught_exceptions_output.json"
     ```
 
+#### Stack trace logging
+
+By default, the Logger will automatically log the full stack trace in JSON format when using `logger.exception`. If you want to disable this feature, set `serialize_stacktrace=False` during initialization."
```
rubenfonseca:
```suggestion
By default, the Logger will automatically includes the full stack trace in JSON format when using `logger.exception`. If you want to disable this feature, set `serialize_stacktrace=False` during initialization."
```

Question: If the new behaviour is True, will this not be a breaking change, breaking the expected format that people are used to?

docs/contributing/documentation/rfcs.md
```diff
@@ -3,6 +3,8 @@ title: Writing Request For Comment (RFC)
 description: Contributing RFCs to Powertools for AWS Lambda (Python)
 ---
 
+<!-- markdownlint-disable MD043 -->
+
```
rubenfonseca, line 7: Is this related to the current issue? Can we remove this change before merging?

rubenfonseca (changes requested):
Left some comments and questions, but overall looks good!

docs/contributing/documentation/rfcs.md
leandrodamascena: Hello @rubenfonseca! This file was added in PR #3133. But it is breaking the local CI because there are no mandatory sections like "Getting started" and others, so I fixed it and we will update this file in another PR.

docs/core/logger.md
leandrodamascena: We are adding a new field called `stack_trace` and preserving the current fields `exception` and `exception_name` using the same format. I don't see a scenario where it could break a customer, perhaps a client making data validation when exporting logs? I see great added value in logging the stack trace as JSON and allowing the customer to parser it OR easily identify errors.

In my opinion, we can apply the principle of “better to ask for forgiveness than permission” here.

WDTY?

rubenfonseca: You are absolutely right, I thought for a moment that we were changing an existing field.

rubenfonseca (dismissed):

docs/core/logger.md
heitorlessa: @leandrodamascena **include**, not **includes**

heitorlessa (approved):

leandrodamascena (dismissed):

PR #3186, opened 2023-10-10
docs(data-masking): add docs for data masking utility

### LINKED ISSUE #3488: Docs: New Data Masking utitlity

How to use the new Data Masking utility

_No response_

### How can we improve?

Creating the documentation

### Got a suggestion in mind?

_No response_

### Acknowledgment

### LINKED ISSUE #3532: Feature request: Expand DataMasking utility to encrypt list, set, and tuple data type for better user excperience
### Use case

Currently, the DataMasking utility only supports dict and string fields. We need to add support for encrypt data in list, tuple and set fields.		

### Solution/User Experience

```python
    provider = AwsEncryptionSdkProvider(["aaaaa"])
    datamasking = DataMasking(provider=provider)

    data = {
        "payload": {"data": ["a", "b"]}    
	}

    encrypt = datamasking.encrypt(data, fields=["payload.data[0]"])
```

### Alternative solutions

_No response_

### Acknowledgment

seshubaws:
**Issue number:** #3488 

## Summary
Add docs for data masking utility

### Changes

Adding documentation for the sensitive data masking utility.

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**: https://github.com/aws-powertools/powertools-lambda-python/issues/1858

Checklist:

</details>

## Acknowledgment


examples/data_masking/src/getting_started_encrypt_data.py
```diff
@@ -0,0 +1,106 @@
...
+            "state": "CA", 
+            "zip": "12345",
+        },
+    }
+
+    encryption_provider = AwsEncryptionSdkProvider(keys=[KMS_KEY_ARN])
+    data_masker = DataMasking(provider=encryption_provider)
+
+    encrypted = data_masker.encrypt(data=data, fields=["email", "address.street"])
+    # encrypted = {
```
seshubaws: What is the recommended way to show output for this? Linting is complaining about "Commented out code" for the current way

examples/data_masking/tests/src/single_mock.py
```diff
@@ -0,0 +1,43 @@
...
+        ciphertext = base64.b64encode(data).decode()
+        return ciphertext
+
+    def decrypt(self, data: bytes, **kwargs) -> Any:
+        ciphertext_decoded = base64.b64decode(data)
+        ciphertext = self.json_deserializer(ciphertext_decoded)
+        return ciphertext
+
+
+def handler(event, context):
```
seshubaws: Not sure if the section with the mock is correct...

examples/data_masking/sam/template.yaml
```diff
@@ -0,0 +1,57 @@
...
+          Resource: "*"
+          Principal:
+            AWS: !Join [ "", [ "arn:aws:iam::", !Ref "AWS::AccountId", ":root" ] ]
+  DataMaskingFunctionExample:
+    Type: AWS::Serverless::Function     # More info about Function Resource: https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html
+    Properties:
+      Handler: data_masking_function_example.lambda_handler
+      CodeUri: ../src
+      Description: Data Masking Function Example
+      MemorySize: 1024                  # TODO: Recommended to use 1024 MB due to
```
seshubaws: Is there we want to make the recommendation of what memory size to use based on our load test results? Should we link the load test numbers or the traces or something?

leandrodamascena:
Hi @seshubaws! Thanks for sending this PR. I will begin this review tomorrow.

examples/data_masking/sam/template.yaml
leandrodamascena: Hi @seshubaws! We normally configure 128Mb of memory to prevent the client from copying and pasting our code and incurring some unexpected costs. If we don't have any problems running with 128 MB, it's ok to set 128.

examples/data_masking/src/getting_started_encrypt_data.py
leandrodamascena: I think we can 1 or examples and its enough. Show this as an extra tab and its ok.

examples/data_masking/tests/src/single_mock.py
leandrodamascena: I think its ok. I need to test.

docs/utilities/data_masking.md
leandrodamascena, line 1: Please add a new section in the `mkdocs.yaml` file.

Like this:
image

leandrodamascena (commented):
Hi @seshubaws! I left some comments and am making some changes to add new content to the documentation.

leandrodamascena (changes requested):
Hi @seshubaws! I left some comments and am making some changes to add new content to the documentation.

examples/data_masking/sam/template.yaml
seshubaws: Ok, I'll change it to 128, but where should we recommend they use 1024MB since that was the optimal size for latency? I believe @heitorlessa wanted to link to our load testing numbers as well?

mypy.ini
```diff
@@ -8,6 +8,7 @@ show_column_numbers = True
 show_error_codes = True
 show_error_context = True
 disable_error_code = annotation-unchecked
+exclude = examples/data_masking/src
```
seshubaws: Added this to get rid of mypy error 
```
poetry run mypy --pretty aws_lambda_powertools examples
examples/data_masking/src/custom_data_masking_provider.py: error: Source file found twice under different module names:
"custom_data_masking_provider" and "examples.data_masking.src.custom_data_masking_provider"
```

examples/data_masking/sam/template.yaml
seshubaws: As per offline conversation with @heitorlessa set to 1024 in the SAM template and added a Note pointing that out and explaining why.

heitorlessa:
Writing down editing tasks we're going through together to explain line editing and tech docs conventions

* [x] Keep SAM/infra under Install section only
* [x] Improve SAM infrastructure example
  * [x] Add code annotation to explain about Function memory size
  * [x] Revisit IAM permission
  * [x] Add code annotation to IAM permission - not many people understand `*` followed by Lambda function least privilege
* [ ] Rename filenames to match section to match documentation convention
* [x] Hide `Create your own provider` section
* [x] Add a banner that `Creating your own provider` is coming in 2024
  * [x] Create issue to track it
* [x] Add banner explaining that Pydantic, Dataclasses and Classes are coming in 2024
  * [x] Create issue for Pydantic models
  * [x] Create issue for Dataclasses
  * [x] Create issue for standard Python classes
* [x] Refactor intro diagram (show provider, find distinction between mask and encrypt/decrypt)
* [x] Line editing for Intro line
* [x] Refactor Key features to emphasize industry & AWS security best practices (envelop encryption etc.)
* [ ] Explain `fields` in more details, e.g., supported syntax, type return
* [x] Sequence diagram to explain Encryption SDK operations provider (envelope encryption)
* [x] Split up Encrypt and Decrypt sections
* [x] Explain encrypted/masked data will lose its type (-> ciphertext, mask value)
    * [ ] Fix return type in `mask`
    * [ ] Overload `encrypt` and `decrypt` with generic ([T,R]) to allow customers having a model with ciphertext
* [ ] Explain that KMS Key ARN is necessary in Decrypt operations
* [x] Improve Encryption SDK Provider options using a table
* [x] Create a `Providers` section to future-proof expansion
* [ ] Revisit `testing your code` section -- test doubles over mock if possible

leandrodamascena:
I'm working on this items

>* [ ] Keep SAM/infra under Install section only
>* [ ] Improve SAM infrastructure example
>  * [ ] Add code annotation to explain about Function memory size
>  * [ ] Revisit IAM permission
>  * [ ] Add code annotation to IAM permission - not many people understand `*` followed by Lambda function least privilege
>* [ ] Rename filenames to match section to match documentation convention
>* [ ] Hide `Create your own provider` section
>* [ ] Add a banner that `Creating your own provider` is coming in 2024
>  * [ ] Create issue to track it
>* [ ] Add banner explaining that Pydantic, Dataclasses and Classes are coming in 2024
>  * [ ] Create issue for Pydantic models
>  * [ ] Create issue for Dataclasses
>  * [ ] Create issue for standard Python classes

heitorlessa:
Intro diagram refactored - here's a before and after. 

The change highlights that we have 3 distinct operations (mask, encrypt, decrypt), and only two of them use an Encryption provider. 

image

heitorlessa:
@seshubaws could you branch out to a dedicated branch instead of `develop` in your fork, please? Can't push due to behind out of sync with `develop` itself (intertwined) - it makes the overall process easier too and less error prone

To help out, I pushed `docs/data_masking` branch with the new diagram in your fork - will push subsequent changes too.

heitorlessa:
@seshu and I discussed and she'd be taking the Provider's section to convert into a table, and follow up with the Encryption SDK team on how to best bring our own provider (build abstraction atop to prevent customers shooting themselves in the foot w/o good practices)

seshubaws:
- [x]  Improve Encryption SDK Provider options using a table
- [x]  Create a Providers section to future-proof expansion
- [x]  Split up Encrypt and Decrypt sections
- [x]  Explain encrypted/masked data will lose its type (-> ciphertext, mask value)

seshubaws:
@heitorlessa For the item "Fix return type in mask", it won't always be a `str` since masking a list/set/tuple will keep the type (ie. [1,2,"string",3] -> [" ***** ", " ****** ", " ***** ", " ***** "]) so it's still a list even though all the items inside are all strings. Currently the return type for the `mask()` method is `Any`, do we want it to be like `Union[str, list, set, tuple]`?

seshubaws:
> follow up with the Encryption SDK team on how to best bring our own provider (build abstraction atop to prevent customers shooting themselves in the foot w/o good practices)

Cut the AWS Encryption SDK team a ticket for this

seshubaws:
- [ ] Explain fields in more details, e.g., supported syntax, type return

Tried to do this [here](https://github.com/aws-powertools/powertools-lambda-python/pull/3186/files#diff-cd781e0608cbb6b7951ca23ae3abbdce5b48bb3461ae7f02f4e48ded64de979eR92-R96), but please check to see if it is what you wanted @heitorlessa

heitorlessa:
re-attaching so GH doesn't hide the comment due to updates.

----

Tasks

* [ ] Rename filenames to match section to match documentation convention
* [ ] Explain encrypted/masked data will lose its type (-> ciphertext, mask value)
* [ ] Explain that KMS Key ARN is necessary in Decrypt operations
* [ ] Explain `fields` in more details, e.g., supported syntax, type return
* [x] Sequence diagram to explain Encryption SDK operations provider (envelope encryption)
* [ ] Revisit `testing your code` section -- test doubles over mock if possible
* [ ] Split up Encrypt and Decrypt sections
    * [ ] Fix return type in `mask`
    * [ ] Overload `encrypt` and `decrypt` with generic ([T,R]) to allow customers having a model with ciphertext
* [ ] Line-editing 
* [x] Keep SAM/infra under Install section only
* [x] Improve SAM infrastructure example
  * [x] Add code annotation to explain about Function memory size
  * [x] Revisit IAM permission
  * [x] Add code annotation to IAM permission - not many people understand `*` followed by Lambda function least privilege
* [x] Hide `Create your own provider` section
* [x] Add a banner that `Creating your own provider` is coming in 2024
  * [x] Create issue to track it
* [x] Add banner explaining that Pydantic, Dataclasses and Classes are coming in 2024
  * [x] Create issue for Pydantic models
  * [x] Create issue for Dataclasses
  * [x] Create issue for standard Python classes
* [x] Refactor intro diagram (show provider, find distinction between mask and encrypt/decrypt)
* [x] Line editing for Intro line
* [x] Refactor Key features to emphasize industry & AWS security best practices (envelop encryption etc.)
* [x] Improve Encryption SDK Provider options using a table
* [x] Create a `Providers` section to future-proof expansion

examples/data_masking/sam/template.yaml
```diff
@@ -0,0 +1,59 @@
...
+Resources:
+  # Lambda Function
+  DataMaskingFunctionExample:
+    Type: AWS::Serverless::Function
+    Properties:
+      Handler: data_masking_function_example.lambda_handler
+      CodeUri: ../src
+      Description: Data Masking Function Example
+      # Cryptographic operations demand more memory usage.
+      # It is recommended to allocate a minimum of 1024MB of memory to your Lambda function
```
heitorlessa: Change to active voice

examples/data_masking/sam/template.yaml
```diff
@@ -0,0 +1,59 @@
...
+    Environment:
+        Variables:
+          POWERTOOLS_SERVICE_NAME: PowertoolsHelloWorld
+          POWERTOOLS_METRICS_NAMESPACE: Powertools
+          LOG_LEVEL: INFO
+          KMS_KEY_ARN: !GetAtt DataMaskingKMSKey.Arn
+
+
+Resources:
+  # Lambda Function
```
heitorlessa: Explain that this is for documentation purposes; in prod, we recommend splitting Encrypt and Decrypt, so one function can act as encryption proxy (HTTP reqs, Data pipeline, etc) while authorized personnel can either call Decrypt from scripts... OR from a separate function altogether.

Same with IAM permissions, we need an example for Decrypt only

examples/data_masking/sam/template.yaml
```diff
@@ -0,0 +1,59 @@
...
+            Action: 'kms:*'
+            Resource: '*'
+          # KMS utilizes resource policies, allowing direct access grant to the Lambda Role on the KMS Key.
+          # For more details: https://docs.aws.amazon.com/kms/latest/developerguide/key-policies.html
+          - Effect: Allow
+            Principal:
+              AWS: !GetAtt DataMaskingFunctionExampleRole.Arn  # Permission for the Lambda role
+            Action:
+              - kms:Decrypt
+              - kms:GenerateDataKey
```
heitorlessa: add code annotation to explain what this is -- envelope encryption requires generating a random key from CMK, so we encrypt in-memory and use it to prevent calling KMS  on every crypto operation

heitorlessa:
> Another PR, not here.

UX improvements before launch to prevent confusion to customers

* Make error message actionable when customers don't use KMS ARN for `decrypt`
    * Error with Key ARN Alias: `aws_encryption_sdk.exceptions.DecryptKeyError: Unable to decrypt any data key`
    * Error with Key alias name only: `aws_encryption_sdk.exceptions.GenerateKeyError: Master Key datamasking unable to generate data key`
    * Error with Key ID only: `raises aws_encryption_sdk.exceptions.DecryptKeyError: Unable to decrypt any data key`

* Make error message actionable when trying to decrypt a field that doesn't have any encrypted ciphertext
    * Error when using the wrong field: `TypeError: argument should be a bytes-like object or ASCII string, not 'dict'`

* Make error message actionable when using an unsupported `field` syntax, e.g., `fields=["addresses[*]]`
    * Error: `KeyError: 'addresses[0]'`, `KeyError: 'addresses[*]'` 

* Improve error message for unsupported data type in encryption/decryption
    * It is unclear for a non-developer and early career devs what this means: `TypeError: Unsupported data type for 'data' parameter. Expected a traversable type, but got <class 'list'>.`

---

> Key variations that fail in `decrypt` operations only
>
> KMS_KEYS = ["arn:aws:kms:eu-west-1:231436140809:alias/datamasking"]  # alias ARN
> KMS_KEYS = ["datamasking"]  # alias name
> KMS_KEYS = ["1053c7d7-97d7-49ca-9763-5214f28999cc"]  # key ID

heitorlessa:
Another one - we should document which algorithm suite we use as the default, since the customer doesn't yet have the option to change it.

Unsure where to best place it (under KMS Provider?) but something we must do regardless.

heitorlessa:
Created two initial sequence diagrams to start demystifying under the hood operations. I'm gonna split caching operation as it became too difficult (cognitive) to read at once.

---

**Mask**

```mermaid
sequenceDiagram
    autonumber
    participant Client
    participant Lambda
    participant DataMasking as Data Masking (in memory)
    Client->>Lambda: Invoke (event)
    Lambda->>DataMasking: mask(data)
    DataMasking->>DataMasking: replaces data with *****
    Note over Lambda,DataMasking: No encryption providers involved.
    DataMasking->>Lambda: data masked
    Lambda-->>Client: Return response
```

**Encryption**

```mermaid
sequenceDiagram
    autonumber
    participant Client
    participant Lambda
    participant DataMasking as Data Masking
    participant EncryptionProvider as Encryption Provider
    Client->>Lambda: Invoke (event)
    Lambda->>DataMasking: encrypt(data)
    DataMasking->>EncryptionProvider: Request unique data key (kms:GenerateDataKey)
    DataMasking->>DataMasking: Encrypt data key with wrapping key (in-memory)
    DataMasking->>DataMasking: Encrypt data with newly encrypted key (in-memory)
    DataMasking->>Lambda: ciphertext containing encrypted data
    Lambda-->>Client: Return response
```

aws_lambda_powertools/utilities/_data_masking/provider/kms/aws_encryption_sdk.py
```diff
@@ -80,10 +85,20 @@ def __init__(
         )
 
     def encrypt(self, data: bytes | str | Dict | int, **provider_options) -> str:
-        return self._key_provider.encrypt(data=data, **provider_options)
+        try:
+            return self._key_provider.encrypt(data=data, **provider_options)
+        except GenerateKeyError:
+            raise DataMaskingEncryptKeyError(
+                "Failed to encrypt data - Please make sure you are using a valid Symmetric AWS MSK Key ARN",
```
seshubaws: nit: I think you mean `KMS` instead of `MSK`

aws_lambda_powertools/utilities/_data_masking/provider/kms/aws_encryption_sdk.py
```diff
@@ -174,19 +166,35 @@ def decrypt(self, data: str, **provider_options) -> Any:
             ciphertext : bytes
                 The decrypted data in bytes
         """
-        ciphertext_decoded = base64.b64decode(data)
+        try:
+            ciphertext_decoded = base64.b64decode(data)
+        except Error:
+            raise DataMaskingDecryptValueError(
+                "Data decryption failed. Please ensure that you are using a field that was previously encrypted.",
```
seshubaws: Maybe we could say something like "Data decryption failed. Please ensure that you are attempting to decrypt data that was previously encrypted." instead of only talking about fields?

heitorlessa:
Got pulled into a customer, but at this point, it'll be easier to create a main issue tracking what's a blocker vs what's post-launch.

I'll create one later tonight

aws_lambda_powertools/utilities/_data_masking/provider/kms/aws_encryption_sdk.py
leandrodamascena: Great catch @seshubaws! Fixing it now!

leandrodamascena:
Adding information here about "Bring your own keys/provider"

https://github.com/aws-powertools/powertools-lambda-python/issues/3472#issuecomment-1856234851

heitorlessa:
Revamped encryption diagram with correct order of operations, and added a better intro to the diagram. I'll have to update Terminologies now to include `Master key`, `Data key`, and `Envelope encryption`

image

heitorlessa:
List so we can assess whether we are ready to ship by Dec 21st

**blocker**

* [ ] Implementation: Docstrings
* [x] Implementation: Debug logging
* [x] Implementation: CMM should accept `Max bytes processed`
* [ ] Docs: Fields syntax
* [ ] Docs: Testing your code
* [ ] Docs: High availability with multiple keys
* [ ] Docs: Add exception details / error handling

**post-launch**

* [ ] Pydantic support
* [ ] Dataclasses support
* [ ] New base class to make it easier to bring your own encryption provider w/ Encryption SDK, e.g., `bring your own key`
* [ ] Multi-keyring for multiple providers to further increase availability in rare cases
* [ ] Blog post
* [ ] What's New

aws_lambda_powertools/utilities/_data_masking/constants.py
```diff
@@ -1,5 +1,9 @@
...
+# The maximum number of entries that can be retained in the local cryptographic materials cache
 CACHE_CAPACITY: int = 100
+# The maximum time (in seconds) that a cache entry may be kept in the cache
 MAX_CACHE_AGE_SECONDS: float = 300.0
-MAX_MESSAGES_ENCRYPTED: int = 200
-# NOTE: You can also set max messages/bytes per data key
+# Maximum number of messages which are allowed to be encrypted under a single cached data key
+MAX_MESSAGES_ENCRYPTED: int = 4294967296  # 2 ** 32
+# Maximum number of bytes which are allowed to be encrypted under a single cached data key
+MAX_BYTES_ENCRYPTED: int = 9223372036854775807  # 2 ** 63 - 1
```
leandrodamascena: We need to review these values and find the right balance between the user experience and best practices.

seshubaws: These are the default values from the [SDK](https://github.com/aws/aws-encryption-sdk-python/blob/93a67d8a3806f560ead950e6d8898e53c4c4f9df/src/aws_encryption_sdk/internal/defaults.py#L39-L41), would it be inconvenient for the customer to use these values as opposed to the previous ones? If so, we can go back to the previous ones, I just got them from an example in the docs. But [best practices](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/best-practices.html) do say that we should use default values.

seshubaws:
> Docs: Testing your code

What should be added for this? Should we show how to test like we did in our integration tests [here](https://github.com/aws-powertools/powertools-lambda-python/blob/49c5ab1232f494ac00da677707e68e5a16920642/tests/functional/data_masking/test_aws_encryption_sdk.py#L17) so they don't need a KMS key or have to call any AWS APIs?

heitorlessa:
> > Docs: Testing your code
> 
> What should be added for this? Should we show how to test like we did in our integration tests [here](https://github.com/aws-powertools/powertools-lambda-python/blob/49c5ab1232f494ac00da677707e68e5a16920642/tests/functional/data_masking/test_aws_encryption_sdk.py#L17) so they don't need a KMS key or have to call any AWS APIs?

We need to show them how to unit test their code when this new feature is used, see [Idempotency](https://docs.powertools.aws.dev/lambda/python/latest/utilities/idempotency/#testing-your-code). 

At the moment, as a customer, if I were to read `Testing your code` section, it'd give me the impression you're showing me how to test Sensitive Data Masking internally, and not the code I'd write.

heitorlessa:
Discovered a new blocker as I write the section about encryption context -- we need to move `provider_options` as a named parameter, and use `**encryption_context` as kwargs for a more fluid API.

> History: The reason we did that was we weren't sure back then whether we could have a custom provider w/o KMS easily -- we proved we can recently. Since encryption context is tied to an encrypted message (it's portable!), we can improve the API now.

**Improved API**

```python
data = {"email": "joe@dot.com"}

data_masker.encrypt(data, fields=["email"], data_classification="red", data_type="customer-data")


# alternative when there are many fields
ctx = {
    "data_classification": "red",
    "data_type": "customer-data",
    "purpose": "compliance",
    "tenant_id": "2294264a-fadd-40a6-a54c-a8f336d3a30b"
}

data_masker.encrypt(data, fields=["email"], **ctx)


```

**Instead of**

```python
data = {"email": "joe@dot.com"}

ctx = {
    "data_classification": "red",
    "data_type": "customer-data",
    "purpose": "compliance",
    "tenant_id": "2294264a-fadd-40a6-a54c-a8f336d3a30b"
}

data_masker.encrypt(data, fields=["email"], encryption_context=ctx)


```

leandrodamascena:
Added support to List/Sequence types:

```python
data = json.dumps(
        {
            "payload": {
                "first": ["value1", "value2"],
                "second": [{"key1": [0, 1]}],
            },
        },
    )

    fields_operation = ["payload.first[0]", "payload.second[0].key1[0]"]
    # WHEN encrypting and then decrypting the encrypted data
    encrypted_data = data_masker.encrypt(data, fields=fields_operation)
    decrypted_data = data_masker.decrypt(encrypted_data, fields=fields_operation)

    # THEN the result is only the specified fields are masked
    assert decrypted_data == json.loads(data)
```

docs/utilities/data_masking.md
```diff
@@ -97,15 +97,19 @@ Before you start, you will need a KMS key to encrypt and decrypt your data. Your
 
 ### Masking data
 
-You can mask data without having to install any encryption library. Masking data will result in the loss of its original type, and the masked data will always be represented as a string.
+!!! note "You can mask data without [installing any dependency](#install)."
+
+Masking will erase the original data and replace with `*****`. This means you cannot recover masked data, and its type will change to `str`.
```
seshubaws: Should we add here that the type will change to `str` unless the data to be masked is of an Iterable type like `list`, `tuple`, or `set`, in which case it will return a new object of the same type as the input data but with each element replaced by the string `*****`?

aws_lambda_powertools/utilities/_data_masking/provider/base.py
```diff
@@ -1,13 +1,70 @@
...
+    default_json_deserializer(data)
+        Default method for JSON deserialization.
+    encrypt(data)
+        Abstract method for encrypting data. Subclasses must implement this method.
+    decrypt(data)
+        Abstract method for decrypting data. Subclasses must implement this method.
+    mask(data)
+        Default method for masking data.
+
+    Examples
```
seshubaws: >  Implementation: Docstrings (some don't match rest of the code base)

Should I get rid of the "Example" parts of the docstrings? I think it's redundant since it's in the docs.

docs/utilities/data_masking.md
heitorlessa: Great catch! I was gonna add that to the `Fields` piece, but you're 100% right.. one might not reach that section and take it for granted.

We need to encapsulate that into the static typing return too.

aws_lambda_powertools/utilities/_data_masking/provider/base.py
heitorlessa: It serves a different purpose -- customers wanting to stay within the IDE. One example with encryption using fields+encryption context should suffice. Like other features, the `Example` should work as-is if copied/pasted.

docs/utilities/data_masking.md
heitorlessa: image

heitorlessa:
Another blocker, encryption context only allows `str` values, and the exception is untreated and confusing -- we must provide a better experience to customers.

---

```python
...
data = {
    "zip": {"postcode": [26134]},
}

masker.encrypt(data, fields=["zip.postcode"], encryption_context={"non_string_value": 10})
```

image

heitorlessa:
I put back one of the return types @leandrodamascena to intentionally fail CI, because the `encrypt/decrypt/mask` are incorrect when the `fields` parameter is used.

heitorlessa:
List so we can assess whether we are ready to ship by Dec 21st

## Blockers

* [ ] Implementation
  * [ ] Docstrings (some don't match rest of the code base)
  * [ ] Review types, as `fields=` looks incorrect return type
  * [ ] Encryption context validation should be an exact match to prevent confused deputy - exclude Encryption SDK reserved key
  * [ ] Review `fields` implementation - it will be difficult to maintain if we were to support a richer data selection
  * [x] Review `fields` unit test to confirm addition of `Sequence` (list,tuple,set)
  * [x] We should be able to access a field within a list, `{"top_key": [ { "key": "value" } ] }`
  * [x] Non-string encryption context exception is untreated, should re-raise as `DataMaskingUnsupportedTypeError`
  * [x] API change to allow for arbitrary encryption context, `encrypt(data, fields, provider_options, tenant_id='a', data_type='c')`
  * [x] Allow customers to use custom JSON serializer and deserializer for field operations in `DataMasking` too. 
  * [x] Debug logging
  * [x] CMM should accept `Max bytes processed`
  * [x] Handle `list/sequences` types
* [ ] Tests
  * [ ] Encryption context (enc/dec, validation)
  * [ ] Confused deputy for decryption context
  * [ ] Accessing fields within a list/container
  * [ ] Propagating provider options correctly
* [ ] Documentation
  * [ ] Testing your code; need a more realistic source code that uses data masker along with unit test for it
  * [x] Fields syntax
  * [x] Add exception details / error handling
  * [x] High availability with multiple keys
  * [x] Exception handling;  e.g., what if a `field` isn't available, what about an unsupported data type, or passed `Key ID` instead of `Key ARN`, or an unsupported `field pattern`, etc.
  * [x] Encryption context; encrypt and decrypt operations, highlight constraints like exact match, string only, non-sensitive, and exception raised

## Post launch

* [ ] Pydantic support
* [ ] Dataclasses support
* [ ] New base class to make it easier to bring your own encryption provider w/ Encryption SDK, e.g., `bring your own key`
* [ ] Multi-keyring for multiple providers to further increase availability in rare cases
* [ ] Blog post
* [ ] What's New

heitorlessa:
Moving serializer/deserializer from provider to `DataMasking()`.

Edit. Correction, adding serializer/deserializer to Data Masking too.

seshubaws:
> Encryption context validation should be an exact match to prevent confused deputy - exclude Encryption SDK reserved key

I don't believe we are checking the for the Encryption SDK reserved key in the enc_ctx currently. We are only going through the items in the `encryption_context` given by the user in the `encrypt` method and checking to see if all those items are in the `decryptor_header.encryption_context` (which is the object returned by the `decrypt` method that includes the Encryption SDK reserved key). So it never checks against the reserved key bc it's only going through the key-val pairs of the enc_ctx given in the `encrypt` [method](https://github.com/aws-powertools/powertools-lambda-python/pull/3186/files#diff-454e6151557effab56c9ca485477fe6ee8fc02f90f26822b2cf210312cbe8fc8R227). But let me know if I am misunderstanding.

heitorlessa:
> > Encryption context validation should be an exact match to prevent confused deputy - exclude Encryption SDK reserved key
> 
> I don't believe we are checking the for the Encryption SDK reserved key in the enc_ctx currently. We are only going through the items in the `encryption_context` given by the user in the `encrypt` method and checking to see if all those items are in the `decryptor_header.encryption_context` (which is the object returned by the `decrypt` method that includes the Encryption SDK reserved key). So it never checks against the reserved key bc it's only going through the key-val pairs of the enc_ctx given in the `encrypt` [method](https://github.com/aws-powertools/powertools-lambda-python/pull/3186/files#diff-454e6151557effab56c9ca485477fe6ee8fc02f90f26822b2cf210312cbe8fc8R227). But let me know if I am misunderstanding.

Thank you so much, Seshu!! I've added them late last night and it's on my lists to increase test coverage -- your tip will save time (and perf!)

heitorlessa:
Working on static typing now

heitorlessa:
@seshubaws we're gonna power through today as much as we can to complete the blockers list to get ready for tomorrow's launch -- I'd send a list by your morning time of areas I'd love to get your review to make sure we haven't missed anything.

heitorlessa:
@leandrodamascena when you have a moment, add an explanation about `jsonpath-ng` to keep Seshu informed on why.

heitorlessa:
Moving current status here due to GitHub hiding that info as updates come in.

---

List so we can assess whether we are ready to ship by Dec 21st

## Blockers

* [ ] Implementation
  * [ ] Docstrings (some don't match rest of the code base) -- @heitorlessa 
  * [ ] Lower CMM defaults
  * [x] Make feature public -- @heitorlessa 
  * [ ] Last security pass -- @sthulb 
  * [x] Encryption context validation should be an exact match to prevent confused deputy - exclude Encryption SDK reserved key -- @heitorlessa 
  * [x] Review `fields` implementation - it will be difficult to maintain if we were to support a richer data selection; move to jsonpath-ng if necessary due to in-place editing.
  * [x] Handle missing keys strict vs non-strict
  * [x] Review types, as `fields=` looks incorrect return type
  * [x] Review `fields` unit test to confirm addition of `Sequence` (list,tuple,set)
  * [x] We should be able to access a field within a list, `{"top_key": [ { "key": "value" } ] }`
  * [x] Non-string encryption context exception is untreated, should re-raise as `DataMaskingUnsupportedTypeError`
  * [x] API change to allow for arbitrary encryption context, `encrypt(data, fields, provider_options, tenant_id='a', data_type='c')`
  * [x] Allow customers to use custom JSON serializer and deserializer for field operations in `DataMasking` too. 
  * [x] Debug logging
  * [x] CMM should accept `Max bytes processed`
  * [x] Handle `list/sequences` types
* [ ] Tests
  * [x] Accessing fields within a list/container -- @leandrodamascena 
  * [ ] Propagating provider options correctly -- @leandrodamascena 
  * [x] Encryption context (enc/dec, validation) -- @heitorlessa 
  * [x] Confused deputy for decryption context -- @heitorlessa 
* [ ] Documentation
  * [ ] Testing your code; need a more realistic source code that uses data masker along with unit test for it -- @leandrodamascena 
  * [ ] Document exception handling for missing keys behavior (e.g., strict, warning) -- @heitorlessa 
  * [x] Fields syntax
  * [x] Add exception details / error handling
  * [x] High availability with multiple keys
  * [x] Exception handling;  e.g., what if a `field` isn't available, what about an unsupported data type, or passed `Key ID` instead of `Key ARN`, or an unsupported `field pattern`, etc.
  * [x] Encryption context; encrypt and decrypt operations, highlight constraints like exact match, string only, non-sensitive, and exception raised

## Post launch

* [ ] Pydantic support
* [ ] Dataclasses support
* [ ] New base class to make it easier to bring your own encryption provider w/ Encryption SDK, e.g., `bring your own key`
* [ ] Multi-keyring for multiple providers to further increase availability in rare cases
* [ ] Blog post
* [ ] What's New

heitorlessa:
> > Encryption context validation should be an exact match to prevent confused deputy - exclude Encryption SDK reserved key
> 
> I don't believe we are checking the for the Encryption SDK reserved key in the enc_ctx currently. We are only going through the items in the `encryption_context` given by the user in the `encrypt` method and checking to see if all those items are in the `decryptor_header.encryption_context` (which is the object returned by the `decrypt` method that includes the Encryption SDK reserved key). So it never checks against the reserved key bc it's only going through the key-val pairs of the enc_ctx given in the `encrypt` [method](https://github.com/aws-powertools/powertools-lambda-python/pull/3186/files#diff-454e6151557effab56c9ca485477fe6ee8fc02f90f26822b2cf210312cbe8fc8R227). But let me know if I am misunderstanding.

There was an issue with the current logic that led to a confused deputy situation -- we now do an **exact match** of the encryption context received during `decrypt` vs what's been injected in the encrypted message.

For that purpose, we now remove the encrypted data key from encryption context before converting to a set for exact comparison (ordered or out of order)

image

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,578 @@
...
+
+## Key features
+
+* Encrypt, decrypt, or irreversibly mask data with ease
+* Remove sensitive information in one or more fields within nested data
+* Seamless integration with [AWS Encryption SDK](https://docs.aws.amazon.com/encryption-sdk/latest/developer-guide/introduction.html){target="_blank"} for industry and AWS security best practices
+
+## Terminology
+
+**Masking** replaces sensitive information **irreversibly** with a non-sensitive placeholder _(`*****`)_. It replaces data in-memory, hence why being irreversible.
```
sthulb: We need to change the language here from "irreversibly" to something else, I think the word probably isn't accessible to some ESL speakers.

sthulb: This sentence probably needs rewriting, it doesn't seem right to me for some reason.
> **Masking** is the act of replacing information with a non-sensitive placeholder _(`*****`)_, this operation is done in memory making it a one-way action

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,578 @@
...
+<center>
+```mermaid
+graph LR
+    M(Master key) --> |Encrypts| D(Data key)
+    D(Data key) --> |Encrypts| S(Sensitive data)
+```
+<i>Envelope encryption visualized.</i>
+</center>
+
+## Getting started
```
sthulb, line 69: We should link to either an encryption 101 doc or something to help people understand the concept if they're not familiar, this jumps to L400 quite quickly

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,578 @@
...
+    --8<-- "examples/data_masking/src/encrypt_data_output.json"
+    ```
+
+=== "getting_started_decrypt_data_output.json"
+
+    ```json hl_lines="5-7 12-17"
+    --8<-- "examples/data_masking/src/getting_started_decrypt_data_output.json"
+    ```
+
+### Encryption context for integrity and authenticity
```
sthulb, line 180: this section should link to something talking about the confused deputy issue, there's probably a good 3rd party source we can link to.

It can help explain why AAD is a good idea

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,578 @@
...
+    --8<-- "examples/data_masking/src/advanced_custom_serializer.py"
+    ```
+
+### Providers
+
+#### AWS Encryption SDK
+
+You can modify the following values when initializing the `AWSEncryptionSDKProvider` to best accommodate your security and performance thresholds.
+
+| Parameter                  | Required | Default               | Description                                                                                   |
```
sthulb: Are any of these parameters required? if not, there's no need to have the column

sthulb (commented):
Overall, this is technical correct but some of the language makes it seem a little rushed – I think it's the reuse of the term "under the hood" (used 4x)

aws_lambda_powertools/utilities/data_masking/provider/kms/aws_encryption_sdk.py
```diff
@@ -0,0 +1,245 @@
...
+        self._validate_encryption_context(encryption_context)
+
+        data_encoded = self.json_serializer(data).encode("utf-8")
+
+        try:
+            ciphertext, _ = self.client.encrypt(
+                source=data_encoded,
+                materials_manager=self.cache_cmm,
+                encryption_context=encryption_context,
+                **provider_options,
```
seshubaws, line 169: Is this supposed to be `provider_options` and `**encryption_context`? That is what I see in most places but in a few instances it is reversed.

heitorlessa: Because this is the actual API call, we pass the actual `dict` to the client. If we were to unpack `**encryption_context`, it would cause a SDK Client ParamError because it wouldn't have those key=value as accepted values. 

`**provider_options` on the other hand would be a 1:1 match with what's accepted by the client so it's safe to unpack - e.g., `provider_options={"algorithm": "..."}`

heitorlessa (changes requested):
We will postpone the launch until further notice. During review with the crypto team today, there's a potential confused deputy scenario due to data locality and encryption context bounding distinct blobs (bc fields).

We'll regroup in January to discuss mitigations, if any, and decide its fate.

aws_lambda_powertools/utilities/_data_masking/constants.py
seshubaws: @heitorlessa @sthulb Do we want to change these values for `max_messages_encrypted` and `max_bytes_encrypted`? The current values are the default ones that came from the encryption SDK docs page, but it could be a security concern bc cxs would be using one cached key to encrypt 4294967296 messages by default... I was thinking we use lower numbers (maybe `200` and `2000` respectively) and explain the range of possible values they take.

docs/utilities/data_masking.md
seshubaws: I don't think AAD (the encryption context in this case) will help to prevent the confused deputy issue here right? Won't only proper authentication and authorization prevent that, especially since the encryption context is plaintext?

heitorlessa:
Quick ones before you merge

* [ ] update `Features` list in the docs homepage
* [ ] update `Features` list in the GitHub homepage

---

Reviewing

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,587 @@
...
+    ```
+
+    1. They must match otherwise the operation will fail with `DataMaskingContextMismatchError`.
+
+### Choosing parts of your data
+
+!!! note "We support `JSON` data types only - see [data serialization for more details](#data-serialization-and-preservation)."
+
+???+ note "Current limitations"
+    1. The `fields` parameter is currently only available to use with the `erase` method, with the potential for it to be added to the `encrypt` and `decrypt` methods in the future.
```
heitorlessa: merge these two and make the `fields` note a single short sentence, please

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,587 @@
...
+
+You can use the `fields` parameter with dot notation `.` to choose one or more parts of your data to `erase`. This is useful when you want to keep data structure intact except the confidential fields.
```
heitorlessa:
```suggestion
You can use the `fields` parameter with the dot notation `.` to choose one or more parts of your data to `erase`. This is useful when you want to keep data structure intact except the confidential fields.
```

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,587 @@
...
+
+When `fields` is present, `erase` behaves differently:
+
+| Operation | Behavior                                                    | Example                 | Obfuscated                      |
```
heitorlessa:
```suggestion
| Operation | Behavior                                                    | Example                 | Result                      |
```

Now that we changed to `erase`, `obfuscated` could give the impression is can be recovered. Different from `encrypt` we had earlier.

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,587 @@
...
+        ```json hl_lines="8 12"
+        --8<-- "examples/data_masking/src/choosing_payload_complex_search.json"
+        ```
+
+    === "Result"
+
+        ```json hl_lines="8 12"
+        --8<-- "examples/data_masking/src/choosing_payload_complex_search_output.json"
+        ```
+For comprehensive guidance on using JSONPath syntax, please refer to the official documentation available at [jsonpath-ng](https://github.com/h2non/jsonpath-ng#jsonpath-syntax){target="_blank" rel="nofollow"}
```
heitorlessa:
```suggestion
    You want to erase data in the `card_number` field.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["card_number"])`

        ```json hl_lines="4"
        --8<-- "examples/data_masking/src/choosing_payload_top_keys.json"
        ```

    === "Result"

        ```json hl_lines="4"
        --8<-- "examples/data_masking/src/choosing_payload_top_keys_output.json"
        ```

=== "Nested key"

    You want to erase data in the `postcode` field.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["address.postcode"])`

        ```json hl_lines="6"
        --8<-- "examples/data_masking/src/choosing_payload_nested_key.json"
        ```

    === "Result"

        ```json hl_lines="6"
        --8<-- "examples/data_masking/src/choosing_payload_nested_key_output.json"
        ```

=== "Multiple keys"

    You want to erase data in both `postcode` and `street` fields.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["address.postcode", "address.street"])`

        ```json hl_lines="6-7"
        --8<-- "examples/data_masking/src/choosing_payload_multiple_keys.json"
        ```

    === "Result"

        ```json hl_lines="6-7"
        --8<-- "examples/data_masking/src/choosing_payload_multiple_keys_output.json"
        ```

=== "All key items"

    You want to erase data under `address` field.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["address"])`

        ```json hl_lines="6-17"
        --8<-- "examples/data_masking/src/choosing_payload_all_nested_keys.json"
        ```

    === "Result"

        ```json hl_lines="6-7"
        --8<-- "examples/data_masking/src/choosing_payload_all_nested_keys_output.json"
        ```

=== "Complex nested key"

    You want to erase data under `name` field.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["category..name"])`

        ```json hl_lines="6"
        --8<-- "examples/data_masking/src/choosing_payload_complex_nested_keys.json"
        ```

    === "Result"

        ```json hl_lines="6"
        --8<-- "examples/data_masking/src/choosing_payload_complex_nested_keys_output.json"
        ```

=== "All fields in a list"

    You want to erase data under `street` field located at the any index of the address list.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["address[*].street"])`

        ```json hl_lines="8 12"
        --8<-- "examples/data_masking/src/choosing_payload_list_all_index.json"
        ```

    === "Result"

        ```json hl_lines="8 12"
        --8<-- "examples/data_masking/src/choosing_payload_list_all_index_output.json"
        ```

=== "Slicing a list"

    You want to erase data by slicing a list.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["address[-1].street"])`

        ```json hl_lines="16"
        --8<-- "examples/data_masking/src/choosing_payload_list_slice.json"
        ```

    === "Result"

        ```json hl_lines="16"
        --8<-- "examples/data_masking/src/choosing_payload_list_slice_output.json"
        ```

=== "Complex expressions"

    You want to erase data by finding for a field with conditional expression.

    === "Data"

        > Expression: `data_masker.erase(data, fields=["$.address[?(@.postcode > 12000)]"])`

        > `$`: Represents the root of the JSON structure.

        > `.address`: Selects the "address" property within the JSON structure.

        > `(@.postcode > 12000)`: Specifies the condition that elements should meet. It selects elements where the value of the `postcode` property is `greater than 12000`.

        ```json hl_lines="8 12"
        --8<-- "examples/data_masking/src/choosing_payload_complex_search.json"
        ```

    === "Result"

        ```json hl_lines="8 12"
        --8<-- "examples/data_masking/src/choosing_payload_complex_search_output.json"
        ```

For comprehensive guidance on using JSONPath syntax, please refer to the official documentation available at [jsonpath-ng](https://github.com/h2non/jsonpath-ng#jsonpath-syntax){target="_blank" rel="nofollow"}
```

replacing `obfuscate` with `erase`. Once `fields` is added back to encryption/decryption we can rollback to the use of `obfuscate` semantic.

docs/utilities/data_masking.md
```diff
@@ -0,0 +1,587 @@
...
+
+When decrypting, we revert the operation to restore the original data structure.
+
+For compatibility or performance, you can optionally pass your own JSON serializer and deserializer to replace `json.dumps` and `json.loads` respectively:
+
+=== "advanced_custom_serializer.py"
+
+    ```python hl_lines="17-18"
+    --8<-- "examples/data_masking/src/advanced_custom_serializer.py"
+    ```
```
heitorlessa:
```suggestion
```python hl_lines="17-18" title="advanced_custom_serializer.py"
--8<-- "examples/data_masking/src/advanced_custom_serializer.py"
```
```

Single snippets don't need Tabs. Please double check if it renders as GitHub preview seems to be removing the closing backtick.

File: docs/utilities/data_masking.md [hunk 23]
```diff
@@ -0,0 +1,587 @@
...
+
+**Changing the default algorithm**
+
+The AWS Encryption SDK defaults to using the `AES_256_GCM_HKDF_SHA512_COMMIT_KEY_ECDSA_P384` algorithm for encrypting your Data Key. If you want, you have the flexibility to customize and choose a different encryption algorithm.
+
+=== "changing_default_algorithm.py"
+
+    ```python hl_lines="5 26"
+    --8<-- "examples/data_masking/src/changing_default_algorithm.py"
+    ```
```
heitorlessa: For completion, we need to explain how to override `EncryptionSDKClient` options (algorithm is one of them).

suggestion: rename to `Passing additional SDK arguments`, and keep the `algorithm` as one of the arguments they can use via `provider_options`.

image

aws_lambda_powertools/utilities/_data_masking/constants.py
seshubaws: Left the eSDK defaults and explained in the docs how to override them

README.md
```diff
@@ -30,6 +30,7 @@ Powertools for AWS Lambda (Python) is a developer toolkit to implement Serverles
 * **[Event source data classes](https://docs.powertools.aws.dev/lambda/python/latest/utilities/data_classes/)** - Data classes describing the schema of common Lambda event triggers
 * **[Parser](https://docs.powertools.aws.dev/lambda/python/latest/utilities/parser/)** - Data parsing and deep validation using Pydantic
 * **[Idempotency](https://docs.powertools.aws.dev/lambda/python/latest/utilities/idempotency/)** - Convert your Lambda functions into idempotent operations which are safe to retry
+* **[Data Masking](https://docs.powertools.aws.dev/lambda/python/latest/utilities/data_masking/)** -  Protect confidential data with easy removal or encryption
```
heitorlessa, line 33: @seshubaws I wanted to say I loved how short and on point this it. This could be the title of a presentation, blog post, and tutorial!

docs/utilities/data_masking.md
```diff
@@ -204,24 +205,23 @@ For a stronger security posture, you can add metadata to each encryption operati
 
 ### Choosing parts of your data
 
-!!! note "We support `JSON` data types only - see [data serialization for more details](#data-serialization-and-preservation)."
-
 ???+ note "Current limitations"
-    1. The `fields` parameter is currently only available to use with the `erase` method, with the potential for it to be added to the `encrypt` and `decrypt` methods in the future.
+    1. The `fields` parameter is currently exclusive to the `erase` method, with potential future inclusion into `encrypt` and `decrypt`.
```
heitorlessa:
```suggestion
    1. The `fields` parameter is not yet supported in `encrypt` and `decrypt` operations.
```

docs/utilities/data_masking.md
```diff
@@ -204,24 +205,23 @@ For a stronger security posture, you can add metadata to each encryption operati
...
+    2. We support `JSON` data types only - see [data serialization for more details](#data-serialization)."
```
heitorlessa:
```suggestion
    2. We support `JSON` data types only - see [data serialization for more details](#data-serialization).
```

docs/utilities/data_masking.md
```diff
@@ -408,9 +409,9 @@ For compatibility or performance, you can optionally pass your own JSON serializ
 
 === "advanced_custom_serializer.py"
```
heitorlessa:
```suggestion
```

docs/utilities/data_masking.md
```diff
@@ -408,9 +409,9 @@ For compatibility or performance, you can optionally pass your own JSON serializ
...
 
-    ```python hl_lines="17-18"
-    --8<-- "examples/data_masking/src/advanced_custom_serializer.py"
-    ```
+```python hl_lines="17-18"
```
heitorlessa:
```suggestion
```python hl_lines="17-18" title="advanced_custom_serializer.py"
```

Same as before but without tabbed content (single example).

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
 | **max_bytes_encrypted**    | `9223372036854775807` | The maximum number of bytes that may be encrypted under a cache entry                         |
 
-**Changing the default algorithm**
+If required, you have the option to customize the default values when initializing the `AWSEncryptionSDKProvider` class.
+
+=== "aws_encryption_provider_example.py"
+
+```python hl_lines="14-19"
+--8<-- "examples/data_masking/src/aws_encryption_provider_example.py"
+```
```
heitorlessa:
```suggestion
```python hl_lines="14-19" title=""aws_encryption_provider_example.py""
--8<-- "examples/data_masking/src/aws_encryption_provider_example.py"
```
```

Removing tab, keeping a single snippet with title for consistency.

File: docs/utilities/data_masking.md [hunk 30]
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
 | **max_messages_encrypted** | `4294967296`          | The maximum number of messages that may be encrypted under a cache entry                      |
 | **max_bytes_encrypted**    | `9223372036854775807` | The maximum number of bytes that may be encrypted under a cache entry                         |
 
-**Changing the default algorithm**
+If required, you have the option to customize the default values when initializing the `AWSEncryptionSDKProvider` class.
```
heitorlessa:
```suggestion
If required, you can customize the default values when initializing the `AWSEncryptionSDKProvider` class.
```

`You have the option to` becomes `You can`

You could go even further in the future when shorter length is necessary by using synonyms, for example.

`You have the option to customize` becomes `You can change`

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+
+=== "aws_encryption_provider_example.py"
+
+```python hl_lines="14-19"
+--8<-- "examples/data_masking/src/aws_encryption_provider_example.py"
+```
 
-The AWS Encryption SDK defaults to using the `AES_256_GCM_HKDF_SHA512_COMMIT_KEY_ECDSA_P384` algorithm for encrypting your Data Key. If you want, you have the flexibility to customize and choose a different encryption algorithm.
+**Passing additional SDK arguments**
```
heitorlessa:
```suggestion
##### Passing additional SDK arguments
```

changing to heading so customers can use deep linking in their content, etc. It's not the best visual but what we've got for deep linking in this case.

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+
+You can pass additional arguments to the `AWSEncryptionSDKProvider` via the `provider_options` parameter. To learn more about the different arguments you can give to the SDK, see the [EncryptionSDKClient's documentation](https://aws-encryption-sdk-python.readthedocs.io/en/latest/generated/aws_encryption_sdk.html#aws_encryption_sdk.EncryptionSDKClient.encrypt){target="_blank"}.
```
heitorlessa:
```suggestion
!!! note "See the [AWS Encryption SDK docs for more details](https://aws-encryption-sdk-python.readthedocs.io/en/latest/generated/aws_encryption_sdk.html#aws_encryption_sdk.EncryptionSDKClient.encrypt){target="_blank"}"

As an escape hatch mechanism, you can pass additional arguments to the `AWSEncryptionSDKProvider` via the `provider_options` parameter.
```

We can shorten the message by using a visual cue and decouple our sentence vs an action - a `note` :)

heitorlessa: **BEFORE**

image


**AFTER**

image

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+
+For example, the AWS Encryption SDK defaults to using the `AES_256_GCM_HKDF_SHA512_COMMIT_KEY_ECDSA_P384` algorithm for encrypting your Data Key. If you want, you have the flexibility to customize and choose a different encryption algorithm.
 
 === "changing_default_algorithm.py"
```
heitorlessa:
```suggestion
```

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
 
-    ```python hl_lines="5 26"
-    --8<-- "examples/data_masking/src/changing_default_algorithm.py"
-    ```
+```python hl_lines="5 26 30"
```
heitorlessa:
```suggestion
```python hl_lines="5 26 30" title="changing_default_algorithm.py"
```

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+--8<-- "examples/data_masking/src/changing_default_algorithm.py"
+```
+
+**Using multiple keys**
```
heitorlessa:
```suggestion
##### Using multiple keys
```

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+
+The `AWSEncryptionSDKProvider` allows you to instantiate it with several KMS keys by passing them all in a `list` to the `keys` parameter. This could be beneficial if you own keys in different regions, enabling you to perform cross-regional encryption and decryption.
```
heitorlessa:
```suggestion
You can use multiple KMS keys from more than one AWS account for higher availability, when instantiating `AWSEncryptionSDKProvider`.
```

You can shorten it by using active voice. A minor fix is that they can't use multi-region yet, because we are not exposing the MRK class yet.

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+
+=== "using_multiple_keys.py"
```
heitorlessa:
```suggestion
```

docs/utilities/data_masking.md
```diff
@@ -425,15 +426,35 @@ You can modify the following values when initializing the `AWSEncryptionSDKProvi
...
+
+```python hl_lines="15"
```
heitorlessa:
```suggestion
```python hl_lines="15" title="using_multiple_keys.py"
```

heitorlessa (changes requested):
last changes ;) look awesome now!

The only piece I wish we had but we can merge without is a sequence diagram to demonstrate what happens when you have multiple KMS keys.

leandrodamascena:
> last changes ;) look awesome now!
> 
> The only piece I wish we had but we can merge without is a sequence diagram to demonstrate what happens when you have multiple KMS keys.

Hello @heitorlessa! I created the diagram but I don't know if it's enough. Can you review, please?

leandrodamascena (dismissed):
APPROVED!!!!!!!!

heitorlessa:
> > last changes ;) look awesome now!
> > The only piece I wish we had but we can merge without is a sequence diagram to demonstrate what happens when you have multiple KMS keys.
> 
> Hello @heitorlessa! I created the diagram but I don't know if it's enough. Can you review, please?

done ;) Changes we made

* We've changed the diagram from `Alt` to `Loop` so it's easier to know this operation will repeat for N keys specified
* We've used KEY_1 and KEY_2 so it's easier to follow which operation happens under which key
* We've updated the SAM template to include `kms:Encrypt` least privilege permission along with a link on why
* We've updated a quick note explaining that ALL Encrypted Data keys will be included in the encrypted message for portability, allowing higher availability.

heitorlessa (dismissed):

heitorlessa:
we're good to go after ensuring deps are included, since Layers will include (and optimize size) Encryption SDK and JSONPath-Ng too.

heitorlessa (approved):

PR #3195, opened 2023-10-12
fix(parameter): improve AppConfig cached configuration retrieval

### LINKED ISSUE #3194: Bug: AppConfig Provider Returning Wrong Configuration
### Expected Behaviour

When using the `AppConfig Provider` to fetch user specific configurations at runtime based on the identity of the requester that invokes our lambda function, we expect the correct configuration to be returned. Our AppConfig Application and Environment names are used to instantiate a global `AppConfig Provider` object at runtime, and the this provider is used to fetch the configuration specific to the requester multiple times throughout the execution of the request in that lambda container. We do not cache the profile and instead rely on the caching that is built into the `AppConfig Provider`. 

### Current Behaviour

What we are seeing is that when multiple requests are being made at the same time from different users (across different lambda invocations), the AppConfig provider will be called to fetch the configuration with the right username, but return the wrong configuration profile (it will be called with user A, but return user B's profile). We have tried turning off caching of the profiles and this did not help. All we are doing is calling that initialized AppConfigProvider with the config_name that is specific to each user when we need to fetch that profile. It is important to note that for each request, we only need to fetch one profile that is scoped to the user who made the request.

### Code snippet

```python
# Global AppConfig Provider Object
appconfig_client = AppConfigClient(
    application=APPLICATION_NAME,
    environment=ENVIRONMENT_NAME,
)

# Method used to fetch the Configuration
def get_config_as_dict(self, config_name: str) -> Dict[str, Any]:
     logger.info(f"Fetching configuration: {config_name}...")
     return cast(
         dict,
         self._app_config_client.get(
             name=config_name, transform="json", max_age=600
         ),
     )
```

### Possible Solution

_No response_

### Steps to Reproduce

Multiple requests attempting to fetch different configurations at the same time should be able to reproduce this error. The TPS for our service should not be more than 2-3 per second at most. Lambda containers being reused across requests most likely does play a part here. 

2.17

3.8

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #3194 

### Changes

This pull request fix an issue related to `AppConfig` caching behavior when multiple configuration calls are made using the same application/environment. 

The expected behavior of `get_latest_configuration` function is to return a `None/null` value when no new configuration is available. This design is intentional and value should be cached on the customer side. We created a dictionary for storing the most recently retrieved value for a specific configuration profile.

For additional context and details regarding the behavior of `get_latest_configuration`, please refer to the [official AWS AppConfig documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/appconfigdata/client/get_latest_configuration.html).

### User experience

There is no change in the user experience, but it fixes the error.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/utilities/parameters/appconfig.py
```diff
@@ -126,10 +126,13 @@ def _get(self, name: str, **sdk_options) -> str:
         return_value = response["Configuration"].read()
         self._next_token[name] = response["NextPollConfigurationToken"]
 
+        # The return of get_latest_configuration can be null because this value is supposed to be cached
+        # on the customer side. We created a dict with the last returned value for the specific configuration
+        # See https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/appconfigdata/client/get_latest_configuration.html
```
heitorlessa, line 133: Add one line about the dict change.

This will help trace back this bug, along with a PR Body update to make it clear what the bug was

heitorlessa (dismissed):
Tiny changes - LGTM!!

1. Update PR body - what's the problem of making multiple calls? This will help improve the PR title in the Changelog too 

2. Code comment on why we need to store last returned as a dict

3. Create an issue to add a E2E test after the release (non-blocking) so we can be confident there's no regression in the future

rubenfonseca:
Looking at this now

aws_lambda_powertools/utilities/parameters/appconfig.py
```diff
@@ -126,10 +127,14 @@ def _get(self, name: str, **sdk_options) -> str:
         return_value = response["Configuration"].read()
         self._next_token[name] = response["NextPollConfigurationToken"]
 
+        # The return of get_latest_configuration can be null because this value is supposed to be cached
+        # on the customer side.
+        # We created dict that stores the most recently retrieved value for a specific configuration.
```
rubenfonseca:
```suggestion
        # We created a dictionary that stores the most recently retrieved value for a specific configuration.
```

rubenfonseca (dismissed):
Just left one tiny thing, but it's ready to go!

aws_lambda_powertools/utilities/parameters/appconfig.py
leandrodamascena: Thanks Ruben! Done.

rubenfonseca (approved):

PR #3241, opened 2023-10-23
fix(parameters): Respect POWERTOOLS_PARAMETERS_SSM_DECRYPT environment variable when getting multiple ssm parameters.

### LINKED ISSUE #3240: Bug: POWERTOOLS_PARAMETERS_SSM_DECRYPT environment variable not respected for get multiple
### Expected Behaviour

When the environment variable `POWERTOOLS_PARAMETERS_SSM_DECRYPT` is set to `true`, a call to `get_multiple` parameters should automatically decrypt the values, as is the way with a call to `get`.

### Current Behaviour

The `POWERTOOLS_PARAMETERS_SSM_DECRYPT` is ignored when calling `get_multiple` and only the explicit `decrypt` argument is respected.

### Code snippet

```python
from aws_lambda_powertools.utilities.parameters.ssm import SSMProvider
import os

os.environ['POWERTOOLS_PARAMETERS_SSM_DECRYPT'] = 'true'
ssm_provider = SSMProvider()

for _, v in ssm_provider.get_multiple('/some/parameters').items():
    print(v) # Still encrypted
```

### Possible Solution

https://github.com/aws-powertools/powertools-lambda-python/pull/3241

### Steps to Reproduce

Please see code snippet

latest

3.10

### Packaging format used

PyPi

### Debugging logs

_No response_

sean-hernon:
When getting multiple ssm parameters by path, the `POWERTOOLS_PARAMETERS_SSM_DECRYPT` environment variable should be respected as to whether to automatically decrypt the values, as is with the single `get`.

**Issue number: 3240**

## Summary

Check the value of the `POWERTOOLS_PARAMETERS_SSM_DECRYPT` and respect that when deciding
whether to decrypt multiple parameters that were loaded.

### Changes

> The `get_multiple` method of the ssm provider will check the value of the environment variable, when deciding
whether to decrypt. The default remains false for the case where none exists, or no explicit parameter is passed.

### User experience

> The user can get multiple parameters, decrypted, by setting the environment variable.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/parameters/ssm.py
+++ b/aws_lambda_powertools/utilities/parameters/ssm.py
@@ -188,7 +188,7 @@ class SSMProvider(BaseProvider):
 
         return self.client.get_parameter(**sdk_options)["Parameter"]["Value"]
 
-    def _get_multiple(self, path: str, decrypt: bool = False, recursive: bool = False, **sdk_options) -> Dict[str, str]:
+    def _get_multiple(self, path: str, decrypt: Optional[bool] = None, recursive: bool = False, **sdk_options) -> Dict[str, str]:
         """
         Retrieve multiple parameter values from AWS Systems Manager Parameter Store
 
@@ -203,6 +203,12 @@ class SSMProvider(BaseProvider):
         sdk_options: dict, optional
             Dictionary of options that will be passed to the Parameter Store get_parameters_by_path API call
         """
+
+        # If decrypt is not set, resolve it from the environment variable, defaulting to False
+        decrypt = resolve_truthy_env_var_choice(
+            env=os.getenv(constants.PARAMETERS_SSM_DECRYPT_ENV, "false"),
+            choice=decrypt,
+        )
 
         # Explicit arguments will take precedence over keyword arguments
         sdk_options["Path"] = path
--- a/tests/functional/test_utilities_parameters.py
+++ b/tests/functional/test_utilities_parameters.py
@@ -960,6 +960,53 @@ def test_ssm_provider_get_sdk_options_overwrite(mock_name, mock_value, mock_vers
         stubber.deactivate()
 
 
+def test_ssm_provider_get_multiple_with_decrypt_environment_variable(monkeypatch, mock_name, mock_value, mock_version, config):
+    """
+    Test SSMProvider.get_multiple() with decrypt value replaced by environment variable
+    """
+
+    # Setting environment variable to override the default value
+    monkeypatch.setenv("POWERTOOLS_PARAMETERS_SSM_DECRYPT", "true")
+
+    mock_param_names = ["A", "B", "C"]
+
+    # Create a new provider
+    provider = parameters.SSMProvider(config=config)
+
+    # Stub the boto3 client
+    stubber = stub.Stubber(provider.client)
+    response = {
+        "Parameters": [
+            {
+                "Name": f"{mock_name}/{name}",
+                "Type": "String",
+                "Value": f"{mock_value}/{name}",
+                "Version": mock_version,
+                "Selector": f"{mock_name}/{name}:{mock_version}",
+                "SourceResult": "string",
+                "LastModifiedDate": datetime(2015, 1, 1),
+                "ARN": f"arn:aws:ssm:us-east-2:111122223333:parameter/{mock_name}/{name}",
+            }
+            for name in mock_param_names
+        ],
+    }
+    expected_params = {"Path": mock_name, "Recursive": False, "WithDecryption": True}
+    stubber.add_response("get_parameters_by_path", response, expected_params)
+    stubber.activate()
+
+    try:
+        values = provider.get_multiple(mock_name)
+
+        stubber.assert_no_pending_responses()
+
+        assert len(values) == len(mock_param_names)
+        for name in mock_param_names:
+            assert name in values
+            assert values[name] == f"{mock_value}/{name}"
+    finally:
+        stubber.deactivate()
+
+
 def test_ssm_provider_get_multiple(mock_name, mock_value, mock_version, config):
     """
     Test SSMProvider.get_multiple() with a non-cached path


aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -188,7 +188,7 @@ def _get(self, name: str, decrypt: bool = False, **sdk_options) -> str:
 
         return self.client.get_parameter(**sdk_options)["Parameter"]["Value"]
 
-    def _get_multiple(self, path: str, decrypt: bool = False, recursive: bool = False, **sdk_options) -> Dict[str, str]:
+    def _get_multiple(self, path: str, decrypt: Optional[bool] = None, recursive: bool = False, **sdk_options) -> Dict[str, str]:
```
leandrodamascena, line 191: I'm not sure we would change this method. I think that due to our error, we did not implement `get_multiple` in the `SSMProvider` class. We already implemented `get` and we should do the same for `get_multiple`. 

Makes more sense for us to put the `get_multiple` method in this class and do the appropriate checks for variables/env and others in it and then call the superclass.

It would be something more or less like this:

```python
def get_multiple( # type: ignore[override]
        self,
        path: str,
        max_age: Optional[int] = None,
        transform: TransformOptions = None,
        raise_on_transform_error: bool = False,
        decrypt: Optional[bool] = None,
        force_fetch: bool = False,
        **sdk_options,
    ) -> Union[Dict[str, str], Dict[str, dict], Dict[str, bytes]]:
        """
        Retrieve multiple parameters based on a path prefix

        Parameters
        ----------
        path: str
            Parameter path used to retrieve multiple parameters
        max_age: int, optional
            Maximum age of the cached value
        transform: str, optional
            Optional transformation of the parameter value. Supported values
            are "json" for JSON strings, "binary" for base 64 encoded
            values or "auto" which looks at the attribute key to determine the type.
        raise_on_transform_error: bool, optional
            Raises an exception if any transform fails, otherwise this will
            return a None value for each transform that failed
        force_fetch: bool, optional
            Force update even before a cached item has expired, defaults to False
        sdk_options: dict, optional
            Arguments that will be passed directly to the underlying API call

        Raises
        ------
        GetParameterError
            When the parameter provider fails to retrieve parameter values for
            a given path.
        TransformParameterError
            When the parameter provider fails to transform a parameter value.
        """
        # If max_age is not set, resolve it from the environment variable, defaulting to DEFAULT_MAX_AGE_SECS
        max_age = resolve_max_age(env=os.getenv(constants.PARAMETERS_MAX_AGE_ENV, DEFAULT_MAX_AGE_SECS), choice=max_age)

        # If decrypt is not set, resolve it from the environment variable, defaulting to False
        decrypt = resolve_truthy_env_var_choice(
            env=os.getenv(constants.PARAMETERS_SSM_DECRYPT_ENV, "false"),
            choice=decrypt,
        )

        # Add to `decrypt` sdk_options to we can have an explicit option for this
        sdk_options["decrypt"] = decrypt

        return super().get_multiple(path, max_age, transform, force_fetch, raise_on_transform_error, **sdk_options)
```

Make sense for you?

leandrodamascena (changes requested):
Hey @sean-hernon! I left a comment to change this PR.

aws_lambda_powertools/utilities/parameters/ssm.py
sean-hernon: Yes that makes sense, seems more in-line with the existing pattern for `get`.

Would you like me to make the change, or will you do it?

sean-hernon: `decrypt` is missing from the doc, btw

leandrodamascena: > Yes that makes sense, seems more in-line with the existing pattern for `get`.
> 
> Would you like me to make the change, or will you do it?

If you can do it, please do. :1st_place_medal:

leandrodamascena: > `decrypt` is missing from the doc, btw

Hmm where?

sean-hernon: In

```
Parameters
        ----------
        path: str
            Parameter path used to retrieve multiple parameters
        max_age: int, optional
            Maximum age of the cached value
        transform: str, optional
            Optional transformation of the parameter value. Supported values
            are "json" for JSON strings, "binary" for base 64 encoded
            values or "auto" which looks at the attribute key to determine the type.
        raise_on_transform_error: bool, optional
            Raises an exception if any transform fails, otherwise this will
            return a None value for each transform that failed
        force_fetch: bool, optional
            Force update even before a cached item has expired, defaults to False
        sdk_options: dict, optional
            Arguments that will be passed directly to the underlying API call
```

but as I am making the change, I will add it.

sean-hernon: In `get` we have `decrypt` before `force_fetch`, which breaks liskov substitution for people using the arguments positionally and requiring the base class.

The change that you've suggested above does likewise. I can fix this by putting `decrypt` at the end, before `sdk_options`. Should I do this, or keep it consistent with `get` (breaking liskov substitution)?

leandrodamascena: I think we should keep consistence with `get` because people can easily switch between `ssm_provider.get("/leandro/a", decrypt=True)` and `ssm_provider.get_multipe("/leandro/", decrypt=True)`.

please add a note in the comment to review this in Powertools V3, ok?

sean-hernon: As the method in `base` is not currently overridden, this will be a breaking change for those consumers using the method arguments positionally. Should I proceed?

A non-breaking alternative (less explicit, though), could be to override the method and check `sdk_options['decrypt']`.

Perhaps it's better to do this and leave a note that adding the extra argument should be reviewed in v3. That way we keep backwards-compatibility with anyone not using named arguments.

leandrodamascena: Hi @sean-hernon! I can't thank you enough for investigating as many scenarios as possible and bringing this to the table. In fact, I was testing the code here and any scenario we implement `get_multiple` will be a potential breaking change. So for now let's do the following:

1 - Keep the implementation simple as you proposed.
2 - I'll create a new issue so we can fix this in Powertools v3.

So, if you don't have any other considerations for now, we can approve this PR.

sean-hernon: @leandrodamascena I'm a little confused now. Is the proposal to create a `get_multiple` which overrides the `base`, but read from `sdk_options['decrypt']`, or, to keep the implementation as it already is in the PR?

leandrodamascena: Sorry if I wasn´t clear, we can keep the implementation as it already is in this PR.

sean-hernon: Got it. 👍 

Should I update any comments?

sean-hernon: As far as other considerations, I think it's fine. The `_get_multiple` here is only called in `base`, which is passing **kwargs (and has no understanding of the shape of `decrypt`), so I don't think the minor change in signature is important.

leandrodamascena (approved):
LGTM!

aws_lambda_powertools/utilities/parameters/ssm.py
leandrodamascena: Yeah, in this case this is not a problem at all. But in V3 we need to implement `get_multiple` in the SSM class.

sean-hernon:
> LGTM!

Thank you!

PR #3254, opened 2023-10-25
feat(event_handler): add ability to expose a Swagger UI

### LINKED ISSUE #3253: Feature request: Add ability to expose a Swagger UI on the event handler
### Use case

This is a continuation of the work on the RFC #2421

Now that we've merged the code to add OpenAPI generation and runtime validation, we want to optionally enable a Swagger UI so the consumers can explore the API provided by the Lambda function.

**NOTE**: this ticket is not yet for customers using micro-functions. This is only useful for customers with a single Lambda with many routes. 

### Solution/User Experience

See [here](https://github.com/aws-powertools/powertools-lambda-python/issues/2421#issuecomment-1757645483) for the discussion.

We should create a new `enable_swagger` method on all the API resolvers that will create all the necessary endpoints to server the Swagger UI. Optionally the user can pass `middlewares` if additional security / authorization is needed.

```python
app.enable_swagger(path="/swagger")
```

Additional features include:
- ability to override the Swagger base path, so other versions can be saved
- ability to customize the OpenAPI metadata
- automatic detection of the base path for Rest/Http gateways

### Alternative solutions

```markdown
N/A
```

### Acknowledgment

rubenfonseca:
**Issue number:** #3253

### Changes

This PR adds the ability to expose a Swagger UI from the AWS Lambda. It works with all supported REST gateways.

In addition:
- it adds support for omitting routes from the OpenAPI spec (`include_in_schema`)
- adds a lot of coverage
- fixes some bugs encountered with improving coverage

### User experience

```python

logger = Logger()
app.enable_swagger(path="/swagger")


@app.get("/hello_get")
def get_openapi():
    return "hello_get"

@app.post("/hello_post")
def post_openapi():
    return "hello_post"

# You can continue to use other utilities just as before
@logger.inject_lambda_context(correlation_id_path=correlation_paths.API_GATEWAY_REST, log_event=True)

```

image

## Checklist

- [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
             indent=2,
         )
 
+    def enable_swagger(
+        self,
+        *,
+        path: str = "/swagger",
+        title: str = "Powertools API",
+        version: str = "1.0.0",
+        openapi_version: str = "3.1.0",
```
leandrodamascena: We are using this "3.1.0" default version at least in 2 methods, what do you think about creating a constant?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
             indent=2,
         )
 
+    def enable_swagger(
+        self,
+        *,
+        path: str = "/swagger",
+        title: str = "Powertools API",
+        version: str = "1.0.0",
```
leandrodamascena: We are using this "1.0.0" default version at least in 2 methods, what do you think about creating a constant?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
             indent=2,
         )
 
+    def enable_swagger(
+        self,
+        *,
+        path: str = "/swagger",
+        title: str = "Powertools API",
```
leandrodamascena: Powertools for AWS Lambda (Python)?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
...
+    ],
+    plugins: [
+      SwaggerUIBundle.plugins.DownloadUrl
+    ]
+  }}
+
+  var ui = SwaggerUIBundle(swaggerUIOptions)
+</script>
+</html>
+            """.strip()
```
leandrodamascena: HTML here makes me sad. I know it has to be done, but do you think it's worth moving this function to a file called `swagger.py` and putting it there? Just an idea.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1606,6 +1779,9 @@ def _create_route_key(self, item: str, rule: str):
             )
         self._route_keys.append(route_key)
 
+    def _get_base_path(self) -> str:
+        raise NotImplementedError()
+
```
leandrodamascena, line 1745: We use the same logic for `APIGatewayRestResolver`, `APIGatewayHttpResolver` and `LambdaFunctionUrlResolver`. Why don't we implement the codebase here and just override this method in VpcLattic/AlbResolver?

We can use this code here:

```python
def _get_base_path(self) -> str:
    stage = stage = self.current_event.request_context.stage
    if stage != "$default":
        return stage[len("/" + stage) :]
    return "/"
```

The reason we need to add this is because we can provide the local experience using SAM and SAM doesn't add the stage locally.

Reference: https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/data_classes/api_gateway_proxy_event.py#L251

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
...
+        version: str = "1.0.0",
+        openapi_version: str = "3.1.0",
+        summary: Optional[str] = None,
+        description: Optional[str] = None,
+        tags: Optional[List[str]] = None,
+        servers: Optional[List["Server"]] = None,
+        terms_of_service: Optional[str] = None,
+        contact: Optional["Contact"] = None,
+        license_info: Optional["License"] = None,
+        swagger_base_url: Optional[str] = None,
```
leandrodamascena, line 1536:
```suggestion
        base_url: Optional[str] = None,
```

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
...
+        if not swagger_base_url:
+
+            @self.get("/swagger.js", include_in_schema=False)
+            def swagger_js():
+                body = Path.open(Path(__file__).parent / "openapi" / "swagger_ui" / "swagger-ui-bundle.min.js").read()
+                return Response(
+                    status_code=200,
+                    content_type="text/javascript",
+                    body=body,
+                    compress=True,
```
leandrodamascena: I think we need to add a parameter if the customer wants to compress or not. Forcing this we can have issues with `net :: ERR_CONTENT_DECODING_FAILED 200 (OK)` when running locally.


image

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
...
+        swagger_base_url: str, optional
+            The base url for the swagger UI. If not provided, we will serve a recent version of the Swagger UI.
+        middlewares: List[Callable[..., Response]], optional
+            List of middlewares to be used for the swagger route.
+        """
+        from aws_lambda_powertools.event_handler.openapi.models import Server
+
+        if not swagger_base_url:
+
+            @self.get("/swagger.js", include_in_schema=False)
```
leandrodamascena, line 1575: We need to remember to add in our documentation that the customer will invoke Lambda 3 times each time it renders Swagger:

1x to serve the HTML
1x to serve JS
1x to serve the css

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
...
+                )
+
+            @self.get("/swagger.css", include_in_schema=False)
+            def swagger_css():
+                body = Path.open(Path(__file__).parent / "openapi" / "swagger_ui" / "swagger-ui.min.css").read()
+                return Response(
+                    status_code=200,
+                    content_type="text/css",
+                    body=body,
+                    compress=True,
```
leandrodamascena: I think we need to add a parameter if the customer wants to compress or not. Forcing this we can have issues with net :: ERR_CONTENT_DECODING_FAILED 200 (OK) when running locally.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1492,6 +1517,152 @@ def get_openapi_json_schema(
...
+  var ui = SwaggerUIBundle(swaggerUIOptions)
+</script>
+</html>
+            """.strip()
+
+            return Response(
+                status_code=200,
+                content_type="text/html",
+                body=body,
+                compress=True,
```
leandrodamascena: I think we need to add a parameter if the customer wants to compress or not. Forcing this we can have issues with net :: ERR_CONTENT_DECODING_FAILED 200 (OK) when running locally.

leandrodamascena (changes requested):
I left some comments that we need to change to improve the developer experience and fix small issues.

This is incredible and will work perfectly even locally after these changes.

image

image

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: Time to open a bug report in SAM?

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: Time to open a bug report in SAM?

leandrodamascena: https://github.com/aws/aws-sam-cli/issues/5860

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: I don't believe this is a good change. `base_url` could pass the idea that we're going to prefix the swagger URL with something. But what we're doing here is accepting the swagger JS and CSS base URL, hence I prefixed it with `swagger`. Does it make sense or would you prefer something different?

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: Good idea, moved to a new file

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: Two things:

1. I can not move the code to the base class, because the `current_event` of the ApiGatewayResolver doesn't know about `request_context`, only its subclasses

2. I don't understand what your code is doing differently than mine. Care to explain with some example?

rubenfonseca:
@leandrodamascena addressed your comments, answered the pending questions and re-requested a review

heitorlessa:
Quick one for now: Add the constraint to the linked issue that this is not yet for customers using micro-functions. This is only useful for customers with a single Lambda with many routes.

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: Ok ok, I agree with your point about the confusion between names. Lets keep the `swagger_base_url` name.

guywilsonjr:
I'm currently managing a combination of a hodge podge of solutions to achieve this functionality and this feature would be greatly appreciated by my team :)

@leandrodamascena respectfully, are there any remaining issues here?

leandrodamascena:
> I'm currently managing a combination of a hodge podge of solutions to achieve this functionality and this feature would be greatly appreciated by my team :)
> 
> @leandrodamascena respectfully, are there any remaining issues here?

Hi @guywilsonjr! It's amazing to know that this feature will make your life easier when working with OpenAPI Spec + Swagger. There are no critical issues here and we hope to merge this PR early next week and release it soon.

leandrodamascena (approved):
Just a final comment:

Running this new code with `enable_validation=False`, means we don´t bring Pydantic.

image


Running this new code with `enable_validation=True`, which means we bring in Pydantic+ assets but have no impact because of the assets, only because Pydantic, what is expected.

image

leandrodamascena:
APPROVED @rubenfonseca! :rocket:

PR #3302, opened 2023-11-06
fix(event_handler): Router prefix mismatch regression after Middleware feat

### LINKED ISSUE #3255: Bug: Extended middleware not being called after assigned to route.
### Expected Behaviour

After registering middleware to route, once endpoint is called, registered route is called before endpoint function.

### Current Behaviour

Skips registered middleware on route and calls funvtion

### Code snippet

```python
# middlewares.py

class UserMiddleWare(BaseMiddlewareHandler):
    def __init__(self, allowed_roles: list[UserRoles], authenticate: bool = True):
        super().__init__()
        self._allowed_roles = allowed_roles
        self._authenticate = authenticate

    def handler(self, app: APIGatewayRestResolver, next_middleware: NextMiddleware) -> Response:
        ...
        return next_middleware(app)

# users.py

...
from aws_lambda_powertools.event_handler.router import APIGatewayRouter

router = APIGatewayRouter()


@router.get("/me", cors=True, middlewares=[UserMiddleWare(allowed_roles=[UserRoles.INFLUENCER], authenticate=False)])
def get_user():
    user: UserDict = router.context["user"]
    ...

# main.py
...

app = APIGatewayRestResolver(
    strip_prefixes=["/xxx"], debug=get_settings().STAGE in ["staging", "dev"], cors=cors_config
)


def log_request_response(app: APIGatewayRestResolver, next_middleware: NextMiddleware) -> Response:
    logger.info("Incoming request", method=app.current_event.http_method, path=app.current_event.path)
    logger.append_keys(
        user_id=app.current_event.request_context.authorizer["id"],
        role=app.current_event.request_context.authorizer["role"],
    )
    result = next_middleware(app)
    logger.info("Response received", response=result.__dict__)
    return result


app.use(
    middlewares=[
        log_request_response,
    ]
)

app.include_router(
    users.router,
    prefix="/users",
)

@logger.inject_lambda_context(correlation_id_path=correlation_paths.API_GATEWAY_REST)
@event_source(data_class=APIGatewayProxyEvent)
def wallets_gateway_handler(event: APIGatewayProxyEvent, context: LambdaContext):
    tracer.put_annotation(
        "user", event.request_context.authorizer["role"] + str(event.request_context.authorizer["id"])
    )
```

### Possible Solution

_No response_

### Steps to Reproduce

- Create app
- Create middleware
- Create route in separate file and attach middleware to specific endpoint without using `router.use()`
- Include router to app with prefix

latest

3.11

### Packaging format used

PyPi

### Debugging logs

```python
Processing Route:::get_user (/users/me)

Middleware Stack:
=================
log_request_response
_registered_api_adapter
=================
```

leandrodamascena:
**Issue number:** #3255 

### Changes

When using a route with a prefix the Middleware execution was failing.

### User experience

There are no changes to the user experience other than fixing the bug.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
+        (APIGatewayRestResolver(), API_REST_EVENT),
+        (APIGatewayHttpResolver(), API_RESTV2_EVENT),
+    ],
+)
+def test_api_gateway_middleware_with_include_router_prefix(app: EventHandlerInstance, event):
+    # GIVEN a router instance
+    router = Router()
+
+    def app_middleware(app: EventHandlerInstance, next_middleware: NextMiddleware):
+        # inject a variable into resolver context
```
heitorlessa:
```suggestion
        # AND a variable injected into resolver context
```

this makes it easier to quickly parse the test intent when reading top to bottom

tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
 
 
+@pytest.mark.parametrize(
+    "app, event",
+    [
+        (ApiGatewayResolver(proxy_type=ProxyEventType.APIGatewayProxyEvent), API_REST_EVENT),
+        (APIGatewayRestResolver(), API_REST_EVENT),
+        (APIGatewayHttpResolver(), API_RESTV2_EVENT),
+    ],
+)
```
heitorlessa: Use a single Event/Resolver, we only left the parametrize one as we ran out of time. It makes it difficult to read the test.

```suggestion
```

tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
+def test_api_gateway_middleware_with_include_router_prefix(app: EventHandlerInstance, event):
+    # GIVEN a router instance
+    router = Router()
```
heitorlessa:
```suggestion
def test_api_gateway_middleware_with_include_router_prefix():
    # GIVEN an App and Router instance
    app = APIGatewayRestResolver()
    router = Router()
```

tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
+
+    def app_middleware(app: EventHandlerInstance, next_middleware: NextMiddleware):
+        # inject a variable into resolver context
+        app.append_context(injected="injected_value")
+        response = next_middleware(app)
+
+        return response
```
heitorlessa:
```suggestion
        return next_middleware(app)
```

suggestion: we can shorten the test since we're not altering the response.

tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
+
+    # WHEN we register a route with a middleware
+    @router.get("/path", middlewares=[app_middleware])
+    def dummy_route():
+        # if the middleware is executed, the context variable will be available
+        # Otherwise, an exception will be raised
```
heitorlessa:
```suggestion
        # THEN we should have access to the middleware's injected variable 
```

tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
+        injected_context = app.context["injected"]
+
+        assert injected_context == "injected_value"
```
heitorlessa:
```suggestion
        assert app.context["injected"] == "injected_value"
```

suggestion: this will now fail with KeyError Exception instead AssertionError, making it more explicit the behaviour we're testing and easier to identify why that's the case (back to the comment being THEN too).

Mental model is that the `app.context[...]` is the key we're supposedly injecting. If the key isn't injected, then the exception alone will tell us the problem.

tests/functional/event_handler/test_api_middlewares.py
```diff
@@ -397,6 +397,46 @@ def dummy_route():
...
+
+        return Response(status_code=200, body="works!")
+
+    # WHEN register the route with a prefix
+    app.include_router(router, prefix="/my")
+
+    # THEN resolving a request must execute the middleware
+    # THEN return a successful response http 200 status code
```
heitorlessa:
```suggestion
    # THEN resolving a request must execute the middleware
    # and return a successful response http 200 status code
```

suggestion(non-blocking): makes it easier to parse that we're asserting two behaviours only: 1/ Route runs successfully (non-404), and 2/ Middleware runs when a prefix is in place

heitorlessa (changes requested):
GREAT catch! Made suggestions to clean up the test. We ran out of time back then to get rid of parametrize, so we can do better now.

One additional suggestion I couldn't due to the UI limiting it... is to add a comment in the for loop block so we don't lose that much time figuring out that logic..

https://github.com/aws-powertools/powertools-lambda-python/pull/3302/files#diff-bdb4b43087f89fdf398381f3d88048f6182562ca67e8907259d5c56d658465c1R2012

It isn't immediately clear this block: https://github.com/aws-powertools/powertools-lambda-python/pull/3302/files#diff-bdb4b43087f89fdf398381f3d88048f6182562ca67e8907259d5c56d658465c1R2012

heitorlessa:
There's another test we're missing to catch a regression like this --- we don't have a test where we have both strip_prefix + router prefix. 

Could you add that one in too, please?

tests/functional/event_handler/test_api_middlewares.py
leandrodamascena: Make sense! I removed the parameters and created the app instance inside the test.

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -1197,6 +1197,26 @@ def base():
...
+    def base():
+        return {}
+
+    # THEN we can register the router with a prefix
+    app.include_router(router, prefix="/bar")
+
+    result = app(event, {})
+
+    # THEN process event correctly
+    assert result["statusCode"] == 200
```
heitorlessa:
```suggestion
    # GIVEN all routes are stripped from its version e.g., /v1

    app = ApiGatewayResolver(strip_prefixes=["/v1"])
    router = Router()

    event = {"httpMethod": "GET", "path": "/v1/users/leandro", "resource": "/users"}

    @router.get("<user_id>")
    def base(user_id: str):
        return {"user": user_id}

    # WHEN a router is included prefixing all routes with "/users"
    app.include_router(router, prefix="/users")

    result = app(event, {})

    # THEN route correctly to the registered route after stripping each prefix (global + router)
    assert result["statusCode"] == 200
```

Two changes to make it easier to document its behaviour we're testing

1. Paths to be more realistic using API versioning `/v1` and service specific routes that are often split `/users`
2. Adjust GIVEN WHEN THEN to reflect this accordingly

heitorlessa (changes requested):
tiny change in the latest test to make it easier to go back in the future.

Can be merged after that 🎉 !!!!!!

tests/functional/event_handler/test_api_gateway.py
leandrodamascena: Yeah! Thanks for the feedback :heart:

heitorlessa (approved):

heitorlessa:
GREEEEEAAAAT JOB.

That was a pesky regression to detect

PR #3392, opened 2023-11-22
fix(event_handler): apply serialization as the last operation for middlewares

### LINKED ISSUE #3391: Bug: JSON serialization should be the last operation before returning a response from an event handler
### Expected Behaviour

Middleware writers should access the original return of the handler, before serialization happens.

### Current Behaviour

When writing your own middleware, you only get access to a serialized version of the handler response, and not the original. This introduces challenges when you want to manipulate the original response.

### Code snippet

```python
class MyMiddleware(BaseMiddlewareHandler):
    def handler(self, app: EventHandlerInstance, next_middleware: NextMiddleware) -> Response:
        # Call the handler by calling the next middleware
        response = next_middleware(app)

        response.body # this should be the original response, not a serialized version
```

### Possible Solution

Move the serialization as the last operation to be done before returning, after running all the middlewares.

### Steps to Reproduce

See code snippet above

latest

3.7

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3391

### Changes

This PR moves the serialization on event handler to the very last step before returning to the client. This allows all middlewares to see the original returns from the handler.

The serialization is also only applied if the content type of the response is JSON. This means we don't try to apply the serialization to HTML or PLAIN responses.

### User experience

After this PR, a user can again use a custom serializer, even when using data validation.

## Checklist

- [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -783,7 +783,10 @@ class ResponseBuilder(Generic[ResponseEventT]):
             logger.debug("Encoding bytes response with base64")
             self.response.base64_encoded = True
             self.response.body = base64.b64encode(self.response.body).decode()
-        elif self.response.is_json():
+
+        # We only apply the serializer when the content type is JSON and the
+        # body is not a str, to avoid double encoding
+        elif self.response.is_json() and not isinstance(self.response.body, str):
             self.response.body = self.serializer(self.response.body)
 
         return {
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -702,7 +702,7 @@ class ResponseBuilder(Generic[ResponseEventT]):
     def __init__(
         self,
         response: Response,
-        serializer: Optional[Callable[[Any], str]] = json.dumps,
+        serializer: Callable[[Any], str] = json.dumps,
         route: Optional[Route] = None,
     ):
         self.response = response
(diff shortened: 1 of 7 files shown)


aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -782,6 +783,8 @@ def build(self, event: ResponseEventT, cors: Optional[CORSConfig] = None) -> Dic
             logger.debug("Encoding bytes response with base64")
             self.response.base64_encoded = True
             self.response.body = base64.b64encode(self.response.body).decode()
+        elif self.response.is_json():
+            self.response.body = self.serializer(self.response.body)
```
heitorlessa, line 795:
```suggestion
            self.response.body = body if isinstance(self.response.body, str) else self.serializer(self.response.body)
```

suggestion so we can revert tests and don't accidentally double encode a JSON that would break a front-end

heitorlessa (approved):
pending CI checks

tests/functional/event_handler/test_api_gateway.py
```diff
@@ -681,7 +681,7 @@ def another_one():
 def test_no_content_response():
     # GIVEN a response with no content-type or body
     response = Response(status_code=204, content_type=None, body=None, headers=None)
-    response_builder = ResponseBuilder(response)
+    response_builder = ResponseBuilder(response, serializer=json.dumps)
```
heitorlessa: can we make serializer optional so our tests remain intact?

Hyrum law... we dunno if any "crazy" person ended up taking dependency on an internal response builder to create their own framework atop.

heitorlessa (changes requested):
tiny change to prevent a possible hyrum law case of an unknown internal dependency

heitorlessa (approved):

PR #3394, opened 2023-11-22
fix(event_handler): allow fine grained Response with data validation

### LINKED ISSUE #3393: Bug: Data validation doesn't support the return of Response objects
### Expected Behaviour

Even though I'm using data validation, I would like to still be able to return `Response` objects, which means I can still use all the other features of the event handler, including compression, headers, status codes, etc.

### Current Behaviour

Right now, I can't return a `Response` object from my API. To do so, I would have to change the method signature, which would defeat the data validation middleware.

### Code snippet

```python
class Model(BaseModel):
    name: str
    age: int

# WHEN a handler is defined with a body parameter
@app.post("/")
def handler(user: Model) -> Model:
    return user # can't return anything else
```

### Possible Solution

```python
class Model(BaseModel):
    name: str
    age: int

@app.post("/")
def handler(user: Model) -> Response[Model]:
    return Response(body=user, status_code=200)
```

Where Response becomes a generic class, and mypy will validate that it matches the type in the `body`.

### Steps to Reproduce

Check the code snippet above.

latest

3.7

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3393

### Changes

This PR adds support for returning Response objects when using data validation.

### User experience

This is what it looks like:

```python
class User(BaseModel):
    name: str
    age: int

@app.post("/")
def handler(user: User) -> Response[User]:
    return Response(body=user, status_code=200)
```

## Checklist

- [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/tests/functional/event_handler/test_openapi_validation_middleware.py
+++ b/tests/functional/event_handler/test_openapi_validation_middleware.py
@@ -6,7 +6,7 @@ from typing import List, Tuple
 
 from pydantic import BaseModel
 
-from aws_lambda_powertools.event_handler import APIGatewayRestResolver
+from aws_lambda_powertools.event_handler import APIGatewayRestResolver, Response
 from aws_lambda_powertools.event_handler.openapi.params import Body
 from aws_lambda_powertools.shared.types import Annotated
 from tests.functional.utils import load_event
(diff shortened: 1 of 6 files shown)


tests/functional/event_handler/test_openapi_validation_middleware.py
```diff
@@ -330,3 +330,33 @@ def handler(user: Annotated[Model, Body(embed=True)]) -> Model:
...
+        age: int
+
+    # WHEN a handler is defined with a body parameter
+    @app.post("/")
+    def handler(user: Annotated[Model, Body(embed=True)]) -> Response[Model]:
+        return Response(body=user, status_code=200)
+
+    LOAD_GW_EVENT["httpMethod"] = "POST"
+    LOAD_GW_EVENT["path"] = "/"
+    LOAD_GW_EVENT["body"] = json.dumps({"name": "John", "age": 30})
```
heitorlessa, line 350: for later.. but mutating this could cause problems when running tests in parallel (we don't **yet**).

tests/functional/event_handler/test_openapi_validation_middleware.py
```diff
@@ -330,3 +330,33 @@ def handler(user: Annotated[Model, Body(embed=True)]) -> Model:
...
+    # THEN the body must be a dict
+    result = app(LOAD_GW_EVENT, {})
+    assert result["statusCode"] == 422
+    assert "missing" in result["body"]
+
+    # THEN the handler should be invoked and return 200
+    # THEN the body must be a dict
+    LOAD_GW_EVENT["body"] = json.dumps({"user": {"name": "John", "age": 30}})
+    result = app(LOAD_GW_EVENT, {})
+    assert result["statusCode"] == 200
```
heitorlessa: can we split this into a separate test?

heitorlessa (changes requested):
split tests; all great otherwise!

heitorlessa (approved):

PR #3397, opened 2023-11-22
fix(event_handler): lazy load Pydantic to improve cold start

### LINKED ISSUE #3370: Bug: Pydantic is being imported when using Event Handler without validation
### Expected Behaviour

I'm using `ApiGatewayResolver` without enabling validation and it shouldn't import Pydantic. Even if a customer has Pydantic installed, they should not import Pydantic without explicitly enabling validation: `app = ApiGatewayResolver(enable_validation=True)`

### Current Behaviour

Profiling the application with tuna, I can see Pydantic being imported:

image

### Code snippet

```python
from aws_lambda_powertools.event_handler import ApiGatewayResolver

app = ApiGatewayResolver(enable_validation=False)

requirements.txt
aws-lambda-powertools==2.27.0
aws-xray-sdk==2.12.1
botocore==1.32.3
fastjsonschema==2.19.0
jmespath==1.0.1
pydantic==1.10.13
python-dateutil==2.8.2
six==1.16.0
tuna==0.5.11
typing_extensions==4.8.0
urllib3==2.0.7
wrapt==1.16.0
```

### Possible Solution

Move [these imports](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/api_gateway.py#L36) to an standalone file or import them only when enable_validation is True.

### Steps to Reproduce

1 - Create a file
2 - pip install aws-lambda-powertools[all] tuna
3 - Use the code snippet provided
4 - python -X importtime test.py 2> prof1.log
5 - tuna prof1.log  

latest

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3370

### Changes

This PR fixes the problem where Pydantic was being loaded when using the event handler, even though no data validation was being used.

### User experience

#### BEFORE

image

#### AFTER

image

## Checklist

- [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/openapi/types.py
+++ b/aws_lambda_powertools/event_handler/openapi/types.py
@@ -16,13 +16,6 @@ COMPONENT_REF_PREFIX = "#/components/schemas/"
 COMPONENT_REF_TEMPLATE = "#/components/schemas/{model}"
 METHODS_WITH_BODY = {"GET", "HEAD", "POST", "PUT", "DELETE", "PATCH"}
 
-try:
-    from pydantic.version import VERSION as PYDANTIC_VERSION
-
-    PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
-except ImportError:
-    PYDANTIC_V2 = False
-
 
 validation_error_definition = {
     "title": "ValidationError",
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -790,6 +790,8 @@ class ResponseBuilder(Generic[ResponseEventT]):
             logger.debug("Encoding bytes response with base64")
             self.response.base64_encoded = True
             self.response.body = base64.b64encode(self.response.body).decode()
+        elif self.response.is_json():
+            self.response.body = self.serializer(self.response.body)
 
         # We only apply the serializer when the content type is JSON and the
         # body is not a str, to avoid double encoding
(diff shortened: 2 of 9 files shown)


aws_lambda_powertools/event_handler/openapi/types.py
```diff
@@ -16,13 +16,6 @@
 COMPONENT_REF_TEMPLATE = "#/components/schemas/{model}"
 METHODS_WITH_BODY = {"GET", "HEAD", "POST", "PUT", "DELETE", "PATCH"}
 
-try:
-    from pydantic.version import VERSION as PYDANTIC_VERSION
-
-    PYDANTIC_V2 = PYDANTIC_VERSION.startswith("2.")
-except ImportError:
-    PYDANTIC_V2 = False
-
```
heitorlessa, line 25: for my own understanding... the reason this "lazy" works is that:

* Event Handler entry point is importing `openapi/types.py`
* We were importing Pydantic to confirm it was present at the global scope
* This change moves the try/catch to a separate module which is not requested by `openapi/types.py`
* When `enable_validation=True` is used, we effectively import the rest which will import the new module that imports Pydantic to confirm is installed or not

image


image

heitorlessa (approved):

heitorlessa:
for my own understanding in the future... the reason this "lazy" works is that:

* Event Handler entry point is importing `openapi/types.py`
* We were importing Pydantic to confirm it was present at the global scope
* This change moves the try/catch to a separate module which is not requested by `openapi/types.py`
* When `enable_validation=True` is used, we effectively import the rest which will import the new module that imports Pydantic to confirm is installed or not

image


image

PR #3446, opened 2023-12-04
feat(idempotency): leverage new DynamoDB Failed conditional writes behavior with ReturnValuesOnConditionCheckFailure

### LINKED ISSUE #3327: Feature request: Leverage new DynamoDB Failed Conditional Writes behavior
### Use case

The new idempotency utility uses conditional writes against DynamoDB for its locking mechanism.

When the conditional write fails, the record is considered locked and the request is already processed. In that case, a query is issued against DynamoDB to fetch the cached result and return it.

This means for this scenario that customers must wait and pay for [both a write and read](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/WorkingWithItems.html#WorkingWithItems.ConditionalWrites.ReturnConsumedCapacity), even though the write fails.

On June 30, 2023, [DynamoDB announced](https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-dynamodb-cost-failed-conditional-writes/) the ability for a failed conditional write to return a copy of the existing record. This would allow the package to skip performing the query, improving performance and lowering cost.

### Solution/User Experience

In the idempotency utility, use the new [DynamoDB conditional write behavior](https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-dynamodb-cost-failed-conditional-writes/) to return the current state of the item in the same conditional write operation. Remove the follow-up query.

### Alternative solutions

_No response_

### Acknowledgment

dastra:
**Issue number:** #3327

### Changes

This feature passes ReturnValuesOnConditionCheckFailure when putting the idempotency record in DynamoDB.  If the put fails, then a copy of the item as it was during the failed write attempt is returned in the response.  This avoids having to perform a separate get request to retrieve the item.

Detail:
* ReturnValuesOnConditionCheckFailure is only supported in boto3 >=1.26.164.  As the [default version of boto3](https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html) installed within lambda for pythons 3.7-3.11 is 1.26.90, I've made the new functionality conditional on a library version check rather than requiring an upgrade of boto3
* aws_lambda_powertools.utilities.idempotency.persistence.dynamodb._put_record now checks whether boto3's version is >=1.26.164, and if so passes ReturnValuesOnConditionCheckFailure.  If the conditional check fails, then it adds the item returned to the IdempotencyItemAlreadyExistsError exception before raising it.
* aws_lambda_powertools.utilities.idempotency.base._process_idempotency extracts the item from the exception.  Where it doesn't exist, it calls _get_idempotency_record as before to get the item.
* These changes meant that I had to move the DataRecord class into its utilities/idempotency/persistence/datarecord.py to avoid a circular dependency, as it's now imported into aws_lambda_powertools.utilities.idempotency.exceptions
* I've updated the documentation to reflect that there's no RCU consumed.  To keep the docs simple, I've not added the detail that an RCU is consumed when boto3 < 1.26.164.  Happy to add this detail if you think it's worth this level of clarity.  

### User experience

The user experience is identical before and after the change.  However as there is no subsequent get when the conditional check fails, there is no associated RCU consumed.

## Checklist

* [X] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
hey @dastra stellar work - setting expectation that we might be able to review and give you feedback next week only; we're catching up to the re:Invent backlog as the team was out.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -234,16 +235,51 @@ def _put_record(self, data_record: DataRecord) -> None:
...
+        version = boto3_version.split(".")
+        # Only supported in boto3 1.26.164 and above
+        if len(version) >= 3 and int(version[0]) == 1 and int(version[1]) == 26 and int(version[2]) >= 164:
+            return True
+        if len(version) >= 2 and int(version[0]) == 1 and int(version[1]) > 26:
+            return True
+        if int(version[0]) > 1:
+            return True
+
+        return False
```
leandrodamascena:
```suggestion
    @staticmethod
    def _boto3_supports_condition_check_failure() -> bool:
        """
        Check if the installed boto3 version supports condition check failure.

        Returns
        -------
        bool 
            True if the boto3 version supports condition check failure, False otherwise.
        """
        major, minor, patch = map(int, (boto3.__version__.split('.')))

        # Condition check failure support introduced in boto3 1.26.164
        if (major, minor) >= (1, 26) and (minor, patch) >= (26, 164):
            return True

        return False
```

We may remove the parameter and get the boto3 directly.

aws_lambda_powertools/utilities/idempotency/persistence/base.py
```diff
@@ -260,6 +182,27 @@ def _validate_payload(self, data: Dict[str, Any], data_record: DataRecord) -> No
...
+
+        Raises
+        ----------
+        IdempotencyValidationError
+            Payload doesn't match the stored record for the given idempotency key
+
+        """
+        if self.payload_validation_enabled:
+            if old_data_record.payload_hash != data_record.payload_hash:
+                raise IdempotencyValidationError("Hashed payload does not match stored record for this event key")
```
leandrodamascena: We already have a [function to validate](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L241) the payload and raise an exception in case the hashed payload does not match the one that is stored. Do you think we can modify the existing function to accept more types/arguments instead of creating a new one?

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -234,16 +235,51 @@ def _put_record(self, data_record: DataRecord) -> None:
...
             if error_code == "ConditionalCheckFailedException":
+                old_data_record = self._item_to_data_record(exc.response["Item"]) if "Item" in exc.response else None
+                if old_data_record is not None:
+                    logger.debug(
+                        f"Failed to put record for already existing idempotency key: "
+                        f"{data_record.idempotency_key} with status: {old_data_record.status}, "
+                        f"expiry_timestamp: {old_data_record.expiry_timestamp}, "
+                        f"and in_progress_expiry_timestamp: {old_data_record.in_progress_expiry_timestamp}",
+                    )
+                    self._save_to_cache(data_record=old_data_record)
```
leandrodamascena, line 262: Great catch! Since we are no longer calling [get_record](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L421) in case of returning from an exception, we would lose the ability to cache without it.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -234,16 +235,51 @@ def _put_record(self, data_record: DataRecord) -> None:
...
+
+                    try:
+                        self._validate_hashed_payload(old_data_record=old_data_record, data_record=data_record)
+                    except IdempotencyValidationError as ive:
+                        raise ive from exc
```
leandrodamascena: Maybe a more explicit variable name than `ive`?

docs/utilities/idempotency.md
```diff
@@ -100,8 +100,8 @@ If you're not [changing the default configuration for the DynamoDB persistence l
 
 ???+ info "Info: DynamoDB"
     Each function invocation will generally make 2 requests to DynamoDB. If the
-    result returned by your Lambda is less than 1kb, you can expect 2 WCUs per invocation. For retried invocations, you will
-    see 1WCU and 1RCU. Review the [DynamoDB pricing documentation](https://aws.amazon.com/dynamodb/pricing/){target="_blank"} to
+    result returned by your Lambda is less than 1kb, you can expect 2 WCUs per invocation.
+    Review the [DynamoDB pricing documentation](https://aws.amazon.com/dynamodb/pricing/){target="_blank"} to
     estimate the cost.
```
leandrodamascena: We need to improve the documentation to show the new behavior, and we can even add an X-Ray image to show how the client will see it when the tracker is enabled.

But I can do it and we review it, what do you think?

image

aws_lambda_powertools/utilities/idempotency/exceptions.py
```diff
@@ -30,6 +32,20 @@ class IdempotencyItemAlreadyExistsError(BaseError):
...
+        self.old_data_record = old_data_record
+
+    def __str__(self):
+        """
+        Return all arguments formatted or original message
+        """
+        old_data_record = f" from [{(str(self.old_data_record))}]" if self.old_data_record else ""
+        details = f" - ({self.details})" if self.details else ""
+
+        return f"{self.message}{details}{old_data_record}"
```
leandrodamascena: Is there a way to optimize this function? I'm still trying to validate this, but I'm not sure we need all of these things.

leandrodamascena (changes requested):
Hello @dastra! This is a great job and a great improvement for our customers, after all, we are reducing costs by removing a `get_item` call and even reducing network latency, as now the data stored in the DDB is returned in a single call.

I made some comments but this PR is super! I'll make sure we have this merged as soon as possible.

docs/utilities/idempotency.md
dastra: Yes please do!

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
leandrodamascena: After a conversation with @dastra, we updated the code:

```python
@staticmethod
def boto3_supports_condition_check_failure(boto3_version: str) -> bool:
    """
    Check if the installed boto3 version supports condition check failure.

    Params
    ------
    boto3_version: str
        The boto3 version

    Returns
    -------
    bool 
        True if the boto3 version supports condition check failure, False otherwise.
    """
    # Only supported in boto3 1.26.164 and above
    major, minor, *patch = map(int, boto3_version.split('.'))
    return (major, minor, *patch) >= (1, 26, 164)
```

aws_lambda_powertools/utilities/idempotency/persistence/base.py
dastra: Sure! I've  updated _validate_payload to accept a DataRecord as the payload and removed _validate_hashed_payload:

```python    
def _validate_payload(self, data_payload: Union[Dict[str, Any], DataRecord], stored_data_record: DataRecord) -> None:
        """
        Validate that the hashed payload matches data provided and stored data record

        Parameters
        ----------
        data_payload: Union[Dict[str, Any], DataRecord]
            Payload
        stored_data_record: DataRecord
            DataRecord fetched from Dynamo or cache

        Raises
        ----------
        IdempotencyValidationError
            Payload doesn't match the stored record for the given idempotency key

        """
        if self.payload_validation_enabled:
            if isinstance(data_payload, DataRecord):
                data_hash = data_payload.payload_hash
            else:
                data_hash = self._get_hashed_payload(data=data_payload)

            if stored_data_record.payload_hash != data_hash:
                raise IdempotencyValidationError("Payload does not match stored record for this event key")
```

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
dastra: Sure - changes made:

```python
                    except IdempotencyValidationError as idempotency_validation_error:
                        raise idempotency_validation_error from exc
```

aws_lambda_powertools/utilities/idempotency/exceptions.py
dastra: I've called the super methods to re-use code:

```python

    def __init__(self, *args: Optional[Union[str, Exception]], old_data_record: Optional[DataRecord] = None):
        self.old_data_record = old_data_record
        super().__init__(*args)

    def __str__(self):
        """
        Return all arguments formatted or original message
        """
        old_data_record = f" from [{(str(self.old_data_record))}]" if self.old_data_record else ""
        message = super().__str__()
        return f"{message}{old_data_record}"
```

tests/functional/idempotency/test_idempotency.py
```diff
@@ -514,6 +493,55 @@ def lambda_handler(event, context):
     stubber.deactivate()
 
 
+@pytest.mark.parametrize(
+    "idempotency_config",
+    [{"use_local_cache": True, "payload_validation_jmespath": "requestContext"}],
+    indirect=True,
+)
+def test_idempotent_lambda_already_completed_with_validation_bad_payload_from_local_cache(
```
leandrodamascena, line 501: Great catch to improve the coverage of our tests.

tests/functional/idempotency/test_idempotency.py
```diff
@@ -502,6 +478,9 @@ def test_idempotent_lambda_expired_during_request(
 
     stubber.activate()
 
+    submethod_mocked = mocker.patch.object(DynamoDBPersistenceLayer, "boto3_supports_condition_check_failure")
```
leandrodamascena:
```suggestion
    submethod_mocked = mocker.patch.object(persistence_store, "boto3_supports_condition_check_failure")
```

I was curious about this change, could you please explain the effect of this to me? Out of curiosity, I ran some local tests with the True and False values and the test passed normally.

leandrodamascena (changes requested):
Hello @dastra! Thanks for addressing the last few review comments! I just left a comment about a small change. 

I'm still working on our documentation because now we have 3 possible workflows for successful executions, and each of them makes a different number of requests. So we must document it properly to make it clear to our customers.

### First execution
`PutItem` (1 DDB request) -> execute Lambda -> `UpdateItem` with Lambda response (1 DDB request) = 2 requests total

### Second execution (record exists and boto3 supports `ReturnValuesOnConditionCheckFailure`)
`PutItem` (1 DDB request) -> The conditional check fails with existing record -> Return Item from exception = 1 request total

### Second execution with legacy behavior (record exists and boto3 doesn't support `ReturnValuesOnConditionCheckFailure`)
`PutItem` (1 request) - The conditional check fails with the existing record -> `GetItem` (1 DDB request) = 2 requests total

leandrodamascena:
Items missing before we merge this:

- [x] Modify our documentation
- [x] Validate this changes with end-2-end tests in different versions of Lambda Runtime/boto3
- [ ] Final review

tests/functional/idempotency/test_idempotency.py
dastra: The idea with the test was that when testing locally, you will probably have a version of the boto3 >= 1.26.164 installed.  So unless we have boto3_supports_condition_check_failure return false, the path for what happens when boto3 <  1.26.164 will not be tested locally.

leandrodamascena:
This is ready for review @rubenfonseca! Please review it.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -15,8 +15,9 @@
 from aws_lambda_powertools.utilities.idempotency.exceptions import (
     IdempotencyItemAlreadyExistsError,
     IdempotencyItemNotFoundError,
+    IdempotencyValidationError,
 )
-from aws_lambda_powertools.utilities.idempotency.persistence.base import (
+from aws_lambda_powertools.utilities.idempotency.persistence.datarecord import (
```
rubenfonseca, line 20: Just picking this example to demonstrate how this change is a huge breaking change. If someone has implemented its own persistence store, the code will now break. Can we keep compatibility?

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -221,6 +222,15 @@ def _put_record(self, data_record: DataRecord) -> None:
...
                 f"{idempotency_key_not_exist} OR {idempotency_expiry_expired} OR ({inprogress_expiry_expired})"
             )
+
+            # Use DynamoDB's ReturnValuesOnConditionCheckFailure to optimize put and get operations and optimize costs.
+            # This feature is supported in boto3 versions 1.26.164 and later.
+            support_return_on_check_failure = (
+                {"ReturnValuesOnConditionCheckFailure": "ALL_OLD"}
+                if self.boto3_supports_condition_check_failure(boto3.__version__)
+                else {}
+            )
```
rubenfonseca: It seems to me this code is going to run on every single idemotent request. Can we memorize this check so we’re not wasting cycles all the time?

rubenfonseca (changes requested):

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
leandrodamascena: Great point, Ruben!! But we are i[mporting DataRecord](https://github.com/aws-powertools/powertools-lambda-python/pull/3446/files#diff-becc2a8b1233bddbb0d38b13a8634ab96bdd5d00f16f0514f6ea1df509c3b1d6R26) in `aws_lambda_powertools.utilities.idempotency.persistence.base` and customers can continue importing in the same way, this will not break this import. Pls let me know if I'm missing something here.

```python
from aws_lambda_powertools.utilities.idempotency.persistence.base import DataRecord as DR1 # OLD IMPORT
from aws_lambda_powertools.utilities.idempotency.persistence.datarecord import DataRecord as DR2 # NEW IMPORT

dr1 = DR1("idkey","pending",1,1,"response", "payload") # OLD IMPORT

dr2 = DR2("idkey","pending",1,1,"response", "payloadxxxxxxxxxxx") # NEW IMPORT

# Values are different, this must fail
assert dr1.__dict__ == dr2.__dict__, "AssertionError - Values are different"
```

```sh
❯ python test.py
  File "/tmp/ddbcondition/test.py", line 9, in <module>
    assert dr1.__dict__ == dr2.__dict__, "AssertionError - Values are different"
AssertionError: AssertionError - Values are different
```

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
leandrodamascena: Yes!! It makes perfect sense, I changed it to the Class init method, that way we keep it in memory.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
rubenfonseca: Ah I see, good explanation, thanks!

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -114,6 +115,14 @@ def __init__(
         self.data_attr = data_attr
         self.validation_key_attr = validation_key_attr
 
+        # Use DynamoDB's ReturnValuesOnConditionCheckFailure to optimize put and get operations and optimize costs.
+        # This feature is supported in boto3 versions 1.26.164 and later.
+        self.support_return_on_check_failure = (
```
rubenfonseca: One last thing, when reading this variable my mind thinks it will be a boolean (supports or not supports) but that's not the case. What about:

```suggestion
        self.return_value_on_condition = (
```

? what do you think?

rubenfonseca (changes requested):
Sorry, just one more tiny thing

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
leandrodamascena: Yes! It makes sense! I changed.

rubenfonseca (approved):

leandrodamascena (approved):
Thank you for this stellar work @dastra! 

APPROVED!

PR #3514, opened 2023-12-15
fix(event_handler): allow responses and metadata when using Router

### LINKED ISSUE #3506: Bug: APIGatewayRestResolver - openAPI metadata responses / tags on @router fails with unhashable type: dict
### Expected Behaviour

When adding `responses` or `tags` metadata to a router this should behave in the same was as when it is added to the app.

### Current Behaviour

When the routers are added to the app the metadata is used as a key. As the metadata fields for `responses` and `tags` are Dict and List then this produces the following error:

```  
File "/Users/my_path/venv/lib/python3.10/site-packages/aws_lambda_powertools/event_handler/api_gateway.py", line 2160, in register_route
    self._routes_with_middleware[route_key] = []
TypeError: unhashable type: 'dict'
```

### Code snippet

```python
from aws_lambda_powertools.event_handler.api_gateway import Router

router = Router()

app.include_router(router)


@router.put(
    "/example-resource",
    responses={200: {"description": "Todo item found"}},
    tags=["Examples"]
)
def put():
    """"""


if __name__ == "__main__":
    """"""
```

### Possible Solution

_No response_

### Steps to Reproduce

- Create APIGatewatRestResolver.
- Create api_gateway Router.
- Add method to Router with `responses` or `tags` metadata.
- Add Router to app.

2.29.1

3.10

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3506

### Changes

Some of the new additions related to OpenAPI support added non-hashable elements to the route key when using the built-in Router.
This made the code crash. With this change, we convert all those new parameters into hashable equivalents.

### User experience

The following code now works:

```python
router = Router()

@router.put("/example-resource", responses={200: {"description": "Custom response"}}, tags=["example"])
def handler():
    pass

app.include_router(router)

schema = app.get_openapi_schema()
put = schema.paths["/example-resource"].put
assert len(put.tags) == 1
assert puts.tags[0].name == "example"
assert 200 in put.responses.keys()
assert put.responses[200].description == "Custom response"
```

## Checklist

- [ ] Changes are documented

<details>
  **RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
looking... 👀

heitorlessa (changes requested):
let's add a note to refactor that tuple into a class for better encapsulation and growth

heitorlessa (approved):

PR #3528, opened 2023-12-18
fix(event-handler): operation tags are now string to correct schema generation

### LINKED ISSUE #3520: Bug: openAPI json spec generation - path tags validation failure - property type must be string
### Expected Behaviour

When generating the openAPI json schema using `get_openapi_json_schema()` and specifying path tags, the produced schema should validate against OAS 3.0.0 or 3.1.0 and path tags should be an array of strings.

[OAS3.1.0](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml)
'  operation:
    $comment: https://spec.openapis.org/oas/v3.1.0#operation-object
    type: object
    properties:
      tags:
        type: array
        items:
          type: string'

### Current Behaviour

When generating the openAPI json schema using `get_openapi_json_schema()` and specifying path tags, the produced schema raises a validation error against OAS 3.0.0 or 3.1.0 with the following error:

`property type must be string`

### Code snippet

```python

@app.put(
    "/example-resource",
    tags=["Examples"]
)
def put():
    pass


if __name__ == "__main__":
    print(app.get_openapi_json_schema())
```

### Possible Solution

_No response_

### Steps to Reproduce

- Create a simple path with tags and attach to an APIGatewayRestResolver.
- Create and print the openapi_json_schema.
- Compare the output to OAS 3.1.0 or 3.0.0 
- Output shows:

`"paths": {
    "/example-resource": {
      "put": {
        "tags": [
          {
            "name": "Examples"
          }
        ],
        "summary": "PUT /example-resource",`

Which is not correct as OAS 3.1.0 (and 3.0.0) ([OAS3.1.0](https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/schema.yaml)) specify an array of strings:

`  operation:
    $comment: https://spec.openapis.org/oas/v3.1.0#operation-object
    type: object
    properties:
      tags:
        type: array
        items:
          type: string`

2.30.1

3.10

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3520

### Changes

This PR fixes an issue with spec compliance when generating operation tags.

### User experience

Before this PR, our OpenAPI generation was generating opeation tags as dictionaries, similar to the [Tag Object](https://swagger.io/specification/v3/#tag-object). However, in the context of an Operation, this is wrong, as [tags should just be a list of strings](https://swagger.io/specification/v3/#operation-object).

image

This was causing any OpenAPI schema validation to fail. This PR fixes this.

## Checklist

- [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/openapi/models.py
+++ b/aws_lambda_powertools/event_handler/openapi/models.py
@@ -380,7 +380,7 @@ class Tag(BaseModel):
 
 # https://swagger.io/specification/#operation-object
 class Operation(BaseModel):
-    tags: Optional[List[Tag]] = None
+    tags: Optional[List[str]] = None
     summary: Optional[str] = None
     description: Optional[str] = None
     externalDocs: Optional[ExternalDocumentation] = None
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -559,7 +559,7 @@ class Route:
 
         # Ensure tags is added to the operation
         if self.tags:
-            operation["tags"] = [{"name": tag for tag in self.tags}]
+            operation["tags"] = self.tags
 
         # Ensure summary is added to the operation
         operation["summary"] = self._openapi_operation_summary()
--- a/tests/functional/event_handler/test_openapi_params.py
+++ b/tests/functional/event_handler/test_openapi_params.py
@@ -10,7 +10,6 @@ from aws_lambda_powertools.event_handler.openapi.models import (
     Parameter,
     ParameterInType,
     Schema,
-    Tag,
 )
 from aws_lambda_powertools.event_handler.openapi.params import (
     Body,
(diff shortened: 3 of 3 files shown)


aws_lambda_powertools/event_handler/openapi/models.py
```diff
@@ -380,7 +380,7 @@ class Config:
 
 # https://swagger.io/specification/#operation-object
 class Operation(BaseModel):
-    tags: Optional[List[Tag]] = None
+    tags: Optional[List[str]] = None
```
heitorlessa, line 383: To be triple sure -- that's not a breaking change, right?

I didn't see in the docs an example where they could pass an Operation obj along with a Tag obj

rubenfonseca: It can be considered a breaking change since we're changing the type, but it's also a bug as the original implementation is 100% incorrect and doesn't match the spec.

heitorlessa: > 100% incorrect and doesn't match the spec

Bug then.

heitorlessa (approved):

PR #3548, opened 2023-12-21
feat(event-handler): add description to request body in OpenAPI schema

### LINKED ISSUE #3539: Bug: openAPI generation - operation requestBody description only on schema output, not at higher level needed for html
### Expected Behaviour

When creating an operation such as a PUT with a body parameter, the body parameter description should be populated on the output at the requestBody level (when generating the spec with app.get_openapi_json_schema() ). 

### Current Behaviour

When creating an operation such as a PUT with a body parameter, the body parameter description only appears in the output under schema (when generating the spec with app.get_openapi_json_schema() ).

Current output shows requestBody description only at the schema level, this means that when redocly is used to create an html file from the spec the description does not appear.
For the path parameter the descrition is populated at two levels which means it does appear in the redocly generated html.

Current output:
image

OAS 3.1.0 looks as though it is possible to populate the description on the requestBody object as well:
image

### Code snippet

```python
from typing import Dict, Any

from aws_lambda_powertools.event_handler.openapi.params import Body, Path
from aws_lambda_powertools.shared.types import Annotated

@app.put(
    "/example-resource/<path_parameter>",
)
def put(
        path_parameter: Annotated[str, Path(description="test path description appears at schema and higher level", example="example01")],
        event_body: Annotated[Dict[str, Any], Body(description="test body description only appears at schema level")]
):
    pass


if __name__ == "__main__":
    print(app.get_openapi_json_schema())
```

### Possible Solution

Populate description on the requestBody object as well as the schema.

### Steps to Reproduce

Create an operation with a request body and add a description to the Body object.
Add a path parameter with a description to the operation.
Note on the output where description is populated in two places for the path parameter and only in one place for the body.
Optional -> Can generate a .html using redocly to confirm which of the descriptions makes it into the generated documentation.

2.30.2

3.10

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3539

### Changes

This PR adds support to expose the request body description directly on OpenAPI.

### User experience

Before this PR, certain API browsers like Redoc would fail to render the body description, even though it was part of the schema. After this change, we duplicate the body description on a higher level (still part of the OpenAPI specification) and that fixes the problem with Redoc.

image

## Checklist

- [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -617,6 +617,9 @@ class Route:
         if required:
             request_body_oai["required"] = required
 
+        if field_info.description:
+            request_body_oai["description"] = field_info.description
+
         # Generate the request body media type
         request_media_content: Dict[str, Any] = {"schema": body_schema}
         request_body_oai["content"] = {request_media_type: request_media_content}
--- a/tests/functional/event_handler/test_openapi_params.py
+++ b/tests/functional/event_handler/test_openapi_params.py
@@ -349,6 +349,30 @@ def test_openapi_with_embed_body_param():
     assert body_post_handler_schema.properties["user"].ref == "#/components/schemas/User"
 
 
+def test_openapi_with_body_description():
+    app = APIGatewayRestResolver()
+
+    class User(BaseModel):
+        name: str
+
+    @app.post("/users")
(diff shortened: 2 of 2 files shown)


leandrodamascena (approved):
APPROVED!

PR #3571, opened 2023-12-29
feat(event_handler): add support to download OpenAPI spec file

### LINKED ISSUE #3560: Feature request: Add ability to download OpenAPI spec
### Use case

The Swagger UI feature is nice, but it would be even better to be able to download the OpenAPI specification file so that if users would like to render the documentation in alternative formats, they are easily able to do so.

### Solution/User Experience

This could be achieved by adding another endpoint when Swagger is enabled at `/swagger.json` which would download the spec file.

### Alternative solutions

_No response_

### Acknowledgment

Thomas-McKanna:
**Issue number: https://github.com/aws-powertools/powertools-lambda-python/issues/3560**

## Summary

This feature allows for setting the `enable_download_spec` in the `enable_swagger` function so that the OpenAPI specification file can be download for use generating client libraries, hosted documentation external, or for another reason.

Usage would look like:

```python
app.enable_swagger(path="/_swagger", enable_download_spec=True)
```

As I was working on this task, I came to think that it might be better to just expose the JSON endpoint by default rather
than require the user to opt-in. I don't see how adding this feature as default would break existing deployments, and would cut down the number of steps for a user to use the feature from two steps (adding `enable_download_spec` parameter and adding endpoint in SAM template) to one step (adding endpoint in SAM template). What are thoughts on this?

### Changes

- A new conditional endpoint for downloading the JSON was created.
- Creation of the specification JSON was moved into a separate function to avoid duplicate code.
- Documentation was updated to reflect new feature.

### Update - 09/01/2024

We decided to drop the `enable_download_spec=True` flag. We added a querystring parameter (?format=json) along the /swagger route.

### User experience

Additional flexibility of the OpenAPI feature.

## Checklist

* [ ] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [x] Changes are documented

<details>
<summary>Is this a breaking change? No.</summary>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena:
> As I was working on this task, I came to think that it might be better to just expose the JSON endpoint by default rather than require the user to opt-in. I don't see how adding this feature as default would break existing deployments, and would cut down the number of steps for a user to use the feature from two steps (adding `enable_download_spec` parameter and adding endpoint in SAM template) to one step (adding endpoint in SAM template). What are thoughts on this?

Hi @Thomas-McKanna! I'm unsure if we would set this parameter to `True` by default. I'm thinking of cases where customers are already using this feature, just explicitly exposing `/swagger`, `/swagger.js`, and `/swagger.css` instead of using `{proxy+}` and then Swagger won't work because the path `/swagger.json` is missing/not defined.

We can keep it as False by default and hear from others before merging.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1559,6 +1559,7 @@ def enable_swagger(
         license_info: Optional["License"] = None,
         swagger_base_url: Optional[str] = None,
         middlewares: Optional[List[Callable[..., Response]]] = None,
+        enable_download_spec: Optional[bool] = False,
```
leandrodamascena:
```suggestion
        enable_download_spec: bool = False,
```

False is the default value, so, we don't need `Optional` in this field.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1625,9 +1639,19 @@ def swagger_handler():
                 swagger_js = f"{base_path}/swagger.js"
                 swagger_css = f"{base_path}/swagger.css"
 
-            openapi_servers = servers or [Server(url=(base_path or "/"))]
+            spec = get_swagger_spec()
+            body = generate_swagger_html(spec, swagger_js, swagger_css)
```
leandrodamascena: I think we need to add new parameters to this function, right? We serve the Swagger specification in the function [generate_swagger_html](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/openapi/swagger_ui/html.py#L1) and currently we are using the field `spec` in the [SwaggerUIOptions](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/openapi/swagger_ui/html.py#L36) definition.

docs/core/event_handler/api_gateway.md
```diff
@@ -1019,6 +1019,21 @@ Include extra parameters when exporting your OpenAPI specification to apply thes
...
+
+```python
+app.enable_swagger(path="/_swagger", enable_download_spec=True)
+```
+
+With this setting enabled, you can access the specification at `<path>.json`. So, for
+the snippet above, the specification file would be retrievable at `/_swagger.json`.
+To access this endpoint, you will have to add an additional endpoint in your template
+file.
+
```
leandrodamascena: I'll revisit this section before we merge.

leandrodamascena (changes requested):
Hi @Thomas-McKanna! Thanks a lot for sending this PR, I left some comments.

We need to create some additional tests to check if we propagate/use this new parameter. Can you add this, please?

aws_lambda_powertools/event_handler/api_gateway.py
Thomas-McKanna: I added the `get_swagger_spec` function as a way to remove duplicate code between the new route I added and the existing route that serves the documentation UI. No arguments need to be passed because the function is defined with in the `enable_swagger` function (just like the endpoint functions). If I hadn't moved the shared logic to a shared function (`get_swagger_spec`), the code would have looked like:

```python
def enable_swagger(
        self,
        *,
        path: str = "/swagger",
        title: str = "Powertools for AWS Lambda (Python) API",
        version: str = DEFAULT_API_VERSION,
        openapi_version: str = DEFAULT_OPENAPI_VERSION,
        summary: Optional[str] = None,
        description: Optional[str] = None,
        tags: Optional[List[Union["Tag", str]]] = None,
        servers: Optional[List["Server"]] = None,
        terms_of_service: Optional[str] = None,
        contact: Optional["Contact"] = None,
        license_info: Optional["License"] = None,
        swagger_base_url: Optional[str] = None,
        middlewares: Optional[List[Callable[..., Response]]] = None,
        enable_download_spec: bool = False,
    ):
        """
        ...
        """
        ...

        if enable_download_spec:

            @self.get(f"{path}.json", include_in_schema=False)
            def swagger_json():
                # The following 14 lines are duplicated below
                base_path = self._get_base_path()
                openapi_servers = servers or [Server(url=(base_path or "/"))]
                spec = self.get_openapi_json_schema(
                    title=title,
                    version=version,
                    openapi_version=openapi_version,
                    summary=summary,
                    description=description,
                    tags=tags,
                    servers=openapi_servers,
                    terms_of_service=terms_of_service,
                    contact=contact,
                    license_info=license_info,
                )
                return Response(
                    status_code=200,
                    body=spec,
                )

        @self.get(path, middlewares=middlewares, include_in_schema=False)
        def swagger_handler():
            base_path = self._get_base_path()

            if swagger_base_url:
                swagger_js = f"{swagger_base_url}/swagger-ui-bundle.min.js"
                swagger_css = f"{swagger_base_url}/swagger-ui.min.css"
            else:
                swagger_js = f"{base_path}/swagger.js"
                swagger_css = f"{base_path}/swagger.css"

            base_path = self._get_base_path()
            openapi_servers = servers or [Server(url=(base_path or "/"))]
            spec = self.get_openapi_json_schema(
                title=title,
                version=version,
                openapi_version=openapi_version,
                summary=summary,
                description=description,
                tags=tags,
                servers=openapi_servers,
                terms_of_service=terms_of_service,
                contact=contact,
                license_info=license_info,
            )
            body = generate_swagger_html(spec, swagger_js, swagger_css)

            return Response(
                status_code=200,
                content_type="text/html",
                body=body,
            )
```

leandrodamascena: > I added the `get_swagger_spec` function as a way to remove duplicate code between the new route I added and the existing route that serves the documentation UI. No arguments need to be passed because the function is defined with in the `enable_swagger` function (just like the endpoint functions). If I hadn't moved the shared logic to a shared function (`get_swagger_spec`), the code would have looked like:

I understand your point and with this implementation, the customer can open a tab/curl/wget and now download the OpenAPI specification requesting `https://www.example.com/swagger.json` directly, this works perfectly.

But what I want to say is: we are implementing the OpenAPI + SwaggerUI specification, where customers can access the swagger definition in the browser (using SwaggerUI) and test endpoints, see models definition and others. Another super common thing in SwaggerUI is to display the link to directly access the JSON/YAML of that OpenAPI spec. If we don't make the modification I suggested, the customer will be able to download OpenAPI by directly accessing `https://www.example.com/swagger.json` (if they know), but this link will not appear in the SwaggerUI interface, which may confuse some customers.
image

We would do something like this:

**api_gateway.py**
```python
spec = get_swagger_spec()
body = generate_swagger_html(spec, swagger_js, swagger_css, enable_download_spec, path)
```

**html.py**

```python
def generate_swagger_html(spec: str, js_url: str, css_url: str, enable_download_spec: bool, path: str) -> str:

	if enable_download_spec:
	    js_json_spec = f'url: "{path}.json",'
	else:
	    js_json_spec = f"""
	    spec: JSON.parse(`
	            {spec}
	        `.trim()),
	    """
```

```js
<script>
  var swaggerUIOptions = {{
    dom_id: "#swagger-ui",
    docExpansion: "list",
    deepLinking: true,
    filter: true,
    layout: "BaseLayout",
    showExtensions: true,
    showCommonExtensions: true,
    {js_json_spec}
    presets: [
      SwaggerUIBundle.presets.apis,
      SwaggerUIBundle.SwaggerUIStandalonePreset
    ],
    plugins: [
      SwaggerUIBundle.plugins.DownloadUrl
    ]
  }}

  var ui = SwaggerUIBundle(swaggerUIOptions)
</script>
```

This code is just an example, we can improve it. Does this make sense to you? What do you think?

Thomas-McKanna:
@leandrodamascena I added a couple of tests to ensure this feature is working with the default path (`/swagger`) and a custom path.

aws_lambda_powertools/event_handler/api_gateway.py
Thomas-McKanna: Ah, I see what you were originally getting at. Yes, I will make an adjustment similar to the one you mentioned. 

From a user experience perspective, it feels a little burdonsome to have to set the `enable_download_spec` parameter to `True` _and_ to add an additional API endpoint if I'm not using `proxy+`. It's doable, but it would be better if the Swagger UI had a link to the spec JSON that just worked for everyone.

Idea for you: what if the `/swagger` endpoint took an optional query string of `?view=json`:

```python
@self.get(path, middlewares=middlewares, include_in_schema=False)
def swagger_handler(view: Annotated[Optional[str], Query()] = None):
```

This query string would be used to conditionally return the documentation as JSON instead of the HTML.

With something like this, every user would get the ability to download the spec, and the download link would be like `https://www.example.com/swagger?view=json`. And in this case, `enable_download_spec` could be removed.

Let me know your thoughts and I will implement the download link one way or the other.

leandrodamascena:
Hi @Thomas-McKanna! Comment here instead of the [initial conversation](https://github.com/aws-powertools/powertools-lambda-python/pull/3571#discussion_r1438615599) because it is more visible to everyone.

I had a conversation with @rubenfonseca - who created the OpenAPI feature in Powertools - and the main reason we inject OpenAPI spec into the configuration (using `spec` field) instead of requesting a URL is because it's faster and more cost-effective. So, we can keep the field `spec` in the SwaggerUI configuration, but we can add a new line to update the URL and enable the link to download:

```javascript
<script>
  var swaggerUIOptions = {{
    dom_id: "#swagger-ui",
	.
	.
    spec: JSON.parse(`
                {spec}
            `.trim()),
	.
	.    
  }}

  var ui = SwaggerUIBundle(swaggerUIOptions)
  ui.specActions.updateUrl('{URL}/{PATH}.json'); // NEW LINE
</script>
```

We agree that enabling it with a conditional query string is a good approach and we can remove the `enable_download_spec` parameter. I tested it with a real deployment in API Gateway and it works perfectly. Please go ahead with this implementation.

Thanks for addressing all the feedback and participating in this discussion, Thomas! We always try to find the best experience for our customers and avoid breaking anything.

Thomas-McKanna:
Sounds good @leandrodamascena, I will get these changes committed soon!

Thomas-McKanna:
@leandrodamascena I have implemented the requested changes. Now the OpenAPI spec JSON can be fetched by setting the `?format=json` query string on the documentation path. No more need for `enabled_download_spec` and no additional paths need to be created in the SAM template - it should just work for everyone!

Screenshot 2024-01-05 at 9 41 57 AM

I also removed the documentation section I had added.

Let me know if you'd like me to change anything.

leandrodamascena:
> @leandrodamascena I have implemented the requested changes. Now the OpenAPI spec JSON can be fetched by setting the `?format=json` query string on the documentation path. No more need for `enabled_download_spec` and no additional paths need to be created in the SAM template - it should just work for everyone!
> 
> Screenshot 2024-01-05 at 9 41 57 AM
> I also removed the documentation section I had added.
> 
> Let me know if you'd like me to change anything.

Hey @Thomas-McKanna! Thanks for addressing the feedback! I'll review this tomorrow and I expect to merge this before our next release!

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1640,7 +1641,28 @@ def swagger_handler():
...
-            body = generate_swagger_html(spec, swagger_js, swagger_css)
+            # The .replace('</', '<\\/') part is necessary to prevent a potential issue where the JSON string contains
+            # </script> or similar tags. Escaping the forward slash in </ as <\/ ensures that the JSON does not
+            # inadvertently close the script tag, and the JSON remains a valid string within the JavaScript code.
+            escaped_spec = model_json(
+                spec,
+                by_alias=True,
+                exclude_none=True,
+                indent=2,
+            ).replace("</", "<\\/")
```
leandrodamascena, line 1652: Hello @rubenfonseca! I moved this logic to this file because we can use the same specification in JSON View and SwaggerUI rendering.
Please let me know if make sense.

leandrodamascena (commented):
@Thomas-McKanna I resolved a conflict we had due to a PR we merged today and made a small change, but nothing that changes your logic.

@rubenfonseca please review this PR.

leandrodamascena (approved):

rubenfonseca (approved):

PR #3638, opened 2024-01-16
chore(ci): drop support for Python 3.7

### LINKED ISSUE #3615: Maintenance: Deprecate Python 3.7
### Why is this needed?

AWS Lambda runtime will fully deprecate [Python 3.7 on February 8th](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtime-support-policy). This means, 3.7 functions will be unable to be updated.

Additionally, AWS SDK for Python (Boto3 and Botocore), AWS Command Line Interface (AWS CLI), and AWS CDK [no longer offers support for Python 3.7](https://aws.amazon.com/pt/blogs/developer/python-support-policy-updates-for-aws-sdks-and-tools/) as well. The same happens with typing extensions, a library used extensively throughout this project that dropped support for Python3.7 in version [4.7.1](https://github.com/python/typing_extensions/releases/tag/4.7.1). 

We need to measure the impact of this and make the decision to remove support for Python 3.7. Keep the Python3.7 support directly impacts our supply chain in terms of preventing us from receiving new features and potential security bug fixes.

### Minor or major release?
We started writing the roadmap for Powertools V3, however, we are still many steps away from having a roadmap to define new features/codebase for v3, write the RFC for V3, listen to the community, and refactor/rewrite everything necessary. So, will we deprecate Python 3.7 in a v2 minor release or wait until Powertools v3?

While I agree that this changes behavior and may prevent someone from receiving new Powertools updates, I think the impact is minimal for anyone using Managed Runtime Python 3.7. They will not be able to update their Python 3.7 functions from Feb/8, so those who need to update their functions will be forced to update Python to 3.8 and will not be affected by this.

### Running Powertools on Lambda with OCI, and other containers platforms in general
These customers may be impacted by being blocked from receiving new Powertools updates if they use Python 3.7.
I don't think they will have problems building the containers in general, because package management tools like pip, poetry, anaconda and others will pin Powertools dependency on the latest version supported by Python 3.7.

### Other runtimes
Powertools for AWS Lambda (TypeScript) [dropped support for nodejs14.x](https://github.com/aws-powertools/powertools-lambda-typescript/issues/1664) along with the deprecation of Lambda nodejs14.x.

I may be missing something in my points and I would like to hear your considerations before we move forward with this decision.

### Update 11/01

- [x] #3617
- [x] #3637

Other

### Solution

Drop support for Python 3.7

### Acknowledgment

### LINKED ISSUE #3637: Maintenance: Drop Python 3.7 support in our pipelines, pyproject.toml and clean up the code
We need to drop support for Python 3.7 and clean up the code with Python version checks + compat files that are no longer needed.

leandrodamascena:
**Issue number:** #3637 

### Changes

AWS Lambda runtime will fully deprecate [Python 3.7 on February 8th](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtime-support-policy). This means, 3.7 functions will be unable to be updated. Inline with this, Powertools releases will stop supporting it.

### User experience

Customers will no longer be able to use Powertools for AWS Lambda (Python) with Python 3.7.

### Improvements 

In this pull request, we've made some enhancements, taking into account the removal of Python 3.7 support.

1. We've eliminated Python 3.7-specific code from the Logger utility to maintain compatibility.
2. We are importing directly from `typing` new types added in Python 3.8 and previously only available in the `typing_extensions` library.
3. Now we are caching properties in the event source data class with the new `cached_property` decorator.
4. We removed some code from the Streaming utility and added an alias to use the version supported by `botocore`.

#### Before
```python
"""
Currently, the same as https://github.com/boto/botocore/blob/b9c540905a6c9/botocore/response.py
We created this because the version of StreamingBody included with the Lambda Runtime is too old, and
doesn't support many of the standard IO methods (like readline).

As soon as the version of botocore included with the Lambda runtime is equal or greater than 1.29.13, we can drop
this file completely. See https://docs.aws.amazon.com/lambda/latest/dg/lambda-python.html.
"""
import logging
from io import IOBase
from typing import Optional

import botocore
from botocore import endpoint
from botocore.compat import set_socket_timeout
from botocore.exceptions import (
    IncompleteReadError,
    ReadTimeoutError,
    ResponseStreamingError,
)
from urllib3.exceptions import ProtocolError as URLLib3ProtocolError
from urllib3.exceptions import ReadTimeoutError as URLLib3ReadTimeoutError

# Splitting the botocore version string into major, minor, and patch versions,
# and performing a conditional check based on the extracted versions.
major, minor, patch = map(int, botocore.__version__.split("."))

if major == 1 and (minor < 29 or patch < 13):

    class PowertoolsStreamingBody(IOBase):
        """Wrapper class for a HTTP response body.

        This provides a few additional conveniences that do not exist
        in the urllib3 model:
            * Set the timeout on the socket (i.e read() timeouts)
            * Auto validation of content length, if the amount of bytes
              we read does not match the content length, an exception
              is raised.
        """

        _DEFAULT_CHUNK_SIZE = 1024

        def __init__(self, raw_stream, content_length):
            self._raw_stream = raw_stream
            self._content_length = content_length
            self._amount_read = 0

        def __del__(self):
            # Extending destructor in order to preserve the underlying raw_stream.
            # The ability to add custom cleanup logic introduced in Python3.4+.
            # https://www.python.org/dev/peps/pep-0442/
            pass

        def set_socket_timeout(self, timeout):
            """Set the timeout seconds on the socket."""
            # The problem we're trying to solve is to prevent .read() calls from
            # hanging.  This can happen in rare cases.  What we'd like to ideally
            # do is set a timeout on the .read() call so that callers can retry
            # the request.
            # Unfortunately, this isn't currently possible in requests.
            # See: https://github.com/kennethreitz/requests/issues/1803
            # So what we're going to do is reach into the guts of the stream and
            # grab the socket object, which we can set the timeout on.  We're
            # putting in a check here so in case this interface goes away, we'll
            # know.
            try:
                set_socket_timeout(self._raw_stream, timeout)
            except AttributeError:
                logger.error(
                    "Cannot access the socket object of "
                    "a streaming response.  It's possible "
                    "the interface has changed.",
                    exc_info=True,
                )
                raise

        def readable(self):
            try:
                return self._raw_stream.readable()
            except AttributeError:
                return False

        def read(self, amt=None):
            """Read at most amt bytes from the stream.
            If the amt argument is omitted, read all data.
            """
            try:
                chunk = self._raw_stream.read(amt)
            except URLLib3ReadTimeoutError as e:
                raise ReadTimeoutError(endpoint_url=e.url, error=e)
            except URLLib3ProtocolError as e:
                raise ResponseStreamingError(error=e)
            self._amount_read += len(chunk)
            if amt is None or (not chunk and amt > 0):
                # If the server sends empty contents or
                # we ask to read all of the contents, then we know
                # we need to verify the content length.
                self._verify_content_length()
            return chunk

        def readlines(self, hint: Optional[int] = -1):
            return self._raw_stream.readlines(hint)

        def __iter__(self):
            """Return an iterator to yield 1k chunks from the raw stream."""
            return self.iter_chunks(self._DEFAULT_CHUNK_SIZE)

        def __next__(self):
            """Return the next 1k chunk from the raw stream."""
            current_chunk = self.read(self._DEFAULT_CHUNK_SIZE)
            if current_chunk:
                return current_chunk
            raise StopIteration()

        def __enter__(self):
            return self._raw_stream

        def __exit__(self, *args):
            self._raw_stream.close()

        next = __next__  # noqa: A003, VNE003

        def iter_lines(self, chunk_size=_DEFAULT_CHUNK_SIZE, keepends=False):
            """Return an iterator to yield lines from the raw stream.
            This is achieved by reading chunk of bytes (of size chunk_size) at a
            time from the raw stream, and then yielding lines from there.
            """
            pending = b""
            for chunk in self.iter_chunks(chunk_size):
                lines = (pending + chunk).splitlines(True)
                for line in lines[:-1]:
                    yield line.splitlines(keepends)[0]
                pending = lines[-1]
            if pending:
                yield pending.splitlines(keepends)[0]

        def iter_chunks(self, chunk_size=_DEFAULT_CHUNK_SIZE):
            """Return an iterator to yield chunks of chunk_size bytes from the raw
            stream.
            """
            while True:
                current_chunk = self.read(chunk_size)
                if current_chunk == b"":
                    break
                yield current_chunk

        def _verify_content_length(self):
            # See: https://github.com/kennethreitz/requests/issues/1855
            # Basically, our http library doesn't do this for us, so we have
            # to do this ourself.
            if self._content_length is not None and self._amount_read != int(self._content_length):
                raise IncompleteReadError(
                    actual_bytes=self._amount_read,
                    expected_bytes=int(self._content_length),
                )

        def tell(self):
            return self._raw_stream.tell()

        def close(self):
            """Close the underlying http response stream."""
            self._raw_stream.close()

    def convert_to_response_dict(http_response, operation_model):
        """Convert an HTTP response object to a request dict.

        This converts the requests library's HTTP response object to
        a dictionary.

        :type http_response: botocore.vendored.requests.model.Response
        :param http_response: The HTTP response from an AWS service request.

        :rtype: dict
        :return: A response dictionary which will contain the following keys:
            * headers (dict)
            * status_code (int)
            * body (string or file-like object)

        """
        response_dict = {
            "headers": http_response.headers,
            "status_code": http_response.status_code,
            "context": {
                "operation_name": operation_model.name,
            },
        }
        if response_dict["status_code"] >= 300:
            response_dict["body"] = http_response.content
        elif operation_model.has_event_stream_output:
            response_dict["body"] = http_response.raw
        elif operation_model.has_streaming_output:
            length = response_dict["headers"].get("content-length")
            response_dict["body"] = PowertoolsStreamingBody(http_response.raw, length)
        else:
            response_dict["body"] = http_response.content
        return response_dict

    # monkey patch boto3
    endpoint.convert_to_response_dict = convert_to_response_dict
else:
    from botocore.response import StreamingBody

    # Expose PowertoolsStreamingBody as StreamingBody
    vars()["PowertoolsStreamingBody"] = StreamingBody
```

#### After
```python
from botocore.response import StreamingBody

PowertoolsStreamingBody = StreamingBody
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


rubenfonseca (dismissed):

heitorlessa:
On `2`, IIRC we can rely on `typing_extensions` for everything and stop using `from typing` altogether - it'll handle versions gracefully.

On 4, could you add the botocore version (before/after) for completeness?

Just in case.. are there any `# maintenance:` notes about 3.7 that we can do as part of this PR too?

---

```bash
> grep -i "# maintenance" -R aws_lambda_powertools/ -R tests | grep "3.7"

aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key
tests/e2e/idempotency_redis/conftest.py:    # MAINTENANCE: Add the Stack constructor when Python 3.7 is dropped
```

leandrodamascena:
> On `2`, IIRC we can rely on `typing_extensions` for everything and stop using `from typing` altogether - it'll handle versions gracefully.

There are many files where we import from `typing` and `typing_extensions`. To keep the proposal of this PR, I will maintain the imports we are using today, but I will open a new PR to resolve this, there are at least 30 files to change and maintain this standard.

> 
> On 4, could you add the botocore version (before/after) for completeness?

Adding now.

> 
> Just in case.. are there any `# maintenance:` notes about 3.7 that we can do as part of this PR too?
> 
> ```shell
> > grep -i "# maintenance" -R aws_lambda_powertools/ -R tests | grep "3.7"
> 
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key to work
> aws_lambda_powertools//logging/logger.py:        # Maintenance: We can drop this upon Py3.7 EOL. It's a backport for "location" key
> tests/e2e/idempotency_redis/conftest.py:    # MAINTENANCE: Add the Stack constructor when Python 3.7 is dropped
> ```

I've already removed them all, you're probably running this grep on a branch other than this one, right?

Thanks

heitorlessa:
Yeah different branch, it was just for illustration ;-)

Will review as soon as I have a break in a customer meeting

aws_lambda_powertools/utilities/streaming/compat.py
```diff
@@ -1,206 +1,3 @@
...
-        return response_dict
-
-    # monkey patch boto3
-    endpoint.convert_to_response_dict = convert_to_response_dict
-else:
-    from botocore.response import StreamingBody
-
-    # Expose PowertoolsStreamingBody as StreamingBody
-    vars()["PowertoolsStreamingBody"] = StreamingBody
+PowertoolsStreamingBody = StreamingBody
```
heitorlessa, line 4:
```suggestion
# aliasing as 3.7 is no longer supported but unsure if anyone took a dependency on it (hyrum's law)
PowertoolsStreamingBody = StreamingBody
```

heitorlessa: a comment so we skip git blame if one ever needs to know what's this; later can be converted to `_compay.py` in v3.

examples/event_handler_graphql/src/assert_async_graphql_response_module.py
```diff
@@ -13,6 +6,7 @@
 from aws_lambda_powertools import Logger, Tracer
 from aws_lambda_powertools.event_handler import AppSyncResolver
 from aws_lambda_powertools.logging import correlation_paths
+from aws_lambda_powertools.shared.types import TypedDict
```
heitorlessa, line 9: was line highlighting affected in the docs?

heitorlessa (changes requested):
one tiny suggestion to add a comment, and one question as I suspect docs code highlighting will be impacted with the change.

examples/event_handler_graphql/src/assert_async_graphql_response_module.py
leandrodamascena: You are always saving our lives, Heitor! Yes, this refactors messed up the examples.

Done!

aws_lambda_powertools/utilities/streaming/compat.py
leandrodamascena: Added

heitorlessa (approved):
SHIIIIIPP IT! :) great work as always @leandrodamascena

Issue #3732, opened 2024-02-07
Bug: APIGatewayHttpResolver with validation enabled returns HTTP 422 when using single value query parameters


DariusKunce:
### Expected Behaviour

When using Amazon Api Gateway REST API to invoke an AWS Lambda using proxy integration, calling API with a single query parameter should return HTTP 200 with Lambda Powertools validation enabled. 

This is also how it is documented here: https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#validating-query-strings

### Current Behaviour

Calling API with `GET /my-path?productType=Category1` returns HTTP 422:
```json
{
    "statusCode": 422,
    "detail": [{
        "loc": ["query", "productType"], 
        "type": "type_error.str"
    }]
}
```

### Code snippet

```python
Bug can be reproduced with this code inside a Lambda function:


app = api_gateway.APIGatewayRestResolver(
    enable_validation=True
)

@app.get("/my-path")
def my_handler(
    product_type: Annotated[str | None, Query(alias="productType")] = None,
) -> api_gateway.Response[api_model.GetAvailableProductsResponse]:
    ...
```

### Possible Solution

Can be fixed by changing query param type to list:

```python
@app.get("/my-path")
def my_handler(
    product_type: Annotated[list[str] | None, Query(alias="productType")] = None,
) -> api_gateway.Response[api_model.GetAvailableProductsResponse]:
    """Lists available products"""
    ...
```

### Steps to Reproduce

In AWS:
1. Create a REST API in the Amazon Api Gateway Service.
2. Create a Lambda function with Lambda Powertools configured as in the code examples.
3. Configure the API Gateway with proxy integration to the Lambda function.
4. Invoke the API with a query parameter.

Locally:
1. Create unit tests for a Lambda function as described here: https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#testing-your-code
2. Run the test using REST API proxy integration payload. This payload always contains 2 attributes for the query string parameters, regardless if it's single value or multi value in REST API proxy integration:
    ```json
    {
        ...
        "queryStringParameters": { "productType": "Category1" },
        "multiValueQueryStringParameters": { "productType": ["Category1"] }
    }
    ```

The issue is this line of code: https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/data_classes/api_gateway_proxy_event.py#L123 

[...]

PR #3667, opened 2024-01-23
feat(event_handler): add support for multiValueQueryStringParameters in OpenAPI schema

leandrodamascena:
**Issue number:** #3622 

### Changes

This pull request enhances the OpenAPI utility by introducing support for `multiValueQueryStringParameters` in the OpenAPI schema. Previously, our utility encountered issues parsing multi-value query fields, leading to errors for customers attempting to utilize such fields. Work did:

- Added support for multiValueQueryStringParameters in the OpenAPI schema.
- Resolved parsing issues previously encountered when dealing with multi-value fields in different Resolvers.
- Added new tests

### User experience

Using the code below now we can validate multi value param.

```python
from enum import Enum
from http import HTTPStatus

from aws_lambda_powertools.event_handler import APIGatewayRestResolver
from aws_lambda_powertools.event_handler.openapi.params import Query
from aws_lambda_powertools.shared.types import Annotated
from aws_lambda_powertools.utilities.typing import LambdaContext

app = APIGatewayRestResolver(enable_validation=True)


class ExampleEnum(Enum):
    """Example of an Enum class."""

    ONE = "value_one"
    TWO = "value_two"
    THREE = "value_three"


@app.get("/example-resource")
def get(
    example_multi_value_param: Annotated[
        ExampleEnum,
        Query(
            description="This should be a multi value query parameter."
        ),  # TODO: How should this represent multi-value query string parameters?
    ]
):
    """Return validated multi-value param values."""
    return example_multi_value_param


def lambda_handler(event: dict, context: LambdaContext):
    """Handler to route API Gateway events."""
    return app.resolve(event, context)
```

image

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented
* [x] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #3795, opened 2024-02-19
fix(event-handler): multi-value query string and validation of scalar parameters

rubenfonseca:
**Issue number:** #3732

### Changes

This PR fixes a bug when validating scalar parameters. After [the code we've introduced ](https://github.com/aws-powertools/powertools-lambda-python/pull/3667) to handle multiple parameters, the code become inconsistent (notice the `Any` type) in that some resolvers return scalar parameters and some others returned a list of parameters.

This caused problems when handling validation of scalar parameters. This code fixes that by ensuring the same behaviour in all event handlers, and updating the types accordingly.

### User experience

This PR will fix the issue on #3732

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
+++ b/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
@@ -369,7 +369,7 @@ def _get_embed_body(
 
 
 def _normalize_multi_query_string_with_param(
-    query_string: Optional[Dict[str, List[str]]],
+    query_string: Dict[str, List[str]],
     params: Sequence[ModelField],
 ) -> Dict[str, Any]:
     """
@@ -386,18 +386,15 @@ def _normalize_multi_query_string_with_param(
     -------
     A dictionary containing the processed multi_query_string_parameters.
     """
-    if not query_string:
-        return {}
-    else:
-        resolved_query_string: Dict[str, Any] = query_string
-        for param in filter(is_scalar_field, params):
-            try:
-                # if the target parameter is a scalar, we keep the first value of the query string
-                # regardless if there are more in the payload
-                resolved_query_string[param.alias] = query_string[param.alias][0]
-            except KeyError:
-                pass
-        return resolved_query_string
+    resolved_query_string: Dict[str, Any] = query_string
+    for param in filter(is_scalar_field, params):
+        try:
+            # if the target parameter is a scalar, we keep the first value of the query string
+            # regardless if there are more in the payload
+            resolved_query_string[param.alias] = query_string[param.alias][0]
+        except KeyError:
+            pass
+    return resolved_query_string
 
 
 def _normalize_multi_header_values_with_param(headers: Optional[Dict[str, str]], params: Sequence[ModelField]):
--- a/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
+++ b/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
@@ -371,7 +371,7 @@ def _get_embed_body(
 def _normalize_multi_query_string_with_param(
     query_string: Optional[Dict[str, List[str]]],
     params: Sequence[ModelField],
-):
+) -> Optional[Dict[str, Any]]:
     """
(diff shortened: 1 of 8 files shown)


aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
```diff
@@ -383,15 +386,18 @@ def _normalize_multi_query_string_with_param(query_string: Optional[Dict[str, st
...
         for param in filter(is_scalar_field, params):
             try:
                 # if the target parameter is a scalar, we keep the first value of the query string
                 # regardless if there are more in the payload
-                query_string[param.alias] = query_string[param.alias][0]
+                resolved_query_string[param.alias] = query_string[param.alias][0]
             except KeyError:
                 pass
-    return query_string
+        return resolved_query_string
```
rubenfonseca, line 398: I've changed these lines to make typing work, but this is not how the bug was fixed. The bug was fixed by making sure all event handler types return `Dict[str, List[str]]` when resolving query string parameters.

aws_lambda_powertools/utilities/data_classes/api_gateway_proxy_event.py
```diff
@@ -319,14 +323,12 @@ def header_serializer(self):
...
+    def resolved_query_string_parameters(self) -> Optional[Dict[str, List[str]]]:
         if self.query_string_parameters is not None:
-            query_string = {
-                key: value.split(",") if "," in value else value for key, value in self.query_string_parameters.items()
-            }
+            query_string = {key: value.split(",") for key, value in self.query_string_parameters.items()}
             return query_string
 
-        return {}
+        return None
```
heitorlessa: perhaps `{}` so you remove additional checks on `None` and type annotations `Optional` upstream?

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -104,15 +104,19 @@ def query_string_parameters(self) -> Optional[Dict[str, str]]:
...
+    def resolved_query_string_parameters(self) -> Optional[Dict[str, List[str]]]:
         """
         This property determines the appropriate query string parameter to be used
         as a trusted source for validating OpenAPI.
 
         This is necessary because different resolvers use different formats to encode
         multi query string parameters.
         """
-        return self.query_string_parameters
+        if self.query_string_parameters is not None:
```
heitorlessa, line 115: maybe we keep in common to avoid forgetting a resolver when upgrading/improving the resolution?

tests/functional/event_handler/test_openapi_validation_middleware.py
```diff
@@ -1040,3 +1069,24 @@ def my_path(
 
     result = app(event, {})
     assert result["statusCode"] == 200
+
+
+def test_validation_with_http_single_param():
+    # GIVEN a HTTP API V2 proxy type event
+    app = APIGatewayHttpResolver(enable_validation=True)
+    event = deepcopy(LOAD_GW_EVENT_HTTP)
```
heitorlessa: can we perhaps use a plain object to make it more explicit we are testing a single queryString with multiple values?

the `LOAD_GW_EVENT_HTTP` or `load_event(...)` will hide that explicitness. When that happens, the test name will need adjustment too.

thank you <3

heitorlessa (changes requested):
tiny changes

heitorlessa (dismissed):
pre-approving, lemme know if more changes and I can approve again (I think only testing globals are left)

tests/functional/event_handler/test_openapi_validation_middleware.py
rubenfonseca: I've moved everything to pytest fixtures and put them on `conftest.py` so they can be reused on all the event handler tests.

rubenfonseca:
@heitorlessa reasking for a review since I did a lot of changes

heitorlessa (approved):

PR #3822, opened 2024-02-21
fix(idempotency): validate before saving to cache

### LINKED ISSUE #3821: Bug: idempotency logic should first validate the payload and then cache
### Expected Behaviour

The idempotency utility should only cache the results from the persistence layer if the retrieved record passes payload tampering validation.

### Current Behaviour

The results from the persistence layer are always written to the cache, even if it later fails tampering validation. This should be the other way around.

### Code snippet

```python
https://github.com/aws-powertools/powertools-lambda-python/blob/8ed384d6b58819843bcd9df8caeefa9e7001e450/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L386-L388
```

### Possible Solution

Change the logic	

### Steps to Reproduce

https://github.com/aws-powertools/powertools-lambda-python/blob/8ed384d6b58819843bcd9df8caeefa9e7001e450/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L386-L388

latest

3.12

### Packaging format used

PyPi

### Debugging logs

_No response_

heitorlessa:
**Issue number:** #3821

## Summary

As recently fixed in TS, we should be validating _(payload tampering)_ before saving to cache.

### Changes

* [x] Change order
* [x] Investigate redis tests failure
* [x] Add missing test to prevent future regression

**Unrelated changes**

* [x] Fix multi-processing serialization on race condition Redis test
  * [x] Add `multiprocess` dev dependency to use `dill` over `pickle` for multi-process serialization 

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/idempotency/persistence/base.py
+++ b/aws_lambda_powertools/utilities/idempotency/persistence/base.py
@@ -1,6 +1,7 @@
 """
 Persistence layers supporting idempotency
 """
+
 import datetime
 import hashlib
 import json
@@ -383,9 +384,9 @@ class BasePersistenceLayer(ABC):
 
         record = self._get_record(idempotency_key=idempotency_key)
 
+        self._validate_payload(data_payload=data, stored_data_record=record)
         self._save_to_cache(data_record=record)
 
-        self._validate_payload(data_payload=data, stored_data_record=record)
         return record
 
     @abstractmethod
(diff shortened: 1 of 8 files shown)


rubenfonseca (dismissed):

heitorlessa:
@roger-zhangg and @leandrodamascena unrelated to this PR, Redis race condition tests were failing due to serialization in multi-processing -- only on M1, but not on CI.

Diving into this I found that the issue is due to the standard library `multiprocessing` using `pickle` for serializing functions to run (the target!). Because you're using a local function (closure) to run an idempotent function in a multiprocessing scenario (you're rightly so), `pickle` fails to serialize randomly.

To fix it, I've added `multiprocess` library. Two quick notes for you:

1. Use `from multiprocess` instead of `from multiprocessing` (ING) to avoid this issue
2. Please spend a bit more time annotating tests with `GIVEN/WHEN/THEN` to make it easier to grasp later. This is a complex test and it was difficult to track it down what it was doing and later the cryptic error

image

sthulb (approved):

roger-zhangg:
Thanks @heitorlessa , for the finding and fix. I wasn't aware of this M1 behavior difference before

PR #3602, opened 2024-01-08
docs(event_handler): add bedrock agent resolver documentation

rubenfonseca:
**Issue number:** #3326

### Changes

> Please provide a summary of what's being changed

This PR adds documentation to the Bedrock Agent resolver feature.

Note: to allow sharing content between REST and Bedrock Agents resolvers, some content was moved to shared files and imported in both sides.

### User experience

> Please share what the user experience looks like before and after this change

image

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

- [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
- [x] I have performed a self-review of this change
- [x] Changes have been tested
- [x] Changes are documented
- [x] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

- [ ] Migration process documented
- [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #3860, opened 2024-02-27
fix(event_handler): OpenAPI schema version respects Pydantic version

### LINKED ISSUE #3731: Bug: OpenAPI schema invalid when using Pydantic v2 (latest)
### Expected Behaviour

Using `ApiGatewayResolver.get_openapi_json_schema` should produce a valid OpenAPI schema

### Current Behaviour

When using Pydantic version 2.6.1 (latest at the time of writing) the OpenAPI spec produced is invalid. This is _not_ the case when using Pydantic version 1.10.14.

Errors seen in schema:

```json
...
  "info": {
    "title": "TODO's API",
    "summary": "API to manage TODOs",
    "version": "1.0.0"
  },
...
```

> Structural error at info
> should NOT have additional properties
> additionalProperty: summary

**Note:** This error is present when using both v1 and v2 of Pydantic.

```json
...
      "Todo": {
        "properties": {
          "type": {
            "const": "ingest",
            "title": "Type"
          },
...
```

> Structural error at components.schemas.Todo.properties.type
> should NOT have additional properties
> additionalProperty: const

```json
...
          "id": {
            "anyOf": [
              {
                "type": "integer"
              },
              {
                "type": "null"
              }
            ],
...
```

> Structural error at components.schemas.TodoAttributes.properties.id.anyOf.1.type
> should be equal to one of the allowed values
> allowedValues: array, boolean, integer, number, object, string

### Code snippet

<details><summary>Python code: handler.py</summary>
<p>

```python
from typing import Literal

from aws_lambda_powertools.event_handler.openapi.models import Contact, Server
from aws_lambda_powertools.event_handler.openapi.types import OpenAPIResponse
from pydantic import BaseModel, Field

app.enable_swagger(path="/swagger", title="TODO")


class TodoAttributes(BaseModel):
    userId: int
    id_: int | None = Field(alias="id", default=None)
    title: str
    completed: bool


class Todo(BaseModel):
    type: Literal["ingest"]
    attributes: TodoAttributes


class TodoEnvelope(BaseModel):
    data: Todo


@app.get("/todos/<todo_id>", tags=["todo"])
def get_todo_by_id(todo_id: int) -> TodoEnvelope: ...


@app.post(
    "/todos",
    tags=["todo"],
    responses={
        204: OpenAPIResponse(
            description="Successful creation",
            content={"": {"schema": {}}},
        )
    },
)
def create_todo(todo: TodoEnvelope): ...


if __name__ == "__main__":
    print(
        app.get_openapi_json_schema(
            title="TODO's API",
            summary="API to manage TODOs",
        ),
    )
```

</p>
</details> 

### Possible Solution

_No response_

### Steps to Reproduce

Run the provided script: `python handler.py > swagger.json`

Then paste the JSON data in https://editor.swagger.io/ to see the errors (there are also tools like [`spectral`](https://github.com/stoplightio/spectral) which can be used to validate the output). 

latest

3.12

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3731

### Changes

This PR forces the OpenAPI schema version to follow the Pydantic version. Pydantic v1 only generates OpenAPI Schema 3.0.x, and Pydantic v2 only generates OpenAPI Schema 3.1.x.

List of tasks in this PR:

* Determines the default OpenAPI schema version according to the installed Pydantic version
* Emits a warning if the user requests an OpenAPI schema version that is incompatible with the current installed Pydantic version
* Adapts the schema to version 3.0.x and 3.1.x, removing things that are unnecessary
* Add automated tests for both pydantic v1 and v2 that generate simple and complex schemas, and validate them against the official OpenAPI spec (using fastjsonschema)
* Refactored the pytest fixtures to only load them once per session, making tests much faster

Pending items:
- [x] Change event handler documentation to explain when the different schema versions are generated
- [ ] ~Updated Bedrock Agents documentation to explain why it will not work if you use Pydantic v2 (Agents for Amazon Bedrock seems to only support 3.0.x OpenAPI schemas)~ moved this checklist to the [Bedrock Agents PR](https://github.com/aws-powertools/powertools-lambda-python/pull/3602)
- [x] Upgrade Swagger-UI to support 3.1 OpenAPI schemas

### User experience

After this change, the generated OpenAPI schema should always validate correctly against the official schemas

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


kbakk:
@rubenfonseca Note that the swagger-ui version that aws-lambda-powertools ships (4.x) doesn't support OpenAPI schema version 3.1.x – you'll get an error when using it (the entire ui is inaccessible):

> `Please indicate a valid Swagger or OpenAPI version field. Supported version fields are swagger: "2.0" and those that match openapi: 3.0.n (for example, openapi: 3.0.0).`

It does seem like passing `swagger_base_url="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/5.11.8"` to `enable_swagger` mitigates this (swagger-ui is then showing), so you might want to consider updating swagger-ui as well. 

See compatibility table here: https://github.com/swagger-api/swagger-ui?tab=readme-ov-file#compatibility

rubenfonseca:
> @rubenfonseca Note that the swagger-ui version that aws-lambda-powertools ships (4.x) doesn't support OpenAPI schema version 3.1.x – you'll get an error when using it (the entire ui is inaccessible):
> 
> > `Please indicate a valid Swagger or OpenAPI version field. Supported version fields are swagger: "2.0" and those that match openapi: 3.0.n (for example, openapi: 3.0.0).`
> 
> It does seem like passing `swagger_base_url="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/5.11.8"` to `enable_swagger` mitigates this (swagger-ui is then showing), so you might want to consider updating swagger-ui as well.
> 
> See compatibility table here: [swagger-api/swagger-ui?tab=readme-ov-file#compatibility](https://github.com/swagger-api/swagger-ui?tab=readme-ov-file#compatibility)

This is awesome feedback, thank you so much @kbakk !

rubenfonseca:
I've updated Swagger-UI to 5.x and verified that it works when using Pydantic v2 (which generated a 3.1.0 schema). @leandrodamascena this is now ready for review

docs/core/event_handler/api_gateway.md
```diff
@@ -961,6 +961,9 @@ This will enable full tracebacks errors in the response, print request and respo
 
 When you enable [Data Validation](#data-validation), we use a combination of Pydantic Models and [OpenAPI](https://www.openapis.org/){target="_blank"} type annotations to add constraints to your API's parameters.
 
+???+ warning "OpenAPI schema version depends on the installed version of Pydantic"
+    Pydantic v1 generates valid OpenAPI 3.0.3 schemas, and Pydantic v2 generates valid OpenAPI 3.1.0 schemas.
+
```
leandrodamascena: @rubenfonseca this is just a suggestion... what do you think about adding the documentation link to make it explicit why we are forcing OpenAPI versions?

https://docs.pydantic.dev/latest/why/#json-schema

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1455,10 +1455,25 @@ def get_openapi_schema(
...
+                "You are using Pydantic v2, which is incompatible with OpenAPI schema 3.0. Forcing OpenAPI 3.1",
+                stacklevel=2,
+            )
+            openapi_version = "3.1.0"
+        elif not PYDANTIC_V2 and not openapi_version.startswith("3.0"):
+            warnings.warn(
+                "You are using Pydantic v1, which is incompatible with OpenAPI schema 3.1. Forcing OpenAPI 3.0",
+                stacklevel=2,
+            )
+            openapi_version = "3.0.3"
```
leandrodamascena, line 1475: Our default OpenAPI version is [3.0.0](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/openapi/constants.py#L2), so what makes more sense: Changing this to 3.0.0 or changing our default to 3.0.3? I think we should have consistency in the default versions.

docs/core/event_handler/api_gateway.md
```diff
@@ -1042,18 +1045,18 @@ Below is an example configuration for serving Swagger UI from a custom path or C
...
+| ------------------ | -------------- |------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `title`            | `str`          | The title for your API. It should be a concise, specific name that can be used to identify the API in documentation or listings.                                                                                       |
+| `version`          | `str`          | The version of the API you are documenting. This could reflect the release iteration of the API and helps clients understand the evolution of the API.                                                                 |
+| `openapi_version`  | `str`          | Specifies the version of the OpenAPI Specification on which your API is based. When using Pydantic v1 it defaults to 3.0.3, and when using Pydantic v2, it defaults to 3.1.0.                                          |
```
leandrodamascena, line 1052: Our default OpenAPI version is [3.0.0](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/openapi/constants.py#L2), so what makes more sense: Changing this to 3.0.0 or changing our default to 3.0.3? I think we should have consistency in the default versions.

leandrodamascena (changes requested):
Hey @rubenfonseca! I left some comments and we need to resolve them before I review this PR again and merge. Thanks for fixing this issue.

aws_lambda_powertools/event_handler/api_gateway.py
rubenfonseca: Good point, changed the default to 3.0.3

docs/core/event_handler/api_gateway.md
rubenfonseca: Changed the default to 3.0.3

leandrodamascena (approved):
APPROVED!

sthulb (approved):

PR #3995, opened 2024-03-21
fix(event_handler): always add 422 response to the schema

### LINKED ISSUE #3600: Feature: Event handler Swagger doesnt add 422 response to openapi when adding custom responses
### Expected Behaviour

422 is a built in schema that i as a user don't control, i expect it to be there even though I'm adding other responses (501 , 404 etc.).

### Current Behaviour

422 is omitted from schema when adding a custom response.
In addition, the 422 schema says there's a msg: string in the schema but in my tests i didnt get that fields, maybe it's optional?

### Code snippet

```python
See:
https://github.com/ran-isenberg/aws-lambda-handler-cookbook/pull/784/files#diff-03a550a2c5158da1b1459e42e5799bdfe3809d2563dfb16014d3ea3bcd7cd043R20
```

### Possible Solution

_No response_

### Steps to Reproduce

add the snippet, go to /swagger

latest

3.12

### Packaging format used

PyPi

### Debugging logs

_No response_

rubenfonseca:
**Issue number:** #3600

### Changes

This PR makes sure that the 422 error is always added to the OpenAPI schema, even when the user is customizing the response.
It also fixes the default schema of the 422 error to not have a "msg" field, for security reasons.
Also took the opportunity to simplify the code to make it easier to read.

### User experience

After this PR, the user can still customize the openapi responses, while maintaining the 422 error that is generated when there is a validation error.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/openapi/types.py
+++ b/aws_lambda_powertools/event_handler/openapi/types.py
@@ -28,7 +28,7 @@ validation_error_definition = {
             "type": "array",
             "items": {"anyOf": [{"type": "string"}, {"type": "integer"}]},
         },
-        "msg": {"title": "Message", "type": "string"},
+        # For security reasons, we hide **msg** details (don't leak Python, Pydantic or filenames)
         "type": {"title": "Error Type", "type": "string"},
     },
     "required": ["loc", "msg", "type"],
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -503,8 +503,23 @@ class Route:
             if request_body_oai:
                 operation["requestBody"] = request_body_oai
 
+        # Validation failure response (422) will always be part of the schema
+        default_responses: Dict[int, OpenAPIResponse] = {
+            422: {
+                "description": "Validation Error",
+                "content": {
+                    "application/json": {
+                        "schema": {"$ref": COMPONENT_REF_PREFIX + "HTTPValidationError"},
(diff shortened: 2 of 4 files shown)


aws_lambda_powertools/event_handler/openapi/types.py
```diff
@@ -28,7 +28,7 @@
             "type": "array",
             "items": {"anyOf": [{"type": "string"}, {"type": "integer"}]},
         },
-        "msg": {"title": "Message", "type": "string"},
+        # For security reasons, we hide **msg** details (don't leak Python, Pydantic or filenames)
```
rubenfonseca, line 31: This follows what we've done here https://github.com/aws-powertools/powertools-lambda-python/blob/0082a670f8a212ed654dbcf3c9ee06133429b7cd/aws_lambda_powertools/event_handler/api_gateway.py#L2066

leandrodamascena (changes requested):
Hey @rubenfonseca! The PR appears to be in a very good shape, with all functionality working as expected. I'm wondering if it's beneficial to include additional tests to safeguard against unintentional overriding of the client-defined 422 response when setting it specifically in the route definition. 
What do you think?

rubenfonseca:
> additional tests to safeguard against unintentional overriding of the client-defined 422 response when setting it specifically in

great idea

leandrodamascena (approved):
APPROVED!

PR #4005, opened 2024-03-22
fix(tracer): add name sanitization for X-Ray subsegments

### LINKED ISSUE #3981: Bug: Generates valid subsegment names that comply with X-Ray SDK restrictions
### Expected Behaviour

When decorating a method with `tracer.capture_method`, it should generate names that are [compatible with the AWS X-Ray Python SDK](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/entity.py#L18C1-L18C32).

### Current Behaviour

In some circumstances, it generates subsegment names that are [incompatible with the AWS X-Ray Python SDK](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/entity.py#L18C1-L18C32). This will continually cause AWS X-Ray Python SDK to produce a [WARNING log message](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/entity.py#L43).

### Code snippet

```python
def get_inner_func():
    @tracer.capture_method
    def inner_func():
        pass

    return inner_func

if __name__ == "__main__":
    method = get_inner_func()
    print(f"{method.__module__}.{method.__qualname__}")
```

OUTPUTS:
```
__main__.get_inner_func.<locals>.inner_func
```

which is an invalid name for a segment and generates a warning.

### Possible Solution

The name of the subsegment is taken from the method name using [method_name = f"{method.__module__}.{method.__qualname__}"](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/tracing/tracer.py#L523) but does nothing to sanitize the result to make it compatible with AWS X-Ray or AWS X-Ray Python SDK.

[AWS X-Ray docs](https://docs.aws.amazon.com/xray/latest/devguide/xray-api-segmentdocuments.html) say the name can be:
> The logical name of the service that handled the request, up to 200 characters. For example, your application's name or domain name. Names can contain Unicode letters, numbers, and whitespace, and the following symbols: _, ., :, /, %, &, #, =, +, \\, -, @

[AWS X-Ray Python SDK](https://github.com/aws/aws-xray-sdk-python/blob/master/aws_xray_sdk/core/models/entity.py#L18C1-L18C32) strips the following characters:
> _common_invalid_name_characters = '?;*()!$~^<>'

From [PEP 3155 - Qualified name for classes and functions](https://peps.python.org/pep-3155/#proposal)
> This PEP proposes the addition of a __qualname__ attribute to functions and classes. For top-level functions and classes, the __qualname__ attribute is equal to the __name__ attribute. For nested classes, methods, and nested functions, the __qualname__ attribute contains a dotted path leading to the object from the module top-level. A function’s local namespace is represented in that dotted path by a component named \<locals\>.

In our case, our segment is being affected by "A function’s local namespace is represented in that dotted path by a component named \<locals\>." where the "<" and ">" are invalid characters and raising a WARNING in our logs.

Add logic to sanitize the value of `method_name` to be compatible with AWS X-Ray or AWS X-Ray Python SDK by restricting the length to 200 characters and removing any invalid character.

### Steps to Reproduce

Create a locally defined function and decorate it with `@tracer.capture_method` and receive WARNING in logs.

2.26.0

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #3981 

### Changes

This pull request addresses the concern raised by @loesak regarding subsegment names generated when using our Tracer utility. While there's currently no loss of tracer or subsegment data, it's crucial to ensure that the generated names comply with the X-Ray specification to prevent any potential issues or warnings.

### User experience

By implementing name sanitization, we'll ensure that X-Ray no longer emits warnings due to non-compliant subsegment names. This bug fix will result in customers no longer receiving X-Ray warnings.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/shared/functions.py
```diff
@@ -275,13 +275,11 @@ def abs_lambda_path(relative_path: str = "") -> str:
         If the path is empty, it will return the current working directory.
     """
     # Retrieve the LAMBDA_TASK_ROOT environment variable or default to an empty string
-    current_working_directory = os.environ.get("LAMBDA_TASK_ROOT", "")
+    current_working_directory = os.environ.get("LAMBDA_TASK_ROOT", "") or str(Path.cwd())
```
rubenfonseca, line 279:
```suggestion
    current_working_directory = os.environ.get("LAMBDA_TASK_ROOT", str(Path.cwd()))
```

Do we really need to "or" ?

aws_lambda_powertools/shared/functions.py
```diff
@@ -275,13 +275,11 @@ def abs_lambda_path(relative_path: str = "") -> str:
...
-    if not current_working_directory:
-        current_working_directory = str(Path.cwd())
+    return str(Path(current_working_directory, relative_path))
 
-    # Combine the current working directory and the relative path to get the absolute path
-    absolute_path = str(Path(current_working_directory, relative_path))
 
-    return absolute_path
+def sanitize_xray_segment_name(name: str) -> str:
+    _common_invalid_name_characters = "?;*()!$~^<>"
```
rubenfonseca, line 284: Can we move this to a constant so it's re-evaluated on every call?

aws_lambda_powertools/shared/functions.py
```diff
@@ -275,13 +275,11 @@ def abs_lambda_path(relative_path: str = "") -> str:
...
+    return "".join(char for char in name if char not in _common_invalid_name_characters)
```
rubenfonseca: Actually, now that I'm thinking on the code, I believe this would be faster:

```python
INVALID_CHARS_PATTERN = r"[?;*()!$~^<>]"
def sanitize_xray_segment_name(name: str) -> str:
    return re.sub(invalid_chars_pattern, "", name)
 ```

Comment (rubenfonseca) [inline aws_lambda_powertools/shared/functions.py:285]: Actually, now that I'm thinking on the code, I believe this would be faster:

```python
INVALID_CHARS_PATTERN = r"[?;*()!$~^<>]"
def sanitize_xray_segment_name(name: str) -> str:
    return re.sub(invalid_chars_pattern, "", name)
 ```

**Review by rubenfonseca — CHANGES_REQUESTED** [review]

File: aws_lambda_powertools/shared/functions.py (hunk 1 above)
Comment (leandrodamascena) [reply aws_lambda_powertools/shared/functions.py:279]: Good catch!!!

File: aws_lambda_powertools/shared/functions.py (hunk 3 above)
Comment (leandrodamascena) [reply aws_lambda_powertools/shared/functions.py]: I'm not sure about this. With the `re` approach, we need to import the module and the execution time of `re.sub` might be slightly slower compared to a simple iteration.

Look at these very simple example.

#### re_test.py
```python
import re

INVALID_CHARS_PATTERN = r"[?;*()!$~^<>]"
def sanitize_xray_segment_name(name: str) -> str:
    return re.sub(INVALID_CHARS_PATTERN, "", name)

name = sanitize_xray_segment_name("ijheirkhwekwd098;92093kjwd()!kjnsdkjwhjewjieowueuwuieuwioe~^<>")
```
image


#### for_test.py
```python
INVALID_CHARS_PATTERN = r"[?;*()!$~^<>]"
def sanitize_xray_segment_name(name: str) -> str:
    return "".join(char for char in name if char not in INVALID_CHARS_PATTERN)

name = sanitize_xray_segment_name("ijheirkhwekwd098;92093kjwd()!kjnsdkjwhjewjieowueuwuieuwioe~^<>")
```
image

I think we can use iteration instead of regex.

rubenfonseca: Sure, but can you run the example 1000 times? It's unrealistic to only measure the cold start :)

aws_lambda_powertools/shared/functions.py
leandrodamascena: Ops sorry! I think we need to keep the `or` because `LAMBDA_TASK_ROOT` may exist, but it may be an empty string. Keeping as is.

aws_lambda_powertools/shared/functions.py
leandrodamascena: Yeah, sure! Using hyperfine..

```sh
❯ hyperfine --runs 1000  'python for_test.py' -i
Benchmark 1: python for_test.py
  Time (mean ± σ):       7.4 ms ±   0.7 ms    [User: 5.8 ms, System: 1.5 ms]
  Range (min … max):     6.3 ms …  15.6 ms    1000 runs
```
 
```sh
❯ hyperfine --runs 1000  'python re_test.py' -i
Benchmark 1: python re_test.py
  Time (mean ± σ):      11.5 ms ±   0.8 ms    [User: 9.4 ms, System: 2.0 ms]
  Range (min … max):    10.0 ms …  20.7 ms    1000 runs
```

rubenfonseca: Code:

```python
import re

INVALID_CHARS_PATTERN = "?;*()!$~^<>"
INVALID_CHARS_RE_PATTERN = r"[?;*()!$~^<>]"


def using_for(name: str) -> str:
    return "".join(char for char in name if char not in INVALID_CHARS_PATTERN)


def using_re(name: str) -> str:
    return re.sub(INVALID_CHARS_RE_PATTERN, "", name)


if __name__ == "__main__":
    import timeit

    print(
        timeit.timeit("using_for('ijheirkhwekwd098;92093kjwd()!kjnsdkjwhjewjieowueuwuieuwioe~^<>')", globals=locals()),
    )
    print(timeit.timeit("using_re('ijheirkhwekwd098;92093kjwd()!kjnsdkjwhjewjieowueuwuieuwioe~^<>')", globals=locals()))

```

Results:

```
✦ ❯ python bench.py
2.6501595829613507
1.1926250829128549
```

using re.sub vs iteration is **at least** two times faster

leandrodamascena: Thank you @rubenfonseca! I changed the function.

rubenfonseca (approved):

PR #4006, opened 2024-03-22
feat(metrics): allow custom timestamps for metrics

### LINKED ISSUE #3429: Feature request: allow custom timestamps for metrics
### Use case

I'm creating a lambda to report metrics from an external source to Cloudwatch for observability within AWS. I'm looking for a way to set a custom timestamp for the metrics emitted each invocation.

The metrics I'm querying from the external source will basically be a list of (metric name, value) tuples for a specific time. This specific time will be in the recent past, maybe a few minutes old, but I'm not sure that matters here.

I can easily add the key and values to metrics, however I cannot set an explicit time because Metrics does not expose a way and the the time ultimately always set to [`datetime.now()`](https://github.com/aws-powertools/powertools-lambda-python/blob/56c2ee846124895ce3de19011aec3424ff0e0712/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L233) when rendered to emf. 

### Solution/User Experience

I'm imagining a simple api to set the timestamp on the Metrics object. If a timestamp is not specific then the current time should be used.

```
metrics = Metrics()

@metrics.log_metrics 
    time = ...
    metric_values: list[(string, string)] = ...
    for (name, value) in metric_values:
            metrics.add_metric(name=name unit=MetricUnit.Count, value=value)
   metrics.set_timestamp(time) # <---
```

### Alternative solutions

[aws-embedded-metrics-python](https://github.com/awslabs/aws-embedded-metrics-python) supports setting the timestamp on their MetricLogger class. 

### Acknowledgment

leandrodamascena:
**Issue number:** #3429 

### Changes

This pull request adds a functionality to allow custom dates instead of the current timestamp for Metrics EMF, while still retaining epoch time as expected value. Previously, only the current timestamp was used by default.

### Questions

1 - Timestamp type
> According to the [AWS Documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html) timestamp is a number representing the time stamp used for metrics extracted from the event. Values MUST be expressed as the number of milliseconds after Jan 1, 1970 00:00:00 UTC.

That said, should we only allow `int` inputs with epoch value, or should customers be able to pass datetime objects and have them converted internally? In our Datadog Metrtics provider we just allow `int` values.

2 - Invalid values
If a customer passes an invalid number, we will emit a warning and fallback to the current timestamp instead of raising an exception.

### Update 26/03/2024

1 - Timestamp is either a `Datetime` object or an integer representing an epoch time.
2 - Any `Datetime` object will be converted into epoch time.
3 - The `timestamp` should not exceed 14 days in the past or be more than 2 hours in the future. Reference: [Amazon CloudWatch Documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html)
4 - If any metric fails to meet the requirements, a warning will be raised, and the metric will be flushed accordingly.
5 - We will not revert the timestamp to the current timestamp under any circumstances to prevent potential side effects.

### User experience

```python
import datetime

metrics = Metrics(namespace="Powertools")


@metrics.log_metrics 

    # Calculate the metric timestamp as 2 days before the current time
    metric_timestamp = int((datetime.datetime.now() - datetime.timedelta(days=2)).timestamp() * 1000)
    metrics.set_timestamp(metric_timestamp)
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena:
Hello everyone! This PR is ready for review. Here are some notes to facilitate the review process:

1 - Timestamp is either a `Datetime` object or an integer representing an epoch time.
2 - Any `Datetime` object will be converted into epoch time.
3 - The `timestamp` should not exceed 14 days in the past or be more than 2 hours in the future. Reference: [Amazon CloudWatch Documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/AgentReference.html)
4 - If any metric fails to meet the requirements, a warning will be raised, and the metric will be flushed accordingly.
5 - We will not revert the timestamp to the current timestamp under any circumstances to prevent potential side effects.

@rubenfonseca I need your review here.

@dreamorosi considering your additional experience with metrics and EMF, I'd appreciate it if you could take a look at this PR and provide any thoughts or feedback you have.

aws_lambda_powertools/metrics/functions.py
```diff
@@ -69,3 +72,63 @@ def extract_cloudwatch_metric_unit_value(metric_units: List, metric_valid_option
         unit = unit.value
 
     return unit
+
+
+def validate_emf_timestamp(timestamp: int | datetime) -> bool:
```
dreamorosi, line 77: Is this validation time zone aware? Do we know which time zone CloudWatch uses to decide whether to drop or keep the metric?

I see that you're using UTC by default, which I think is sensible, however what happens if the customer has changed the `TZ` value?

dreamorosi (commented):
Left a comment about time zones, just want to make sure we are covering this aspect since we support it in other utilities.

Other than that, it looks great!

rubenfonseca:
Reviewing now

aws_lambda_powertools/metrics/base.py
```diff
@@ -296,6 +299,21 @@ def add_metadata(self, key: str, value: Any) -> None:
         else:
             self.metadata_set[str(key)] = value
 
+    def set_timestamp(self, timestamp: int | datetime.datetime):
+        # The timestamp must be a Datetime object or an integer representing an epoch time.
+        # This should not exceed 14 days in the past or be more than 2 hours in the future.
+        # any metrics failing to meet this criteria will be skipped by Amazon CloudWatch.
```
rubenfonseca:
```suggestion
        # Any metrics failing to meet this criteria will be skipped by Amazon CloudWatch.
```

aws_lambda_powertools/metrics/functions.py
```diff
@@ -69,3 +72,63 @@ def extract_cloudwatch_metric_unit_value(metric_units: List, metric_valid_option
...
+    bool
+        Valid or not timestamp values
+    """
+
+    if not isinstance(timestamp, (int, datetime)):
+        return False
+
+    if isinstance(timestamp, datetime):
+        # Converting timestamp to integer
+        timestamp = int(timestamp.timestamp() * 1000)
```
rubenfonseca: I'm not sure you're covering the scenario where a use passes a epoch "int". I don't see it being converted to milliseconds anywhere, as you are doing here when the user passes a "datetime". Can you please check?

aws_lambda_powertools/metrics/functions.py
```diff
@@ -69,3 +72,63 @@ def extract_cloudwatch_metric_unit_value(metric_units: List, metric_valid_option
...
+        The timestamp to convert. If already in milliseconds format, returns it as is.
+        If datetime object, converts it to milliseconds since Unix epoch.
+
+    Returns:
+    --------
+    int
+        The timestamp converted to EMF compatible format (milliseconds since Unix epoch).
+    """
+    if isinstance(timestamp, int):
+        return timestamp
```
rubenfonseca, line 129: I believe use is passing an epoch here (in seconds) and this needs to be converted to milliseconds

rubenfonseca (changes requested):
I thing you're treating "ints" as seconds and converting "datetimes" to milliseconds. I believe everything needs to be milliseconds. Can you please check?

aws_lambda_powertools/metrics/functions.py
leandrodamascena: Thanks for catching that specific and critical detail, @dreamorosi! We should definitely account for timezones. I've updated the code accordingly

leandrodamascena:
> I thing you're treating "ints" as seconds and converting "datetimes" to milliseconds. I believe everything needs to be milliseconds. Can you please check?

Thanks for the feedback @rubenfonseca! I've updated the documentation, docstring, and comments to make it clear that we expect epoch time in milliseconds. 

Ready to review again.

aws_lambda_powertools/metrics/functions.py
leandrodamascena: Changed

dreamorosi (approved):

aws_lambda_powertools/metrics/functions.py
```diff
@@ -69,3 +72,66 @@ def extract_cloudwatch_metric_unit_value(metric_units: List, metric_valid_option
...
+
+    if not isinstance(timestamp, (int, datetime)):
+        return False
+
+    if isinstance(timestamp, datetime):
+        # Converting timestamp to epoch time in milliseconds
+        timestamp = int(timestamp.timestamp() * 1000)
+
+    # Consider current timezone when working with date and time
+    current_timezone = datetime.now().astimezone().tzinfo
```
rubenfonseca, line 104: I'm thinking that this line could be moved outside the function, as it will always be the same during a Lambda session. I've done a quick benchmark and calling `astimezone()` twice on each call is by far the most expensive part of the operation. What do you think?

docs/core/metrics.md
```diff
@@ -224,14 +239,15 @@ You can add high-cardinality data as part of your Metrics log with `add_metadata
     --8<-- "examples/metrics/src/add_metadata_output.json"
     ```
 
-### Single metric with a different dimension
+### Single metric
 
-CloudWatch EMF uses the same dimensions across all your metrics. Use `single_metric` if you have a metric that should have different dimensions.
+CloudWatch EMF uses the same dimensions and timestamp across all your metrics. Use `single_metric` if you have a metric that should have different dimensions or timestamp.
```
rubenfonseca, line 244: This sounds weird to me. If I just use the metrics powertools module without specifying any timestamp, and my function takes 60 seconds, will all the metrics generated have the same millisecond?

rubenfonseca (changes requested):
Two small notes, feel free to ping me offline

aws_lambda_powertools/metrics/functions.py
dreamorosi: What if for some reason the customer sets a different timezone at runtime? For example, in the case of a multi-tenant application.

Do you think it’d be less expensive to move the function outside and initialize the value, but then within the function check if the `TZ` value has changed, and only thenupdate the `current_timezone` value?

rubenfonseca: Ah mutable global TZ settings, what can possibly go wrong? :D 

Yes, you're right, feel free to ignore my request @leandrodamascena!

docs/core/metrics.md
dreamorosi: If I understand [the spec](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html) correctly, they will have the timestamp of the time the EMF blob was created & flushed out of the buffer. Unless you’re using the `SingleMetric` - which is flushed immediately - when calling `add_metric` we put these metric in a buffer. 

At that stage the metric doesn’t have any timestamp but only the values for name, unit, etc. then, later, when they are flushed either manually by calling the appropriate method or automatically by the decorator logic, the EMF blob is emitted:

```json
{
  "_aws": {
    "Timestamp": 1559748430481,
    "CloudWatchMetrics": [ ... ]
  }
}
```

Dimensions behave similarly in the sense that they’re not specific to a single metric but apply to the entire EMF blob:

```json
{
  "_aws": {
    "Timestamp": 1559748430481,
    "CloudWatchMetrics": [
      {
        "Dimensions": [ [ "functionVersion" ] ],
        "Namespace": "serverlessAirline ",
        "Metrics": [ … ] <- actual metrics from add_metric go here
      }
    ]
  }
}
```

aws_lambda_powertools/metrics/functions.py
leandrodamascena: >Ah mutable global TZ settings, what can possibly go wrong? :D

"Beautiful and scary things in Python - Part 1" haha.

Yeah, I agree with @dreamorosi! Let's keep it as is.

dreamorosi: I think if we cache both `current_timezone` (which is the expensive part) and `TZ` outside of the function, we could shave some ms by checking `TZ` against `os.getEnv(‘TZ’)` and running `datetime.now().astimezone().tzinfo` again only if it changed.

Although that would make the implementation a bit more complex..

docs/core/metrics.md
rubenfonseca: Great point. I guess the only thing to think about (outside of the scope) is when we are batching a big number of metrics, we are generating a slightly different timestamp for each batch. But should all be inside the same second (thinking of high resolution metrics) so it should be fine.

rubenfonseca (approved):

PR #4037, opened 2024-03-28
feat(Idempotency): add feature for manipulating idempotent responses

### LINKED ISSUE #2164: Manipulation of Idempotent response
### Use case

We have a client-server web application.
The server side is built using AWS SAM with an API Gateway and multiple lambda functions. The server side laso executes an asynchronous pipeline.

We want to prevent cases where a duplicated message is sent to the backend and triggers the pipeline multiple time.
We also want to notify the user if a duplication occurred, and hence his request was not processed, this is the main issue.

### Solution/User Experience

Add an `X-Idempotent: True` header to the response. 

### Alternative solutions

A hook for changing the response (e.g. adding a field).

### Acknowledgment

- [x] Should this be considered in other Lambda Powertools languages? i.e. [Java](https://github.com/awslabs/aws-lambda-powertools-java/), [TypeScript](https://github.com/awslabs/aws-lambda-powertools-typescript/)

walmsles:
**Issue number:** #2164

## Summary

I have added the capability for developers to provide a response hook, which is called when the idempotency handler finds a valid idempotent response.

### Changes

Addition of "response_hook" in IdempotentConfig class to enable developers to register a repsonse_hook

### User experience

```python
import datetime
import uuid
from typing import Dict

    DynamoDBPersistenceLayer,
    IdempotencyConfig,
    idempotent_function,
)
from aws_lambda_powertools.utilities.idempotency.persistence.base import (
    DataRecord,
)

logger = Logger()

def my_response_hook(response: Dict, idempotent_data: DataRecord) -> Dict:
    # Return inserted Header data into the Idempotent Response
    response["x-idempotent-key"] = idempotent_data.idempotency_key

    # expiry_timestamp could be None so include if set
    expiry_timestamp = idempotent_data.expiry_timestamp
    if expiry_timestamp:
        expiry_time = datetime.datetime.fromtimestamp(int(expiry_timestamp))
        response["x-idempotent-expiration"] = expiry_time.isoformat()

    # Must return the response here
    return response

dynamodb = DynamoDBPersistenceLayer(table_name="IdempotencyTable")
config = IdempotencyConfig(response_hook=my_response_hook)

@idempotent_function(data_keyword_argument="order", config=config, persistence_store=dynamodb)
def process_order(order: dict) -> dict:
    # create the order_id
    order_id = str(uuid.uuid4())

    # create your logic to save the order
    # append the order_id created
    order["order_id"] = order_id

    # return the order
    return {"order": order}


    config.register_lambda_context(context)  # see Lambda timeouts section
    try:
        logger.info(f"Processing order id {event.get('order_id')}")
        return process_order(order=event.get("order"))
    except Exception as err:
        return {"status_code": 400, "error": f"Erro processing {str(err)}"}
```
#### Response hooks
We are introducing support for hooks in the Idempotency utility. Customers can now create a function and pass it as the hook in the IdempotencyConfig. This hook will be invoked every time a request is idempotent. If the request is not idempotent, the hook will not execute.

```python
from aws_lambda_powertools.utilities.idempotency import IdempotencyConfig
def my_response_hook(response: Dict, idempotent_data: DataRecord): ...
config = IdempotencyConfig(response_hook=my_response_hook)
```

#### Manipulating response
This hook will be able to access both: the response from the invocation and the `DataRecord` from the Idempotency Store. This means customers can inject new keys into the response or utilize attributes from the DataRecord to log information or create metrics, for example.

#### Persistent Store
This hook is compatible with any Persistent Store. Currently, we support DynamoDB and Redis for this implementation.

#### Handling exceptions
Customers are responsible for handling exceptions when using Response hooks. During the implementation of this new feature, we considered the possibility of implementing soft failure or raising warnings to ignore hook execution. For instance, this approach could prevent customers from having exceptions in the hook, thereby avoiding potential malfunctions in the Lambda function execution. However, we opted against this approach because it could result in unexpected behaviors and potentially breaking the contract that customers expect for `Response hooks`. Let me provide some context with an example.
Consider a scenario where a customer creates a hook to inject a specific key into the response when the operation is idempotent and wants to use it to trigger an alert for third-party integration. The code might appear like this:

```python
def my_response_hook(response: Dict, idempotent_data: DataRecord) -> Dict:
    # Return inserted Header data into the Idempotent Response
    response["x-idempotent-key"] = idempotent_data.idempotency_key

    imagine_an_exception_here()

    # expiry_timestamp could be None so include if set
    expiry_timestamp = idempotent_data.expiry_timestamp
    if expiry_timestamp:
        expiry_time = datetime.datetime.fromtimestamp(int(expiry_timestamp))
        response["x-idempotent-expiration"] = expiry_time.isoformat()

    # Must return the response here
    return response

dynamodb = DynamoDBPersistenceLayer(table_name="IdempotencyTable")
config = IdempotencyConfig(response_hook=my_response_hook)

@idempotent_function(data_keyword_argument="order", config=config, persistence_store=dynamodb)
def process_order(order: dict) -> dict: ...

    config.register_lambda_context(context)  # see Lambda timeouts section
    process_order = process_order(event.get("order"))

    if process_order["x-idempotent-key"] is not None:
        trigger_some_sns_alert()
```

If we choose to ignore the hook execution exception and only raise a warning instead of propagating the exception, customers may encounter difficulties in debugging. They may struggle to understand why the `if process_order["x-idempotent-key"] is not None:` condition is failing and why the `x-idempotent-key` doesn't exist in the response.

#### Documentation
The documentation will be updated to inform customers on the best practices for using Response hooks. Additionally, it will feature an example demonstrating how to utilize Response hooks and effectively handle exceptions.

## Checklist

* [x] Changes are documented

<details>
<summary>Is this a breaking change?</summary> No, not a breaking change.  Configuration is optional and implementation checked for existence of the hook

**RFC issue number**: 2164

Checklist:

</details>

## Acknowledgment


walmsles:
Still need to work on documentation - will do this next.

leandrodamascena:
Hi @walmsles! Your contributions continue to shine brightly. Access to the Idempotency operation results has long been on our backlog and I'm so happy that we will add this new feature. If you need any assistance with documentation, testing, or anything else, please don't hesitate to let me know.

Thanks

walmsles:
I added documentation for review. I also changed the order so that response_hook processes **after** the custom de-serializer if it is used. This makes the most sense to developers so the response_hook is passed an expected structure rather than a raw event format.

aws_lambda_powertools/utilities/idempotency/__init__.py
```diff
@@ -2,6 +2,7 @@
 Utility for adding idempotency to lambda functions
 """
 
+from aws_lambda_powertools.utilities.idempotency.config import IdempotentHookData, IdempotentHookFunction
```
leandrodamascena, line 5: Would it be better to move these classes to `hook.py` or something like that?

aws_lambda_powertools/utilities/idempotency/config.py
```diff
@@ -1,8 +1,45 @@
...
+    def idempotency_key(self) -> str:
+        return self._idempotency_key
+
+    @property
+    def status(self) -> str:
+        return self._status
+
+    @property
+    def expiry_timestamp(self) -> Optional[int]:
+        return self._expiry_timestamp
```
leandrodamascena: I'm considering adding a field named `raw_data_record` to return the `data_record` as JSON. We can't predict if the customer might access  `in_progress_expiry_timestamp` or `payload_hash`, for instance. Make sense?

aws_lambda_powertools/utilities/idempotency/__init__.py
leandrodamascena: Would it be better to move these classes to `hook.py` or something like that?

docs/utilities/idempotency.md
```diff
@@ -909,6 +910,22 @@ You can create your own persistent store from scratch by inheriting the `BasePer
 
     For example, the `_put_record` method needs to raise an exception if a non-expired record already exists in the data store with a matching key.
 
+### Modifying the Idempotent Repsonse
```
leandrodamascena: Not everyone will want to modify the response. Perhaps `Accessing the Idempotent Response` would be more appropriate?

leandrodamascena (changes requested):
Hey, @walmsles! I looked at the code and ran some tests, and everything looked good to me at first glance! I left some comments for us to discuss before a final review. Thank you so much!

aws_lambda_powertools/utilities/idempotency/__init__.py
walmsles: Maybe let's jump on a call - placed where they are to avoid circular references between base.py, config.py and hook.py - happy for guidance  to improve - that's where they started.

leandrodamascena: Sure thing! Let's schedule a call to review and potentially merge this PR before the next release. I'll ping you on Discord and we can find a slot to join in a meeting.

Just out of curiosity, why would we have issues with circular references here? We're importing from different files, so it shouldn't be a problem. I might be missing something, but I can't see any potential problems with circular references here.

I've created a patch with this modification, and all tests are passing, mypy is also happy, and I couldn't spot any issues with circular references. 
Could you please try applying this patch in your local environment?

[hook_patch.txt](https://github.com/aws-powertools/powertools-lambda-python/files/14844033/hook_patch.txt)

aws_lambda_powertools/utilities/idempotency/base.py
```diff
@@ -227,6 +230,11 @@ def _handle_for_status(self, data_record: DataRecord) -> Optional[Any]:
             )
         response_dict: Optional[dict] = data_record.response_json_as_dict()
         if response_dict is not None:
+            if self.config.response_hook is not None:
+                return self.config.response_hook(
+                    self.output_serializer.from_dict(response_dict),
+                    IdempotentHookData(data_record),
+                )
             return self.output_serializer.from_dict(response_dict)
```
rubenfonseca: Suggestion just to keep the code cleaner

```suggestion
            serialized_output = self.output_serializer.from_dict(response_dict)
            if self.config.response_hook is not None:
                return self.config.response_hook(
                    serialized_output
                    IdempotentHookData(data_record),
                )
            return serialized_output
```

aws_lambda_powertools/utilities/idempotency/config.py
rubenfonseca: I kinda agree, but if we expose `raw_data_record` why do we need this layer of indirection in the first place? Can't we just send the entire DataRecord directly on the hook functions?

docs/utilities/idempotency.md
```diff
@@ -699,15 +699,16 @@ For advanced configurations, such as setting up SSL certificates or customizing
...
+| **use_local_cache**             | `False` | Whether to locally cache idempotency results                                                                                                                                                                                        |
+| **local_cache_max_items**       | 256     | Max number of items to store in local cache                                                                                                                                                                                         |
+| **hash_function**               | `md5`   | Function to use for calculating hashes, as provided by [hashlib](https://docs.python.org/3/library/hashlib.html){target="_blank" rel="nofollow"} in the standard library.                                                           |
+| **response_hook**               | `None`  | Function to use for processing the stored Idempotent response.  This function hook is called when an already returned response is found.  See [Modifying The Idempotent Response](idempotency.md#modifying-the-idempotent-repsonse) |
```
rubenfonseca:
```suggestion
| **response_hook**               | `None`  | Function to use for processing the stored Idempotent response.  This function hook is called when an existing idempontent response is returned.  See [Modifying The Idempotent Response](idempotency.md#modifying-the-idempotent-repsonse) |
```

docs/utilities/idempotency.md
rubenfonseca: What about `Manipulating the Idempotent Response`?

examples/idempotency/src/working_with_response_hook.py
```diff
@@ -0,0 +1,27 @@
...
+)
+from aws_lambda_powertools.utilities.typing import LambdaContext
+
+
+def my_response_hook(response: Dict, idempotent_data: IdempotentHookData) -> Dict:
+    # How to add a field to the response
+    response["is_idempotent_response"] = True
+
+    # Must return the response here
+    return response
```
rubenfonseca: Can we expand the sample to actually make use of `idempotent_data`?

rubenfonseca (changes requested):

leandrodamascena:
Hey @walmsles! I've updated the documentation and example. I believe it's ready for the final review now.

I'll also be updating the issue with the decisions regarding exception/error handling, including why we allow customers to handle their own exceptions and why we fail in case of an exception.

walmsles:
I love the **being a Good Citizen** sections!  That makes so much more sense - I knew it was missing something.

rubenfonseca:
Reviewing now

examples/idempotency/src/working_with_response_hook.py
```diff
@@ -1,27 +1,56 @@
...
-@idempotent(persistence_store=persistence_layer, config=config)
-def lambda_handler(event: dict, context: LambdaContext) -> dict:
-    return event
+def lambda_handler(event: dict, context: LambdaContext):
+    config.register_lambda_context(context)  # see Lambda timeouts section
+    try:
+        logger.info(f"Processing order id {event.get('order_id')}")
+        return process_order(order=event.get("order"))
+    except Exception as err:
+        return {"status_code": 400, "error": f"Erro processing {str(err)}"}
```
rubenfonseca:
```suggestion
        return {"status_code": 400, "error": f"Error processing {str(err)}"}
```

I loved this example, you can really understand how powerful the feature can be

docs/utilities/idempotency.md
```diff
@@ -454,6 +454,40 @@ sequenceDiagram
...
+        Lambda-->>Client: Response sent to client
+    else retried request
+        Client->>Lambda: Invoke (event)
+        Lambda->>Persistence Layer: Get or set idempotency_key=hash(payload)
+        activate Persistence Layer
+        Persistence Layer-->>Response hook: Already exists in persistence layer.
+        deactivate Persistence Layer
+        Note over Response hook,Persistence Layer: Record status is COMPLETE and not expired
+        Response hook->>Lambda: Response hook invoked
+        Lambda-->>Client: Same response sent to client
```
rubenfonseca:
```suggestion
        Lambda-->>Client: Manipulated idempotent response sent to client
```

Otherwise it makes me thing that the Response Hook can't change anything

docs/utilities/idempotency.md
```diff
@@ -454,6 +454,40 @@ sequenceDiagram
...
+    end
+```
+<i>Idempotent successful request with response hook</i>
```
rubenfonseca:
```suggestion
<i>Successful idempotent request with a response hook</i>
```

docs/utilities/idempotency.md
```diff
@@ -910,22 +944,38 @@ You can create your own persistent store from scratch by inheriting the `BasePer
 
     For example, the `_put_record` method needs to raise an exception if a non-expired record already exists in the data store with a matching key.
 
-### Modifying the Idempotent Repsonse
-
-The IdempotentConfig allows you to specify a _**response_hook**_ which is a function that will be called when an idempotent response is loaded from the PersistenceStore.
+### Manipulating the Idempotent Response
 
-You can provide the response_hook using _**IdempotentConfig**_.
+You can set up a `response_hook` in the `IdempotentConfig` class to access the returned data when an operation is idempotent. The hook function will be called with the current deserialized response object and the Idempotency record.
```
rubenfonseca:
```suggestion
You can set up a `response_hook` in the `IdempotentConfig` class to manipulate the returned data when an operation is idempotent. The hook function will be called with the current deserialized response object and the Idempotency record.
```

docs/utilities/idempotency.md
```diff
@@ -910,22 +944,38 @@ You can create your own persistent store from scratch by inheriting the `BasePer
...
+
+Using Response hooks can add subtle improvements to manipulating returned data from idempotent operations, but also add significant complexity if you're not careful.
+
+Keep the following in mind when authoring hooks for Idempotency utility:
+
+1. **Response hook works exclusively when operations are idempotent.** Carefully consider the logic within the `Response hook` and prevent any attempt to access the key from relying exclusively on idempotent operations.
+
+2. **Catch your own exceptions.** Catch and handle known exceptions to your logic.
+
+3. **Watch out when you are decorating the Lambda Handler and using the Response hook.** If you don't catch and handle exceptions in your `Response hook`, your function might not run properly.
```
rubenfonseca: A suggestion to make this more clear:

1- response hooks only work on idempotent responses
2- catch exceptions
3- keep it simple

Is this better?

```suggestion
When using response hooks to manipulate returned data from idempotent operations, it's important to follow best practices to avoid introducing complexity or issues. Keep these guidelines in mind:

1. **Response hook works exclusively when operations are idempotent.** The hook will not be called when an operation is not idempotent, or when the idempotent logic fails.

2. **Catch and Handle Exceptions** Your response hook code should catch and handle any exceptions that may arise from your logic. Unhandled exceptions will cause the Lambda function to fail unexpectedly.

3. **Keep Hook Logic Simple** Response hooks should contain minimal, straightforward logic for manipulating response data. Avoid complex conditional branching and aim for hooks that are easy to reason about.
```

rubenfonseca (changes requested):

docs/utilities/idempotency.md
leandrodamascena: This is much better than the one I wrote! Thank you for this very useful review and suggestion @rubenfonseca

leandrodamascena:
@rubenfonseca ready to one more review.

rubenfonseca (approved):

rubenfonseca:
Great job @walmsles @leandrodamascena pulling this off!

leandrodamascena (approved):

Issue #4074, opened 2024-04-05
Maintenance: Monthly reminder to update roadmap


heitorlessa:
### Why is this needed?

As part of governance to keep public roadmap up to date.

### Which area does this relate to?

Other

### Acknowledgment

- [X] This request meets [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [X] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

heitorlessa:
Reopening as there's some fixes to port from another branch, and to share lessons learned in this.

heitorlessa:
Main lesson learned is that when using cross-repo workflows that use local scripts, we must change the `checkout` in the target workflow to explicitly `checkout` itself.

**What do I mean by that?** 

Take the following job in mind

```yaml
jobs:
    call-workflow-passing-data:
        permissions:
            contents: read
            pull-requests: read
            issues: write  # create monthly roadmap report
        uses: aws-powertools/actions/.github/workflows/monthly_roadmap_reminder.yml@main
        secrets:
            token: ${{ secrets.GITHUB_TOKEN }}
```

There are two implicit designs that aren't easy to understand at first and docs don't call that out:

1. `aws-powertools/actions` is where the workflow is
2. `monthly_roadmap_reminder.yml` is the workflow name
4. The workflow runs in **this** repository

With that in mind, the target workflow naturally would try to do a code checkout as its first step. However, that code checkout will checkout **this** repository (Python source code), not the shared actions we want.

For this to work, the target workflow owned by `aws-powertools/actions` must explicitly pass which repository it will checkout code -- it can't be via `${{github.repo}}` because it would resolve to `powertools-lambda-python`.

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633  # v4.1.2
        with:
            # NOTE Since it's called from another repository
            # we must checkout the repository where the github-script is (aws-powertools/actions)
            # otherwise the caller check out their own repository code (e.g., Python/TS), where the github-scripts aren't available
            repository: aws-powertools/actions
```

[...]

PR #4078, opened 2024-04-05
fix(ci): apply lessons learned to monthly roadmap reminder cross-repo

heitorlessa:
Signed-off-by: heitorlessa <lessa@amazon.co.uk>
**Issue number:** #4074

## Summary

Main lesson learned is that when using cross-repo workflows that use local scripts, we must change the `checkout` in the target workflow to explicitly `checkout` itself.

**What do I mean by that?** 

Take the following job in mind

```yaml
jobs:
    call-workflow-passing-data:
        permissions:
            contents: read
            pull-requests: read
            issues: write  # create monthly roadmap report
        uses: aws-powertools/actions/.github/workflows/monthly_roadmap_reminder.yml@main
        secrets:
            token: ${{ secrets.GITHUB_TOKEN }}
```

There are two implicit designs that aren't easy to understand at first and docs don't call that out:

1. `aws-powertools/actions` is where the workflow is
2. `monthly_roadmap_reminder.yml` is the workflow name
4. The workflow runs in **this** repository

With that in mind, the target workflow naturally would try to do a code checkout as its first step. However, that code checkout will checkout **this** repository (Python source code), not the shared actions we want.

For this to work, the target workflow owned by `aws-powertools/actions` must explicitly pass which repository it will checkout code -- it can't be via `${{github.repo}}` because it would resolve to `powertools-lambda-python`.

```yaml
jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633  # v4.1.2
        with:
            # NOTE Since it's called from another repository
            # we must checkout the repository where the github-script is (aws-powertools/actions)
            # otherwise the caller check out their own repository code (e.g., Python/TS), where the github-scripts aren't available
            repository: aws-powertools/actions
```

A nicer benefit of this is that we don't need to control from the target who can call it and try to abuse creating GitHub issues in another place, because it's the caller repository that runs the workflow (it's not a fire-n-forget).

### Changes

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
Don't merge - I forgot two things:

 
PR review status column, and point to main over SHA until we have actions releases

leandrodamascena (approved):

PR #4114, opened 2024-04-11
fix(typing): improve overloads to ensure the return type follows the default_value type

### LINKED ISSUE #4113: Static typing: missing @overload to ensure return type is a str when default_value is set
### Static type checker used

pyright/pylance

3.12

latest

### Static type checker info

Expression of type "str | None" cannot be assigned to declared type "str"

Like in #3839 I found some more places for missing overloads.

### Code snippet

```python
header_value: str = event.get_header_value("name", "")
```

### Possible Solution

_No response_

tonnico:
**Issue number: #4113**

### Changes

Overload functions with `default_value` when it becomes not `Optional` anymore.

### User experience

Ensure return type without casting.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -88,13 +88,7 @@ class DictWrapper(Mapping):
     def _properties(self) -> List[str]:
         return [p for p in dir(self.__class__) if isinstance(getattr(self.__class__, p), property)]
 
-    @overload
-    def get(self, key: str, default: T) -> T: ...
-
-    @overload
-    def get(self, key: str, default: Optional[T] = None) -> Optional[T]: ...
-
-    def get(self, key: str, default: Optional[T] = None) -> Optional[T]:
+    def get(self, key: str, default: Optional[Any] = None) -> Optional[Any]:
         return self._data.get(key, default)
 
     @property
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -2,7 +2,7 @@ import base64
 import json
 from collections.abc import Mapping
 from functools import cached_property
-from typing import Any, Callable, Dict, Iterator, List, Optional, TypeVar, overload
+from typing import Any, Callable, Dict, Iterator, List, Optional, overload
 
 from aws_lambda_powertools.shared.headers_serializer import BaseHeadersSerializer
 from aws_lambda_powertools.utilities.data_classes.shared_functions import (
@@ -10,8 +10,6 @@ from aws_lambda_powertools.utilities.data_classes.shared_functions import (
     get_multi_value_query_string_values,
     get_query_string_value,
 )
-
-T = TypeVar("T")
 
 
 class DictWrapper(Mapping):
--- a/aws_lambda_powertools/utilities/data_classes/common.py
+++ b/aws_lambda_powertools/utilities/data_classes/common.py
@@ -2,7 +2,7 @@ import base64
 import json
 from collections.abc import Mapping
 from functools import cached_property
-from typing import Any, Callable, Dict, Iterator, List, Optional, overload
+from typing import Any, Callable, Dict, Iterator, List, Optional, TypeVar, overload
 
 from aws_lambda_powertools.shared.headers_serializer import BaseHeadersSerializer
 from aws_lambda_powertools.utilities.data_classes.shared_functions import (
@@ -10,6 +10,8 @@ from aws_lambda_powertools.utilities.data_classes.shared_functions import (
     get_multi_value_query_string_values,
     get_query_string_value,
 )
+
+T = TypeVar("T")
 
 
 class DictWrapper(Mapping):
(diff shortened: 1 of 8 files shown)


aws_lambda_powertools/utilities/data_classes/shared_functions.py
```diff
@@ -21,12 +21,30 @@ def base64_decode(value: str) -> str:
     return base64.b64decode(value).decode("UTF-8")
 
 
+@overload
 def get_header_value(
     headers: dict[str, Any],
     name: str,
-    default_value: str | None,
-    case_sensitive: bool | None,
-) -> str | None:
```
tonnico, line 29: This should not work with Python < 3.10

rubenfonseca: I believe the presence of `from __future__ import annotations` makes it work :)

Our CI runs on all versions of the supported Python runtimes, starting at 3.8

rubenfonseca:
@Wurstnase it seems some tests are failing, can you please look at them?

aws_lambda_powertools/utilities/data_classes/shared_functions.py
tonnico: Oh, didn't know, that `annotations` make this possible.

Should I revert this part?

tonnico: I looked into some other files. There are already a lot of files with `from __future__ import annotations`. I decide to revert the `Optional` change.

tonnico: The doc-string say `Default is None`. Before this commit `case_sensitive` was unset.

My feeling is, it should be `False` by default and never `None`.

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -11,6 +11,8 @@
     get_query_string_value,
 )
 
+T = TypeVar("T")
```
rubenfonseca: I don't thing this is used anymore?

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -2,7 +2,7 @@
 import json
 from collections.abc import Mapping
 from functools import cached_property
-from typing import Any, Callable, Dict, Iterator, List, Optional, overload
+from typing import Any, Callable, Dict, Iterator, List, Optional, TypeVar, overload
```
rubenfonseca: I don't thing this is used anymore?

aws_lambda_powertools/utilities/data_classes/kafka_event.py
```diff
@@ -69,10 +69,26 @@ def decoded_headers(self) -> Dict[str, bytes]:
         """Decodes the headers as a single dictionary."""
         return {k: bytes(v) for chunk in self.headers for k, v in chunk.items()}
 
+    @overload
     def get_header_value(
         self,
         name: str,
-        default_value: Optional[Any] = None,
+        default_value: str,
+        case_sensitive: bool = False,
```
rubenfonseca: I think the default here should be True, since that's the default on the concrete implementation

aws_lambda_powertools/utilities/data_classes/kafka_event.py
```diff
@@ -69,10 +69,26 @@ def decoded_headers(self) -> Dict[str, bytes]:
...
+    ) -> str: ...
+
+    @overload
+    def get_header_value(
+        self,
+        name: str,
+        default_value: Optional[str] = None,
+        case_sensitive: bool = False,
```
rubenfonseca: I think the default here should be True, since that's the default on the concrete implementation

aws_lambda_powertools/utilities/data_classes/shared_functions.py
rubenfonseca: Agree!

aws_lambda_powertools/utilities/data_classes/shared_functions.py
```diff
@@ -103,7 +137,7 @@ def get_multi_value_query_string_values(
 
     Returns
     -------
-    List[str]. optional
+    list[str]. optional
```
rubenfonseca: Can we revert al the `List` --> `list` and `Dict` --> `dict` changes in this file? They are unrelated to the PR and when we do it we should do it project wide.

rubenfonseca (changes requested):
This is amazing work, thank you VERY much!! I left some small nit picks here and there that should be easy to fix. I loved to see the test changing at the end to remove the type ignore!

aws_lambda_powertools/utilities/data_classes/common.py
tonnico: Right, I removed it.

aws_lambda_powertools/utilities/data_classes/common.py
tonnico: Right, I removed it.

aws_lambda_powertools/utilities/data_classes/shared_functions.py
tonnico: Ok

aws_lambda_powertools/utilities/data_classes/shared_functions.py
tonnico: I will keep this weird behaviour because there are a lot of implementations that have the same.

I will move this to a new PR.

rubenfonseca (approved):
Stellar work, thank you so much for spending time improving the experience here. Seriously, thank you!

PR #4295, opened 2024-05-08
fix(logger): correctly pick powertools or custom handler in custom environments

### LINKED ISSUE #4277: Bug: logger_handler argument is ignored if logging.getLogger() already has a handler.
### Expected Behaviour

The `logger_handler` argument for a Logger instance should set the provided logger as the registered handler, regardless of what handlers are already set on `logging.getLogger()` outputs.

### Current Behaviour

When providing a custom logger via the `logger_handler` argument, the custom log handler is ignored when the logger returned by `logging.getLogger(name)` already has a handler configured. This is problematic since environments like Serverless Framework cause `getLogger` to return a logger with a StreamHandler preconfigured.

### Code snippet

```python
from logging import Handler

class CustomHandler(Handler):
    """
    A custom handler class. (Truncated for clarity, contents don't matter)
    """
logger = Logger(logger_handler=CustomHandler())
print(logger.registered_handler) # This prints <StreamHandler <stdout> (NOTSET)>

def handler(event, context):
    print("Lambda running")
```

### Possible Solution

aws_lambda_powertools can override `getLogger` default handlers, with or without user-provided arguments, instead of assuming the returned logger's handler list is empty.

### Steps to Reproduce

1. Create a custom log handler
2. Create a Logger instance using the custom handler in the `logging_handler` parameter.
3. Deploy and run lambda using Serverless Framework with serverless-python-requirements in a lambda layer config.

latest

3.10

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

heitorlessa:
**Issue number:** #4277

## Summary

This PR fixes a bug when a compute environment or customer preconfigures a standard logging logger with the same Powertools logger name.

Previously, we'd always pick the first handler - child or parent. However, it led to pick the incorrect one if a logger we're creating for the first time (`Logger`) was already pre-setup through other means (`logging.getLogger`).

```python
import logging
import os

service_name = os.getenv("POWERTOOLS_SERVICE_NAME")

# A environment, third-party library, or customer intentionally creates a logger with the same name as Powertools
# and add a Handler

logging.getLogger(service_name).addHandler(logging.StreamHandler())

logger = Logger(service=service_name)  # Powertools will ask standard logging to give the previously setup logger, then configure another handler etc as expected 
```

### Changes

- [x] Encapsulate handler selection in `_get_handler` 
- [x] Reverse depend `registered_handler` on `_get_handler`
- [x] Logger with two handlers and log items use both of them
- [x] The above with a child logger
- [x] Add explicit test for orphaned child without handler that will fail with `AttributeError`
- [x] Create a backwards compatible exception to provide actionable error on orphaned child formatter (e.g., child init before parent and append_keys)
- [x] Update docs to clarify child logger and their caveats 

### User experience

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
Docs cleaned up to make it clearer the side effects of child loggers; removed a ton of banners we accrued over the years.

image

.gitignore
```diff
@@ -315,3 +315,4 @@ examples/**/sam/.aws-sam
 cdk.out
 # NOTE: different accounts will be used for E2E thus creating unnecessary git clutter
 cdk.context.json
+bla.py
```
leandrodamascena: Need to remove this.

.markdownlint.yaml
```diff
@@ -182,7 +182,7 @@ MD038: false
 MD039: true
 
 # MD040/fenced-code-language - Fenced code blocks should have a language specified
-MD040: true
+MD040: false
```
leandrodamascena: Linter won't comply if we remove spaces in the Diagram.

leandrodamascena (changes requested):
Hey @heitorlessa! I've pushed a commit with two small things, but this LGTM! Congrats for this great catch and fix.

leandrodamascena (dismissed):

aws_lambda_powertools/logging/constants.py
```diff
@@ -0,0 +1,3 @@
+LOGGER_ATTRIBUTE_POWERTOOLS_HANDLER = "powertools_handler"
+LOGGER_ATTRIBUTE_PRECONFIGURED = "init"
```
heitorlessa, line 4:
```suggestion
# logger.init attribute is set when Logger has been configured 
LOGGER_ATTRIBUTE_PRECONFIGURED = "init"
```

aws_lambda_powertools/logging/constants.py
```diff
@@ -0,0 +1,3 @@
+LOGGER_ATTRIBUTE_POWERTOOLS_HANDLER = "powertools_handler"
```
heitorlessa, line 2:
```suggestion
# logger.powertools_handler is set with Powertools Logger handler; useful when there are many handlers
LOGGER_ATTRIBUTE_POWERTOOLS_HANDLER = "powertools_handler"
```

leandrodamascena (approved):

PR #4298, opened 2024-05-08
fix(event-sources): sane defaults for authorizer v1 and v2

### LINKED ISSUE #4102: Bug: APIGateWayProxyEvent.request_context.authorizer results in KeyError
### Expected Behaviour

returns authorizer dictionary

### Current Behaviour

KeyError

### Code snippet

powertools-lambda-python/aws_lambda_powertools/utilities/data_classes/api_gateway_proxy_event.py
line 94 in class APIGatewayEventRequestContext

```python
 def authorizer(self) -> APIGatewayEventAuthorizer:
        return APIGatewayEventAuthorizer(self._data["requestContext"]["authorizer"])
```

### Possible Solution

```python
 def authorizer(self) -> APIGatewayEventAuthorizer:
        return APIGatewayEventAuthorizer(self.["requestContext"].get("authorizer"))
```

### Steps to Reproduce

```python
@event_source(data_class=APIGatewayProxyEvent)
def handler(event: APIGatewayProxyEvent, context):
    authorizer = event.request_context.authorizer
```

2.36.0

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

heitorlessa:
**Issue number:** #4102

## Summary

This PR returns default container types when key is `null` or not present (test environments). While it's not an issue at runtime, it aims to reduce the number of false positive issues, until we change event-source feature to **always** return a sane default _(where correct)_.

### Changes

* [x] Return empty `dict` instead of `None` in Authorizers v1 and v2
* [x] Return empty `list` instead of `None` in Authorizers v1 and v2
* [x] Correct tests assertion from `is type` to `==` as it's no longer None
* [x] Address mypy issues where certain keys have `null` as the value e.g., `key.get(name) or {}` over `key.get(name, {}`

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena (approved):
APPROVED!

PR #4385, opened 2024-05-21
fix(event_handler): CORS Origin for ALBResolver multi-headers 

### LINKED ISSUE #4327: Bug: ALB with multiValueHeaders will not return correct CORS headers in OPTIONS preflight
### Expected Behaviour

An OPTIONS preflight should return all "Access-Control-*" headers.

### Current Behaviour

https://github.com/aws-powertools/powertools-lambda-python/blob/87d580af0dfaa8007d4e6f6318eef09a3dad206d/aws_lambda_powertools/event_handler/api_gateway.py#L2132

will only fetch `"headers"` from the event.  This will return `None` in an ALBEvent with `multiValueHeaders` enabled.

### Code snippet

```python
cors = CORSConfig(
    allow_origin="*",
    allow_credentials=True,
)

app = ALBResolver(cors=cors)
```

### Possible Solution

```python
_origin_header = self.current_event.resolved_headers_field.get("origin")  # case insensitive?!?
_origin = _origin_header if isinstance(_origin_header, str) else _origin_header.pop(None)
headers.update(self._cors.to_dict(_origin))
```

### Steps to Reproduce

```
curl -X OPTIONS https://xxx.amazonaws.com -H "origin: http://example.com"
```

```
... snip
< HTTP/2 204
< server: awselb/2.0
< date: Tue, 14 May 2024 09:06:13 GMT
< access-control-allow-methods: CONNECT,DELETE,GET,HEAD,OPTIONS,PATCH,POST,PUT,TRACE
```

latest

3.12

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #4327 

### Changes

AWS offers various integration options for Lambda, such as ALB, REST API, HTTP API, Lambda URL, and others. These integrations have different payloads, and when working with ALB and multi-headers, we were unable to add CORS as it only considered a header single value.

With this fix, we can now work with both single and multi-header scenarios without any issues. Additionally, the fix looks for Origin/origin header in both single-value and multi-value headers, providing a more comprehensive and robust solution for handling different types of request headers.

### User experience

There is no change in the user experience, this is a bug.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -217,6 +217,30 @@ def to_dict(self, origin: Optional[str]) -> Dict[str, str]:
...
+        The 'origin' or 'Origin' header can be either a single header or a multi-header.
+
+        Args:
+            resolver_headers (Dict): A dictionary containing the headers.
+
+        Returns:
+            Union[str, List[str], None]: The value(s) of the 'origin' or 'Origin' header.
+                If the header is a single header, a string is returned.
+                If the header is a multi-header, a list of strings is returned.
+                If the header is not present, None is returned.
```
tonnico:
```suggestion
            Optional[str]: The value(s) of the origin header or None.
```

heitorlessa:
Hey @leandrodamascena thanks for the fix ;) 

When you've got time could you update the PR body to explain what the actual fix is?

docs/index.md
```diff
@@ -89,7 +89,7 @@ You can install Powertools for AWS Lambda (Python) using your favorite dependenc
             === "CDK"
 
                 ```python hl_lines="13 19"
-                --8<-- "examples/homepage/install/x86_64/cdk.py"
+                --8<-- "examples/homepage/install/x86_64/cdk_x86.py"
```
tonnico, line 92: renaming is probably a different pull request?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -217,6 +217,30 @@ def to_dict(self, origin: Optional[str]) -> Dict[str, str]:
...
+        """
+        resolved_header = resolver_headers.get("origin") or resolver_headers.get("Origin")
```
heitorlessa: can we use our internal utility function to get header and use case_sensitive=False so it lowers it?

That should avoid any other permutations of the header we can't predict

tonnico:
Many thanks. I will take a look into the fix.

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -782,7 +806,8 @@ def __init__(
 
     def _add_cors(self, event: ResponseEventT, cors: CORSConfig):
         """Update headers to include the configured Access-Control headers"""
-        self.response.headers.update(cors.to_dict(event.get_header_value("Origin")))
+        extracted_origin_header = cors.extract_origin_header(event.resolved_headers_field)
```
heitorlessa: Is there a better place to do this extraction logic?

Ideally this wouldn't be tied to CORS only but in the headers boundary.

PS: on my phone so please forgive me if I'm misreading it

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -217,6 +217,30 @@ def to_dict(self, origin: Optional[str]) -> Dict[str, str]:
             headers["Access-Control-Allow-Credentials"] = "true"
         return headers
 
+    @staticmethod
+    def extract_origin_header(resolver_headers: Dict):
```
tonnico:
```suggestion
    def extract_origin_header(resolver_headers: Mapping[str, Union[str, list[str]]) -> str:
```

aws_lambda_powertools/event_handler/api_gateway.py
tonnico: Maybe it should be similar to

https://github.com/aws-powertools/powertools-lambda-python/blob/fc376f90a0779226adac6c8ee6619c30b6b2f11d/aws_lambda_powertools/utilities/data_classes/shared_functions.py#L75-L80

heitorlessa (changes requested):
Two quick changes as I head to the airport.

1. Use our internality utility function to get case insensitive headers 

2. Revisit origin headers extraction logic placement

docs/index.md
leandrodamascena: Hey @Wurstnase! I think its ok keep it here, so we don't need to open a new PR for this.

leandrodamascena:
Just an update: E2E tests are green.

image

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: Good catch @heitorlessa.

leandrodamascena: I refactored this and now using this function: https://github.com/aws-powertools/powertools-lambda-python/blob/fc376f90a0779226adac6c8ee6619c30b6b2f11d/aws_lambda_powertools/utilities/data_classes/shared_functions.py#L42

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: I didn't understand this point. Do you suggest I remove `extract_origin_header` from CORS and move this to another class?

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: Good catch!

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: We are now using the `get_header_value` function and there is no overload for this signature. In the end this is a Dict, so I think we can keep this signature.

aws_lambda_powertools/event_handler/api_gateway.py
tonnico: Probably `get_header_value` needs now some extra overloads?

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: Yeah (I can't see from mobile what "this class" is).

From a responsibility perspective, CORS logic shouldn't be responsible to extract the origin header but the headers logic

leandrodamascena:
E2E tests are green

image

aws_lambda_powertools/event_handler/util.py
```diff
@@ -11,3 +16,24 @@ class _FrozenDict(dict):
...
+
+    The 'origin' or 'Origin' header can be either a single header or a multi-header.
+
+    Args:
+        resolver_headers (Dict): A dictionary containing the headers.
+
+    Returns:
+        Optional[str]: The value(s) of the origin header or None.
+    """
+    resolved_header = get_header_value(resolver_headers, "origin", None, case_sensitive=False)
```
heitorlessa: Thanks a lot 🫶

tip: always use keyword arguments over arguments followed by keyword argument for readability and easy refactoring

aws_lambda_powertools/event_handler/util.py
```diff
@@ -11,3 +16,24 @@ class _FrozenDict(dict):
...
+    if isinstance(resolved_header, str):
+        return resolved_header
+    if isinstance(resolved_header, list):
+        return resolved_header[0]
+
+    return resolved_header
```
heitorlessa:
```suggestion
    if isinstance(resolved_header, list):
        return resolved_header[0]

    return resolved_header
```

Did you mean like this? Since you're always returning at the end

heitorlessa (changes requested):
Tiny change to remove unnecessary branching (if I understood it right)

aws_lambda_powertools/event_handler/util.py
leandrodamascena: Great point!

heitorlessa (approved):

aws_lambda_powertools/event_handler/util.py
```diff
@@ -11,3 +16,27 @@ class _FrozenDict(dict):
...
+    Args:
+        resolver_headers (Dict): A dictionary containing the headers.
+
+    Returns:
+        Optional[str]: The value(s) of the origin header or None.
+    """
+    resolved_header = get_header_value(
+        headers=resolver_headers,
+        name="origin",
+        default_value=None,
```
heitorlessa, line 36: tip: you can set a same default like "". That way the caller doesn't need to check for None and instead always works with a string.

The same way you'd prefer to have an empty list or empty duct instead of optional list/dict, like the v3 for event source data classes.

Pre-approved

leandrodamascena: In this case I think there is no problem in working with `None` because the function is already checking `None` values...

```python

def to_dict(self, origin: Optional[str]) -> Dict[str, str]:
        """Builds the configured Access-Control http headers"""

        # If there's no Origin, don't add any CORS headers
        if not origin:
            return {}
```

heitorlessa: It's less about being a problem but removing the need for checking for it and typing.

I'm gonna merge

heitorlessa (approved):

pseudochaos:
Hi guys! Your changelog indicates that this issue was resolved in the [release 2.39.0](https://docs.powertools.aws.dev/lambda/python/latest/changelog/#v2390-2024-06-13). However, the fix is not present in 2.39.0 or in 2.39.1. Can you clarify when we can expect it?

leandrodamascena:
> Hi guys! Your changelog indicates that this issue was resolved in the [release 2.39.0](https://docs.powertools.aws.dev/lambda/python/latest/changelog/#v2390-2024-06-13). However, the fix is not present in 2.39.0 or in 2.39.1. Can you clarify when we can expect it?

Hello @pseudochaos! I see this fix in version 2.39.0, but do you mean it's not working as expected? Can you give more details about this please?

pseudochaos:
> > Hi guys! Your changelog indicates that this issue was resolved in the [release 2.39.0](https://docs.powertools.aws.dev/lambda/python/latest/changelog/#v2390-2024-06-13). However, the fix is not present in 2.39.0 or in 2.39.1. Can you clarify when we can expect it?
> 
> Hello @pseudochaos! I see this fix in version 2.39.0, but do you mean it's not working as expected? Can you give more details about this please?

Hi @leandrodamascena! Yes, it's not working as expected or as it was working in `1.x`. After upgrading from `1.31.1` to `2.39.1` I stopped receiving the `Access-Control-Allow-Origin: *` header. My code looks like this:

```python
app = APIGatewayRestResolver(strip_prefixes=['/foo'], cors=CORSConfig())
``` 
and this is how I specify the dependency - `aws-lambda-powertools[tracer]==2.39.1`.
To temporarily workaround the issue, I had to explicitly return a `Response` object with the header like this:

```python
return Response(headers={'Access-Control-Allow-Origin': '*'}, body=...)
```

leandrodamascena:
> Yes, it's not working as expected or as it was working in 1.x. After upgrading from 1.31.1 to 2.39.1

Just to make sure I understand the point here: you are migrating from Porwertools v1 to v2, correct?

I'm opening an issue on your behalf and we can discuss more there. It might be interesting if you provide some code snippet and the result you expect from it.

leandrodamascena:
@pseudochaos can you please add more information in this issue?

https://github.com/aws-powertools/powertools-lambda-python/issues/4589

PR #4391, opened 2024-05-22
fix(idempotency): POWERTOOLS_IDEMPOTENCY_DISABLED should respect truthy values

### LINKED ISSUE #4390: Bug: POWERTOOLS_IDEMPOTENCY_DISABLED does not respect truthy values
### Expected Behaviour

I would expect truthy values to be respected for `POWERTOOLS_IDEMPOTENCY_DISABLED`:

- True: ["1", "y", "yes", "t", "true", "on"]
- False: ["0": "n", "no", "f", "false", "off"]

### Current Behaviour

The logic below is accessing the environment variable and bypassing the idempotency layer if it the condition evaluates to `True`. But it's always evaluating to `True` unless `os.getenv` returns `None`.

```
if os.getenv(constants.IDEMPOTENCY_DISABLED_ENV):
    return function(*args, **kwargs)
```

### Code snippet

```python
>>> import os
>>> os.environ['POWERTOOLS_IDEMPOTENCY_DISABLED'] = 'false'
>>> bool(os.getenv('POWERTOOLS_IDEMPOTENCY_DISABLED'))
True
>>> bool(None)
False
```

### Possible Solution

Leverage the `strtobool` function in `aws_lambda_powertools.shared.functions` to convert the environment variable value into a `bool`.

### Steps to Reproduce

I would recommend setting the value of `POWERTOOLS_IDEMPOTENCY_DISABLED` to `false`, which implies the user does not want idempotency to be disabled. In this situation, the Idempotency layer will be bypassed and no records will be recorded in the idempotency store.

latest

3.11

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

stevenhoelscher:
**Issue number:** [4390](https://github.com/aws-powertools/powertools-lambda-python/issues/4390)

### Changes

- When evaluating the `POWERTOOLS_IDEMPOTENCY_DISABLED` environment variable, first transform the string truthy value into a `bool`.

### User experience

* **Before**: A user sets `POWERTOOLS_IDEMPOTENCY_DISABLED=false` but the idempotency layer would actually be disabled
* **After**: A user sets `POWERTOOLS_IDEMPOTENCY_DISABLED=false` and the idempotency layer would not be disabled

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
Looking into it shortly (in transit)

leandrodamascena:
Hi @stevenhoelscher! Thank you so much for submitting this PR. It's a great catch, and I'm glad you discovered this long-standing bug. We really appreciate you taking the time to fix it!

I made only two changes in this PR:
1 - I set the default value of `os.getenv` to `false`, making the behavior more explicit.
2 - I included a test case to ensure the correct behavior.

Thanks

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -66,7 +67,9 @@ def idempotent(
         >>>     return {"StatusCode": 200}
     """
 
-    if os.getenv(constants.IDEMPOTENCY_DISABLED_ENV):
+    # Check if the IDEMPOTENCY_DISABLED_ENV environment variable is set to True
+    # If IDEMPOTENCY_DISABLED_ENV is not set or set to any false value, the function will be idempotent by default
```
heitorlessa:
```suggestion
    # Skip idempotency controls when POWERTOOLS_IDEMPOTENCY_DISABLED has a truthy value
```

more concise

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -66,7 +67,9 @@ def idempotent(
...
+    if strtobool(os.getenv(constants.IDEMPOTENCY_DISABLED_ENV, "false")):
         return handler(event, context, **kwargs)
```
heitorlessa, line 81: let's add a warning as it's not something customers would naturally want it disabled in production. It's easy to set and forget as we've learned from Metrics and other features.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -150,7 +153,9 @@ def process_order(customer_id: str, order: dict, **kwargs):
 
     @functools.wraps(function)
     def decorate(*args, **kwargs):
-        if os.getenv(constants.IDEMPOTENCY_DISABLED_ENV):
+        # Check if the IDEMPOTENCY_DISABLED_ENV environment variable is set to True
+        # If IDEMPOTENCY_DISABLED_ENV is not set or set to any false value, the function will be idempotent by default
```
heitorlessa:
```suggestion
        # Skip idempotency controls when POWERTOOLS_IDEMPOTENCY_DISABLED has a truthy value
```

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -150,7 +153,9 @@ def process_order(customer_id: str, order: dict, **kwargs):
...
+        if strtobool(os.getenv(constants.IDEMPOTENCY_DISABLED_ENV, "0")):
             return function(*args, **kwargs)
```
heitorlessa, line 173: warning (as earlier)

heitorlessa (changes requested):
let's add a warning to prevent customers forgetting it's disabled in dev but promoting it to prod.

aws_lambda_powertools/utilities/idempotency/idempotency.py
leandrodamascena: Hey @heitorlessa! I have created a new class, PowertoolsWarning, which inherits from UserWarning. This new class enables customers to filter and handle specific warnings raised by Powertools.

heitorlessa: niiice! need to think of a good naming so the intent is clear. I'll come back today

aws_lambda_powertools/warnings/__init__.py
```diff
@@ -0,0 +1,21 @@
+"""Shared warnings that don't belong to a single utility"""
+
+
+class PowertoolsWarning(UserWarning):
```
heitorlessa:
```suggestion
class PowertoolsUserWarning(UserWarning):
```

This makes it more aligned with `UserWarning` as that's broadly applied to many types of warning. 

We should clean up the docstring to make it clearer and shorter -- what kind of use cases we raise this warning etc.

heitorlessa (changes requested):

aws_lambda_powertools/warnings/__init__.py
leandrodamascena: Done.

leandrodamascena: Done

leandrodamascena:
@heitorlessa ready to review again.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -66,7 +69,16 @@ def idempotent(
         >>>     return {"StatusCode": 200}
     """
 
-    if os.getenv(constants.IDEMPOTENCY_DISABLED_ENV):
+    # Skip idempotency controls when POWERTOOLS_IDEMPOTENCY_DISABLED has a truthy value
+    # Raises a warning if not running in development mode
+    if strtobool(os.getenv(constants.IDEMPOTENCY_DISABLED_ENV, "false")):
+        if not powertools_dev_is_set():
```
heitorlessa: Think is necessary? It's another if branching here to maintain and higher memory

I'd personally warn either way. If they're explicitly in dev, other warnings will appear, and they can also suppress them.

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -150,7 +162,16 @@ def process_order(customer_id: str, order: dict, **kwargs):
 
     @functools.wraps(function)
     def decorate(*args, **kwargs):
-        if os.getenv(constants.IDEMPOTENCY_DISABLED_ENV):
+        # Skip idempotency controls when POWERTOOLS_IDEMPOTENCY_DISABLED has a truthy value
+        # Raises a warning if not running in development mode
+        if strtobool(os.getenv(constants.IDEMPOTENCY_DISABLED_ENV, "false")):
+            if not powertools_dev_is_set():
```
heitorlessa: Same as before

aws_lambda_powertools/warnings/__init__.py
```diff
@@ -0,0 +1,23 @@
+"""Shared warnings that don't belong to a single utility"""
+
+
+class PowertoolsUserWarning(UserWarning):
+    """
+    This class provides a custom Warning tailored for better clarity when certain situations occur.
```
heitorlessa, line 6: That's a great docstring - straight to the point and actionable!

Well done 👏 

cc @sthulb who's learning the code style

heitorlessa (commented):
Last question to clarify a piece I missed (changes are great!!)

aws_lambda_powertools/utilities/idempotency/idempotency.py
leandrodamascena: Agree. I'm sure we have situations like this in other parts of the codebase and I will review them.

heitorlessa (approved):

heitorlessa:
hey @stevenhoelscher **thank you** so much for kicking this off <3

PR #4421, opened 2024-05-28
fix(event_handler): security scheme unhashable list when working with router

### LINKED ISSUE #4375: Security configuration for route does not work for APIGatewayHttpRouter
### Expected Behaviour

Method `get_openapi_json_schema` from `APIGatewayHttpResolver` should work for routes from `APIGatewayHttpRouter` with the `security` parameter.

### Current Behaviour

When the `security` parameter is a part of the router (`APIGatewayHttpRouter`), it doesn't work - `/openapispec`  has: `{"message": "Internal Server Error"}`.

If I remove the `security` parameter from the router file, it works, and it means that the security configuration is correct only for `APIGatewayHttpResolver`.

Error in CloudWatch after `/openapispec ` execution:

`[ERROR] TypeError: unhashable type: 'list'
  File "/var/lang/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/var/task/app.py", line 14, in <module>
    from project_member_v1 import router as project_member_router
  File "/var/task/project_member_v1/router.py", line 27, in <module>
    @router.get(
  File "/opt/python/aws_lambda_powertools/event_handler/api_gateway.py", line 2412, in register_route
    self._routes_with_middleware[route_key] = []`

### Code snippet

```python
`
app = APIGatewayHttpResolver(enable_validation=True)
logger = Logger()
metrics = Metrics(namespace="Powertools")

OpenApiConfiguration.configure_swagger(app)

app.include_router(test_router.router, prefix="/v1/test")

@app.get("/hello", tags=["Test"], summary="An endpoint to test", security=[{"TestAuth": []}],)
def hello():
    return {"message": f"Hello!"}

@app.get("/openapispec", tags=["Docs"], summary="Get the OpenAPI specification")
def get_openapi():
    return app.get_openapi_json_schema(**OpenApiConfiguration.get_meta_data())


@logger.inject_lambda_context(correlation_id_path=correlation_paths.API_GATEWAY_REST)
@metrics.log_metrics(capture_cold_start_metric=True)
`

Router file:
`
router = APIGatewayHttpRouter()

@router.get(
    "/<test_id>",
    tags=["Test"],
    summary="Test endpoint",
    security=[{"TestAuth": []}],
)
def get_test_item(test_id: str) -> TestItemResponse:
    return TestItemResponse(test_id= test_id)
`

`
class OpenApiConfiguration:
    @staticmethod
    def get_meta_data():
        return {
            "title": "TestAPI",
            "version": "0.0.1",
            Contact(name="Test", url=Url("test.com")),
            "description": "TestAPI",
            "terms_of_service": None,
            "license_info": None,
            "openapi_version": "3.0.1",
            "security_schemes": {
                "TestAuth": HTTPBearer(scheme="bearer", bearerFormat="JWT", description="JWT token authentication"),
            },
        }

    @staticmethod
    def configure_swagger(resolver: ApiGatewayResolver) -> None:
        resolver.enable_swagger(
            **OpenApiConfiguration.get_meta_data(),
            path="/swagger",
            swagger_base_url="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/5.11.0"
        )
`

Route `openspec` should returns a definition of 3 endpoints:
- `/hello` with `TestAuth` security
- `/openapispec` without security
- `/test/< test_id >`  with `TestAuth` security
```

### Possible Solution

_No response_

### Steps to Reproduce

1. Create instance of `APIGatewayHttpResolver` 
2. Create instance of `APIGatewayHttpRouter`
3. Implement `OpenApiConfiguration` and define security_schemes
4. Add route to APIGatewayHttpRouter with security 
5. Add route to openapispec: 
`@app.get("/openapispec", tags=["Docs"], summary="Get the OpenAPI specification")
def get_openapi():
    return app.get_openapi_json_schema(**OpenApiConfiguration.get_meta_data())`
6. Execute `/openapispec`

latest

3.12

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

### LINKED ISSUE #4460: Bug: Security schemes are not validated when using operation level
### Expected Behaviour

The OpenAPI specification expects `security schemes` to define what type of authentication that API will support and then customers can define the operations that will use that authentication or even whether all operations will use the same authentication.

When using this code, the OpenAPI utility should validate the security defined in operation level and throw an exception as it is a schema validation error. But this is not happening.

```python

@app.get("/hello", security=[{"apiKey": []}])
def get_schema():
    return app.get_openapi_json_schema()

```

### Current Behaviour

This is currently not being validated and customers receive an error when trying to use the OpenAPI specification.

### openapischema

```json
{
   "openapi":"3.0.3",
   "info":{
      "title":"Powertools API",
      "version":"1.0.0"
   },
   "servers":[
      {
         "url":"/"
      }
   ],
   "paths":{
      "/hello":{
         "get":{
            "summary":"GET /hello",
            "operationId":"get_schema_hello_get",
            "responses":{
               "200":{
                  "description":"Successful Response",
                  "content":{
                     "application/json":{
                        
                     }
                  }
               },
               "422":{
                  "description":"Validation Error",
                  "content":{
                     "application/json":{
                        "schema":{
                           "$ref":"#/components/schemas/HTTPValidationError"
                        }
                     }
                  }
               }
            },
            "security":[
               {
                  "apiKey":[
                     
                  ]
               }
            ]
         }
      }
   },
   "components":{
      "schemas":{
         "HTTPValidationError":{
            "properties":{
               "detail":{
                  "items":{
                     "$ref":"#/components/schemas/ValidationError"
                  },
                  "type":"array",
                  "title":"Detail"
               }
            },
            "type":"object",
            "title":"HTTPValidationError"
         },
         "ValidationError":{
            "properties":{
               "loc":{
                  "items":{
                     "anyOf":[
                        {
                           "type":"string"
                        },
                        {
                           "type":"integer"
                        }
                     ]
                  },
                  "type":"array",
                  "title":"Location"
               },
               "type":{
                  "type":"string",
                  "title":"Error Type"
               }
            },
            "type":"object",
            "required":[
               "loc",
               "msg",
               "type"
            ],
            "title":"ValidationError"
         }
      }
   }
}
```

Error:

image

### Code snippet

```python

@app.get("/hello", security=[{"apiKey": []}])
def get_schema():
    return app.get_openapi_json_schema()

```

### Possible Solution

_No response_

### Steps to Reproduce

Use this code

```python

@app.get("/hello", security=[{"apiKey": []}])
def get_schema():
    return app.get_openapi_json_schema()

```

latest

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #4375 

### Changes

This pull request ensures that the `get_openapi_json_schema` method can correctly generate the OpenAPI JSON schema for routes that include security parameters when using the `APIGatewayHttpRouter`.

Previously, when the security parameter was included in a router, any endpoint would return an "Internal Server Error".

```sh
[ERROR] TypeError: unhashable type: 'list' Traceback (most recent call last): File "/var/lang/lib/python3.12/importlib/__init__.py", line 90, in import_module return _bootstrap._gcd_import(name[level:], package, level) File "<frozen importlib._bootstrap>", line 1387, in _gcd_import File "<frozen importlib._bootstrap>", line 1360, in _find_and_load File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked File "<frozen importlib._bootstrap>", line 935, in _load_unlocked File "<frozen importlib._bootstrap_external>", line 995, in exec_module File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed File "/var/task/app.py", line 14, in <module> from project_member_v1 import router as project_member_router File "/var/task/project_member_v1/router.py", line 27, in <module> @router.get( File "/opt/python/aws_lambda_powertools/event_handler/api_gateway.py", line 2412, in register_route self._routes_with_middleware[route_key] = []
```

#### app.py
```python
from router import OpenApiConfiguration, router


app = APIGatewayHttpResolver(enable_validation=True)
OpenApiConfiguration.configure_swagger(app)


app.include_router(router, prefix="/v1/test")

@app.get("/openapispec", tags=["Docs"], summary="Get the OpenAPI specification", security=[{"TestAuth": []}])
def get_openapi():
    return app.get_openapi_json_schema(**OpenApiConfiguration.get_meta_data())

```

#### router.py
```python
from aws_lambda_powertools.event_handler.api_gateway import Router
from aws_lambda_powertools.event_handler.openapi.models import HTTPBearer

router = Router()

@router.get(
    "/<test_id>",
    tags=["Test"],
    summary="Test endpoint",
    security=[{"TestAuth": ["a"]}],
)
def get_test_item(test_id: str) :
    return "OK"

class OpenApiConfiguration:
    @staticmethod
    def get_meta_data():
        return {
            "title": "TestAPI",
            "version": "0.0.1",
            "description": "TestAPI",
            "terms_of_service": None,
            "license_info": None,
            "openapi_version": "3.0.1",
            "security_schemes": {
                "TestAuth": HTTPBearer(scheme="bearer", bearerFormat="JWT", description="JWT token authentication"),
            },
        }

    @staticmethod
    def configure_swagger(resolver: APIGatewayHttpResolver) -> None:
        resolver.enable_swagger(
            **OpenApiConfiguration.get_meta_data(),
            path="/swagger",
            swagger_base_url="https://cdnjs.cloudflare.com/ajax/libs/swagger-ui/5.11.0"
        )
```

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -33,7 +33,7 @@ from typing import (
 from aws_lambda_powertools.event_handler import content_types
 from aws_lambda_powertools.event_handler.exceptions import NotFoundError, ServiceError
 from aws_lambda_powertools.event_handler.openapi.constants import DEFAULT_API_VERSION, DEFAULT_OPENAPI_VERSION
-from aws_lambda_powertools.event_handler.openapi.exceptions import RequestValidationError
+from aws_lambda_powertools.event_handler.openapi.exceptions import RequestValidationError, SchemaValidationError
 from aws_lambda_powertools.event_handler.openapi.types import (
     COMPONENT_REF_PREFIX,
     METHODS_WITH_BODY,
@@ -43,7 +43,12 @@ from aws_lambda_powertools.event_handler.openapi.types import (
     validation_error_definition,
     validation_error_response_definition,
 )
-from aws_lambda_powertools.event_handler.util import _FrozenDict, _FrozenListDict, extract_origin_header
+from aws_lambda_powertools.event_handler.util import (
+    _FrozenDict,
+    _FrozenListDict,
+    extract_origin_header,
+    validate_openapi_security_parameters,
+)
 from aws_lambda_powertools.shared.cookies import Cookie
 from aws_lambda_powertools.shared.functions import powertools_dev_is_set
 from aws_lambda_powertools.shared.json_encoder import Encoder
@@ -1589,6 +1594,15 @@ class ApiGatewayResolver(BaseRouter):
 
         # Add routes to the OpenAPI schema
         for route in all_routes:
+
+            if route.security and not validate_openapi_security_parameters(
+                security=route.security,
+                security_schemes=security_schemes,
+            ):
+                raise SchemaValidationError(
+                    "Security configuration was not found in security_schemas or security_schema was not defined.",
+                )
+
             if not route.include_in_schema:
                 continue
 
@@ -1631,15 +1645,14 @@ class ApiGatewayResolver(BaseRouter):
         security: Optional[List[Dict[str, List[str]]]],
         security_schemes: Optional[Dict[str, "SecurityScheme"]],
     ) -> Optional[List[Dict[str, List[str]]]]:
+
         if not security:
             return None
 
-        if not security_schemes:
-            raise ValueError("security_schemes must be provided if security is provided")
-
-        # Check if all keys in security are present in the security_schemes
-        if any(key not in security_schemes for sec in security for key in sec):
-            raise ValueError("Some security schemes not found in security_schemes")
+        if not validate_openapi_security_parameters(security=security, security_schemes=security_schemes):
+            raise SchemaValidationError(
+                "Security configuration was not found in security_schemas or security_schema was not defined.",
+            )
 
         return security
 
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -703,6 +703,7 @@ class Route:
         from aws_lambda_powertools.event_handler.openapi.params import Param
 
         parameters = []
+        parameter: Dict[str, Any]
         for param in all_route_params:
             field_info = param.field_info
             field_info = cast(Param, field_info)
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -1600,7 +1600,8 @@ class ApiGatewayResolver(BaseRouter):
                 security_schemes=security_schemes,
             ):
                 raise SchemaValidationError(
-                    "Security configuration was not found in security_schemas or security_schema was not defined.",
+                    "Security configuration was not found in security_schemas or security_schema was not defined. "
+                    "See: https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#security-schemes",
                 )
 
             if not route.include_in_schema:
@@ -1651,7 +1652,8 @@ class ApiGatewayResolver(BaseRouter):
 
         if not _validate_openapi_security_parameters(security=security, security_schemes=security_schemes):
             raise SchemaValidationError(
-                "Security configuration was not found in security_schemas or security_schema was not defined.",
+                "Security configuration was not found in security_schemas or security_schema was not defined. "
+                "See: https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#security-schemes",
             )
 
         return security
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -46,8 +46,8 @@ from aws_lambda_powertools.event_handler.openapi.types import (
 from aws_lambda_powertools.event_handler.util import (
     _FrozenDict,
     _FrozenListDict,
+    _validate_openapi_security_parameters,
     extract_origin_header,
-    validate_openapi_security_parameters,
 )
 from aws_lambda_powertools.shared.cookies import Cookie
 from aws_lambda_powertools.shared.functions import powertools_dev_is_set
@@ -1595,7 +1595,7 @@ class ApiGatewayResolver(BaseRouter):
         # Add routes to the OpenAPI schema
         for route in all_routes:
 
-            if route.security and not validate_openapi_security_parameters(
+            if route.security and not _validate_openapi_security_parameters(
                 security=route.security,
                 security_schemes=security_schemes,
             ):
@@ -1649,7 +1649,7 @@ class ApiGatewayResolver(BaseRouter):
         if not security:
             return None
 
-        if not validate_openapi_security_parameters(security=security, security_schemes=security_schemes):
+        if not _validate_openapi_security_parameters(security=security, security_schemes=security_schemes):
             raise SchemaValidationError(
                 "Security configuration was not found in security_schemas or security_schema was not defined.",
             )
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -43,7 +43,7 @@ from aws_lambda_powertools.event_handler.openapi.types import (
     validation_error_definition,
     validation_error_response_definition,
 )
-from aws_lambda_powertools.event_handler.util import _FrozenDict, extract_origin_header
+from aws_lambda_powertools.event_handler.util import _FrozenDict, _FrozenListDict, extract_origin_header
 from aws_lambda_powertools.shared.cookies import Cookie
 from aws_lambda_powertools.shared.functions import powertools_dev_is_set
 from aws_lambda_powertools.shared.json_encoder import Encoder
@@ -2386,6 +2386,7 @@ class Router(BaseRouter):
             methods = (method,) if isinstance(method, str) else tuple(method)
             frozen_responses = _FrozenDict(responses) if responses else None
             frozen_tags = frozenset(tags) if tags else None
+            frozen_security = _FrozenListDict(security) if security else None
 
             route_key = (
                 rule,
@@ -2400,7 +2401,7 @@ class Router(BaseRouter):
                 frozen_tags,
                 operation_id,
                 include_in_schema,
-                security,
+                frozen_security,
             )
 
             # Collate Middleware for routes
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -1,5 +1,6 @@
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
+from aws_lambda_powertools.event_handler.openapi.models import SecurityScheme
 from aws_lambda_powertools.utilities.data_classes.shared_functions import get_header_value
 
 
@@ -48,11 +49,15 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
 
     The 'origin' or 'Origin' header can be either a single header or a multi-header.
 
-    Args:
-        resolver_headers (Dict): A dictionary containing the headers.
+    Parameters
+    ----------
+    resolver_headers: Dict
+        A dictionary containing the headers.
 
-    Returns:
-        Optional[str]: The value(s) of the origin header or None.
+    Returns
+    -------
+    Optional[str]
+        The value(s) of the origin header or None.
     """
     resolved_header = get_header_value(
         headers=resolver_headers,
@@ -64,3 +69,30 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
         return resolved_header[0]
 
     return resolved_header
+
+
+def validate_openapi_security_parameters(
+    security: List[Dict[str, List[str]]],
+    security_schemes: Optional[Dict[str, "SecurityScheme"]],
+) -> bool:
+    """
+    Validates the security parameters based on the provided security schemes.
+
+    Parameters
+    ----------
+    security: List[Dict[str, List[str]]]
+        A list of security requirements
+    security_schemes: Optional[Dict[str, "SecurityScheme"]]
+        A dictionary mapping security scheme names to their corresponding security scheme objects.
+
+    Returns
+    -------
+    bool
+        True if all security scheme names in the `security` parameter are present in the `security_schemes` parameter,
+        False otherwise.
+
+    """
+
+    return bool(
+        security_schemes and all(key in security_schemes for sec in security for key in sec),
+    )
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict, FrozenSet, List
+from typing import Any, Dict, List
 
 from aws_lambda_powertools.utilities.data_classes.shared_functions import get_header_value
 
@@ -36,7 +36,7 @@ class _FrozenListDict(List[Dict[str, List[str]]]):
     """
 
     def __hash__(self):
-        return hash(FrozenSet({_FrozenDict({key: FrozenSet(self) for key, self in item.items()}) for item in self}))
+        return hash(frozenset({_FrozenDict({key: frozenset(self) for key, self in item.items()}) for item in self}))
 
 
 def extract_origin_header(resolver_headers: Dict[str, Any]):
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -1,4 +1,4 @@
-from typing import Any, Dict
+from typing import Any, Dict, FrozenSet, List
 
 from aws_lambda_powertools.utilities.data_classes.shared_functions import get_header_value
 
@@ -18,7 +18,7 @@ class _FrozenDict(dict):
         return hash(frozenset(self.keys()))
 
 
-class _FrozenListDict(list[dict[str, list[str]]]):
+class _FrozenListDict(List[Dict[str, List[str]]]):
     """
     Freezes a list of dictionaries containing lists of strings.
 
@@ -36,7 +36,7 @@ class _FrozenListDict(list[dict[str, list[str]]]):
     """
 
     def __hash__(self):
-        return hash(frozenset({_FrozenDict({key: frozenset(self) for key, self in item.items()}) for item in self}))
+        return hash(FrozenSet({_FrozenDict({key: FrozenSet(self) for key, self in item.items()}) for item in self}))
 
 
 def extract_origin_header(resolver_headers: Dict[str, Any]):
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -36,7 +36,10 @@ class _FrozenListDict(List[Dict[str, List[str]]]):
     """
 
     def __hash__(self):
-        return hash(frozenset({_FrozenDict({key: frozenset(self) for key, self in item.items()}) for item in self}))
+        hashable_items = []
+        for item in self:
+            hashable_items.extend((key, frozenset(value)) for key, value in item.items())
+        return hash(frozenset(hashable_items))
 
 
 def extract_origin_header(resolver_headers: Dict[str, Any]):
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -18,6 +18,27 @@ class _FrozenDict(dict):
         return hash(frozenset(self.keys()))
 
 
+class _FrozenListDict(list[dict[str, list[str]]]):
+    """
+    Freezes a list of dictionaries containing lists of strings.
+
+    This function takes a list of dictionaries where the values are lists of strings and converts it into
+    a frozen set of frozen sets of frozen dictionaries. This is done by iterating over the input list,
+    converting each dictionary's values (lists of strings) into frozen sets of strings, and then
+    converting the resulting dictionary into a frozen dictionary. Finally, all these frozen dictionaries
+    are collected into a frozen set of frozen sets.
+
+    This operation is useful when you want to ensure the immutability of the data structure and make it
+    hashable, which is required for certain operations like using it as a key in a dictionary or as an
+    element in a set.
+
+    Example: [{"TestAuth": ["test", "test1"]}]
+    """
+
+    def __hash__(self):
+        return hash(frozenset({_FrozenDict({key: frozenset(self) for key, self in item.items()}) for item in self}))
+
+
 def extract_origin_header(resolver_headers: Dict[str, Any]):
     """
     Extracts the 'origin' or 'Origin' header from the provided resolver headers.
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -71,12 +71,14 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
     return resolved_header
 
 
-def validate_openapi_security_parameters(
+def _validate_openapi_security_parameters(
     security: List[Dict[str, List[str]]],
     security_schemes: Optional[Dict[str, "SecurityScheme"]],
 ) -> bool:
     """
-    Validates the security parameters based on the provided security schemes.
+    This function checks if all security requirements listed in the 'security'
+    parameter are defined in the 'security_schemes' dictionary, as specified
+    in the OpenAPI schema.
 
     Parameters
     ----------
@@ -88,11 +90,11 @@ def validate_openapi_security_parameters(
     Returns
     -------
     bool
-        True if all security scheme names in the `security` parameter are present in the `security_schemes` parameter,
-        False otherwise.
-
+        Whether list of security schemes match allowed security_schemes.
     """
 
-    return bool(
-        security_schemes and all(key in security_schemes for sec in security for key in sec),
-    )
+    security_schemes = security_schemes or {}
+
+    security_schema_match = all(key in security_schemes for sec in security for key in sec)
+
+    return bool(security_schema_match and security_schemes)
--- a/aws_lambda_powertools/event_handler/util.py
+++ b/aws_lambda_powertools/event_handler/util.py
@@ -73,7 +73,7 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
 
 def _validate_openapi_security_parameters(
     security: List[Dict[str, List[str]]],
-    security_schemes: Optional[Dict[str, "SecurityScheme"]],
+    security_schemes: Optional[Dict[str, "SecurityScheme"]] = None,
 ) -> bool:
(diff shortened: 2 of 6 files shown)


aws_lambda_powertools/event_handler/util.py
```diff
@@ -18,6 +18,27 @@ def __hash__(self):
...
+
+    This operation is useful when you want to ensure the immutability of the data structure and make it
+    hashable, which is required for certain operations like using it as a key in a dictionary or as an
+    element in a set.
+
+    Example: [{"TestAuth": ["test", "test1"]}]
+    """
+
+    def __hash__(self):
+        return hash(frozenset({_FrozenDict({key: frozenset(self) for key, self in item.items()}) for item in self}))
```
heitorlessa: break this into multiple lines to make it easier to read plz

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,31 +30,67 @@ def handler():
...
+    # THEN the resulting schema should have security defined at the operation level, not the top level
+    security = schema.security
+    assert security is None
+
+    operation = schema.paths["/"].get
+    security = operation.security
+    assert security is not None
+
+    assert len(security) == 1
+    assert security[0] == {"apiKey": []}
```
heitorlessa:
```suggestion
    top_level_security = schema.security
    path_level_security = schema.paths["/"].get.security

    assert top_level_security is None
    assert path_level_security[0] == {"apiKey": []}
```

* tip: careful with shadowing variables with the same name. 
* tip: when you have multiple assertion, often times you might be able to narrow down to one or two, otherwise split tests. For example, if you're asserting operation level security scheme for a given value, you are already implicitly asserting that `is not None`, `it has at least 1 item` - you can therefore keep it down to a single assert.

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,31 +30,67 @@ def handler():
...
+    @router.get("/", security=[{"apiKey": []}])
+    def handler():
+        raise NotImplementedError()
+
+    app.include_router(router)
+
+    # WHEN the get_openapi_schema method is called with security defined at the operation level in the Router
+    schema = app.get_openapi_schema(
+        security_schemes={
+            "apiKey": APIKey(name="X-API-KEY", description="API Key", in_=APIKeyIn.header),
```
heitorlessa: is that necessary for the test?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -2386,6 +2387,7 @@ def register_route(func: Callable):
             methods = (method,) if isinstance(method, str) else tuple(method)
             frozen_responses = _FrozenDict(responses) if responses else None
             frozen_tags = frozenset(tags) if tags else None
+            frozen_security = _FrozenListDict(security) if security else None
```
heitorlessa, line 2405: do we anticipate another non-hashable parameter we need to resolve now too?

I remember `tags` was one of them, now `security`. I dunno the implementation yet so I wonder if we need to stop and build a more robust solution OR if `security` is the last piece we overlooked

heitorlessa (changes requested):
one tiny change, a few suggestions to improve tests, and one major question to prevent another bug

aws_lambda_powertools/event_handler/util.py
leandrodamascena: I understand what you mean, but black is formatting this line, so it makes sense to follow the linter formatting. WDYT?

tests/functional/event_handler/test_openapi_security.py
leandrodamascena: Yep, this is because we need to configure the route with the security scheme.

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: At this moment, security is the last piece we overlooked. I can't say that we won't change this function or that we will never add another parameter, but at the moment we have the OpenAPI implementation almost complete, missing things like vendor extension and cookies, for example. But I don't think both will cause problems here.

tests/functional/event_handler/test_openapi_security.py
leandrodamascena: Thanks a lot for this tips, @heitorlessa! I fixed this here and in another test.

aws_lambda_powertools/event_handler/util.py
heitorlessa: use multiple expressions in this case. I'm gonna pull the PR to send a suggestion as Web version is tricky

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: gotcha. Let's do this, if it happens with yet another parameter, we will stop and prioritize investigating a refactoring. The encoder is a good example of an area that we should've used a singledispatch to avoid a myriad of if/elif we overlooked at launch (issue with extremely large PRs).

heitorlessa:
Adding a summary shortly after our call that led to important discoveries

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -1,23 +1,27 @@
...
     @app.get("/")
     def handler():
         raise NotImplementedError()
 
+    # WHEN the get_openapi_schema method is called with a security scheme
     schema = app.get_openapi_schema(
         security_schemes={
             "apiKey": APIKey(name="X-API-KEY", description="API Key", in_=APIKeyIn.header),
         },
         security=[{"apiKey": []}],
```
heitorlessa: **change**: move these into a top-level variable like `top_level_security_schemes` to ease reading the intent (pre-setup), please.

Later it'll make it even clearer when we have a new API to configure OpenAPI.

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,37 +30,63 @@ def handler():
...
 
     @app.get("/", security=[{"apiKey": []}])
     def handler():
         raise NotImplementedError()
 
+    # WHEN the get_openapi_schema method is called with security defined at the operation level
     schema = app.get_openapi_schema(
         security_schemes={
             "apiKey": APIKey(name="X-API-KEY", description="API Key", in_=APIKeyIn.header),
         },
```
heitorlessa: **change**: move these into a top-level variable like `top_level_security_schemes` to ease reading the intent (pre-setup), please.

Later it'll make it even clearer when we have a new API to configure OpenAPI.

tests/functional/event_handler/test_openapi_security.py
heitorlessa: **change**: move these into a top-level variable like `top_level_security_schemes` to ease reading the intent (pre-setup), please.

Later it'll make it even clearer when we have a new API to configure OpenAPI.

aws_lambda_powertools/event_handler/util.py
heitorlessa: Best I could tinker right now - got rid of the extra wrapping as it was unnecessary. Tests passing but lemme know otherwise.

I broke by:

1. Group all items that are hashable
2. We're interested in immutable values so only use `frozenset` in the value itself
3. Create a `frozenset` out of all hashable items
4. Hash it

That way we also don't need to use `_FrozenDict` to wrap some of it too.

---

```suggestion
        hashable_items = []
        for item in self:
            hashable_items.extend((key, frozenset(value)) for key, value in item.items())
        
        return hash(frozenset(hashable_items))
```

heitorlessa (changes requested):
last changes before merging the PR.

* Suggested a multi-expression and slightly better perf for hashing (along w/ explanation)
* Suggested a change in tests to make it more explicit that there is a `setup` for OpenAPI until we have a new API to solve the ambiguity
* We're missing a test for Security Scheme mismatch. Top-level says `ApiKey` but path level uses a `JWT`.

If you've got time, it'd be great to already address other pieces we missed prior to this PR

* An OpenAPI Schema specific exception, not `ValueError`. The exceptions we're returning can be more actionable too (we can start with the security scheme)

Thank you so much for the hard work on this <3

heitorlessa:
For anyone reading this later...

As part of the security scheme bug fix review (#4421), we found that the responsibilities between OpenAPI and SwaggerUI are ambiguous, leading to confusion among customers and maintainers _at times_.

As of now, these are common use cases:

1. **Setting up OpenAPI**. Customers use a combination of implicit type annotations, OpenAPI params at the route level, and extra params via `enable_swagger`.
2. **Setting up SwaggerUI**. Customers use `enable_swagger` to render the implicit  OpenAPI Schema in a web app _(SwaggerUI)_, customize assets, etc.
3. **Serializing OpenAPI JSON Schema**. Customers create an explicit route _e.g.,_ `/schema`, and use [`app.get_openapi_json_schema`](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#global-openapi-security-schemes) to serialize it. Those using SwaggerUI can download it at the homepage _(https://endpoint.url/swagger?format=json)_.

We're confusing customers on 1 and 2, because `enable_swagger` or `get_openapi_*schema` is the only way to further extend / override OpenAPI. Customers **should not** need to use `enable_swagger` or `get_openapi_*schema` to further customize OpenAPI.

**How do we fix it in a backwards compatible way?**

> need two new issues

* Introduce a new API to configure / extend OpenAPI in the resolver instance _(`app`)_
* Improve docs by surfacing that API instead of `get_openapi_*schema` and `enable_swagger` - [example](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#customizing-openapi-metadata)
* Rename and revamp OpenAPI docs sections to match the use-case driven docs  we have. For now, it should be easier to find how to
  * [ ] extend OpenAPI spec
  * [ ] export OpenAPI schema programmatically
  * [ ] override OpenAPI schema when exporting
  * [ ] when to override OpenAPI metadata in SwaggerUI

aws_lambda_powertools/event_handler/util.py
leandrodamascena: good catch!

tests/functional/event_handler/test_openapi_security.py
leandrodamascena: Done

tests/functional/event_handler/test_openapi_security.py
leandrodamascena: Done

tests/functional/event_handler/test_openapi_security.py
leandrodamascena: Done

leandrodamascena (commented):
Thanks for the review and suggestions @heitorlessa!

heitorlessa:
checking in 👀 ;)

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1630,15 +1645,14 @@ def _get_openapi_security(
...
             return None
 
-        if not security_schemes:
-            raise ValueError("security_schemes must be provided if security is provided")
-
-        # Check if all keys in security are present in the security_schemes
-        if any(key not in security_schemes for sec in security for key in sec):
-            raise ValueError("Some security schemes not found in security_schemes")
+        if not validate_openapi_security_parameters(security=security, security_schemes=security_schemes):
+            raise SchemaValidationError(
```
leandrodamascena, line 1654: Hey @heitorlessa! I've updated the exception from `ValueError` to `SchemaValidationError` as it is a schema validation and we should use the appropriate exception name. I don't think that customers are using this, but we can revert in any case. WDYT?

heitorlessa: great thanks! I'll look into the messaging (what we tell customers in the exception) after lunch

aws_lambda_powertools/event_handler/util.py
```diff
@@ -40,3 +69,30 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
...
 
     return resolved_header
+
+
+def validate_openapi_security_parameters(
+    security: List[Dict[str, List[str]]],
+    security_schemes: Optional[Dict[str, "SecurityScheme"]],
+) -> bool:
+    """
+    Validates the security parameters based on the provided security schemes.
```
heitorlessa: plz add more details on the logic here, it's easy to miss the difference between the two by looking at the `all` logic.

aws_lambda_powertools/event_handler/util.py
```diff
@@ -40,3 +69,30 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
         return resolved_header[0]
 
     return resolved_header
+
+
+def validate_openapi_security_parameters(
```
heitorlessa:
```suggestion
def _validate_openapi_security_parameters(
```

make it private to avoid breakage, as we should move these type of validations into a Schema Validator (missing?) once we have one

aws_lambda_powertools/event_handler/util.py
```diff
@@ -40,3 +69,30 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
...
+    -------
+    bool
+        True if all security scheme names in the `security` parameter are present in the `security_schemes` parameter,
+        False otherwise.
+
+    """
+
+    return bool(
+        security_schemes and all(key in security_schemes for sec in security for key in sec),
+    )
```
heitorlessa:
```suggestion
    
    security_schemes = security_schemes or {}

    security_schema_match = all(
        key in security_schemes for sec in security for key in sec
    )

    return bool(security_schema_match and security_schemes)
```

A few changes:

* **Always guard against a `None`**. Otherwise the `all` will raise `TypeError: argument of type 'NoneType' is not iterable`
* **Break statements when there are two four loops**. It's easy to accidentally get into a [quadratic situation](https://accidentallyquadratic.tumblr.com/). The exercise of breaking it allows you to have another look at the logic to see if there's another way to do it w/o a double loop (not in this case). At worst, it improves readability
* **Use intentional var names on validation**. For schema validation logic, it's easy to get lost when coming back to code as most of them look the same _(e.g., iterating through a data structure)_. Breaking statements, using var names for conditions, or breaking into multiple functions based on conditions make it easier to maintain later.

Q: is there a better naming we can use to make it less ambiguous for these function parameters?

aws_lambda_powertools/event_handler/util.py
```diff
@@ -40,3 +69,30 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
...
+    ----------
+    security: List[Dict[str, List[str]]]
+        A list of security requirements
+    security_schemes: Optional[Dict[str, "SecurityScheme"]]
+        A dictionary mapping security scheme names to their corresponding security scheme objects.
+
+    Returns
+    -------
+    bool
+        True if all security scheme names in the `security` parameter are present in the `security_schemes` parameter,
```
heitorlessa:
```suggestion
        Whether list of security schemes match allowed security_schemes.
```

is that what you meant?

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1588,6 +1594,15 @@ def get_openapi_schema(
 
         # Add routes to the OpenAPI schema
         for route in all_routes:
+
+            if route.security and not validate_openapi_security_parameters(
```
heitorlessa: QQ: As a rule of thumb, if `route.security` is None we don't want to proceed with any validation? 

If the answer is yes, I'd suggest encapsulating that as part of the validator function saving an extra `if` since `and` is an extra if..

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -1588,6 +1594,15 @@ def get_openapi_schema(
...
+                security=route.security,
+                security_schemes=security_schemes,
+            ):
+                raise SchemaValidationError(
+                    "Security configuration was not found in security_schemas or security_schema was not defined.",
```
heitorlessa: anything else we think it'd be useful to tell them? e.g., anything to look to fix it in the route X?

If not, leave as is.

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,37 +25,105 @@ def handler():
...
+    # WHEN the get_openapi_schema method is called with security defined without security schemes
+    # THEN a SchemaValidationError should be raised
+    with pytest.raises(SchemaValidationError):
         app.get_openapi_schema(
             security=[{"apiKey": []}],
         )
 
 
-def test_openapi_operation_security():
+def test_openapi_top_level_security_mismatch(security_scheme):
```
heitorlessa, line 43: Niiiice!!

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,37 +25,105 @@ def handler():
...
+    # GIVEN an APIGatewayRestResolver instance
+    app = APIGatewayRestResolver()
+
+    @app.get("/")
+    def handler():
+        raise NotImplementedError()
+
+    # WHEN the get_openapi_schema method is called with security defined security schemes as APIKey
+    # WHEN top level security is defined as HTTPBearer
```
heitorlessa:
```suggestion
    # AND top level security is defined as HTTPBearer
```

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,37 +25,105 @@ def handler():
...
-    security = operation.security
-    assert security is not None
 
-    assert len(security) == 1
-    assert security[0] == {"apiKey": []}
+def test_openapi_operation_level_security_missing():
+    # GIVEN an APIGatewayRestResolver instance
+    app = APIGatewayRestResolver()
+
+    # WHEN we define a security in operation
```
heitorlessa:
```suggestion
    # AND a route with a security scheme defined
```

tests/functional/event_handler/test_openapi_security.py
```diff
@@ -26,37 +25,105 @@ def handler():
...
+    # THEN a SchemaValidationError should be raised
+    with pytest.raises(SchemaValidationError):
+        app.get_openapi_schema()
+
+
+def test_openapi_operation_level_security_mismatch(security_scheme):
+    # GIVEN an APIGatewayRestResolver instance
+    app = APIGatewayRestResolver()
+
+    # WHEN we define a security in operation with value HTTPBearer
```
heitorlessa:
```suggestion
    # AND a route with a security scheme using HTTPBearer
```

heitorlessa (changes requested):
THANK YOU for working out the missing tests <3

Made last round of suggestions and another accidental bug in the validator. We can merge afterwards

aws_lambda_powertools/event_handler/util.py
leandrodamascena: >Always guard against a None. Otherwise the all will raise TypeError: argument of type 'NoneType' is not iterable

In this case will not because the first expression test if security_schemes is defined. But your code makes it more clear.

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: In this case we will have `if` here or there, right? So, maybe it doesn't make much difference because we will always have if to check whether the route is `None` or not. I think the code and the intention of what we want to do will be more explicit now. WDYT?

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: I think this message is very explicit. Maybe add a `See: https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/api_gateway/#security-schemes`?

aws_lambda_powertools/event_handler/util.py
leandrodamascena: Great catch!

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: Yes, let's add that. The section itself deserves improvement too, as we can now mention that a mismatch will raise a validation error.

aws_lambda_powertools/event_handler/api_gateway.py
heitorlessa: > in this case we will have `if` here or there, right?

You're 10000000% right. I assumed Python would not short-circuit and waste resources evaluating both.. but [IT DOES](https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not).

I'm gonna resolve it.

My personal preference is to always encapsulate validation within the validator to avoid surprises, but changing or not here doesn't lead to any bug :)

aws_lambda_powertools/event_handler/util.py
```diff
@@ -71,12 +71,14 @@ def extract_origin_header(resolver_headers: Dict[str, Any]):
     return resolved_header
 
 
-def validate_openapi_security_parameters(
+def _validate_openapi_security_parameters(
     security: List[Dict[str, List[str]]],
     security_schemes: Optional[Dict[str, "SecurityScheme"]],
```
heitorlessa:
```suggestion
    security_schemes: Optional[Dict[str, "SecurityScheme"]] = None,
```

If the type annotation is correct, you're missing a `None`

aws_lambda_powertools/event_handler/util.py
heitorlessa: DUh, of course, I missed the `and` again.

docs/core/event_handler/api_gateway.md
```diff
@@ -1033,7 +1033,7 @@ Below is an example configuration for serving Swagger UI from a custom path or C
     No. Powertools adds support for generating OpenAPI documentation with [security schemes](https://swagger.io/docs/specification/authentication/), but it doesn't implement any of the security schemes itself, so you must implement the security mechanisms separately.
 
 OpenAPI uses the term security scheme for [authentication and authorization schemes](https://swagger.io/docs/specification/authentication/){target="_blank"}.
-When you're describing your API, declare security schemes at the top level, and reference them globally or per operation.
+When you're describing your API, declare security schemes at the top level, and reference them globally or per operation. Remember to have the security configurations defined in the `security_schemes`, otherwise a `SchemaValidationError` will be raised.
```
heitorlessa:
```suggestion

Security schemes are declared at the top-level first. You can reference them globally or on a per path _(operation)_ level. **However**, if you reference security schemes that are not defined at the top-level it will lead to a `SchemaValidationError` _(invalid OpenAPI spec)_.
```

heitorlessa (approved):

heitorlessa:
suuuuperb working in going through bugs, finding new ones, and improving that section of the docs. It's also great that we identified the whole OpenAPI docs need a big revamp, **and** that we need a new API to make this easier and less confusing <3

who's awesome!?

PR #2917, opened 2023-08-03
feat(event_handler): add Middleware support for REST Event Handler

walmsles:
**Issue number:** #953 #1955 #1696

## Summary

Adds middleware support to event_handlers for REST API Gateway resolvers (all types).  First pass POC implementation following a similar implementation style to Chalice and Starlette

- need to add documentation
- functional tests in place
- Reference middleware implementations provided for CORSMiddleware, CacheControlMiddleware as an alternative to existing ResponseBuilder implementation - existing ResponseBuilder implementation unchanged

### Changes

- BaseRouter route functions have additional `middleware` parameter expecting a list of `Callable` middleware handlers
- Added special case `registered_api_middleware`, which is added as the **last** middleware handler internally (maintain backward compatibility) - without this function, the signature for REST routes would change to match the middleware signature, which feels like a breaking change.  This special middleware also ensures all middleware will get a Response object as the response, which is essential for being able to write middleware (no type guessing).
- Added example Core middleware as Class based middleware - CORSMiddleware, CacheControlMiddleware
- Middleware can be any Python `Callable`
- Capability for adding **global** middleware (all routes) and route-specific middleware.  Global middleware is always executed first since it represents cross-cutting concerns; anything else should be route specific.  
- No ordering gymnastics - Middleware is applied in Order of definition, router-middleware + route-middleware

### User experience

Users can now add middleware either for a specific route or centrally for ALL routes:

```python
import requests

from aws_lambda_powertools import Logger
from aws_lambda_powertools.event_handler import APIGatewayRestResolver, Response
from aws_lambda_powertools.event_handler.middlewares import NextMiddleware

app = APIGatewayRestResolver()
logger = Logger()


def inject_correlation_id(app: APIGatewayRestResolver, next_middleware: NextMiddleware) -> Response:
    request_id = app.current_event.request_context.request_id

    # Use API Gateway REST API request ID if caller didn't include a correlation ID
    correlation_id = app.current_event.headers.get("x-correlation-id", request_id)

    # Inject correlation ID in shared context and Logger
    app.append_context(correlation_id=correlation_id)
    logger.set_correlation_id(request_id)

    # Get response from next middleware OR /todos route
    result = next_middleware(app)

    # Include Correlation ID in the response back to caller
    result.headers["x-correlation-id"] = correlation_id
    return result


@app.get("/todos", middlewares=[inject_correlation_id])
def get_todos():
    todos: Response = requests.get("https://jsonplaceholder.typicode.com/todos")
[...]

PR #4492, opened 2024-06-12
fix(event_handler): do not skip middleware and exception handlers on 404 error

### LINKED ISSUE #3916: Bug: Event handler middleware not triggered for not found routes
### Expected Behaviour

I *think* I would expect (global) middleware to be triggered when a not-found route is used.   I have middleware that implements desirable behaviour even when the route is not found.

### Current Behaviour

Currently global middleware seems to be bypassed when a not found route is hit.

### Code snippet

```python
from aws_lambda_powertools.event_handler import APIGatewayRestResolver, Response
from aws_lambda_powertools.event_handler.middlewares import NextMiddleware


def test_resolve():
    def log_middleware(
        app: APIGatewayRestResolver, next_middleware: NextMiddleware
    ) -> Response:
        print("middleware before")
        return next_middleware(app)

    app.use(middlewares=[log_middleware])

    @app.get("/listItems")
    def list_items():
        return [{"id": 1}, {"id": 2}]

    @app.not_found
    def not_found(exc):
        return Response(status_code=404)

    # calling this triggers middleware
    app.resolve(
        {
            "path": "/listItems",
            "httpMethod": "GET",
            "requestContext": {},
        },
        {},
    )

    # calling this does not trigger middleware
    app.resolve(
        {
            "path": "/junk",
            "httpMethod": "GET",
            "requestContext": {},
        },
        {},
    )
```

### Possible Solution

The solution I've used currently is to use catch all route handlers to sweep up any unexpected routes, which looks to app like a "valid" route.  This means middleware is triggered.

Another solution would be to trigger global middleware before route resolution.  

### Steps to Reproduce

I think I've supplied enough context.  The above code snippet should run under `pytest -rP` Please let me know if more is needed.

latest

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

heitorlessa:
**Issue number:** #3916

## Summary

This PR addresses the bug where non-matched routes (404) did not kick off the request chain and instead immediately process a 404 response.

The reasons this PR do not reach for a 404 middleware solution are:

1. Middlewares are tightly coupled with a Route, so request processing is kicked off from a route (`__call__()`)
2. Moving as a middleware would require a dummy/no-op route to be called only to be overwritten, causing unnecessary CPU cycle and memory usage (tbh, it's somewhat a minor and we can revert later).

> NOTE. Nothing stops us for making it a middleware w/ a dummy route to kick the request chain when we have more time.  This was an issue ever since middleware was introduced in #2917, so I'd favoured keeping the core the same and focusing on covering any edge cases _(e.g., what if a middleware short-circuit)_.

### Changes

* [x] Removed duplicated CORS Origin Header logic
* [x] Removed unnecessary CORS injection within 404 logic
* [x] Creates 404 route handler and call it to kick off request chain
* [x] Test FastAPI and Flask to be triple sure it's a standard 
* [x] Test pre-flight for not found route with middleware
* [ ] Enrich docstring for resolve() to share request flow

**unrelated**

* [x] Moved special sorted set to list str for CORS Allow Methods to CORSConfig

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


heitorlessa:
@leandrodamascena can I get your help in doing your crazy QA skills just in case I missed any edge case?

leandrodamascena:
> @leandrodamascena can I get your help in doing your crazy QA skills just in case I missed any edge case?

Hey @heitorlessa, I've spent some time testing various scenarios with different kinds of Resolvers, including Bedrock, and everything seems to be working fine for me. CORS  is working correctly also, and it's also running Middleware with the OPTIONS method/route.

e2e tests are also green. A note for the future is to create e2e tests with Middleware to ensure that we are testing those scenarios.


image

leandrodamascena (approved):
Thanks for this hard and amazing work @heitorlessa!

APPROVED!

PR #4508, opened 2024-06-13
feat(validation): support JSON Schema referencing in validation utils

### LINKED ISSUE #4499: Support JSON Schema referencing in validation utils (by using fastjsonschema's handlers param)
### Use case

`fastjsonschema`, which is used internally by [aws-lambda-powertools' validation utilities](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/validation/base.py#L32), supports [JSON Schema's schema referencing](https://json-schema.org/understanding-json-schema/structuring#dollarref), but it by default launches an http request to the url provided. However, in the case that users don't want to launch an http request to fetch the referenced schemas (for example, because the references are stored locally), `fastjsonschema` also provides the `handlers` parameter to its validation methods ([docs](https://horejsek.github.io/python-fastjsonschema/#fastjsonschema.compile))

> You can pass mapping from URI to function that should be used to retrieve remote schemes used in your definition in parameter handlers.

### Solution/User Experience

Update aws-lambda-powertools' validation functions to support passing the `handlers` param to fastjsonschema's `validate` call. As far as I can tell, this would require adding the param to [validate_data_against_schema](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/validation/base.py#L32) and to the user-facing methods/decorators [validate and validator](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/validation/validator.py)

### Alternative solutions

_No response_

### Acknowledgment

danieljnchen:
**Issue number:** https://github.com/aws-powertools/powertools-lambda-python/issues/4499

### Changes

This change updates aws-lambda-powertools' validation functions to support passing the `handlers` param to fastjsonschema's validate call.

Customers now can use this new parameter

```python
from aws_lambda_powertools.utilities.validation import validate, validator

PARENT_SCHEMA = {
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "https://example.com/schemas/parent.json",
  "type": "object",
  "properties": {
    "ParentSchema": {
      "$ref": "https://SCHEMA"
    }
  }
}

CHILD_SCHEMA = {
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "https://example.com/schemas/child.json",
  "type": "object",
  "properties": {
    "project": {
      "type": "string"
    }
  },
  "required": ["project"]
}

def get_child_schema(uri):
    return CHILD_SCHEMA

@validator(inbound_schema=PARENT_SCHEMA,inbound_handlers={"https": get_child_schema})
    return "OK"
```

### User experience

Users previously were unable to use schemas with [JSON Schema references](https://json-schema.org/understanding-json-schema/structuring#dollarref)

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/utilities/validation/base.py
```diff
@@ -8,7 +8,9 @@
 logger = logging.getLogger(__name__)
 
 
-def validate_data_against_schema(data: Union[Dict, str], schema: Dict, formats: Optional[Dict] = None):
+def validate_data_against_schema(
+    data: Union[Dict, str], schema: Dict, formats: Optional[Dict] = None, handlers: Optional[Dict] = None
+):
```
leandrodamascena: Hello @danieljnchen! I'm looking at the [Validation API](https://horejsek.github.io/python-fastjsonschema/#fastjsonschema.validate) and it seems to accept other parameters like `use_default`, `use_formats` and they may add others in the future. So what do you think about adding `**kwargs` and propagating that to the `validate` function?

leandrodamascena (changes requested):
Hello @danieljnchen! Thans a lot for this super quick and very useful PR! We really need to offer this support to our customers.
I left a comment before I started doing a more detailed review.

Thank you and please ping me if you have any difficulties and I can help.

aws_lambda_powertools/utilities/validation/base.py
heitorlessa: `**provider_options` so it's more explicit. ideally, we'd encapsulate all that in a `provider`, so a customer could swap `fastjsonschema` with `jsonschema` and whatnot too; it'd be too much work for Daniel, so just having an explicit kwargs name + docs help

danieljnchen: Makes sense, and thanks for the feedback! I'll add some `**provider_options` to make sure we can pass other parameters in the future.

danieljnchen: Alright, thanks for your patience! I've added the `**provider_options` arg to `validate_data_against_schema` and fixed the tests that were failing based on what I could see in the CI pipeline. Still haven't gotten tests running in my dev environment so I may have missed something ;)

leandrodamascena: Hello @danieljnchen! Thanks a lot for addressing our feedback and updating this PR! I see some items failing in our CI, but I'm sure it's because you couldn't run local tests. I know you're blocked because the environment we create in gitpod is broken.

We are fixing this and we can come back with the review of this PR next week, ok?

danieljnchen: No problem, thank you for your feedback and help with the environment issue! Happy to pick this back up next week :)

leandrodamascena: Hello @danieljnchen! We have good news. We fixed the Gitpod environment and you can use it now. You may need to rebuild your existing workspace with this new configuration.

https://github.com/aws-powertools/powertools-lambda-python/pull/4532

danieljnchen: Awesome, thank you! Will update my environment and see if I can get tests running

aws_lambda_powertools/utilities/validation/base.py
```diff
@@ -29,7 +39,9 @@ def validate_data_against_schema(data: Union[Dict, str], schema: Dict, formats:
     """
     try:
         formats = formats or {}
-        fastjsonschema.validate(definition=schema, data=data, formats=formats)
+        handlers = handlers or {}
+        provider_options = provider_options or {}
```
danieljnchen, line 44: So... it looks like this last line
```
provider_options = provider_options or {}
```
is what bumps the xenon complexity from A to B.

Any thoughts on how to resolve this? I haven't worked with this sort of constraint -- this additional line seems pretty reasonable to me.

danieljnchen: Looks like the `complexity-baseline` task in the Makefile manually excludes `aws_lambda_powertools/shared/json_encoder.py`? Seems a bit hacky to skip the check that way, though the complexity checker also seem a bit aggressive. I can add this file to the skipped files if that's how we want to resolve this failed check.

danieljnchen:
Alright, my development environment is all set up, and I've run the `make pr` step. The only check that's failing is the complexity check that I left a note about. Let me know if I need to make any more changes to get this ready to go in!

aws_lambda_powertools/utilities/validation/base.py
leandrodamascena: Hey @danieljnchen, this is because you added this new field (`provider_options = provider_options or {}`) with a conditional and Xenon/Radon is complaining about cyclomatic complexity.
I agree that the check is a bit aggressive in this case and we can skip it manually. Let me review the PR and I'll leave a message with the decision, ok?

leandrodamascena:
> Alright, my development environment is all set up, and I've run the `make pr` step. The only check that's failing is the complexity check that I left a note about. Let me know if I need to make any more changes to get this ready to go in!

Hi @danieljnchen! This is amazing! I'm reviewing this PR and if I find something small and easy to fix, I'll push some commits to get it merged quickly!

Thank you very much for taking the time to make this improvement!

leandrodamascena (dismissed):
Hey @danieljnchen! I've made some minor changes to this PR, let me explain them:

1 - To reduce the cyclomatic complexity, I created a new function, and now we are testing if variables are `None` or not through this function. This approach allows us to avoid direct if statements in the specific code, thereby reducing complexity and making Xenon happy.

2 - I refactored the example to make it more clear for customers that want just to copy and past and test this in their Lambda functions.

This PR is happily approved ❤️ .

aws_lambda_powertools/utilities/validation/base.py
leandrodamascena: > **provider_options so it's more explicit. ideally, we'd encapsulate all that in a provider, so a customer could swap fastjsonschema with jsonschema and whatnot too; it'd be too much work for Daniel, so just having an explicit kwargs name + docs help

Hello @heitorlessa! I'm going to open an issue to implement this because we need to refactor a lot of things to make this idea work.

danieljnchen:
> Hey @danieljnchen! I've made some minor changes to this PR, let me explain them:
> 
> 1 - To reduce the cyclomatic complexity, I created a new function, and now we are testing if variables are `None` or not through this function. This approach allows us to avoid direct if statements in the specific code, thereby reducing complexity and making Xenon happy.
> 
> 2 - I refactored the example to make it more clear for customers that want just to copy and past and test this in their Lambda functions.
> 
> This PR is happily approved ❤️ .

@leandrodamascena (and @heitorlessa) thanks for your help getting this PR all cleaned up! I'm really excited to see this go in -- this is my first time contributing to an open-source project (I tried with python-dateutil a while back, but my MR never got reviewed). I appreciate the responsiveness and encouragement!

aws_lambda_powertools/utilities/validation/base.py
```diff
@@ -19,6 +26,10 @@ def validate_data_against_schema(data: Union[Dict, str], schema: Dict, formats:
         JSON Schema to validate against
     formats: Dict
         Custom formats containing a key (e.g. int64) and a value expressed as regex or callback returning bool
+    handlers: Dict
+        Custom methods to retrieve remote schemes, keyed off of URI scheme
+    provider_options: Dict
+        Arguments that will be passed directly to the underlying validate call
```
heitorlessa: suggestion: Until we have a providers implementation, add info that these are kwargs supported by `fastjsonschema` validate function, followed by a link to their API.

That will help customers know what's possible w/o opening the docs. 

As a rule of thumb, what we strive is to make sure a customer doesn't feel like "hmm I must be dumb as I don't know how to use this" <-- in short, reduce cognitive load, make the implicit explicit.

aws_lambda_powertools/utilities/validation/exceptions.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any, List, Optional
 
-from ...exceptions import InvalidEnvelopeExpressionError
+from aws_lambda_powertools.exceptions import InvalidEnvelopeExpressionError
```
heitorlessa, line 3: is this correct?

aws_lambda_powertools/utilities/validation/validator.py
```diff
@@ -1,10 +1,9 @@
 import logging
 from typing import Any, Callable, Dict, Optional, Union
 
+from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
```
heitorlessa, line 4: love the tech debt fixes from relative to absolute imports <3

aws_lambda_powertools/utilities/validation/validator.py
```diff
@@ -44,6 +47,14 @@ def validator(
...
     outbound_formats: Dict
         Custom formats containing a key (e.g. int64) and a value expressed as regex or callback returning bool
+    inbound_handlers: Dict
+        Custom methods to retrieve remote schemes, keyed off of URI scheme
+    outbound_handlers: Dict
+        Custom methods to retrieve remote schemes, keyed off of URI scheme
+    inbound_provider_options: Dict
+        Arguments that will be passed directly to the underlying validate call for the inbound event
+    outbound_provider_options: Dict
+        Arguments that will be passed directly to the underlying validate call for the outbound event
```
heitorlessa: genuine question - is there a reason why we can't rename `**kwargs` to `**provider_options` instead of adding new ones?

More options typically means a customer will have more room to make mistake or be paralyzed about what to do. 

PS: If you think we must have the ability to have different provider options for inbound **and** outbound, I'll support you; mostly looking to understand what I'm missing.

aws_lambda_powertools/utilities/validation/validator.py
```diff
@@ -161,6 +186,10 @@ def validate(
         Alternative JMESPath options to be included when filtering expr
     formats: Dict
         Custom formats containing a key (e.g. int64) and a value expressed as regex or callback returning bool
+    handlers: Dict
+        Custom methods to retrieve remote schemes, keyed off of URI scheme
+    provider_options: Dict
+        Arguments that will be passed directly to the underlying validate call
```
heitorlessa, line 195: same as earlier, let's help customers understand what's available. We wouldn't be able to type `TypedDict`, as we can't control their API.

docs/utilities/validation.md
```diff
@@ -199,3 +199,31 @@ You can use our built-in [JMESPath functions](./jmespath_functions.md){target="_
 
 ???+ info
     We use these for [built-in envelopes](#built-in-envelopes) to easily to decode and unwrap events from sources like Kinesis, CloudWatch Logs, etc.
+
+### Validating with external references
+
+JSON schema [allows schemas to reference other schemas](https://json-schema.org/understanding-json-schema/structuring#dollarref) using the `$ref` keyword. The value of `$ref` is a URI reference, but you likely don't want to launch an HTTP request to resolve this URI. Instead, you can pass resolving functions through the `handlers` parameter:
```
heitorlessa:
```suggestion
JSON Schema [allows schemas to reference other schemas](https://json-schema.org/understanding-json-schema/structuring#dollarref) using the `$ref` keyword with a URI value. By default, `fastjsonschema` will make a HTTP request to resolve this URI. 

You can use `handlers` parameter to have full control over how references schemas are fetched. This is useful when you might want to optimize caching, reducing HTTP calls, or fetching them from non-HTTP endpoints.

```

changes:

* **might not** over **likely don't**, as we can't assume their preferences given how diverse our customers are.
* _typo_: **JSON Schema** over **JSON schema**.
* _line editing_: merged sentences to give enough context on what happens by default first, followed by why they might want to use this.
* _addition_: added use cases so customers can make an informed decision whether this is for them or skip reading this section. Also helps searching later.

examples/validation/src/custom_handlers.py
```diff
@@ -0,0 +1,13 @@
+from custom_handlers_schema import CHILD_SCHEMA, PARENT_SCHEMA
+
+from aws_lambda_powertools.utilities.typing import LambdaContext
+from aws_lambda_powertools.utilities.validation import validator
+
+
+def get_child_schema(uri):
```
heitorlessa: nitpick: can we type `uri` parameter so customers know what they're dealing with?

aws_lambda_powertools/utilities/validation/base.py
```diff
@@ -28,8 +39,10 @@ def validate_data_against_schema(data: Union[Dict, str], schema: Dict, formats:
         When JSON schema provided is invalid
     """
     try:
-        formats = formats or {}
-        fastjsonschema.validate(definition=schema, data=data, formats=formats)
+        formats = get_field_or_empty_dict(formats)
+        handlers = get_field_or_empty_dict(handlers)
+        provider_options = get_field_or_empty_dict(provider_options)
```
heitorlessa:
```suggestion
        formats = formats or {}
        handlers = handlers or {}
        provider_options = provider_options or {}
```

I see why you did that, but because it's built-in and it's a short statement, we can keep relying on Python's expressiveness here over a function, as there's not much else besides an `or`.

heitorlessa (changes requested):
This is looking GREAT!!! 

Made final comments, one potential error on `exceptions`, a small refactoring, line editing for docs along with explanation as to why each of them.

Looking forward to this - thank you both!!

aws_lambda_powertools/utilities/validation/base.py
leandrodamascena: Reverted.

examples/validation/src/custom_handlers.py
leandrodamascena: great catch.

aws_lambda_powertools/utilities/validation/validator.py
leandrodamascena: >PS: If you think we must have the ability to have different provider options for inbound and outbound, I'll support you; mostly looking to understand what I'm missing.

Yes, that's the reason why we should keep this implementation as is, because we never know if customers want to use the custom handler for inbound, outbound or both.

aws_lambda_powertools/utilities/validation/base.py
leandrodamascena: Great catch.

aws_lambda_powertools/utilities/validation/exceptions.py
leandrodamascena: I don't think this is correct, but we shouldn't remove this. Now we don't know if customers are importing `InvalidEnvelopeExpressionError` from the validator feature.

aws_lambda_powertools/utilities/validation/base.py
```diff
@@ -19,6 +25,11 @@ def validate_data_against_schema(data: Union[Dict, str], schema: Dict, formats:
         JSON Schema to validate against
     formats: Dict
         Custom formats containing a key (e.g. int64) and a value expressed as regex or callback returning bool
+    handlers: Dict
+        Custom methods to retrieve remote schemes, keyed off of URI scheme
+    provider_options: Dict
+        Arguments that will be passed directly to the underlying validation call, in this case fastjsonchema.validate.
```
heitorlessa, line 31: niiice!

heitorlessa (dismissed):

leandrodamascena (dismissed):

sthulb (approved):

aws_lambda_powertools/utilities/validation/validator.py
danieljnchen: FYI I chose not to use the `kwargs` value here because it is already being used to pass a custom set of arguments to the `handler` method. We wouldn't have a way to distinguish between arguments the user wants to pass to the json schema validation provider and arguments the user wants to pass to the `handler` method.

PR #4606, opened 2024-06-22
feat(data_classes): return empty dict or list instead of None

### LINKED ISSUE #1359: Feature request: query_string_parameters should never be None
### Use case

When there are no query string params, `app.current_event.query_string_parameters` is `None` which is  unexpected.

### Solution/User Experience

Proper solution is to make it to behave like an empty dict or to be an actual empty dict.
- avoids extra None checks
- removes one point of surprise which was totally unnecessary

### Alternative solutions

_No response_

### Acknowledgment

### LINKED ISSUE #2605: Feature request: Return empty Dict or List in Event Source Data Classes instead of None
### Use case

Arguably, it might be easier for users if an empty Dict is returned instead of `None` for path_parameters in APIGatewayProxyEventV2, for example.

Instead of:
```python
@event_source(data_class=APIGatewayProxyEventV2)
def lambda_handler(event: APIGatewayProxyEventV2, context) -> None:
    parameters = event.path_parameters or {}
    handle(parameters["id"])
```
it would become:
```python
@event_source(data_class=APIGatewayProxyEventV2)
def lambda_handler(event: APIGatewayProxyEventV2, context) -> None:
    handle(event.path_parameters["id"])
```

Ultimately this approach could be followed for any return value that is either an Optional[Dict] or an Optional[List], returning an empty Dict or List instead of `None` in all cases:

* utilities/data_classes/api_gateway_proxy_event.py
  * APIGatewayEventAuthorizer
    * `def claims(self) -> Dict[str, Any]`
    * `def scopes(self) -> List[str]`
  * APIGatewayProxyEvent
    * `def multi_value_query_string_parameters(self) -> Dict[str, List[str]]`
    * `def path_parameters(self) -> Dict[str, str]`
    * `def stage_variables(self) -> Dict[str, str]`
  * RequestContextV2AuthorizerIam
  	* `def cognito_amr(self) -> List[str]`
  * RequestContextV2Authorizer
    * `def jwt_claim(self) -> Dict[str, Any]`
    * `def jwt_scopes(self) -> List[str]`
    * `def get_lambda(self) -> Dict[str, Any]`
  * APIGatewayProxyEventV2
    * `def cookies(self) -> List[str]`
    * `def path_parameters(self) -> Dict[str, str]`
    * `def stage_variables(self) -> Dict[str, str]`

and so on...

The bottom line is: Does it matter for users to differentiate between `None` and an empty Dict or List for any Event Source Data Classe property? If it does for a particular one, it an exception to a rule or the rule?

This change would be backwards compatible with the previous code.

This approach was already implemented in only one place so far, as far as I could check:

* utilities/data_classes/common.py
  * BaseProxyEvent
    * `def headers(self) -> Dict[str, str]`

### Solution/User Experience

Solution can be the same as for the headers in the BaseProxyEvent:

https://github.com/aws-powertools/powertools-lambda-python/blob/1905e4cb022b5547d75064a643358c2343977be5/aws_lambda_powertools/utilities/data_classes/common.py#L98-L100

### Alternative solutions

_No response_

### Acknowledgment

ericbn:
**Issue number:** #2605

### Changes

This simplifies the code internally and also for users.

Also wrap all headers in CaseInsensitiveDict from requests.

These changes replace the need of utility functions like get_header_value, get_query_string_value or get_multi_value_query_string_values, which are removed.

### User experience

Before:
```python
encoding = event.get_header_value(name="accept-encoding", default_value="")
params = event.query_string_parameters or {}
id = params["id"]
```

After:
```python
encoding = event.headers.get("accept-encoding", "")
id = event.query_string_parameters["id"]
```

## Checklist

* [X] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py
```diff
@@ -184,71 +183,32 @@ def identity(self) -> Union[None, AppSyncIdentityIAM, AppSyncIdentityCognito]:
...
     @property
-    def prev_result(self) -> Optional[Dict[str, Any]]:
+    def prev_result(self) -> Dict[str, Any]:
         """It represents the result of whatever previous operation was executed in a pipeline resolver."""
         prev = self.get("prev")
         if not prev:
-            return None
-        return prev.get("result")
+            return {}
+        return prev.get("result") or {}
```
ericbn: This might be the only case where returning `Optional[Dict[str, Any]]` could make sense to distinguish between "no previous operation result" (`None`) and "previous operation result is empty" (`{}`). Not sure how the AppSync Resolver Events data looks like.

aws_lambda_powertools/utilities/data_classes/common.py
```diff
@@ -117,14 +113,10 @@ def resolved_query_string_parameters(self) -> Dict[str, List[str]]:
...
-        if self.query_string_parameters is not None:
-            query_string = {key: value.split(",") for key, value in self.query_string_parameters.items()}
-            return query_string
-
-        return {}
+        return {k: v.split(",") for k, v in self.query_string_parameters.items()}
 
     @property
-    def resolved_headers_field(self) -> Dict[str, Any]:
+    def resolved_headers_field(self) -> MutableMapping[str, str]:
```
ericbn: Since we're into breaking changes and tidying up, I would argue that `resolved_headers` is a better name for this property. The "field" in the name is redundant, and not even being used in the related `headers` and `multi_value_headers` properties or the similar `resolved_query_string_parameters`.

leandrodamascena:
Hello @ericbn! I am positively surprised by this PR. This is something we really need to fix in V3 and it would take us days to do it, but I'm glad you're shortening the path!

I started taking a quick look at this PR and it looks like we're headed in the right direction! I saw small things that perhaps we can improve or understand better. I will start reviewing this PR on Monday and hope to get it merged next week!

Really thanks for this amazing work! 🚀

leandrodamascena:
Updating this PR: I started the review last week, but as there are a lot of details I plan to finish it this week.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@
 
 ## Features
 
+* **data_classes:** return empty dict or list in Event Source Data Classes instead of None, and return case-insensitive dict for headrs. ([#2605](https://github.com/aws-powertools/powertools-lambda-python/issues/2605))
```
leandrodamascena: I wouldn't add that. I mean, yes, that's what we're doing in this PR, but it doesn't provide much information for the customer looking at this documentation.

To be honest, our documentation on the EventSource data class is really bad, we need to improve it. I'll take care of this next month.

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -126,7 +126,7 @@ def handler(event, context: LambdaContext):
 
         class MyCustomModel(AppSyncResolverEvent):
             @property
-            def country_viewer(self) -> str:
+            def country_viewer(self):
```
leandrodamascena: isn't that str?

aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
```diff
@@ -410,7 +410,7 @@ def _normalize_multi_query_string_with_param(
     return resolved_query_string
 
 
-def _normalize_multi_header_values_with_param(headers: Dict[str, Any], params: Sequence[ModelField]):
+def _normalize_multi_header_values_with_param(headers: MutableMapping[str, Any], params: Sequence[ModelField]):
```
leandrodamascena, line 413: I'm wondering what the benefit is of having `MutableMapping` here instead of Dict. Are there any additional benefits?

aws_lambda_powertools/utilities/data_classes/alb_event.py
```diff
@@ -1,4 +1,6 @@
-from typing import Any, Dict, List, Optional
+from typing import Any, Dict, List, MutableMapping
+
+from requests.structures import CaseInsensitiveDict
```
leandrodamascena: I really like this idea, but I'm not sure we can do it. We have not included the `requests` library in our layer/package and this will force the client to add the requests library, which can lead to challenges about versioning, conflicts and others.

I created a Lambda, added the Powertools Layer and look at this error

```python
import json
from requests.structures import CaseInsensitiveDict

    # TODO implement
    return {
        'statusCode': 200,
        'body': json.dumps('Hello from Lambda!')
    }

```

image

leandrodamascena (changes requested):
Hello @ericbn! I have started reviewing this PR and have initial feedback that may be a block for further review. Can you check my comments and address them?

I would like to inform you that reviewing this PR will be an excellent source of new learning for me, I am very excited about it. Thanks!

CHANGELOG.md
ericbn: Should I remove this line from the CHANGELOG.md then so this can be taken care of later?

aws_lambda_powertools/event_handler/appsync.py
ericbn: No. That's actually returning `Optional[str]`. I thought it would be simpler to just omit the type. Let me know if I should add ` -> Optional[str]` and the respective Optional import.

aws_lambda_powertools/event_handler/middlewares/openapi_validation.py
ericbn: This is because the `CaseInsensitiveDict` implementation used in the PR for headers is a subclass of `MutableMapping`, but not of `Dict`.

EDIT: `Dict` is also a subclass of `MutableMapping`:
```
>>> from typing import MutableMapping
>>> isinstance(dict(), MutableMapping)
True
```

aws_lambda_powertools/utilities/data_classes/alb_event.py
ericbn: Oh yeah, my fault! boto3 uses urllib3 directly, but not requests. Maybe we should replace requests' CaseInsensitiveDict by urllib3's [HTTPHeaderDict](https://github.com/urllib3/urllib3/blob/79a2a30f06b61127dc138158b82ffa791ab16733/src/urllib3/_collections.py#L206), also a subclass of `MutableMapping` (so my comment regarding `MutableMapping` still applies).

ericbn: @leandrodamascena, I just found an issue with urllib3's HTTPHeaderDict that might not serve all our requirements: the values can only be of type `str`. And we do have cases where values are of type `byte` (in KafkaEventRecord) or `List[str]` (multiValueHeaders).

CHANGELOG.md
leandrodamascena: Yeah, please remove.

aws_lambda_powertools/utilities/data_classes/alb_event.py
leandrodamascena: I'm thinking if we need to rely on some external dependencies or we can create our own `SomenameDict` class and implement the init, get, iter methods. We don't need to add or delete items as this is immutable. Does it make sense to you? If so, go ahead and let's focus on unblocking this item before we move on, okay? I'm focused on merging this PR by next week, so please let me know if I can help with anything.

Once again, thank you very much for your effort in working on such important PR.

ericbn: Cool, let me add our own implementation of `CaseInsensitiveDict` then. We still need the implementation to be mutable actually, exactly because of the _normalize_multi_header_values_with_param function.

ericbn: Actually, let me ask: do you prefer we add a new dependency: https://github.com/DeveloperRSquared/case-insensitive-dict, or add our own implementation, which on the other hand means more code to maintain and test?

leandrodamascena: Tbh, I don't like the idea of adding this new dependency, and let me explain why:

1. The premise of the EventSource DataClass is to be a very fast way to transform JSON into an object using "pure" Python, with no dependencies required. If we bring in a new dependency, we will need to add it to our layer and pip package and force customers to install that, what is not ideal right now.

2. We do open source, and we love to adopt other open-source libraries, but we need to choose libraries very carefully. Supply chain is something we need to be mindful of every day, and I look at this library and see PRs opened since 2023 without a merge, and the last commit was 8 months ago. If we choose to adopt this library and, for some reason, the developers archive this project or even remove the PyPI packages, we are in trouble.

3 - This change is a must because we are wanting to reuse code, as today it's too verbose to extract headers. So, I think we can create a tiny class to do that with only the necessary methods we need, and then reuse this class in our code.

Please let me know if make sense for you all these points.

ericbn:
@leandrodamascena, PR updated.

leandrodamascena:
> @leandrodamascena, PR updated.

Perfect @ericbn! I'll continue the review today!

leandrodamascena:
Just a quick note: I'm still reviewing this PR.

aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -126,7 +126,7 @@ def handler(event, context: LambdaContext):
...
                 return self.request_headers.get("cloudfront-viewer-country")
```
leandrodamascena:
```suggestion
                return self.request_headers.get("cloudfront-viewer-country") or ""
```

aws_lambda_powertools/event_handler/appsync.py
leandrodamascena:
```suggestion
            def country_viewer(self) -> str:
```

aws_lambda_powertools/utilities/data_classes/alb_event.py
```diff
@@ -37,25 +38,15 @@ def multi_value_query_string_parameters(self) -> Dict[str, List[str]]:
 
     @property
     def resolved_query_string_parameters(self) -> Dict[str, List[str]]:
-        if self.multi_value_query_string_parameters:
-            return self.multi_value_query_string_parameters
-
-        return super().resolved_query_string_parameters
+        return self.multi_value_query_string_parameters or super().resolved_query_string_parameters
```
leandrodamascena, line 41: Nice catch

aws_lambda_powertools/utilities/data_classes/alb_event.py
```diff
@@ -37,25 +38,15 @@ def multi_value_query_string_parameters(self) -> Dict[str, List[str]]:
...
-        headers: Dict[str, Any] = {}
-
-        if self.multi_value_headers:
-            headers = self.multi_value_headers
-        else:
-            headers = self.headers
-
-        return {key.lower(): value for key, value in headers.items()}
+    def multi_value_headers(self) -> Dict[str, List[str]]:
+        return CaseInsensitiveDict(self.get("multiValueHeaders"))
```
leandrodamascena, line 45: `CaseInsensitiveDict` class always returns the key in lowercase, right?

aws_lambda_powertools/utilities/data_classes/alb_event.py
```diff
@@ -37,25 +38,15 @@ def multi_value_query_string_parameters(self) -> Dict[str, List[str]]:
...
 
     @property
-    def multi_value_headers(self) -> Optional[Dict[str, List[str]]]:
-        return self.get("multiValueHeaders")
+    def resolved_headers_field(self) -> Dict[str, Any]:
+        return self.multi_value_headers or self.headers
```
leandrodamascena, line 49: Super nice!

aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py
leandrodamascena: Yeah, I agree with you! Please revert to Optional.

aws_lambda_powertools/utilities/data_classes/code_pipeline_job_event.py
```diff
@@ -177,7 +176,7 @@ def user_parameters(self) -> Optional[str]:
         return self.data.action_configuration.configuration.user_parameters
 
     @property
-    def decoded_user_parameters(self) -> Optional[Dict[str, Any]]:
+    def decoded_user_parameters(self) -> Dict[str, Any]:
         """Json Decoded action configuration user parameters"""
         return self.data.action_configuration.configuration.decoded_user_parameters
```
leandrodamascena, line 181: We don't need `or {}` because it will return in configurations, right?

leandrodamascena (changes requested):
I did another review. It will take a few more rounds before we approve it because this is huge and there are a lot of changes.

One thing I need to test a little more is where we return `CaseInsensitiveDict` instead of `self.get("headers"`), for example. This is returning the dictionary without lower in the keys, correct? Or do you always do lower?

Thank you very much for your patience, Eric. This PR has a lot of improvement for Powertools v3, it is, without a doubt, one of the great milestones of this Major release.

aws_lambda_powertools/utilities/data_classes/alb_event.py
ericbn: Yes. The CaseInsensitiveDict is basically a Dict that transforms the key to lower in all operations. The side-effect is that the keys are case-insensitive. E.g. doing `header["Authorization"] = "Bearer x"` or `header.get("Authorization")` will result the same as `header["Authorization".lower()] = "Bearer x"` or `header.get("Authorization".lower())` because of how keys are transformed internally.

aws_lambda_powertools/utilities/data_classes/code_pipeline_job_event.py
ericbn: Correct. `configuration.decoded_user_parameters` already guarantees that at least a `{}` is returned, never `None`. Mypy would also help us here and complain if `configuration.decoded_user_parameters` was annotated to return `Optional[Dict[str, Any]]`, with something like:
```
aws_lambda_powertools/utilities/data_classes/code_pipeline_job_event.py: note: In member "decoded_user_parameters" of class "CodePipelineJobEvent":
aws_lambda_powertools/utilities/data_classes/code_pipeline_job_event.py:181:16: error: Incompatible return value type (got "dict[str, Any] | None", expected "dict[str, Any]")  [return-value]
            return self.data.action_configuration.configuration.decoded_user_parameters
                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ericbn:
@leandrodamascena, I've updated the PR with the changes you proposed. Super productive conversation during the review BTW!

leandrodamascena:
> @leandrodamascena, I've updated the PR with the changes you proposed. Super productive conversation during the review BTW!

Thanks for addressing them so quickly! And I say the same about the conversation. Working with open source projects sometimes take time to approve things, but working collaboratively makes things easier. I will continue the review. 🚀

aws_lambda_powertools/utilities/data_classes/common.py
ericbn: This is a suggestion I'd like you to still consider. Hope it does not get lost by being on top of an outdated commit.

docs/utilities/data_classes.md
```diff
@@ -175,7 +175,7 @@ Use **`APIGatewayAuthorizerRequestEvent`** for type `REQUEST` and **`APIGatewayA
 
     @event_source(data_class=APIGatewayAuthorizerRequestEvent)
     def handler(event: APIGatewayAuthorizerRequestEvent, context):
-        user = get_user_by_token(event.get_header_value("Authorization"))
+        user = get_user_by_token(event.headers["Authorization"])
```
leandrodamascena, line 178: How about using `event.headers.get()`? So we prevent the customer from having problems with `KeyError`.

docs/utilities/data_classes.md
```diff
@@ -263,7 +263,7 @@ See also [this blog post](https://aws.amazon.com/blogs/compute/introducing-iam-a
 
     @event_source(data_class=APIGatewayAuthorizerEventV2)
     def handler(event: APIGatewayAuthorizerEventV2, context):
-        user = get_user_by_token(event.get_header_value("x-token"))
+        user = get_user_by_token(event.headers["x-token"])
```
leandrodamascena, line 266: How about using `event.headers.get()`? So we prevent the customer from having problems with `KeyError`.

examples/event_handler_graphql/src/custom_models.py
```diff
@@ -26,11 +26,11 @@ class Location(TypedDict, total=False):
 class MyCustomModel(AppSyncResolverEvent):
     @property
     def country_viewer(self) -> str:
-        return self.get_header_value(name="cloudfront-viewer-country", default_value="", case_sensitive=False)
+        return self.request_headers.get("cloudfront-viewer-country", "")
 
     @property
     def api_key(self) -> str:
-        return self.get_header_value(name="x-api-key", default_value="", case_sensitive=False)
+        return self.request_headers.get("x-api-key", "")
```
leandrodamascena, line 33:
```suggestion
        return self.request_headers.get("x-api-key") or ""
```

examples/event_handler_graphql/src/custom_models.py
```diff
@@ -26,11 +26,11 @@ class Location(TypedDict, total=False):
 class MyCustomModel(AppSyncResolverEvent):
     @property
     def country_viewer(self) -> str:
-        return self.get_header_value(name="cloudfront-viewer-country", default_value="", case_sensitive=False)
+        return self.request_headers.get("cloudfront-viewer-country", "")
```
leandrodamascena, line 29:
```suggestion
        return self.request_headers.get("cloudfront-viewer-country") or ""
```

examples/event_handler_rest/src/accessing_request_details.py
```diff
@@ -16,12 +16,12 @@
 @app.get("/todos")
 @tracer.capture_method
 def get_todos():
-    todo_id: str = app.current_event.get_query_string_value(name="id", default_value="")
+    todo_id: str = app.current_event.query_string_parameters["id"]
```
leandrodamascena, line 19: How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

examples/event_handler_rest/src/accessing_request_details.py
```diff
@@ -16,12 +16,12 @@
...
     # alternatively
     _: Optional[str] = app.current_event.query_string_parameters.get("id")
 
     # or multi-value query string parameters; ?category="red"&?category="blue"
-    _: List[str] = app.current_event.get_multi_value_query_string_values(name="category")
+    _: List[str] = app.current_event.multi_value_query_string_parameters["category"]
```
leandrodamascena, line 24: How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

examples/event_handler_rest/src/accessing_request_details_headers.py
```diff
@@ -16,7 +16,7 @@
 def get_todos():
     endpoint = "https://jsonplaceholder.typicode.com/todos"
 
-    api_key: str = app.current_event.get_header_value(name="X-Api-Key", case_sensitive=True, default_value="")
+    api_key = app.current_event.headers["X-Api-Key"]
```
leandrodamascena, line 19: How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

examples/event_handler_rest/src/split_route_module.py
```diff
@@ -13,7 +13,7 @@
 @router.get("/todos")
 @tracer.capture_method
 def get_todos():
-    api_key: str = router.current_event.get_header_value(name="X-Api-Key", case_sensitive=True, default_value="")
+    api_key = router.current_event.headers["X-Api-Key"]
```
leandrodamascena, line 16: How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

examples/event_handler_rest/src/split_route_module.py
```diff
@@ -25,11 +25,7 @@ def get_todos():
 @router.get("/todos/<todo_id>")
 @tracer.capture_method
 def get_todo_by_id(todo_id: str):  # value come as str
-    api_key: str = router.current_event.get_header_value(
-        name="X-Api-Key",
-        case_sensitive=True,
-        default_value="",
-    )  # noqa: E501
+    api_key = router.current_event.headers["X-Api-Key"]
```
leandrodamascena, line 28: How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

examples/event_handler_rest/src/split_route_prefix_module.py
```diff
@@ -13,7 +13,7 @@
 @router.get("/")
 @tracer.capture_method
 def get_todos():
-    api_key: str = router.current_event.get_header_value(name="X-Api-Key", case_sensitive=True, default_value="")
+    api_key = router.current_event.headers["X-Api-Key"]
```
leandrodamascena, line 16: How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

tests/unit/data_classes/test_api_gateway_authorizer_event.py
```diff
@@ -52,16 +52,16 @@ def test_api_gateway_authorizer_v2():
     assert parsed_event.path_parameters == raw_event["pathParameters"]
     assert parsed_event.stage_variables == raw_event["stageVariables"]
 
-    assert parsed_event.get_header_value("Authorization") == "value"
-    assert parsed_event.get_header_value("authorization") == "value"
-    assert parsed_event.get_header_value("missing") is None
+    assert parsed_event.headers["Authorization"] == "value"
```
leandrodamascena, line 55: I was thinking about using `.get()`, but this is a test I think doesnt matter here. What do you think?

leandrodamascena (changes requested):
Hello @ericbn! I left some other comments but we're almost there!
There's only one thing I'm testing in the `CaseInsensitiveDict` class, but I don't think we'll have any new thing.

I hope to merge this PR this week! Thanks a lot!

ericbn:
Hi @leandrodamascena.

Some key points worth considering regarding your comments. In no particular order:
* Code is supposed to communicate intent to other programmers reading it later. I like to think that doing `my_dict["key"]` communicates "I expect 'key' to exist in my_dict" and doing `my_dict.get("key")` communicates "'key' might not exist in my_dict sometimes and it's fine"...
* We should get an unhandled exception when something unexpected happens, at least in the Python mindset (other languages like JavaScript were started with the idea of failing silently on the other hand...)
* Failing fast is better than hiding errors.
* There's a preference to use the ["Easier to Ask Forgiveness Than Permission"](https://stackoverflow.com/a/12270384/2654518) than the "Look Before You Leap" style in Python.

This is a good example of the first points above:
```python
assert parsed_event.headers["Authorization"] == "value"
```
It's saying "I expect the Authorization header to exist and to have a value equal to 'value'". If the header didn't exist, the test would fail with KeyError. If the header did exist and had a different value, the assertion would fail with AssertionError. If it was a get instead, we would always get an AssertionError, but lose the ability to distinguish between "header did not exist" and "header exists and has a different value". Also, we would be delaying the error to the assert statement, arguably hiding the original cause of the error. And using get instead wound, I think, communicate a confusing intent.

> How about using `.get()`? So we prevent the customer from having problems with `KeyError`.

I didn't use get in the cases where I thought getting a KeyError would be a better outcome than hiding the fact the the header didn't exist and replace that by a default value. I also didn't use get where the key is expected and would make the subsequent code fail otherwise. This is another good example of some of the points above:

```python
api_key = app.current_event.headers["X-Api-Key"]
todos: Response = requests.get(endpoint, headers={"X-Api-Key": api_key})
todos.raise_for_status()
```
If we used get here instead, we would make a call to the endpoint with an empty X-Api-Key. This would fail at the moment of trying to do the request, which hides the original cause of the error: the X-Api-Key header didn't exist in app.current_event.headers.

> ```diff
> -return self.request_headers.get("x-api-key", "")
> +return self.request_headers.get("x-api-key") or ""
> ```

Here it's good to think about the difference between `my_dict.get("key" , "")` and `my_dict.get("key") or ""`. The former will return `""` when "key" is not found and the latter will result in `""` when "key" either was not found or had itself an empty value (`None`, `""`, etc.). If the key existed and had a value `""`, doing `"" or ""` is redundant. And header values cannot be `None` by definition, so `None or ""` will never happen. Given all this, the latter is overkill and arguably communicates a confusing intent.

I think we're using `d.get(key) or {}` in the data classes to avoid having to think about the nature of each specific code and keep the methods homogeneous. Not doing so would have the risk of bad copy-and-paste of a `d.get(key, {})` code when a `d.get(key) or {}` was expected because the new code has a different nature than the code copied from.

The specific case here has a clear nature: since header values cannot be `None`, we can generalize all header get code to `header.get(key, "")` when an empty default value is better than failing because the key was not found.

Does it make sense and can we revisit your comments with these ideas in mind?

tests/unit/data_classes/test_api_gateway_authorizer_event.py
```diff
@@ -52,16 +52,16 @@ def test_api_gateway_authorizer_v2():
...
+    assert parsed_event.headers["authorization"] == "value"
+    assert parsed_event.headers.get("missing") is None
```
ericbn, line 57:
```suggestion
    assert "missing" not in parsed_event.headers
```

examples/event_handler_rest/src/middleware_global_middlewares_module.py
```diff
@@ -34,7 +34,7 @@ def inject_correlation_id(app: APIGatewayRestResolver, next_middleware: NextMidd
 
 def enforce_correlation_id(app: APIGatewayRestResolver, next_middleware: NextMiddleware) -> Response:
     # If missing mandatory header raise an error
-    if not app.current_event.get_header_value("x-correlation-id", case_sensitive=False):
+    if not app.current_event.headers.get("x-correlation-id"):
```
ericbn, line 37:
```suggestion
    if "x-correlation-id" not in app.current_event.headers:
```

ericbn (commented):
A couple of places where we could do `"key" not in headers` instead, now that this is possible with the new API.

leandrodamascena:
> Does it make sense and can we revisit your comments with these ideas in mind?

You have great strengths here, @ericbn! Thank you for sharing so much knowledge and experience, I agree with the experience and intention to fail first. My argument was because we could prevent the client from failing in the Lambda environment, as it might be an event-driven architecture, for example, and failing because of a key is the kind of thing that could perhaps be avoided. On the other hand, if the code does not fail, it may create some side effects in the business logic.

Anyway, if the customers wants to fail silently or avoid exception, there is still the option to use get and solve the problem.

I'm going to resolve all comments.

leandrodamascena (approved):
Hi @ericbn! Thank you for your patience after the long back and forth until we got here. This PR is approved and we plan to create an alpha release of V3 this week!
Once again, thank you very much for your work here!

PR #4613, opened 2024-06-24
docs(idempotency): improve navigation, wording, and new section on guarantees

### LINKED ISSUE #4197: Docs: Explain how to find IdempotencyValidationError

(from Discord)

When searching on how to validate an idempotency request, there's a note:

> If we try to send the same request but with a different amount, we will raise IdempotencyValidationError.

However, it's not explained where the `IdempotencyValidationError` comes from.

https://docs.powertools.aws.dev/lambda/python/latest/utilities/idempotency/#payload-validation

### How can we improve?

The sentence could be expanded to show it comes from 

```python
from aws_lambda_powertools.utilities.idempotency.exceptions import IdempotencyValidationError
```

Or a simple example could be created

### Got a suggestion in mind?

_No response_

### Acknowledgment

heitorlessa:
**Issue number:** #4197

## Summary

This PR refreshes Idempotency docs with newer features like cards, improves wording and navigation from ServerlessDays ANZ feedback, and add a new section to call out guarantees/caveats upfront.

### Changes

* [ ] Create `Guarantee` or `Caveats` section
* [x] Cover `IdempotencyValidationError` in code snippet
* [x] Fix links no longer exist/renamed
* [x] Move expiring records to getting started (rename it too)
* [x] Move composite key under DDB persistence layer
* [x] Common use cases / Compatibility seems to be duplicated
* [x] Fix examples not using table env var _(most are hardcoded)_
* [x] cleanup intro and key features
* [x] cleanup getting started ddb vs redis
* [x] break iam permissions into table; IAM permission to clipboard
* [x] cleanup dynamodb required resource; break subsections and update nav
* [x] cleanup redis usage and link with setup/infra
* [x] make terminologies crispier
* [x] line editing before decorators
* [x] cleanup idempotent decorator; inline admonitions
* [x] cleanup idempotent_decorator section
* [x] cleanup serialization, fields subset, move batch to new common use case section
* [x] cleanup timeout section
* [x] use cards for required resources to showcase DDB and Redis better
* [x] note to skip timeout section when using handler decorator
* [x] remove tabbed content for single timeout snippet
* [x] cleanup handling exceptions
* [x] cleanup persistence layers attrs, snippet titles etc
* [x] typo in batch integration
* [x] rename batch integration to actual use case name
* [x] cleanup default behavior section
* [x] add ability to dismiss announcement
* [x] move bold to draw attention to whole event as idempotency key
* [x] move caching to getting started
* [x] fix redis example parameters typing
* [x] fix batch example using newer mechanism (simpler)
* [x] fix local cache example to use `powertools_json` as it was incorrect

**Unrelated**

* [x] navigation: add social links
* [x] navigation: allow returning customers to dismiss announcement banner
* [x] navigation: increase table of contents depth to 5 to help redis and other subsections

**Tech Debt**

* [ ] New method to more easily calculate remaining time _(-> datetime)_ for customers using response hooks _(e.g., Event Handler middleware)_

### User experience

image

image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


bla.py
```diff
@@ -0,0 +1,78 @@
+# KMS_KEYS = ["arn:aws:kms:eu-west-1:231436140809:key/1053c7d7-97d7-49ca-9763-5214f28999cc"]
```
leandrodamascena: `bla.py` again 🤣 🤣 🤣 🤣 🤣

heitorlessa:
mypy is failing after correcting snippets; will look at that last as there's one major piece of doc to create yet

examples/idempotency/src/using_redis_client_with_local_certs.py
```diff
@@ -9,7 +11,7 @@
     RedisCachePersistenceLayer,
 )
 
-redis_values: Any = parameters.get_secret("redis_info", transform="json")  # (1)!
+redis_values: dict[str, Any] = parameters.get_secret("redis_info", transform="json")  # (1)!
 
 
 redis_client = Redis(
```
leandrodamascena, line 17: host and port must have default values other than `None`.

```
host=redis_values.get("REDIS_HOST", "localhost"),
port=redis_values.get("REDIS_PORT", 6379),
```

https://github.com/redis/redis-py/blob/master/redis/client.py#L176

examples/idempotency/src/using_redis_client_with_aws_secrets.py
```diff
@@ -8,7 +10,7 @@
     RedisCachePersistenceLayer,
 )
 
-redis_values: Any = parameters.get_secret("redis_info", transform="json")  # (1)!
+redis_values: dict[str, Any] = parameters.get_secret("redis_info", transform="json")  # (1)!
 
 redis_client = Redis(
```
leandrodamascena, line 15: host and port must have default values other than `None`.

```
host=redis_values.get("REDIS_HOST", "localhost"),
port=redis_values.get("REDIS_PORT", 6379),
```

https://github.com/redis/redis-py/blob/master/redis/client.py#L176

leandrodamascena:
Hey @heitorlessa! 90% of the errors are related to the table_name variable. 

> Argument "table_name" to "DynamoDBPersistenceLayer" has incompatible type "str | None"; expected "str"

This is because the `DynamoDBPersistenceLayer` class doesnt accept `None` for [table_name parameter](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py#L35) and the default return for `os.getenv` is `None` when it don't find the env variable.

I think we can do this:
`table = os.getenv("IDEMPOTENCY_TABLE", "")`

WDYT?

docs/utilities/idempotency.md
```diff
@@ -5,26 +5,26 @@ description: Utility
 
 <!-- markdownlint-disable MD051 -->
 
-The idempotency utility provides a simple solution to convert your Lambda functions into idempotent operations which are safe to retry.
+The idempotency utility allows you to retry operations with the same input within a time window, producing the same output.
```
leandrodamascena: At this point, the customer may not be familiar with the concept of idempotency and the idea of "producing the same output" could lead to misunderstandings. It's important to clarify that the same output refers to the first sucessful invocation. 

To avoid confusion, why not use the same explanation as before and just add the time window sentence?

docs/utilities/idempotency.md
```diff
@@ -5,26 +5,26 @@ description: Utility
...
 
 ## Key features
 
-* Prevent Lambda handler from executing more than once on the same event payload during a time window
-* Ensure Lambda handler returns the same result when called with the same payload
-* Select a subset of the event as the idempotency key using JMESPath expressions
-* Set a time window in which records with the same payload should be considered duplicates
-* Expires in-progress executions if the Lambda function times out halfway through
-* Support Amazon DynamoDB and Redis as persistence layers
+* Produces the same result when a function is called repeatedly with the same idempotency key
```
leandrodamascena: Same as before here.. Same result for a failed execution? And if I update the code to fix the bug and send the same payload?

docs/utilities/idempotency.md
```diff
@@ -5,26 +5,26 @@ description: Utility
...
+* Choose your idempotency key from one or more fields, or entire payload
+* Safeguard concurrent requests, timeouts, missing idempotency keys, and payload tampering
+* Support for Amazon DynamoDB, Redis, and bring your own persistence layer
```
leandrodamascena: You explain in the "Terminology" section that "in-memory caching" is supported. Does it make sense to add it here?

docs/utilities/idempotency.md
```diff
@@ -5,26 +5,26 @@ description: Utility
...
 
-**Idempotent operations will return the same result when they are called multiple times with the same parameters**. This makes idempotent operations safe to retry.
+**Idempotency key** is a combination of **(a)** Lambda function name, **(b)** fully qualified name of your function, and **(c)** a hash of the entire payload or part(s) of the payload you specify.
 
-**Idempotency key** is a hash representation of either the entire event or a specific configured subset of the event, and invocation results are **JSON serialized** and stored in your persistence storage layer.
+**Idempotent request** is an operation with the same input previously processed that is not expired in your persistent storage or in-memory cache.
 
-**Idempotency record** is the data representation of an idempotent request saved in your preferred  storage layer. We use it to coordinate whether a request is idempotent, whether it's still valid or expired based on timestamps, etc.
+**Persistence layer** is a storage we use to read, create, expire, and delete idempotency records.
```
leandrodamascena:
```suggestion
**Persistence layer** is a storage we use to create, read, expire, and delete idempotency records.
```
CRUD like? 😄

docs/utilities/idempotency.md
```diff
@@ -5,26 +5,26 @@ description: Utility
...
+
+**Idempotency record** is the data representation of an idempotent request in its various status. We use it to coordinate  whether **(a)** a request is idempotent, **(b)** it's not expired, **(c)** JSON response to return, and more.
```
leandrodamascena:
```suggestion
**Idempotency record** is the data representation of an idempotent request saved in the persistent layer and in its various status. We use it to coordinate  whether **(a)** a request is idempotent, **(b)** it's not expired, **(c)** JSON response to return, and more.
```

docs/utilities/idempotency.md
```diff
@@ -52,33 +52,60 @@ classDiagram
...
 
-| Configuration      | Value        | Notes                                                                               |
-| ------------------ | ------------ |-------------------------------------------------------------------------------------|
-| Partition key      | `id`         |                                                                                     |
-| TTL attribute name | `expiration` | This can only be configured after your table is created if you're using AWS Console |
+#### DynamoDB table
 
-???+ tip "Tip: You can share a single state table for all functions"
-    You can reuse the same DynamoDB table to store idempotency state. We add `module_name` and [qualified name for classes and functions](https://peps.python.org/pep-3155/){target="_blank" rel="nofollow"} in addition to the idempotency key as a hash key.
+!!! tip "You can share a single state table for all functions"
```
leandrodamascena:
```suggestion
!!! tip "You can share a single state table for all Lambda functions"
```

docs/utilities/idempotency.md
```diff
@@ -52,33 +52,60 @@ classDiagram
...
+
+Unless you're looking to use an [existing table or customize each attribute](#dynamodbpersistencelayer), you only need the following:
+
+| Configuration      | Value        | Notes                                                                                    |
+| ------------------ | ------------ | ---------------------------------------------------------------------------------------- |
+| Partition key      | `id`         | Format: <br> `{lambda_fn_name}.{module_name}.{fn_qualified_name}#{idempotency_key_hash}` |
```
leandrodamascena: Imagine I'm kind of new to the concept of DynamoDB, where do I configure this format in DynamoDB?
It's better to add something like: it will be saved in the format...

If you do not want to accept this suggestion, please correct the format. The correct one is: `{module_name}.{fn_qualified_name}#{idempotency_key_hash}`. Reference: https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/idempotency/base.py#L94

docs/utilities/idempotency.md
```diff
@@ -52,33 +52,60 @@ classDiagram
...
+| TTL attribute name | `expiration` | Using AWS Console? this is configurable after table creation                             |
```
leandrodamascena:
```suggestion
| TTL attribute name | `expiration` | Using AWS Console? This is configurable after table creation                             |
```

docs/utilities/idempotency.md
```diff
@@ -97,66 +124,80 @@ If you're not [changing the default configuration for the DynamoDB persistence l
...
+* **Expect 2 WCU per non-idempotent call**. During the first invocation, we use `PutItem` for locking and `UpdateItem` for completion. Consider reviewing [DynamoDB pricing documentation](https://aws.amazon.com/dynamodb/pricing/){target="_blank"} to estimate cost.
 
-<!-- markdownlint-disable MD013 -->
-???+ info "Info: DynamoDB"
+* **Old boto3 versions can increase costs**. For cost optimization, we use a conditional `PutItem` to always lock a new idempotency record. If locking fails, it means we already have an idempotency record saving us an additional `GetItem` call. However, this is only supported in boto3 `1.26.194` and higher _([June 30th 2023](https://aws.amazon.com/about-aws/whats-new/2023/06/amazon-dynamodb-cost-failed-conditional-writes/){target="_blank"})_.
 
-    During the first invocation with a payload, the Lambda function executes both a `PutItem` and an `UpdateItem` operations to store the data in DynamoDB. If the result returned by your Lambda is less than 1kb, you can expect 2 WCUs per Lambda invocation.
+#### Redis cluster
```
leandrodamascena: I don't think we should call this section `Redis cluster`. Redis connection/service and can be a standalone instance or a cluster. Why not `Redis Service`?

If you accept this suggestion, please change the link reference in the preview changes.

docs/utilities/idempotency.md
```diff
@@ -97,66 +124,80 @@ If you're not [changing the default configuration for the DynamoDB persistence l
...
+
+We recommend you start with a Redis compatible management services such as [Amazon ElastiCache for Redis](https://aws.amazon.com/elasticache/redis/){target="_blank"} or [Amazon MemoryDB for Redis](https://aws.amazon.com/memorydb/){target="_blank"}.
+
+In both services and self-hosting Redis, you'll need to configure [VPC access](https://docs.aws.amazon.com/lambda/latest/dg/configuration-vpc.html){target="_blank"} to your AWS Lambda.
```
leandrodamascena: We need to review this. If I have a self-hosted solution or any other Redis SaaS solution, I can connect them using the public IP provided by Lambda, for example.

docs/utilities/idempotency.md
```diff
@@ -97,66 +124,80 @@ If you're not [changing the default configuration for the DynamoDB persistence l
...
+For full flexibility, you can use the `idempotent_function` decorator for any synchronous Python function.
 
-When using `idempotent_function`, you must tell us which keyword parameter in your function signature has the data we should use via **`data_keyword_argument`**.
+When using this decorator, you **must** call your decorated function using keyword arguments.
 
-!!! tip "We support JSON serializable data, [Python Dataclasses](https://docs.python.org/3.12/library/dataclasses.html){target="_blank" rel="nofollow"}, [Parser/Pydantic Models](parser.md){target="_blank"}, and our [Event Source Data Classes](./data_classes.md){target="_blank"}."
-
-???+ warning "Limitation"
-    Make sure to call your decorated function using keyword arguments.
+You can use `data_keyword_argument` to tell us the argument to extract an idempotency key.  We support JSON serializable data, [Dataclasses](https://docs.python.org/3.12/library/dataclasses.html){target="_blank" rel="nofollow"}, Pydantic Models, and [Event Source Data Classes](./data_classes.md){target="_blank"}
```
leandrodamascena, line 191:
```suggestion
You can use `data_keyword_argument` to tell us the argument to extract an idempotency key.  We support JSON serializable data, [Dataclasses](https://docs.python.org/3.12/library/dataclasses.html){target="_blank" rel="nofollow"}, Pydantic Models, and [Event Source Data Classes](./data_classes.md){target="_blank"}.
```

docs/utilities/idempotency.md
```diff
@@ -166,15 +207,15 @@ By default, `idempotent_function` serializes, stores, and returns your annotated
 
 The output serializer supports any JSON serializable data, **Python Dataclasses** and **Pydantic Models**.
 
-!!! info "When using the `output_serializer` parameter, the data will continue to be stored in DynamoDB as a JSON object."
+!!! info "When using the `output_serializer` parameter, the data will continue to be stored in DynamoDB as a JSON string."
```
leandrodamascena:
```suggestion
!!! info "When using the `output_serializer` parameter, the data will continue to be stored in your persistent storage as a JSON string."
```

docs/utilities/idempotency.md
leandrodamascena: We support to serialize other type of data using CustomDictSerializer, no?

docs/utilities/idempotency.md
```diff
@@ -204,61 +245,61 @@ The output serializer supports any JSON serializable data, **Python Dataclasses*
...
 
-#### Batch integration
+### Using in-memory cache
 
-You can can easily integrate with [Batch utility](batch.md){target="_blank"} via context manager. This ensures that you process each record in an idempotent manner, and guard against a [Lambda timeout](#lambda-timeouts) idempotent situation.
+!!! note "In-memory cache is local to each Lambda execution environment."
 
-???+ "Choosing an unique batch record attribute"
-    In this example, we choose `messageId` as our idempotency key since we know it'll be unique.
+You can enable caching with the `use_local_cache` parameter in `IdempotencyConfig`. When enabled, you can adjust cache capacity _(256)_ with `local_cache_max_items`.
```
leandrodamascena, line 276: 1024?

image

docs/utilities/idempotency.md
```diff
@@ -278,82 +319,106 @@ If we were to treat the entire request as our idempotency key, a simple HTTP hea
     --8<-- "examples/idempotency/src/working_with_payload_subset_payload.json"
     ```
 
-### Lambda timeouts
+### Adjusting expiration window
 
-???+ note
-    This is automatically done when you decorate your Lambda handler with [@idempotent decorator](#idempotent-decorator).
+!!! note "We expire idempotency records after **an hour** (3600 seconds). After that, a transaction with the same payload [will not be considered idempotent](#expired-idempotency-records)."
```
leandrodamascena:
```suggestion
!!! note "By default, we expire idempotency records after **an hour** (3600 seconds). After that, a transaction with the same payload [will not be considered idempotent](#expired-idempotency-records)."
```

docs/utilities/idempotency.md
```diff
@@ -369,22 +434,102 @@ When using DynamoDB as the persistence layer, you can customize the attribute na
 
 #### RedisPersistenceLayer
 
-This persistence layer is built-in, allowing you to use an existing Redis service. For optimal performance and compatibility, it is strongly recommended to use a Redis service version 7 or higher.
+!!! info "We recommend Redis version 7 or higher for optimal performance."
+
+For simple setups, initialize `RedisCachePersistenceLayer` with your cluster endpoint and port to connect.
```
leandrodamascena:
```suggestion
For simple setups, initialize `RedisCachePersistenceLayer` with your Redis endpoint and port to connect.
```

leandrodamascena (changes requested):
Hey @heitorlessa! This is the kind of thoughtful and remarkable work I aspire to learn and can do this in the future. The changes here make the concept of Idempotency much clearer for the customer, explaining what it is, how it works, its limitations, its potential, what to do, what not to do and other information hard to find in any other documentation.

I would like to see here 2 additional things:

1. Tell customers to be cautious when dealing with payloads containing dynamic HTTP requests or timestamps and using then without extracting specific keys. It's a common error that customers often overlook, and it's crucial to handle such payloads appropriately to ensure consistency and accuracy.

2. It's important to note that we do not store the payload in persistent storage and there is no way to the customer access the payload in the persistent layer.

Thanks a lot man! 🚀

heitorlessa:
adding a `do-not-merge` as I can't work on this in the next two weeks. Thanks a lot for those suggestions @leandrodamascena - I didn't touch those before as they were unrelated. 

I'll do a full sweep instead word for word after this review then

heitorlessa:
Agreed with Leandro that he'll take over

leandrodamascena:
Hello @heitorlessa! Can you rebase this PR? I don't know why, but it's showing that approximately 3k commits will be merged.

image

heitorlessa:
crazy, I'll try it today.

heitorlessa:
Fixed the history. Let's jump on a call to fast track this review

docs/utilities/idempotency.md
heitorlessa: good idea. I was trying to not turn into a kitchen sink. Adding this here instead of creating another line 

image

docs/utilities/idempotency.md
heitorlessa: by `functions` here, we're referring to both Lambda handler and any Python synchronous function - unless you think it's wrong and we should rethink the messaging altogether

docs/utilities/idempotency.md
heitorlessa: I think this fits betters, but I'll try a few more this week.

```
The idempotency utility allows you to retry operations within a time window with the same input, producing the same output.
```

docs/utilities/idempotency.md
heitorlessa: fixed it

docs/utilities/idempotency.md
heitorlessa: will review later, I don't know how to solve it.

docs/utilities/idempotency.md
heitorlessa: Good idea! Changed to `Redis database` to follow `DynamoDB table`

docs/utilities/idempotency.md
heitorlessa: Good catch, removed `self-hosting Redis` to remove ambiguity.

docs/utilities/idempotency.md
heitorlessa: not here, because it's an input (`fn(order: Order)`). Which is a **GREAT** catch because we're missing that functionality today -- similar to Event Handler we should allow customers to bring a custom serializer for input in case they want to bring in a binary of sorts and convert it to JSON before storing in the persistent storage.

`CustomDictSerializer` is for the **output**, which we document under `Any type`

image

docs/utilities/idempotency.md
heitorlessa: `LRUDict` implementation is 1024, but we reduce in Idempotency to default to 256 to reduce memory footprint.

https://github.com/aws-powertools/powertools-lambda-python/blob/c77869a89c1ebb87718b78b2027a2f0978512142/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L84

docs/utilities/idempotency.md
heitorlessa: crux: we use this for any persistent storage so we need to call it out, but not here.

note to self: test whether we include the Lambda function name

heitorlessa (commented):
todo: find out a better place to call out the primary key we use for persistence storage, and confirm Lambda function name placement/usage.

docs/utilities/idempotency.md
heitorlessa: [found the fn name](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/utilities/idempotency/persistence/base.py#L116) (more reason why we need to refactor to split responsibilities.. that's supposed to be Idempotency Handler not persistence storage responsibility)

image

leandrodamascena (approved):
Amazing work @heitorlessa!!!! APPROVED!!

PR #4692, opened 2024-07-04
refactor(typing): enable boto3 implicit type annotations

### LINKED ISSUE #4607: Tech debt: Inconsistent use of from __future__ import annotations
### Why is this needed?

Source files that have `from __future__ import annotations` are not taking full advantage of the import. Namely, this import enables [PEP 563](https://peps.python.org/pep-0563/) (Postponed Evaluation of Annotations), which allows:

1. usage of `if typing.TYPE_CHECKING:` for types to be used only by the type checker and no need to quote type names (unless being used in forward references)
1. use of [PEP 585](https://peps.python.org/pep-0585/) (Type Hinting Generics In Standard Collections) in Python < 3.9
2. use of [PEP 604](https://peps.python.org/pep-0604/) (Allow writing union types as X | Y) in Python < 3.10

Looking at the files with this import we see that not all of them have `if typing.TYPE_CHECKING:`, which could have been the initial motivation to add the import, and they have mixed usages of PEP 585 and PEP 604.

Some other files do have `if typing.TYPE_CHECKING:` without the import and are quoting the type names instead.

Also, I see some `if typing.TYPE_CHECKING:` is being used for the types defined in the  mypy_boto3_* packages, which are dev dependencies and in these cases this makes sense.

Static typing

### Suggestion

We already have a constraint:

1. The import of types from the mypy_boto3_* packages must be inside `if typing.TYPE_CHECKING:`.

And here are suggestions (see https://stackoverflow.com/a/53455562/2654518):

1. Use `from __future__ import annotations` instead of quoting types imported inside `if typing.TYPE_CHECKING:`. 
2. Using both `from __future__ import annotations` and `if typing.TYPE_CHECKING:` can improve the startup performance since less packages are imported during runtime. Although optimizing performance really only makes sense if we have benchmarks first.

Also:

3. The `from __future__ import annotations` enables the use of [PEP 585](https://peps.python.org/pep-0585/) and [PEP 604](https://peps.python.org/pep-0604/) in Python < 3.10.

Looks like the current code does not have a clear direction on how to follow these suggestions. I'd say it's more consistent and cleaner to have a guideline. So, here are some decisions that could be made:

1. Do we use `from __future__ import annotations` instead of quoting types that have been imported inside `if typing.TYPE_CHECKING:`? I'd say YES since we're already using the import in some files.
2. Do we use `from __future__ import annotations` and `if typing.TYPE_CHECKING:` in all files?
   1. Even if we choose YES, should we at least skip the example files?
3. Do we use [PEP 585](https://peps.python.org/pep-0585/) and [PEP 604](https://peps.python.org/pep-0604/) in files using `from __future__ import annotations`? I'd say definitely YES if we choose YES for question 2.

### Acknowledgment

ericbn:
**Issue number:** #4607 

### Changes

Add boto3-stubs and botocore-stubs to enable implicit type annotations with `mypy_boto3_*` packages. See https://youtype.github.io/boto3_stubs_docs/mypy_boto3_s3/usage/#implicit-type-annotations

This enables removing very few unneeded explicit type annotations and helps detect some new errors automatically.

No need to do

    boto3_config = boto3_config or Config()

before passing the boto3_config to the boto3_session client or resource methods, as None is an accepted value.

Refactor the aws_lambda_powertools.utilities.parameters.base.BaseProvider class: move the instantiation of clients in _build_boto3_client and _build_boto3_resource_client to the subclasses and the user_agent method calls to the constructor. Keeping correct typing information with the original implementation was going to be too complex and hopefully the refactor even made the code tidier.

Keep imports consistent: use boto3.session.Session, `mypy_boto3_*.client`, `mypy_boto3_*.service_resource` and `mypy_boto3_*.type_defs` as documented at https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html and https://youtype.github.io/boto3_stubs_docs/mypy_boto3_s3/usage/#explicit-type-annotations

### User experience

The initial motivation of this PR is what we agreed as step 1 of #4607:

> 1. Review the use of mypy_boto3_* types, and try to keep their imports to a minimum, to only when those types cannot be discovered automatically by mypy.

In the process of reviewing the respective code, I've noticed that implicit type annotations for `mypy_boto3_*` packages was not enabled, due to missing dev dependencies as mentioned above. Adding this allows both reducing the need to explicitly annotate these types -- which we were not doing much anyway -- and letting types be discovered automatically -- which we were missing, given some errors were discovered. All new errors reported by mypy were fixed, unless for one. I'm not sure if this is intentional in the code:

```
aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py: note: In member "_put_record" of class "DynamoDBPersistenceLayer":
aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py:251:74: error: TypedDict "_ClientErrorResponseTypeDef" has no key "Item"  [typeddict-item]
                    old_data_record = self._item_to_data_record(exc.response["Item"]) if "Item" in exc.response else None
                                                                             ^~~~~~
Found 1 error in 1 file (checked 616 source files)
make: *** [mypy] Error 1
```

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena:
Hello @ericbn! I'm planning to review this PR right after we finish #4606!

leandrodamascena:
I will start the review tomorrow morning. 🚀

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -91,11 +91,8 @@ def __init__(
             >>>     return {"StatusCode": 200}
         """
         if boto3_client is None:
-            self._boto_config = boto_config or Config()
-            self._boto3_session: boto3.Session = boto3_session or boto3.session.Session()
-            self.client: "DynamoDBClient" = self._boto3_session.client("dynamodb", config=self._boto_config)
-        else:
-            self.client = boto3_client
+            boto3_client = (boto3_session or boto3.session.Session()).client("dynamodb", config=boto_config)
```
leandrodamascena: Good catch! `Config()` or `None` have the same result for a boto3 session! The defaults are the same and we avoid creating another object in memory! Excellent!

leandrodamascena (commented):
Hi @ericbn! I've started looking into this PR. Everything looks great, but I see that you refactored an important part of the code where we register the `UserAgent` in the session. This code is critical and I'll need a few days to do some exploratory testing to see if we won't have any side effects.

As always, you have great attention to detail and your code is elegant and clean, but I need to be 100% sure that refactoring this part of the code won't have any issues that our tests haven't caught.

leandrodamascena:
mypy is complaining about something, but it seems easy to fix. Can you take a look, please @ericbn?

ericbn:
Cool! @leandrodamascena, I've updated the code to ignore that mypy error, since this was a case for a DynamoDB request with `ReturnValuesOnConditionCheckFailure`, for which the response type was not covered by `mypy_boto3_dynamodb`.

examples/parameters/src/custom_boto3_all_providers.py
```diff
@@ -7,7 +7,7 @@
 
 # construct boto clients with any custom configuration
 ssm = boto3.client("ssm", config=config)
-secrets = boto3.client("secrets", config=config)
+secrets = boto3.client("secretsmanager", config=config)
```
leandrodamascena, line 10: Impressive attention to detail! 😱

examples/tracer/src/disable_capture_response_streaming_body.py
```diff
@@ -12,15 +12,15 @@
...
+session = boto3.session.Session()
 s3 = session.client("s3")
 
 
 @tracer.capture_method(capture_response=False)
 def fetch_payment_report(payment_id: str) -> StreamingBody:
     ret = s3.get_object(Bucket=BUCKET, Key=f"{REPORT_KEY}/{payment_id}")
     logger.debug("Returning streaming body from S3 object....")
-    return ret["body"]
+    return ret["Body"]
```
leandrodamascena, line 23: Nice catch!

examples/idempotency/src/bring_your_own_persistent_store.py
```diff
@@ -27,9 +27,7 @@ def __init__(
         boto_config: Optional[Config] = None,
         boto3_session: Optional[boto3.session.Session] = None,
     ):
-        boto_config = boto_config or Config()
-        session = boto3_session or boto3.session.Session()
-        self._ddb_resource = session.resource("dynamodb", config=boto_config)
+        self._ddb_resource = (boto3_session or boto3.session.Session()).resource("dynamodb", config=boto_config)
```
leandrodamascena: I think that splitting this into two lines makes the code more "understandable" for someone reading it for the first time. What do you think?

```suggestion
        self._boto3_session = boto3_session or boto3.session.Session()
        self._ddb_resource = resource("dynamodb", config=self._boto3_session)
```

examples/feature_flags/src/custom_s3_store_provider.py
```diff
@@ -22,10 +22,8 @@ def __init__(self, bucket_name: str, object_key: str):
 
     def _get_s3_object(self) -> Dict[str, Any]:
         # Retrieve the object content
-        parameters = {"Bucket": self.bucket_name, "Key": self.object_key}
-
         try:
-            response = self.client.get_object(**parameters)
+            response = self.client.get_object(Bucket=self.bucket_name, Key=self.object_key)
```
leandrodamascena, line 26: Much more explicit! 🚀

aws_lambda_powertools/utilities/parameters/ssm.py
```diff
@@ -25,11 +25,11 @@
...
 )
 from aws_lambda_powertools.utilities.parameters.exceptions import GetParameterError, SetParameterError
-from aws_lambda_powertools.utilities.parameters.types import PutParameterResponse, TransformOptions
+from aws_lambda_powertools.utilities.parameters.types import TransformOptions
 
 if TYPE_CHECKING:
-    from mypy_boto3_ssm import SSMClient
-    from mypy_boto3_ssm.type_defs import GetParametersResultTypeDef
+    from mypy_boto3_ssm.client import SSMClient
+    from mypy_boto3_ssm.type_defs import GetParametersResultTypeDef, PutParameterResultTypeDef
```
leandrodamascena, line 32: The only thing here is that we are now forcing customers to have these stubs in their local environment in order to have type hints in the IDE. But I think that's ok, since that's the idea of ​​a development environment.

aws_lambda_powertools/utilities/parameters/base.py
```diff
@@ -262,78 +253,6 @@ def _build_cache_key(
         """
         return (name, transform, is_nested)
 
-    @staticmethod
```
leandrodamascena, line 265: I love it when we can remove about 70 lines of code without breaking anything!! 🚀

leandrodamascena (changes requested):
Hello @ericbn! The only thing missing is to review the Idempotency change in the `put_record` method, but the other changes are fantastic! Thank you for improving the quality of the project in development environments, making the project more compatible with python styles and removing unnecessary code.

examples/tracer/src/disable_capture_response_streaming_body.py
ericbn: This and the above where observed actually thanks to the introduced mypy_boto3 implicit type annotations.

aws_lambda_powertools/utilities/parameters/ssm.py
ericbn: Yeah. I saw this is already done for GetParametersResultTypeDef, so I thought it would be fine to introduce another one. Cool!

examples/idempotency/src/bring_your_own_persistent_store.py
ericbn: Good point. Done! I'm not assigning them to `self` as instance attributes because we don't need the references to either boto3_session or boto3_client later.

As a side note, there might other unneeded assignments like these in the code, which communicate the wrong message of "we need this reference later" when we don't. In the worst case, this potentially contributes to memory leaks. Also, the use of "non-private" names like `self.client` instead of `self._client` in other parts of the code communicate that these can be publicly used, when in some cases I'm not sure they should be. But this is all a separate discussion with too many breaking changes involved, so probably too late now to go over all this...

examples/feature_flags/src/custom_s3_store_provider.py
ericbn: It works best this way with the introduced mypy_boto3 implicit type annotations. Otherwise mypy would not be able to check the types of the parameters. I'm happy you agree it looks better too!

aws_lambda_powertools/utilities/parameters/base.py
ericbn: <a href="https://twitter.com/codeinthehole/status/540117725604216832"></a>

leandrodamascena: I agree with the 3rd point! If we delete all the code we won't have bugs, so great deal! HAHAHA

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
```diff
@@ -246,13 +244,20 @@ def _put_record(self, data_record: DataRecord) -> None:
...
-                if old_data_record is not None:
+                try:
+                    item = exc.response["Item"]  # type: ignore[typeddict-item]
+                except KeyError:
+                    logger.debug(
+                        f"Failed to put record for already existing idempotency key: {data_record.idempotency_key}",
+                    )
+                    raise IdempotencyItemAlreadyExistsError() from exc
+                else:
+                    old_data_record = self._item_to_data_record(item)
```
ericbn, line 260: What I've done here was replace the "Look Before You Leap" style by the ["Easier to Ask Forgiveness Than Permission"](https://stackoverflow.com/a/12270384/2654518), which is arguably more pythonic.

Doing:
```python
r = f(d["key"]) if "key" in d else None
if r is not None:
    # key exists
else:
    # key does not exist
```
is the same as (given the function `f` never returns None, which is the case for `self._item_to_data_record`):
```python
try:
    v = d["key"]
except KeyError:
    # key does not exist
else:
    r = f(v)
    # key exists
```
The latter also does one less dict lookup than the former. Also former code has two ifs and latter has none, which I hope you can appreciate is less complex.  :- )

aws_lambda_powertools/utilities/parameters/base.py
ericbn: I love all 4 points there! And here's a great addition to point 2:

<a href="https://twitter.com/jessitron/status/1722302472272650466"></a>

:- )

sthulb (commented):
Overall, I think this PR is a win.

aws_lambda_powertools/utilities/idempotency/persistence/dynamodb.py
leandrodamascena: Thanks again for sharing the idea/concepts behind your code, it's absolutely amazing to educate/teach others in this area.
I was just running your e2e test to make sure everything is working as expected!!

leandrodamascena (approved):
Another amazing PR @ericbn! I'm more than happy to say APPROVED!!!

Issue #4915, opened 2024-08-08
Bug: Swagger UI doesn't work correctly when using custom domain name with mapping key


tlinhart:
### Expected Behaviour

I have an API Gateway HTTP API with route `/foo`. I create a custom domain name for the API `my.domain.com` and create a mapping with a mapping key `api`. Thus I access my foo operation with `https://my.domain.com/api/foo`. In my Lambda function, I create a resolver app and enable swagger on `/swagger` path:

```
app = APIGatewayHttpResolver(enable_validation=True)
app.enable_swagger(
    path="/swagger",
    ...
)
```

I define my operation foo with:

```
@app.get("/foo", ...)
def foo():
    ...
```

I can then access the Swagger UI on `https://my.domain.com/api/swagger`. I also expect:

- I would be able to download the OpenAPI schema via the link under the Swagger UI title, which should be in a form like `https://my.domain.com/api/swagger?format=json`.
- The operation shown would be like `GET /api/foo`.
- When I use _Try it out_ feature, it calls the foo operation through `https://my.domain.com/api/foo`.

### Current Behaviour

At the moment, it works like this:

- The OpenAPI schema link has a form like `https://my.domain.com/swagger?format=json`, i.e. without the mapping key `api`.
- The operation shown is like `GET /foo`, i.e. it lacks the mapping key `api`.
- When I use _Try it out_ feature, it calls the foo operation through `https://my.domain.com/foo`, i.e. without the mapping key `api`.

### Code snippet

```python
The snippet is shown above.
```

### Possible Solution

The last problem with _Try it out_ can be worked around by defining a server with `url="/api"`. The rest of the issues stays.

### Steps to Reproduce

It's described above.

### Powertools for AWS Lambda (Python) version

latest

### AWS Lambda function runtime

3.11

### Packaging format used

PyPi

tlinhart:
The first issue with wrong OpenAPI schema link can be solved [here](https://github.com/aws-powertools/powertools-lambda-python/blob/d86cce1a62210a299f6fd3994e3ad0a2b6a3af7c/aws_lambda_powertools/event_handler/openapi/swagger_ui/html.py#L97) replacing:

```
ui.specActions.updateUrl('{path}?format=json');
```

with:

```
ui.specActions.updateUrl(baseUrl + "?format=json");
```
(to get absolute URL)

or:

```
ui.specActions.updateUrl(currentUrl.pathname + "?format=json");
```
(to get relative URL)

[...]

PR #4930, opened 2024-08-12
fix(event_handler): correct URL for OpenAPI spec in Swagger UI

tlinhart:
**Issue number:** #4915

### Changes

This PR corrects the URL via which the OpenAPI spec in JSON can be downloaded from Swagger UI.

### User experience

Before the fix, the URL was always `/swagger?format=json` regardless of potential custom domain name mapping.

After the change, if I created a custom domain name `custom.example.com` and mapped my API at the `/api` base path, thus accessing the Swagger UI via `https://custom.example.com/api/swagger`, I would get a correct URL for the JSON spec download `/api/swagger?format=json`.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


sthulb:
@tlinhart Thanks for the pull request! 

I have one question: You deleted two test cases, I'm curious about why? Did it break your new implementation somehow?

tlinhart:
> I have one question: You deleted two test cases, I'm curious about why? Did it break your new implementation somehow?

I tried to explain it in the [linked issue](https://github.com/aws-powertools/powertools-lambda-python/issues/4915#issuecomment-2284006939). Yeah, the fix has broken the tests as they relied on the `path` being passed into the `generate_swagger_html`. More importantly, they became superfluous after the fix.

sthulb (approved):
I've confirmed everything still works as expected. Thanks for the PR @tlinhart !

sthulb:
Unrelated to this PR, but `_get_base_path(self)` has an issue too, it'll return the stage variable even if using a custom domain, which might be annoying to some users.

tlinhart:
> Unrelated to this PR, but `_get_base_path(self)` has an issue too, it'll return the stage variable even if using a custom domain, which might be annoying to some users.


Yeah, check out my message on Discord about it - https://discord.com/channels/1006478942305263677/1006527338621710376/1272632575818727476. The method works correctly but covers only the case when one accesses the API via generated API endpoint URL. To cover the custom domain names it would have to be enhanced as I tried to propose.

PR #5261, opened 2024-09-27
feat(ci): GovCloud Layer Workflow

### LINKED ISSUE #1166: Publish layer to Gov Cloud regions
- Consider publishing the layer to the Gov Cloud regions. I checked the [list of regions](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/docs/index.md#lambda-layer) the layer is currently published to but it looks like the Gov Cloud regions are not present. 

- It would be great if the layer could be published and available in the Gov Cloud regions as well

- For now, I am recommending Gov Cloud users perform a copy of the layer to their account and use that route as a temporary work around

**Additional context**

sthulb:
**Issue number:** #1166 

### Changes

This PR aims to extend the availability of the AWS Lambda Powertools layer to AWS GovCloud regions.

### User experience

**Before this PR:**
- The Lambda Powertools layer is currently not published to AWS GovCloud regions.
- GovCloud users have to manually copy the layer to their accounts as a workaround.

**After this PR:**

- Include the following GovCloud regions in our layer publication process:
  - us-gov-east-1
  - us-gov-west-1

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


.github/workflows/layer_govcloud.yml
dreamorosi, line 1: Can we add a header to the file with documentation like we have in almost all other workflows?

dreamorosi (dismissed):

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+      matrix:
+        layer:
+          - AWSLambdaPowertoolsPythonV3-python38
+          - AWSLambdaPowertoolsPythonV3-python39
+          - AWSLambdaPowertoolsPythonV3-python310
+          - AWSLambdaPowertoolsPythonV3-python311
+          - AWSLambdaPowertoolsPythonV3-python312
+        arch:
+          - arm64
+          - x86
```
leandrodamascena:
```suggestion
          - x86_64
```

I understand you are still testing, but please consider changing it to the new name. We will fix the layer name before launching GovCloud.

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+    environment: Prod (Readonly)
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE }}
+          aws-region: us-east-1
+          mask-aws-account-id: true
+      - name: Grab Zip
```
leandrodamascena, line 65:
```suggestion
      - name: Download layer zip file
```

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
+on:
+  workflow_dispatch:
+    inputs:
+      environment:
+        description: Deployment environment
+        type: choice
+        options:
+          - Gamma
+          - Prod
```
leandrodamascena, line 18: This pipeline should be executed in both environments simultaneously, otherwise we may deploy in one environment and forget the other. Is it possible to add a "both" option and make it the default or do we remove it?

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+      - name: Grab Zip
+        run: |
+          aws --region us-east-1 lambda get-layer-version-by-arn --arn arn:aws:lambda:us-east-1:017000801446:layer:${{ matrix.layer }}-${{ matrix.arch }}:${{ inputs.version }} --query 'Content.Location' | xargs curl -L -o ${{ matrix.layer }}_${{ matrix.arch }}.zip
+      - name: Store Zip
+        uses: actions/upload-artifact@50769540e7f4bd5e21e526ee35c689e35e0d6874 # v4.4.0
+        with:
+          name: ${{ matrix.layer }}_${{ matrix.arch }}.zip
+          path: ${{ matrix.layer }}_${{ matrix.arch }}.zip
+          retention-days: 1
+          if-no-files-found: error
```
leandrodamascena: At this step, we need to consider validating the newly downloaded file and avoid issues like:

1 - Connection drop in the middle and zero bytes or incomplete file
2 - No other process changing the file 

So before storing the file you should check if the "CodeSha256" in `aws --region us-east-1 lambda get-layer-version-by-arn` matches the `sha256sum` of the recent download file, otherwise you should abort the pipeline.

Suggestion:

`cat ${{ matrix.layer }}_${{ matrix.arch }}.zip |sha256sum |cut -d' ' -f1 |xxd -r -p |base64`

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+          - AWSLambdaPowertoolsPythonV3-python312
+        arch:
+          - arm64
+          - x86
+    environment: GovCloud ${{ inputs.environment }} (East)
+    steps:
+      - name: Download Zip
+        uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: ${{ matrix.layer }}_${{ matrix.arch }}.zip
```
leandrodamascena, line 107: After restoring the file, please check again if the `sha256sum` of the downloaded file matches with the one returned by the `aws api` and in this way we can ensure that any action is not changing the content. We do the same when we seal the code in the pipeline to have 100% that we are safe.

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+          aws --region us-gov-east-1 lambda publish-layer-version \
+            --layer-name ${{ inputs.environment }}-${{ matrix.layer }}-${{ matrix.arch }} \
+            --zip-file fileb://./${{ matrix.layer }}_${{ matrix.arch }}.zip \
+            --query 'Version' | \
+            xargs aws --region us-gov-east-1 lambda add-layer-version-permission \
+              --layer-name ${{ inputs.environment }}-${{ matrix.layer }}-${{ matrix.arch }} \
+              --statement-id 'PublicLayer' \
+              --action lambda:GetLayerVersion \
+              --principal '*' \
+              --version-number
```
leandrodamascena: If the environment is `Gamma`, please consider to run some canary tests and ensure that the layer is running and we won't have any problems in production.

I suggest creating a simple Lambda (you can use [this example](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/layer_v3/layer/canary/app.py)), attaching the layer and running an invocation with `RequestResponse`, if there are no errors then you can move to production.

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+        arch:
+          - arm64
+          - x86
+    environment: Prod (Readonly)
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502 # v4.0.2
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE }}
+          aws-region: us-east-1
```
leandrodamascena, line 63: Please consider moving this to a ENV variable because we can change in one place and not in many.

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+        with:
+          role-to-assume: ${{ secrets.AWS_IAM_ROLE }}
+          aws-region: us-gov-east-1
+          mask-aws-account-id: true
+      - name: Create Layer
+        run: |
+          aws --region us-gov-east-1 lambda publish-layer-version \
+            --layer-name ${{ inputs.environment }}-${{ matrix.layer }}-${{ matrix.arch }} \
+            --zip-file fileb://./${{ matrix.layer }}_${{ matrix.arch }}.zip \
+            --query 'Version' | \
```
leandrodamascena: Fields such as `Description`, `CompatibleRuntimes`, `LicenseInfo` and `CompatibleArchitectures` are missing, these fields are mandatory because if we don't add them so it can create a bad experience for customers to identify what they are using.

.github/workflows/layer_govcloud.yml
```diff
@@ -0,0 +1,141 @@
...
+            xargs aws --region us-gov-east-1 lambda add-layer-version-permission \
+              --layer-name ${{ inputs.environment }}-${{ matrix.layer }}-${{ matrix.arch }} \
+              --statement-id 'PublicLayer' \
+              --action lambda:GetLayerVersion \
+              --principal '*' \
+              --version-number 
+
+  copy_west:
```
leandrodamascena, line 144: Can't we add those regions in the matrix and avoid duplicating the code? We do this when building and deploying the layer.

leandrodamascena (changes requested):
Hi @sthulb! Thanks for working on this! I've left some comments to make this pipeline even safer and avoid any issues that could cause incidents in production.

Please consider the same comments for both regions.

.github/workflows/layer_govcloud.yml
sthulb: This workflow will be converted to a dispatch workflow so we can use it in the main release workflow eventually. Leaving it as per env is better for that purpose.

.github/workflows/layer_govcloud.yml
sthulb: No, there's no need, this is the copy from region operation – if use-east-1 is down for any reason, we won't be releasing.

.github/workflows/layer_govcloud.yml
sthulb: A `workflow_call` option will be added as well as this so when we're happy with the performance, we can add it to the release workflow. I'm going to leave this as per env for that reason.

.github/workflows/layer_govcloud.yml
sthulb: I'll ignore this until that issue is resolved.

.github/workflows/layer_govcloud.yml
sthulb: good catch

.github/workflows/layer_govcloud.yml
sthulb: No because you can't use multiple environments in the steps

leandrodamascena (approved):

PR #5316, opened 2024-10-05
docs(upgrade_guide): update upgrade guide with Pydantic information

### LINKED ISSUE #5249: Bug: pydantic imports from aws_lambda_powertools.utilities.parser.pydantic no longer available
### Expected Behaviour

As per the [documentation](https://docs.powertools.aws.dev/lambda/python/latest/utilities/parser/#faq) FAQ I would expect to be able to import X from pydantic to include types such as EmailStr

### Current Behaviour

Currently in 3.0.0 we are unable to import pydantic types via the parser.

### Code snippet

```python
from aws_lambda_powertools.utilities.parser.pydantic import EmailStr
```

### Possible Solution

Update documentation to include the recommended way to import X from pydantic, or implement the feature that was available in 2.X.

### Steps to Reproduce

Try and import a pydantic type from aws_lambda_powertools.utilities.parser.pydantic, which no longer exists.

latest

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #5249 

### Changes

In Powertools v3 we have dropped support for Pydantic v1 and are only supporting Pydantic v2 which is why we are no longer exporting pydantic types directly and allowing customers to import them directly from Pydantic.

1 - Pydantic V2 was a complete refactoring of the Pydantic project and they made several changes including, but not limited, to the way they export types, classes, methods, and others. If we continue to expose them directly from Powertools and some export in Pydantic has changed, our customers may get the impression that there is some bug in Powertools and that is not the experience we want to have for our customers.

2 - The way we were doing this was by [exposing everything](https://github.com/aws-powertools/powertools-lambda-python/blob/v2/aws_lambda_powertools/utilities/parser/pydantic.py) exposed from pydantic to our customers, but there is a long discussion in the Pydantic repository about how https://github.com/pydantic/pydantic/discussions/6748 when importing and loading things in this new version. We don't want our customers to inadvertently import everything from Pydantic and this could create some slowdown in their functions.

### User experience

```diff
- #BEFORE - v2
- from aws_lambda_powertools.utilities.parser.pydantic import EmailStr

+ # NOW - v3
+ from pydantic import EmailStr
```
## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


docs/upgrade.md
```diff
@@ -91,6 +92,20 @@ def lambda_handler(event: DynamoDBStreamModel, context: LambdaContext):
 
 ```
 
+## Importing Pydantic objects
+
+We have stopped exporting Pydantic objects directly from `aws_lambda_powertools.utilities.parser.pydantic` to prevent potential confusion caused by Pydantic V2's significant changes in type, class, and method exports.
+
+This change prevents customers from accidentally importing all of Pydantic, which could significantly slow down function startup times.
```
dreamorosi:
```suggestion
This change gives you more control over what you import from Pydantic and minimize its impact on function startup times.
```

docs/upgrade.md
```diff
@@ -91,6 +92,20 @@ def lambda_handler(event: DynamoDBStreamModel, context: LambdaContext):
 
 ```
 
+## Importing Pydantic objects
+
+We have stopped exporting Pydantic objects directly from `aws_lambda_powertools.utilities.parser.pydantic` to prevent potential confusion caused by Pydantic V2's significant changes in type, class, and method exports.
```
dreamorosi:
```suggestion
You can now import Pydantic objects directly rather than from `aws_lambda_powertools.utilities.parser.pydantic`. We've made this change to streamline the import structure and allow you to use new Pydantic features as soon as they become available.
```

leandrodamascena: I like the idea but the `and allow you to use new Pydantic features as soon as they become available.` part is not realistic. 

What about

> You can now import Pydantic objects directly rather than from aws_lambda_powertools.utilities.parser.pydantic. This change provides a more intentional import structure, allowing you to import only the specific Pydantic components you need.

dreamorosi: That's ok - but then I think the next sentence might be somewhat redundant

leandrodamascena: What about this?

image

dreamorosi (dismissed):

dreamorosi (approved):

PR #5317, opened 2024-10-05
fix(appsync): make contextual data accessible for async functions

### LINKED ISSUE #5290: Bug: Router context not being available when using AppSync with async resolvers
### Expected Behaviour

When accessing the router.context in a resolver, with async resolvers, it should not be an empty dict, but much rather expose the context injected with app.append_context[](url)

### Current Behaviour

When using router.context in async resolvers, the router.context is an empty dict

### Code snippet

```python
    token = extract_auth_token_from_event(event)
    logger.info("token", extra={"token": token})
    app.append_context(
        isAdmin=True,
    )
    result = app.resolve(event, context)

    return asyncio.run(result)

@router.resolver(type_name="Mutation", field_name="addPet")
async def addPet(pet: dict):
    logger.info("Handling add pet mutation")
    logger.info("Context", extra={"resolver_context": router.context}) // this shows that router.context is an empty dict.
    
    return { "name": "Peter", "type": "Cat" }
```

### Possible Solution

_No response_

### Steps to Reproduce

Just plug in the lambda_handler and mutation I attached above, deploy it and run the mutation via the AppSync console

latest

3.8

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

leandrodamascena:
**Issue number:** #5290 

### Changes

This PR addresses an issue where the App/Router context was not available when using `AppSyncResolver` with async functions. The problem stemmed from the context being cleared too early in the execution flow, specifically before async functions could access it.

This change allows the context to persist throughout the execution of async functions, ensuring that contextual data remains accessible when needed. However, it shifts the responsibility of context clearing to the developer for async functions to prevent potential data leaks or unintended data persistence between invocations.

We've also updated the documentation to clearly communicate this change to users, providing guidance on how to manually clear the context when using async resolvers.

This bug doesn't affect batch resolvers.

What I fixed:

1. Modifying the `resolve` method in the `AppSyncResolver` class class to detect if the response is a coroutine.
2. If the response is a coroutine (indicating an async function), the context is not cleared automatically.
3. For non-coroutine responses, the context continues to be cleared automatically for safety.

### User experience

```python
def test_route_context_is_manually_cleared_after_resolve_async():
    app = AppSyncResolver()

    mock_event = {"typeName": "Customer", "fieldName": "field", "arguments": {}}

    @app.resolver(field_name="field")
    async def get_async():
        app.context.clear()
        await asyncio.sleep(0.0001)
        return "value"

    # WHEN
    mock_context = LambdaContext()
    app.append_context(is_admin=True)
    result = app.resolve(mock_event, mock_context)

    # THEN
    assert asyncio.run(result) == "value"
    assert app.context == {}
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/event_handler/appsync.py
```diff
@@ -149,7 +149,12 @@ def lambda_handler(event, context):
             Router.current_event = data_model(event)
             response = self._call_single_resolver(event=event, data_model=data_model)
 
-        self.clear_context()
+        # We don't clear the context for coroutines because we don't have control over the event loop.
+        # If we clean the context immediately, it might not be available when the coroutine is actually executed.
+        # For single async operations, the context should be cleaned up manually after the coroutine completes.
+        # See: https://github.com/aws-powertools/powertools-lambda-python/issues/5290
+        if not asyncio.iscoroutine(response):
```
dreamorosi, line 157: As discussed during our sync, before going forward with this implementation I think we could explore calling `self.clear_context()` at the beginning of this function rather than at the end (i.e. ~L141).

My reasoning here is that since the point of this call is to clear the context after each request so that no info is leaked to the next one, then we could move this logic at the beginning and achieve the same outcome.

Since we are explicitly targeting AWS Lambda in this feature, we can assume that we have a single entry point and that all it matters is that at the beginning of a request the context is clean. Cleaning it before ensures this but also removes the issue of awaiting async operations entirely.

Overall, at least based on my mental model of the utility (which might be incorrect), it doesn't matter if we leave the context dirty after a request as long as it's cleaned before the next request since Lambda might recycle the environment anyway.

Let me know if you think this is feasible at all, ultimately I don't have enough context on the full implementation and I might be completely off base.

leandrodamascena: Hi @dreamorosi! I've spent some time investigating this and unfortunately there is no solution for this other than refactoring the way we run the single resolvers in async mode. If I refactor this, I can introduce a major breaking change for those who have control of the async/eventloop event, as I would have to make Powertools have this control and I have no idea of ​​the consequences for those who are using it the current way.

So, I suggest we move on with the current solution.

dreamorosi: Ok, makes sense.

Not sure if you want to track this for a future version, but either way we can resolve the discussion and move forward for now.

dreamorosi (dismissed):

aws_lambda_powertools/event_handler/appsync.py
leandrodamascena: Yep, good point! I'm adding a comment "# TODO" to track that. I'll need a new approval.

dreamorosi (dismissed):

sthulb (approved):

PR #5742, opened 2024-12-16
fix(parser): remove AttributeError validation from event_parser function

### LINKED ISSUE #5705: Bug: event_parser decorator catches an irrelevant AttributeError and raises InvalidModelTypeError exception.
### Expected Behaviour

The expected behavior is that exceptions related to Powertools do not arise from the event_parser decorator in code like the following:

```python
from aws_lambda_powertools.utilities.parser import event_parser


class EventModel(BaseModel):
    """It's just a sample, so the content can be anything."""
    val: str


@event_parser(model=EventModel)
def lambda_handler(event: EventModel, context):
    print(event)  # correctly parsed

    a = "aaa"
    a.property  # raise AttributeError (**NOT related to event_parser**)


# The following is intended for simple reproduction in a local environment:
if __name__ == "__main__":
    lambda_handler({"val": "hello, lambda"}, None)
```

Assume that the `event` is in a format that matches the `EventModel` by pydantic and is correctly parsed.
Therefore, the exception happening in the above code has nothing to do with powertools event_parser.

### Current Behaviour

InvalidModelTypeError occurs within powertools, as shown below:

>     raise InvalidModelTypeError(f"Error: {str(exc)}. Please ensure the type you're trying to parse into is correct")
>     aws_lambda_powertools.utilities.parser.exceptions.InvalidModelTypeError: Error: 'str' object has no attribute 'property'. Please ensure the type you're trying to parse into is correct

I consider it very confusing because even if the event validation in the event_parser decorator succeeds, the above exception is raised.

### Code snippet

```python
Same as described in “Expected Behaviour”.
```

### Possible Solution

Currently (v3.3.0), AttributeError is caught at the error location in powertools as follows:

https://github.com/aws-powertools/powertools-lambda-python/blob/v3.3.0/aws_lambda_powertools/utilities/parser/parser.py#L114

Is it problematic to rewrite this part to catch Pydantic-derived Exceptions (e.g. [pydantic.ValidationError](https://docs.pydantic.dev/2.10/api/pydantic_core/#pydantic_core.ValidationError)) instead of AttributeError?

### Steps to Reproduce

- Python version: 3.12
- architecture: x86_64
- The libraries used are the same versions as those installed by the following requirements.txt file:

```txt
# requirements.txt
annotated-types==0.7.0
aws-lambda-powertools==3.3.0
jmespath==1.0.1
pydantic==2.10.3
pydantic-core==2.27.1
typing-extensions==4.12.2
```

3.3.0

3.12

### Packaging format used

PyPi

### Debugging logs

```python

```

anafalcao:
**Issue number:** #5705 

### Changes

Refine exception handling in the event_parser decorator and parser function. Replace the broad AttributeError catch with specific handling for ValidationError, InvalidModelTypeError, TypeError, and ValueError. This change ensures that Pydantic validation errors and invalid model type errors are raised as-is, while other type and value errors during model instantiation are wrapped in an InvalidModelTypeError. The AttributeError catch has been removed to prevent masking unrelated exceptions.

### User experience

Users encounter confusing InvalidModelTypeError exceptions, even when the event parsing succeeds. Unrelated exceptions within the handler function are incorrectly attributed to event parsing issues. The PR wants to provide the users more accurate error reporting. Exceptions unrelated to event parsing are no longer masked by parser, allowing for easier debugging and a clearer understanding of where errors originate in their code.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


tests/functional/parser/test_parser.py
```diff
@@ -184,17 +252,18 @@ class FailedCallback(pydantic.BaseModel):
...
-            return f"Uh oh. Had a problem: {event.error}"
+        if isinstance(event.data, FailedCallback):
+            return f"Uh oh. Had a problem: {event.data.error}"
 
-        return f"Successfully retrieved {event.breed} named {event.name}"
+        return f"Successfully retrieved {event.data.breed} named {event.data.name}"
 
-    ret = handler(test_input, None)
+    wrapped_input = {"data": test_input}
+    ret = handler(wrapped_input, None)
```
leandrodamascena: Same here.

aws_lambda_powertools/utilities/parser/parser.py
```diff
@@ -204,4 +202,4 @@ def handler(event: Order, context: LambdaContext):
             f"Error: {str(exc)}. Please ensure the Input model inherits from BaseModel,\n"
             "and your payload adheres to the specified Input model structure.\n"
             f"Model={model}",
-        ) from exc
+        )
```
leandrodamascena: Please keep `from exc`, so we can propagate the original error to the customer.

leandrodamascena (changes requested):
Hey @anafalcao! Thanks a lot for working on this. We need to make some small changes before we merge!

aws_lambda_powertools/utilities/parser/parser.py
anafalcao: Thank you @leandrodamascena , reverted it back

tests/functional/parser/test_parser.py
anafalcao: done!

leandrodamascena (approved):
Amazing work @anafalcao! APPROVED!

PR #5801, opened 2024-12-27
fix(appsync): enhance consistency for custom resolver field naming in AppSync

### LINKED ISSUE #5560: Maintenance: AppSyncResolverEvent uses `typeName` instead of `parentTypeName`
### Why is this needed?

I'm new to GraphQL in general, but I think in the original PR #323 the `AppSyncResolverEvent` was based on AWS Amplify, so instead of reading `parentTypeName`, we are reading `typeName`.

It confused me that the code is consistent (in some ways) with AWS AppSync Context info:

```ts
type Info = {
  fieldName: string;
  parentTypeName: string;
  variables: any;
  selectionSetList: string[];
  selectionSetGraphQL: string;
};
```

But we are assigning variables in a way they are not clear:

https://github.com/aws-powertools/powertools-lambda-python/blob/a3f3359bafeba39002a27ee9d02c66c57ed32bbc/aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py#L161

And we define different property names for reading the `parentTypeName`:

https://github.com/aws-powertools/powertools-lambda-python/blob/a3f3359bafeba39002a27ee9d02c66c57ed32bbc/aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py#L121-L124

https://github.com/aws-powertools/powertools-lambda-python/blob/a3f3359bafeba39002a27ee9d02c66c57ed32bbc/aws_lambda_powertools/utilities/data_classes/appsync_resolver_event.py#L165-L168

This has the side-effect that when setting up an AppSync resolved in AWS like the following, it does not work out of the box and you have to pass `typeName` to make it work:

```ts
import { util } from '@aws-appsync/utils';

/**
 * Sends a request to a Lambda function. Passes all information about the request from the `info` object.
 * @param {import('@aws-appsync/utils').Context} ctx the context
 * @returns {import('@aws-appsync/utils').LambdaRequest} the request
 */
export function request(ctx) {
    return {
        operation: 'Invoke',
        payload: {
            fieldName: ctx.info.fieldName,
            parentTypeName: ctx.info.parentTypeName,
            typeName: ctx.info.parentTypeName,  // Extra field which is exactly the same as the one above
            arguments: {}, // Extra field
            variables: ctx.info.variables,
            selectionSetList: ctx.info.selectionSetList,
            selectionSetGraphQL: ctx.info.selectionSetGraphQL,
        },
    };
}

/**
 * Process a Lambda function response
 * @param {import('@aws-appsync/utils').Context} ctx the context
 * @returns {*} the Lambda function response
 */
export function response(ctx) {
    const { result, error } = ctx;
    if (error) {
        util.error(error.message, error.type, result);
    }
    return result;
}

```

Event Source Data Classes

### Solution

I'm not sure if it was the intention and it should stay the same, but I think we can improve this in different ways:

- If the old approach is not desired, we can just deprecate it and mention `parent_type_name` should be used
- If the old approach is still valid, we can try to read `info` from `parentTypeName` after `typeName` is checked. This way it is consistent with the AWS Resolver context [here](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference-js.html) and with the Amplify structure

I can prepare a PR if the suggestion makes sense.

### Acknowledgment

leandrodamascena:
**Issue number:** #5560 

### Changes

This PR aims to bring consistency to scenarios where customers want to use a custom resolver in AppSync to place fields outside of the `info` object while maintaining consistency by using the same field name `parentTypeName`.

- Modified the logic to check for `parentTypeName` both inside and outside the `info` object
- Implemented a fallback to `typeName` if `parentTypeName` is not found

### User experience

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


anafalcao (approved):
great job @leandrodamascena ! approved

PR #5837, opened 2025-01-07
feat(data-masking): add custom mask functionalities

### LINKED ISSUE #5826: Feature request: support for custom masking with regx pattern or custom masking chars
### Use case

In some of case I would like to custom masking like example Aadhar number or SSN number is 1111-1111-1111 then masking should be done like `****-****-6789` or `****-****-****` or regex pattern. 

### Solution/User Experience

currently `erase`, `encrypt` and `decrypt` is supported. Similarly can be introduce `mask` or `custom_mask` method which accept parameter s like `masking_chars` 
This is just an example to explain how it can worked this can move to prebuild function `mask` or `custom_mask` 
```
import re

# Initialize logger with custom masking
logger = Logger()

def mask_sensitive_data(value):
    """Custom function to mask sensitive data"""
    return re.sub(r'\d{4}-\d{4}-\d{4}-(\d{4})', '****-****-****-****', value)

user_data = {
    "username": "john_doe",
    "card_number": "4111-1111-1111-1111"
}

# Apply custom masking
masked_card_number = mask_sensitive_data(user_data["card_number"])
logger.info("Processing payment", extra={"card_number": masked_card_number})

```

### Alternative solutions

```markdown

```

### Acknowledgment

anafalcao:
**Issue number:**
#5826 

## Summary

This PR enhances the data masking tool by introducing flexible masking options. These new features allow for dynamic, pattern-based, and regex-based masking, providing users with greater control over how sensitive data is obscured in using the `erase` method.

### Changes

New flags for `erase()`:

- `dynamic_mask` (bool): Enables dynamic masking behavior when set to True, by maintaining the original length and structure of the text replacing with `*`. 
Example: dynamic_mask = True for 'Avenue St' is '****** **'
- `custom_mask` (str): Specifies a simple pattern for masking data. This pattern is applied directly to the input string, replacing all the original characters.  
For example, with a mask_pattern of "XX-XX" applied to "12345", the result would be "XX-XX".
- `regex_pattern` (str): Defines a regular expression pattern used to identify parts of the input string that should be masked. This allows for more complex and flexible masking rules. It's used in conjunction with mask_format.
- `mask_format` (str): Specifies the format to use when replacing parts of the string matched by regex_pattern. It can include placeholders (like \1, \2) to refer to captured groups in the regex pattern, allowing some parts of the original string to be preserved. 
For example: 'example@email.com' could become 'e*****@email.com'
- `masking_rules` (dict): Apply different rules (formats) for each data field. 

### User experience

Previously, users had limited options for masking sensitive data. The `erase()` function provided basic masking capabilities, typically replacing entire fields or values with a fixed mask (e.g., '*****'). 
With the new masking options, users now have much more control over how their sensitive data is obscured. The enhanced `erase()` function offers a range of flexible masking techniques to suit various use cases, including different techniques for each field:

```python

masker = DataMasking()

masking_rules = {
    "credit_card": {"custom_mask": "XX"},
    "street": {"dynamic_mask": True},
    "email": {"regex_pattern": r"(\w)[\w.-]+@([\w.-]+)", "mask_format": r"\1****@\2"}
}
masked_data = masker.erase(
    data={"credit_card": "1234-5678-9012-3456", "street": "Avenue St", "email": "user@example.com"},
    masking_rules=masking_rules
)
# Result: {"credit_card": "XX", "street": "****** **" , "email": "u****@example.com"}
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


anafalcao:
Hi @leandrodamascena ! Can I have your help here with mypy? Thanks!

anafalcao:
Hi @leandrodamascena!
I just converted to Ready for review. I've been having issues regarding Incompatible types in assignment with mypy. Can you take a look?
I also created some tests for the new functionalities, but I also may need to implement some more after fixing this types issues

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -239,13 +293,82 @@ def _apply_action_to_fields(
...
+        Apply masking rules to data, supporting both simple field names and complex path expressions.
+
+        Args:
+            data: The dictionary containing data to mask
+            masking_rules: Dictionary mapping field names or path expressions to masking rules
+
+        Returns:
+            dict: The masked data dictionary
+        """
+        result = data.copy()
```
leandrodamascena: We need to replace it with `deepcopy`. Just a quick explanation on how Python works with [shallow copies and deepcopies](https://docs.python.org/3/library/copy.html#module-copy).

Imagine this code:
```python
import copy

# Original dictionary with nested structure
original_data = {
    'user': {
        'name': 'John Doe',
        'address': {'street': '123 Main St', 'city': 'Anytown'}
    },
    'preferences': ['dark mode']
}

shallow_copy = original_data.copy()
```

When you write `shallow_copy = original_data.copy()`, python creates a new object (in this case, a new dictionary), but it does not create new copies of the objects inside it. Instead, it only copies references to those inner objects.
So, if you run this code:

```python
import copy

# Original dictionary with nested structure
original_data = {
    'user': {
        'name': 'John Doe',
        'address': {'street': '123 Main St', 'city': 'Anytown'}
    },
    'preferences': ['dark mode']
}

shallow_copy = original_data.copy()

# Shallow copy
shallow_copy['user']['address']['street'] = '456 Elm St'  # Mutates original
shallow_copy['preferences'].append('auto-save')           # Mutates original

print("\nShallow copy:", shallow_copy)
print("Original data:", original_data)
```

The output will be:
```python
Shallow copy: {'user': {'name': 'John Doe', 'address': {'street': '456 Elm St', 'city': 'Anytown'}}, 'preferences': ['dark mode', 'auto-save']}
Original data: {'user': {'name': 'John Doe', 'address': {'street': '456 Elm St', 'city': 'Anytown'}}, 'preferences': ['dark mode', 'auto-save']}
```

You see that Python mutates the original dictionary, thats why because nested keys are not part of the new object, but references to the original one instead.

If you use `deep_copy = copy.deepcopy(original_data)`, it will create a new dictionary and copy all the keys, making it completely isolated object.

The side effect of actual code is because we will mutate the original value of `event` and this can be a huge problems for some customers.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -239,13 +293,82 @@ def _apply_action_to_fields(
...
+                    base_path, field = path.split("..")
+                    jsonpath_expr = parse(f"$.{base_path}..{field}")
+                elif "[" in path:
+                    # Handle array notation paths (e.g., "address[*].street")
+                    jsonpath_expr = parse(f"$.{path}")
+                else:
+                    # Handle simple field names (e.g., "email")
+                    jsonpath_expr = parse(f"$.{path}")
+
+                matches = jsonpath_expr.find(result)
```
leandrodamascena:
```suggestion
                jsonpath_expr = parse(f"$.{path}")
                matches = jsonpath_expr.find(result)
```

You're basically writing if to do the same thing, I think we can remove those.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -239,13 +293,82 @@ def _apply_action_to_fields(
...
+                        value = match.value
+                        if value is not None:
+                            if isinstance(value, dict):
+                                # Handle dictionary values by masking each field
+                                for k, v in value.items():
+                                    if v is not None:
+                                        value[k] = self.provider.erase(str(v), **rule)
+                            else:
+                                masked_value = self.provider.erase(str(value), **rule)
+                                match.full_path.update(result, masked_value)
```
leandrodamascena:
```suggestion
			        masked_value = self.provider.erase(str(value), **rule)
                                match.full_path.update(result, masked_value)
```

I understand why you're doing this, but our original implementation is to remove the field/value, so, I think make sense to keep the same behavior.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -94,15 +94,53 @@ def erase(self, data: tuple, fields: list[str]) -> tuple[str]: ...
     @overload
     def erase(self, data: dict, fields: list[str]) -> dict: ...
 
-    def erase(self, data: Sequence | Mapping, fields: list[str] | None = None) -> str | list[str] | tuple[str] | dict:
-        return self._apply_action(data=data, fields=fields, action=self.provider.erase)
+    @overload
+    def erase(self, data: dict[Any, Any], *, masking_rules: dict[str, object]) -> dict[Any, Any]: ...
```
leandrodamascena: I'm not sure if we need the arg `*` here. Can you try to remove this method signature and see if mypy complains?

docs/utilities/data_masking.md
```diff
@@ -117,6 +117,42 @@ Erasing will remove the original data and replace it with a `*****`. This means
...
+=== "regex_pattern & mask_format"
+
+    (str) `regex_pattern` defines a regular expression pattern used to identify parts of the input string that should be masked. This allows for more complex and flexible masking rules. It's used in conjunction with `mask_format`.
+    `mask_format` specifies the format to use when replacing parts of the string matched by `regex_pattern`. It can include placeholders (like \1, \2) to refer to captured groups in the regex pattern, allowing some parts of the original string to be preserved.
+
+    > Expression: `data_masker.erase(data, fields=["email"], regex_pattern=r"(.)(.*)(@.*)", mask_format=r"\1****\3")`
+
+    > Field result: `'email': 'j****@example.com'`
+
+=== "masking_rules"
```
leandrodamascena, line 149: Please add a tab with the payload.

tests/unit/data_masking/_aws_encryption_sdk/test_unit_data_masking.py
```diff
@@ -205,3 +215,70 @@ def test_parsing_nonexistent_fields_warning_on_missing_field():
 
     # THEN the "erased" payload is the same of the original
     assert masked_json_string == data
+
+
+def test_regex_mask(data_masker):
+    data = "Hello! My name is Fulano Ciclano"
```
leandrodamascena:
```suggestion
    data = "Hello! My name is John Doe"
```

🇧🇷

leandrodamascena (changes requested):
Hey @anafalcao! Another round of review. This is a nice work, we just need to fix some things. 🚀

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Thank you so for detailed explanation!

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Got it! fixed in the new commit

docs/utilities/data_masking.md
anafalcao: Added Input and Output examples

tests/unit/data_masking/_aws_encryption_sdk/test_unit_data_masking.py
anafalcao: changed 🇧🇷

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Tried to remove, but increases the number of mypy errors

leandrodamascena: Lets keep this conversation opened until we find a solution for this.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -94,15 +95,53 @@ def erase(self, data: tuple, fields: list[str]) -> tuple[str]: ...
...
+        self,
+        data: dict,
+        fields: list[str],
+        dynamic_mask: bool | None = None,
+        custom_mask: str | None = None,
+        regex_pattern: str | None = None,
+        mask_format: str | None = None,
+    ) -> dict: ...
+
+    def erase(
```
leandrodamascena: I think we will need to remove all the overloads and keep only this function implementation. Now we have different use cases that I don't know if we will can cover all of them and make mypy happy.

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -94,15 +95,53 @@ def erase(self, data: tuple, fields: list[str]) -> tuple[str]: ...
...
+        self,
+        data: Sequence | Mapping,
```
leandrodamascena:
```suggestion
        data: Any,
```

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -94,15 +95,53 @@ def erase(self, data: tuple, fields: list[str]) -> tuple[str]: ...
...
+        fields: list[str] | None = None,
+        dynamic_mask: bool | None = None,
+        custom_mask: str | None = None,
+        regex_pattern: str | None = None,
+        mask_format: str | None = None,
+        masking_rules: dict | None = None,
+    ) -> str | list[str] | tuple[str] | dict:
```
leandrodamascena:
```suggestion
    ) -> Any:
```

aws_lambda_powertools/utilities/data_masking/base.py
```diff
@@ -94,15 +95,53 @@ def erase(self, data: tuple, fields: list[str]) -> tuple[str]: ...
...
+        if masking_rules:
+            return self._apply_masking_rules(data, masking_rules)
```
leandrodamascena:
```suggestion
            return self._apply_masking_rules(data=data, masking_rules=masking_rules)
```

leandrodamascena (changes requested):
Hello @anafalcao! I left some comments in this review. 

As this PR is more complex, we will do it step by step to have more detailed feedback.

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Thank you again Leo! Makes sense, removed all the overloads

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Done!

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Done!

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Done!

aws_lambda_powertools/utilities/data_masking/base.py
anafalcao: Now it's resolved as removed the overload

leandrodamascena (dismissed):
APPROVING it @anafalcao! THANK YOU SO MUCH!

dreamorosi (approved):

PR #5896, opened 2025-01-22
chore(ci): drop support for Python 3.8

### LINKED ISSUE #5238: Maintenance: drop support for Python 3.8
### Summary

> [!important]
> Starting from February 7, 2025 Powertools for AWS Lambda (Python) will end support for Python 3.8

This action follows [the date when AWS Lambda will mark the `python3.8` as deprecated](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported) and precedes the date after which you will no longer be able to create new functions using the runtime.

We recommend that you upgrade your existing Python 3.8 functions to Python 3.10+ before February 7, 2025.

End of support does not impact existing functions and functions using versions of Powertools for AWS Lambda (Python) released before the date, however starting from this date we will not guarantee compatibility with Python 3.8.

You can use the following AWS CLI command to list all functions in a specific region using Python 3.8, including published function versions:

> [!Note]
> Make sure to replace the value of the `--region` flag and run this command in each region you have Lambda functions workloads

```sh
aws lambda list-functions --function-version ALL --region <Region> --output text --query "Functions[?Runtime=='python3.8'].FunctionArn"
```

### Why is this needed?

Starting from Feb 28, 2025 we will no longer be able to deploy new Lambda functions using the `python3.8` runtime, this means we will no longer be able to run integration tests for new code changes.

With this in mind, we plan on dropping support for the runtime just three weeks before the date to avoid any impact in our releases and end2end tests.

Automation

### Solution

_No response_

### Acknowledgment

leandrodamascena:
**Issue number:** #5238 

### Changes

On 2024-10-25, we announced the deprecation of Python 3.8 in Powertools, scheduled for 2025-02-07. During this time, we continued to release versions for Python 3.8 and allowed customers to migrate to newer versions of Python without any impact.

#### AWS Lambda
AWS Lambda runtime will fully deprecate [Python 3.8 on Mar 31, 2025](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtime-support-policy). This means, 3.8 functions will be unable to be updated. Inline with this, Powertools releases will stop supporting it.

We strongly advice you update your Lambda functions to at least Python 3.10+.

### User experience

Customers will no longer be able to use Powertools for AWS Lambda (Python) with Python 3.8.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


dreamorosi (dismissed):
Thanks for working on this.

Leaving a comment, but feel free to not implement it, if you think it's not useful:

The upgrade guide from v2 to v3 currently says to migrate to at least 3.8. That was true at that time, but if someone follows the guide next week they might get stuck since the version at that time doesn't support it anymore. Not sure you want to change it or call out that we dropped support.

Other than this, good to go.

leandrodamascena:
> The upgrade guide from v2 to v3 currently says to migrate to at least 3.8. That was true at that time, but if someone follows the guide next week they might get stuck since the version at that time doesn't support it anymore. Not sure you want to change it or call out that we dropped support.
> 

You're right here and it's a good point, but I don't think we should be encouraging customers to adopt 3.8 anymore, especially since if someone is migrating from version 2 to version 3 and sticking with Python 3.8, that means they have less than 2 months to get everything done and never change a single line of code in that function again because AWS Lambda will drop update function soon.

What do you think about waiting for someone to open an issue or complain about it? Then we can provide specific answers to specific requests, that is, if they happen.

dreamorosi:
> > The upgrade guide from v2 to v3 currently says to migrate to at least 3.8. That was true at that time, but if someone follows the guide next week they might get stuck since the version at that time doesn't support it anymore. Not sure you want to change it or call out that we dropped support.
> 
> You're right here and it's a good point, but I don't think we should be encouraging customers to adopt 3.8 anymore, especially since if someone is migrating from version 2 to version 3 and sticking with Python 3.8, that means they have less than 2 months to get everything done and never change a single line of code in that function again because AWS Lambda will drop update function soon.
> 
> What do you think about waiting for someone to open an issue or complain about it? Then we can provide specific answers to specific requests, that is, if they happen.

Sorry I haven't been clear maybe.

My suggestion was to update the text in the upgrade guide to say "upgrade to at least 3.9".

leandrodamascena:
> 
> My suggestion was to update the text in the upgrade guide to say "upgrade to at least 3.9".

This: https://github.com/aws-powertools/powertools-lambda-python/pull/5896/files#diff-3ec187be4d5027db52c481841dbbf321b2f4b39dea03033c971aeba9a4a20e51R34?

dreamorosi (approved):
Thanks for the PR and for addressing my comments.

PR #5956, opened 2025-01-30
feat(logger): add clear_state method

### LINKED ISSUE #4134: Feature request: convenience `reset_state` method in Logger
### Use case

> [!note]
> This is a cross-language feature request that was originally brought up under aws-powertools/powertools-lambda-typescript#2337

As a customer who uses canonical logging - aka wide logging - I would like to be able to clear any key I might have appended to the Logger instance regardless of me using the default formatter or the decorator - both of which either expose a method along these lines or can be configured to automatically clear state.

### Solution/User Experience

```py

logger = Logger()

    try:
        logger.append_keys(order_id=event.get('order_id', None)
    finally:
        logger.info('CANONICAL')
        logger.clear_state() 
```

### Alternative solutions

```markdown
N/A
```

### Acknowledgment

### LINKED ISSUE #4458: Docs: impact of `clear_state=True` when using Lambda Layers & Logger

I was reading about `Logger() `

https://docs.powertools.aws.dev/lambda/python/latest/core/logger/#clearing-all-state

### How can we improve?

The documentation contains a warning `Danger: This can have unintended side effects if you use Layers` however I believe this relates to using other layers as opposed to using the Powertools layer itself.

### Got a suggestion in mind?

Something like...

```
Danger: This can have unintended side effects if you use Layers other than Powertools.
```

### Acknowledgment

anafalcao:
**Issue number:**
#4134 
#4458

## Summary
This PR introduces `clear_state()` method to the Logger class. This enhancement provides developers with more fine-grained control over the logger's state within a single Lambda function execution, allowing for more flexible and context-specific logging.

### Changes

- Add a new `clear_state()` method to the Logger class
- For clear_state decorator, the 'danger message' contains more information with what it actually affects.

### User experience

1. clear_state method

image

2. Add more information about clear_state decorator

image

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/logging/logger.py
```diff
@@ -222,6 +222,7 @@ def __init__(
             choice=sampling_rate,
             env=os.getenv(constants.LOGGER_LOG_SAMPLING_RATE),
         )
+        self._default_log_keys: dict[str, Any] = {"service": self.service, "sampling_rate": self.sampling_rate}
```
leandrodamascena, line 225: This already exists in the line [235](https://github.com/aws-powertools/powertools-lambda-python/pull/5956/files#diff-7644429742639d898597d1548e2d8b39bf59a71be9c6d8d143d19d8529cd62cbR235). 

I think it makes sense for you keep type annotation `dict[..` , but please remove one of them.

aws_lambda_powertools/logging/logger.py
```diff
@@ -605,6 +606,15 @@ def append_context_keys(self, **additional_keys: Any) -> Generator[None, None, N
...
             yield
 
+    def clear_state(self) -> None:
+        """Removes all custom keys that were appended to the Logger."""
+        # Clear all custom keys from the formatter
+        self.registered_formatter.clear_state()
+
+        # Reset to default keys
+        default_keys: dict[Any, Any] = dict(self._default_log_keys)
+        self.structure_logs(**default_keys)
```
leandrodamascena:
```suggestion
        self.structure_logs(self._default_log_keys, self.logger_formatter)
```

`self._default_log_keys` is already a dict and immutable for this context, so, we don't need to copy this to prevent mutation and stuff. 

The reason I'm adding `self.logger_formatter` is just to keep the API consistent. It shouldn't make any difference in this context, but since `structure_logs` is the function responsible for resolving whether the logger instance will use the custom `logger_formatter` or the default `LambdaPowertoolsFormatter`, then we keep things in order.

docs/core/logger.md
```diff
@@ -274,13 +274,15 @@ You can remove any additional key from Logger state using `remove_keys`.
...
+##### Decorator with clear_state
+
 Logger is commonly initialized in the global scope. Due to [Lambda Execution Context reuse](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-context.html){target="_blank"}, this means that custom keys can be persisted across invocations. If you want all custom keys to be deleted, you can use `clear_state=True` param in `inject_lambda_context` decorator.
 
 ???+ tip "Tip: When is this useful?"
     It is useful when you add multiple custom keys conditionally, instead of setting a default `None` value if not present. Any key with `None` value is automatically removed by Logger.
 
 ???+ danger "Danger: This can have unintended side effects if you use Layers"
-    Lambda Layers code is imported before the Lambda handler.
+    Lambda Layers code is imported before the Lambda handler. When a Lambda function starts, it first imports and executes all code in the Layers (including any global scope code) before proceeding to the function's own code.
```
leandrodamascena, line 285: SUPER NICE! 🚀

docs/core/logger.md
```diff
@@ -304,6 +306,29 @@ Logger is commonly initialized in the global scope. Due to [Lambda Execution Con
     --8<-- "examples/logger/src/clear_state_event_two.json"
     ```
 
+##### clear_state method
+
+`clear_state()` is a method you can call explicitly within your code to clear appended keys at any point during the execution of a single Lambda invocation.
```
leandrodamascena:
```suggestion
You can call `clear_state()` as a method explicitly within your code to clear appended keys at any point during the execution of your Lambda invocation.
```

It's better to use active voice.

docs/core/logger.md
```diff
@@ -304,6 +306,29 @@ Logger is commonly initialized in the global scope. Due to [Lambda Execution Con
...
+
+This allows for more granular control over the logger's state within a single function execution, enabling you to reset the logger to its initial state before specific logging operations or at the end of certain processes within the same Lambda run.
```
leandrodamascena: I think we can remove this paragraph. 

> state within a single function execution,
> of certain processes within the same Lambda run.

These statements may be invalid due to Lambda's execution model and state persisting between invocations, so this means that if you add a key in one execution, it may persist between executions:

```python

logger = Logger()

    logger.info("No keys")
    logger.append_keys(mykey="key")
    logger.info("With keys")
```

First execution will log `no keys` without mykey
First execution will log `no keys` with mykey

tests/functional/logger/required_dependencies/test_logger.py
```diff
@@ -1232,3 +1232,32 @@ def test_logger_change_level_child_logger(stdout, service_name):
...
+    # WHEN append keys are added
+    logger.append_keys(custom_key="custom_key")
+    logger.info("message with appended keys")
+    logger.clear_state()
+
+    # THEN context keys should be cleared
+    assert "custom_key" not in logger.get_current_keys()
+
+
+def test_clear_state(stdout, service_name):
```
leandrodamascena, line 1250: Super nice test!

Just please add GIVEN WHEN THEN to make it clear to others what we are testing.

tests/functional/logger/required_dependencies/test_logger.py
```diff
@@ -1232,3 +1232,32 @@ def test_logger_change_level_child_logger(stdout, service_name):
     logs = list(stdout.getvalue().strip().split("\n"))
     assert len(logs) == 1
     assert "service" in logs[0]
+
+
+def test_clear_state_with_append_keys():
```
leandrodamascena, line 1237: Please add another test reading the output after you clean the state and log a new info. In this way we can make sure that the key is not present in the output/serialization.

leandrodamascena (changes requested):
Hey @anafalcao! SUPER NICE work!! 🥇 

I left some comments.

aws_lambda_powertools/logging/logger.py
anafalcao: Great, thanks for the heads up!

aws_lambda_powertools/logging/logger.py
anafalcao: If I remove `default_keys: dict[Any, Any] = dict(self._default_log_keys)` and keep only the suggested, it gives an error in mypy. 
Keeping as is for now.

docs/core/logger.md
anafalcao: Perfect!! Thanks, changed.

docs/core/logger.md
anafalcao: Removed the second paragraph, and left the first one with the modification proposed earlier.

tests/functional/logger/required_dependencies/test_logger.py
anafalcao: Thanks!! Added the GIVEN WHEN THEN

tests/functional/logger/required_dependencies/test_logger.py
anafalcao: Added as `test_clear_state_log_output`

aws_lambda_powertools/logging/logger.py
leandrodamascena: Actually you're right. Keys in the `structure_logs` are kwargs arguments, good catch!

But you can simple use `self.structure_logs(**self._default_log_keys)` and mypy will be happy.

leandrodamascena (changes requested):
Hi @anafalcao! We just need to fix one thing and we're ready to merge! Thanks for responding to all the comments.

aws_lambda_powertools/logging/logger.py
anafalcao: Perfect, this works! Thank you again.

leandrodamascena (approved):
Amazing work @anafalcao! Just wait until CI is complete and you can merge!
APPROVED! ✈️

PR #5997, opened 2025-02-03
feat(openapi): enhance support for tuple return type validation

### LINKED ISSUE #5034: Bug: get_field_info_and_type_annotation doesn't respect tuple return types
### Expected Behaviour

The `get_field_info_and_type_annotation` function should respect `tuple` response types since it is supported since [2.7.0](/aws-powertools/powertools-lambda-python/releases/tag/v2.7.0). See #1845 and #1853.

### Current Behaviour

Validation error (HTTP 422) happens when the endpoint function returning type annotation is a tuple, if we had `enable_validation=True` for OpenAPI validation.

### Code snippet

```python

app = ALBResolver(enable_validation=True)

class Result(BaseModel):
    payload: str

@app.post('/foobar')
def foobar() -> tuple[Result, int]:
    return Result(payload='foobar'), 201

def test_foobar():
    response = app.resolve({
        'path': '/foobar',
        'httpMethod': 'POST',
        'queryStringParameters': {},
        'headers': {}
    })
    assert response['statusCode'] == 201
```

Instead of `201 Created`, the endpoint currently returns 422 error with following body:

```json
{"statusCode":422,"detail":[{"loc":["response"],"type":"tuple_type"}]}
```

### Possible Solution

Update here to support tuple type for a response. In addition, distinguishing `is_response_param` is necessary for avoiding changing behaviors for other types of annotations.

https://github.com/aws-powertools/powertools-lambda-python/blob/8251bb09b3c95401faafd45d09cfe9072aed8765/aws_lambda_powertools/event_handler/openapi/params.py#L969-L987

### Steps to Reproduce

Please see above code snippet.

latest

3.11

### Packaging format used

PyPi

### Debugging logs

_No response_

philiptzou:
**Issue number:** #5034

### Changes

This PR fixes the issue that OpenAPI validation failed to validate two-element tuple returns, which is allowed by the `APIGateway`.

### User experience

Below test fails before this PR but passes after this PR:

```python3

app = ALBResolver(enable_validation=True)


class Result(BaseModel):
    payload: str


@app.post('/foobar')
def foobar() -> tuple[Result, int]:
    return Result(payload='foobar'), 201


def test_foobar():
    response = app.resolve({
        'path': '/foobar',
        'httpMethod': 'POST',
        'queryStringParameters': {},
        'headers': {}
    }, None)
    assert response['statusCode'] == 201
```
## Checklist

* [ ] Changes are documented

<details>
<summary><s>Is this a breaking change?</s></summary>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


tests/functional/event_handler/_pydantic/test_openapi_params.py
```diff
@@ -172,6 +172,42 @@ def handler() -> Response[Annotated[str, Body(title="Response title")]]:
     assert response.schema_.type == "string"
 
 
+def test_openapi_with_tuple_returns():
+    app = APIGatewayRestResolver()
+
+    @app.get("/")
+    def handler() -> tuple[str, int]:
```
leandrodamascena: Can you pls import `Tuple` from typing? I need to migrate our tests to adopt the new type annotation, but we have some issues for now.

tests/functional/event_handler/_pydantic/test_openapi_params.py
```diff
@@ -172,6 +172,42 @@ def handler() -> Response[Annotated[str, Body(title="Response title")]]:
...
+    response = get.responses[200].content[JSON_CONTENT_TYPE]
+    assert response.schema_.title == "Return"
+    assert response.schema_.type == "string"
+
+
+def test_openapi_with_tuple_annotated_returns():
+    app = APIGatewayRestResolver()
+
+    @app.get("/")
+    def handler() -> tuple[Annotated[str, Body(title="Response title")], int]:
```
leandrodamascena: Can you pls import `Tuple` from typing? I need to migrate our tests to adopt the new type annotation, but we have some issues for now.

leandrodamascena (changes requested):
Hey @philiptzou! Thanks a lot for submitting this PR. 🚀 

I see we have some problems with type checking that is breaking the CI and we need to address them before I make a final review.

tests/functional/event_handler/_pydantic/test_openapi_params.py
philiptzou: Done.

tests/functional/event_handler/_pydantic/test_openapi_params.py
philiptzou: Done.

leandrodamascena (approved):
Hey @philiptzou! Thanks for working on this PR and addressing the feedback!

I have a feeling that this PR can breaks the validation for people using dict as return. But I think this is really a bug that we need to fix.

e.g
```python
@app.get('/foobar')
def foobar() -> dict:
    return Result(payload='foobar'), 201
```

APPROVED!

philiptzou:
> Hey @philiptzou! Thanks for working on this PR and addressing the feedback!
> 
> I have a feeling that this PR can breaks the validation for people using dict as return. But I think this is really a bug that we need to fix.
> 
> e.g
> 
> ```python
> @app.get('/foobar')
> def foobar() -> dict:
>     return Result(payload='foobar'), 201
> ```
> 
> APPROVED!

Well, I don't think this PR will break dict as return since it still falls in the `else` branch. This example still work, but it irritates mypy anyway.

PR #6088, opened 2025-02-12
feat(batch): raise exception for invalid batch event

### LINKED ISSUE #2394: Feature request: Batch processor should fail early if event payload is incorrect
### Expected Behaviour

@heitorlessa this is a continuation of [23777](https://github.com/awslabs/aws-lambda-powertools-python/issues/2377)

When using the `BatchProcessor` with `process_partial_response` from [here](https://awslabs.github.io/aws-lambda-powertools-python/2.16.1/utilities/batch/) I have noticed that the event does not have to respect the model and it can return an empty list for the `batchItemFailures` which will tell AWS to delete the messages from the queue in case of SQS triggers. 

I'd expect the behavior to be fail-first fail-fast. If I got a bunch of messages to my Lambda that do not have the format of an `EventType.SQS` that should fail right away, in my opinion. Passing a random `dict` to the `process_partial_response` should not be considered a valid event.

### Current Behaviour

The current implementation of `process_partial_response` accepts any `dict` passed in. Since it used the `dict` method `.get` to get the `Records` which is a key in SQS type events. But returns a default empty list in case the key doesn't exist.  Which in turn will tell the context manager to run the processor on an empty list of records, which will always return success.

image

### Code snippet

```python
from aws_lambda_powertools.utilities.batch import (
    BatchProcessor, EventType, process_partial_response
)
import pydantic


class Event(pydantic.BaseModel):
    foo: str
    bar: str


def record_handler(record: Event):
    print(record)

    if record.foo == "barr":
        raise ValueError


def handler(event: dict, context: LambdaContext):

    return process_partial_response(
        event=event,
        record_handler=record_handler,
        processor=BatchProcessor(
            event_type=EventType.SQS,
            model=Event,
        ),
        context=context
    )


# This is an invalid event since it's not of the type SQS
# But the error about missing SQS model field is only thrown after the record_handler is called
# It should be at the very beginning, before any business logic occurs
print(handler({'Records': [{"foo": "bar", "bar": "baz"}, {"foo": "barr", "bar": "bazz"}]}, object))

# This is not valid, but does not throw an error
# And the batchItemFailures returns an empty list which is not accurate.
print(handler({"foo": "bar", "bar": "baz"}, object))

# This throws ValueError: Invalid event format. Please ensure batch event is a valid SQS event.
print(handler(["foo", "bar", "bar", "baz"], object))
```

### Possible Solution

I'm sure there are a variety of reasons for why this is the way it is that I'm not considering. One obvious answer is that SQS event types will always have the 'Records` key hence anything that doesn't have it should answer the question: "How did you get here and should you even be considered?"

But I believe that we can be a bit more strict on enforcing that the `dict` passed is indeed of the type SQS, DynamoDB stream or Kinesis before doing any processing.

For example, in the snippet above, I pass in a `dict` with the `Records` key, but the rest of the event does not have the remaining SQS keys. When the `record_handler` fails, the batch processor looks for the `messageId` field, which doesn't exist. I believe if this was done upfront and fail immediately it would be a better developer experience. 

What's everyone's thoughts?

### Steps to Reproduce

Snippet above should be enough to demonstrate all scenarios brought up.

latest

3.10

### Packaging format used

Lambda Layers

### Debugging logs

_No response_

anafalcao:
**Issue number:** #2394

### Changes

This PR adds fail fast error handling for unexpected batch types in the Batch Processing utility.

- Added `UnexpectedBatchTypeError` to handle invalid batch types in `process_partial_response` and `async_process_partial_response`
- Error messaging to indicate valid event source types:
`Unexpected batch event type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams`

### User experience

```python
from aws_lambda_powertools.utilities.batch import (
    BatchProcessor, 
    EventType,
    process_partial_response,
    UnexpectedBatchTypeError
)

logger = Logger()

def record_handler(record):
    return record["body"]

def handler(event, context):
    processor = BatchProcessor(event_type=EventType.SQS)
    
    try:
        return process_partial_response(
            event=event,
            record_handler=record_handler,
            processor=processor
        )
    except UnexpectedBatchTypeError as e:
        # "Unexpected batch event type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams"
        logger.error(f"Invalid batch event structure: {e}")
        raise
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/utilities/batch/exceptions.py
```diff
@@ -38,6 +38,15 @@ def __str__(self):
...
 
+class UnexpectedBatchTypeError(BatchProcessingError):
+    """Error thrown by the Batch Processing utility when a partial processor receives an unexpected batch type"""
+
+    def __init__(self):
+        msg = "Unexpected batch type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams"
+        super().__init__(msg)
+        self.name = "UnexpectedBatchTypeError"
+
+
```
leandrodamascena:
```suggestion
class UnexpectedBatchTypeError(Exception):
    """Error thrown by the Batch Processing utility when a partial processor receives an unexpected batch type"""
    
    pass
```

This is a simple exception where you don't need to handle messages and stuff like that, so you can inherit from Exception and delegate the message/info/raising to whoever is calling this exception.

aws_lambda_powertools/utilities/batch/decorators.py
```diff
@@ -204,6 +205,9 @@ def handler(event, context):
     """
     try:
         records: list[dict] = event.get("Records", [])
+        if not records or not isinstance(records, list):
+            raise UnexpectedBatchTypeError
```
leandrodamascena:
```suggestion
            raise UnexpectedBatchTypeError("Unexpected batch event type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams")
```

aws_lambda_powertools/utilities/batch/decorators.py
```diff
@@ -268,6 +272,9 @@ def handler(event, context):
...
```
leandrodamascena:
```suggestion
            raise UnexpectedBatchTypeError("Unexpected batch event type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams")
```

tests/functional/batch/required_dependencies/test_utilities_batch.py
```diff
@@ -708,3 +708,69 @@ def test_async_process_partial_response_invalid_input(async_record_handler: Call
...
+            event=event,
+            record_handler=async_record_handler,
+            processor=processor,
+        )
+
+    # THEN the correct error message is raised
+    assert "Unexpected batch type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams" in str(exc_info.value)
+
+
+def test_process_partial_response_should_not_raise_unexpected_batch_type(record_handler):
```
leandrodamascena: I understand the idea of ​​this test, but we don't need it here. We are already testing different batch types in this file and they are passing, what means we are processing events with valid types.

tests/functional/batch/required_dependencies/test_utilities_batch.py
```diff
@@ -708,3 +708,69 @@ def test_async_process_partial_response_invalid_input(async_record_handler: Call
...
+    # WHEN processing an event with invalid Records
+    with pytest.raises(UnexpectedBatchTypeError) as exc_info:
+        process_partial_response(
+            event=event,
+            record_handler=record_handler,
+            processor=processor,
+        )
+
+    # THEN the correct error message is raised
+    assert "Unexpected batch type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams" in str(exc_info.value)
```
leandrodamascena:
```suggestion
    assert "Unexpected batch event type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams" in str(exc_info.value)
```

tests/functional/batch/required_dependencies/test_utilities_batch.py
```diff
@@ -708,3 +708,69 @@ def test_async_process_partial_response_invalid_input(async_record_handler: Call
...
+    # WHEN processing an event with invalid Records asynchronously
+    with pytest.raises(UnexpectedBatchTypeError) as exc_info:
+        await async_process_partial_response(
+            event=event,
+            record_handler=async_record_handler,
+            processor=processor,
+        )
+
+    # THEN the correct error message is raised
+    assert "Unexpected batch type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams" in str(exc_info.value)
```
leandrodamascena:
```suggestion
    assert "Unexpected batch event type. Possible values are: SQS, KinesisDataStreams, DynamoDBStreams" in str(exc_info.value)
```

leandrodamascena (changes requested):
Hey @anafalcao! Thanks for working in this PR! I just left some comments before we merge this.

Please also leave a message in the original issue saying that we will raise an exception because we want to match the TypeScript version and have the same behavior in the API.

Also, in the "User Experience" section, add a complete example of the experience, including the try/except block, so that it's easy to copy when we write the release notes.

aws_lambda_powertools/utilities/batch/exceptions.py
anafalcao: Got it, thanks! Changed it

aws_lambda_powertools/utilities/batch/decorators.py
anafalcao: Resolved!

aws_lambda_powertools/utilities/batch/decorators.py
anafalcao: Resolved!

tests/functional/batch/required_dependencies/test_utilities_batch.py
anafalcao: OK! Removed it

anafalcao:
Hi @leandrodamascena ! Thank you for the feedback. I worked on them in this last commit, and updated the User Experience section for a more complete example. 
I'll leave a message in the original issue.

leandrodamascena (approved):
APPROVED!

PR #6142, opened 2025-02-21
feat(logger): add new logic to sample debug logs

### LINKED ISSUE #6141: Feature request: Calculate Log Sampling Per Request
### Use case

Original discussion: https://github.com/aws-powertools/powertools-lambda-typescript/issues/3278

Copying @dreamorosi's [comment](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3278#issuecomment-2664890356) to add more context to this feature request:

> Had a discussion about this with @am29d yesterday and I thought it'd be useful to bring here the outcomes of the discussion, primarily for future implementation & consideration.

> The log sampling feature changes the log level of a Logger to `debug` for a percentage of requests. Customers can set a rate in the Logger constructor and this rate is used to determine whether or not to change the log level to `debug`. This is useful for customers who want to keep a less verbose log level most of the time, but have more logs emitted for a percentage of their requests.

> As it stands, the feature doesn't exactly behaves as described above. This is because the percentage or ratio is not calculated at the request level, but rather when the Logger class is instantiated, which is usually during the `INIT` phase of the execution environment, i.e.

```python

logger = Logger(sampling_rate=0.5)   # whether or not the log level is switched to `debug` is decided here only

def handler(event, context):
    # ... your logic here
    pass

```

> This means that all the requests served by the same environment will inherit the sampling decision that was made when the environment was initialized, which in turn results in a sampling rate different than the desired one. The degree of this difference will depend on how many environments are spun up and the distribution of requests among them.

> To explain what I mean by that, let's consider this example that has 3 environments/sandboxes and a number of requests distributed across them, and - for the sake of simplicity - a log sampling of 0.5 (aka 50%):

> Image

> Assuming a truly random chance of 50%, one could end up in a situation like the above, which would result in a sample rate of ~85% rather than the expected 50%.

> To get around this, and instead get a more closer rate to the desired 50%, customers can use the [`logger.refresh_sample_rate_calculation()`](https://docs.powertools.aws.dev/lambda/typescript/latest/api/classes/_aws-lambda-powertools_logger.index.Logger.html#refreshsampleratecalculation) method mentioned above at the start/end of each request. 

```python

logger = Logger(sampling_rate=0.5)   # whether or not the log level is switched to `debug` is decided here only

def handler(event, context):
    # ... your logic here
    logger.refresh_sample_rate_calculation()
    pass
```

> When called, this method essentially flips the coin again and decides whether the log level should be switched to `debug` or not. Because this is done at the request level, statistically speaking, the ratio of sampled requests should be much closer to the desired one:

> Image

> With this in mind, we should consider easing this situation for customers by adding an optional flag to our class method decorator and Middy.js middleware so that when this flag is enabled, we'll call the `logger.refresh_sample_rate_calculation()` method for them at the end of each request, as proposed [here](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3278#issuecomment-2497752384).

> The flag would be `false` by default to maintain backward compatibility, although in a future major version we could consider making it enabled by default since this would be a much accurate behavior than the current one. 

> Obviously, as mentioned above, this would work only if we're able to wrap the handler, so customers who are not using either of the two mechanisms just mentioned would have to continue calling the `logger.refresh_sample_rate_calculation()` manually.

### Solution/User Experience

We want to have this experience:

1 - Customers will continue to set `sampling_rate` at the constructor level.

2 - Customers using the `@logger.inject_lambda_context` decorator will observe the sampling rate being recalculated on every request and having the expected result.

3 - Customers not using the decorator must call `refresh_sample_rate_calculation` manually.

4 - Customers not using the decorator or the `refresh_sample_rate_calculation` method will end up with unexpected sampling rates/logs.

5 - We need to change our documentation to make it more clear.

### Alternative solutions

```markdown

```

### Acknowledgment

leandrodamascena:
**Issue number:** #6141 

## Summary

This PR improves the log sampling rate calculation mechanism in our Logger utility. It aims to provide a more consistent and expected behavior for customers using log sampling, especially when reusing Lambda environments.

### Changes
1. Introduced a new method `refresh_sample_rate_calculation()` in the Logger class.
2. Updated the `@logger.inject_lambda_context` decorator to call `refresh_sample_rate_calculation()` at the end of each invocation.
3. Modified the internal sampling logic to use the refreshed calculation.

### Expected Behavior
1. Customers will continue to set `sampling_rate` at the constructor level.
2. When using `@logger.inject_lambda_context`, the sampling rate will be recalculated on every request, providing the expected sampling behavior.
3. Customers not using the decorator can manually call `refresh_sample_rate_calculation()` to achieve the same effect.
4. If neither the decorator nor manual refresh is used, sampling behavior may be unexpected.

### Backwards Compatibility
This change is backwards compatible. Existing code will continue to work, but may benefit from adopting the new best practices.

### Missing
- [ ] Update the documentation

#### Using decorator

```python

logger = Logger(level="INFO", sampling_rate=0.5)

@logger.inject_lambda_context
    logger.debug("test") # expected to be flushed in 50% of the requests
    pass
```

#### Using manual method

```python

logger = Logger(level="INFO", sampling_rate=0.5)

    logger.debug("test") # expected to be flushed in 50% of the requests
    logger.refresh_sample_rate_calculation()
    pass
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


anafalcao (dismissed):
APPROVED

leandrodamascena:
Hey @dreamorosi! Pls I'd like a review in the documentation part. Thanks

docs/core/logger.md
```diff
@@ -627,24 +627,27 @@ Notice in the CloudWatch Logs output how `payment_id` appears as expected when l
 
 ### Sampling debug logs
 
-Use sampling when you want to dynamically change your log level to **DEBUG** based on a **percentage of your concurrent/cold start invocations**.
+Use sampling when you want to dynamically change your log level to **DEBUG** based on a **percentage of the Lambda function executions**.
```
dreamorosi: I would keep `invocations` or `requests` instead of `executions`

docs/core/logger.md
```diff
@@ -627,24 +627,27 @@ Notice in the CloudWatch Logs output how `payment_id` appears as expected when l
...
 
-You can use values ranging from `0.0` to `1` (100%) when setting `POWERTOOLS_LOGGER_SAMPLE_RATE` env var, or `sample_rate` parameter in Logger.
+You can use values ranging from `0.0` to `1` (100%) when setting `POWERTOOLS_LOGGER_SAMPLE_RATE` env var, or `sampling_rate` parameter in Logger.
 
 ???+ tip "Tip: When is this useful?"
-    Let's imagine a sudden spike increase in concurrency triggered a transient issue downstream. When looking into the logs you might not have enough information, and while you can adjust log levels it might not happen again.
+    Log sampling allows you to capture debug information for a fraction of your requests, helping you diagnose rare or intermittent issues without drowning your logs in unnecessary details.
```
dreamorosi: This sentence `without drowning your logs in unnecessary details`, I would change it to either:
- `without flooding your logs with unnecessary details`
- `without increasing the overall verbosity of your logs`

docs/core/logger.md
```diff
@@ -627,24 +627,27 @@ Notice in the CloudWatch Logs output how `payment_id` appears as expected when l
...
 
-    This feature takes into account transient issues where additional debugging information can be useful.
+    Example: Imagine an e-commerce checkout process where you want to understand rare payment gateway errors. With 10% sampling, you'll log detailed information for a small subset of transactions, making troubleshooting easier without generating excessive logs.
 
-Sampling decision happens at the Logger initialization. This means sampling may happen significantly more or less than depending on your traffic patterns, for example a steady low number of invocations and thus few cold starts.
+Sampling decision happens at every execution when using `@logger.inject_lambda_context` decorator or `refresh_sample_rate_calculation` method. If you don't use any of both, you may end up with undesired sampling behavior.
```
dreamorosi: I think I'd consider rewording this to make it clear that:
- sampling decision happens automatically at every request when using the decorator
- when not using the decorator, you're in charge of refreshing it via the method

dreamorosi:
cc @am29d - FYI we can lift the doc updates for the feature in our own PR.

leandrodamascena:
Thanks @dreamorosi! I've incorporated all your suggestions.

docs/core/logger.md
leandrodamascena: Accepted.

docs/core/logger.md
leandrodamascena: Accepted.

docs/core/logger.md
leandrodamascena: Accepted.

dreamorosi (dismissed):
Thanks for working on this and addressing the comments.

destinydick:
great work

leandrodamascena:
Hey @dreamorosi! I need your review here again.

dreamorosi (approved):

PR #6324, opened 2025-03-21
feat(general): make logger, tracer and metrics utilities aware of provisioned concurrency

### LINKED ISSUE #6321: Feature request: make cold start aware of Provisioned Concurrency
### Use case

Original issue: https://github.com/aws-powertools/powertools-lambda-typescript/issues/3723

When using Provisioned Concurrency (PC) in AWS Lambda, the service initializes and keeps active a number of execution environments so that a portion of the requests can be served by these already-warm environments.

As a Powertools for AWS customer using PC, I'd like the cold start metric, log key, and tracer annotation to correctly reflect whether a request is a cold start or not - which is not currently the case.

### Solution/User Experience

Currently Logger, Metrics, and Tracer are not aware of Provisioned Concurrency (PC) because the state required to report cold starts in all of them is maintained internally and is dependent on the amount of requests seen by the utility.

Lambda provides an environment variable `AWS_LAMBDA_INITIALIZATION_TYPE` that can have three values in the Python runtime: `provisioned-concurrency`, `snap-start` or `on-demand`.

We could use this value to tweak our cold start detection logic to report a non-cold start (aka warm start) whenever PC is enabled.

This should not require any code changes from customers besides upgrading the packages to the latest version.

### Alternative solutions

```markdown

```

### Acknowledgment

leandrodamascena:
**Issue number:** #6321 

### Changes

This PR improves cold start detection in AWS Powertools for Python by leveraging the `AWS_LAMBDA_INITIALIZATION_TYPE` environment variable to accurately identify function initialization states across Provisioned Concurrency (PC) environments.

There are some considerations in this PR:

- We currently rely on a global variable when importing a specific module, which creates challenges for cold start detection across modules. To address this properly, we would need to refactor the code to use a single function that can control cold start logic across all modules. However, this would involve significant work and I'd need to invest too much time seeing the impact. To avoid this risk, I will temporarily duplicate code across modules to ensure compatibility and prevent disruptions.

- The cold start detection logic for Metrics is even more problematic because it relies on legacy code that has been in place since Powertools was first released. In the [#2194], we attempted to refactor this logic but realized that customers are directly importing certain variables, and changing them could break existing implementations. As a result, adding tests for Metrics (specific for PC) would require a major refactor, which I don't want to undertake at this time. For now, I will maintain the current implementation to avoid introducing risks while ensuring stability for customers.

- I'm happy to move forward with this PR, even with these limitations.

- I'm adding a new item to the v4 discussion to track these items

### User experience

Doesn't change the customer experience.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena:
Hey @dreamorosi! I'd like your review here. Please read the PR body.

aws_lambda_powertools/logging/logger.py
```diff
@@ -64,14 +64,22 @@ def _is_cold_start() -> bool:
...
-
     global is_cold_start
-    if is_cold_start:
-        cold_start = is_cold_start
+
+    initialization_type = os.getenv(constants.LAMBDA_INITIALIZATION_TYPE)
+
+    # Check for Provisioned Concurrency environment
+    # AWS_LAMBDA_INITIALIZATION_TYPE is set when using Provisioned Concurrency
+    if initialization_type == "provisioned-concurrency":
```
dreamorosi, line 73: If I understand this correctly, we'll return false by default only when in PC, and use the original logic in all other cases.

Why not do the other way around and check explicitly for `on-demand`, and if in that mode use the original logic, else default to `False`?

This way we could also cover SnapStart, as well as other compute platforms (i.e. if running on container)

dreamorosi (commented):
Agree with the overall strategy of postponing the major refactor, I only have a question about the actual logic before we move forward.

After we've clarified that I can approve

aws_lambda_powertools/logging/logger.py
leandrodamascena: > If I understand this correctly, we'll return false by default only when in PC, and use the original logic in all other cases.

Yeah, you're right.

> Why not do the other way around and check explicitly for on-demand, and if in that mode use the original logic, else default to False?
>
> This way we could also cover SnapStart, as well as other compute platforms (i.e. if running on container)

I don't agree with that for some reasons:

**on-demand mode:** AWS Lambda creates a new sandbox from scratch whenever a request requires a new container. After the sandbox is created, the Lambda handler is executed, meaning that cold starts can occur due to the initialization phase - `INIT PHASE`.

**snap-start mode:** AWS Lambda takes a snapshot of the execution environment when creating a new version and stores it. Unlike Provisioned Concurrency, SnapStart does not keep an instance running; instead, it restores the container from the snapshot when a new request arrives. This improves request processing time and introduces a `RESTORE TIME`, which is essentially equivalent to the `INIT PHASE` but uses the snapshot. Cold starts can still occur in this mode as well.

**provisioned-concurrency mode**: is the only scenario where cold starts do not occur, as the container is pre-warmed and kept ready to handle incoming requests.

**Other Compute Platforms**: Cold starts are not typically a concern when running outside of AWS Lambda because containers in other compute environments, such as ECS, EKS, or traditional server-based architectures, are designed to be long-lived rather than ephemeral. These platforms maintain persistent containers or instances, avoiding the need to initialize a new environment for each concurrent request.

Additionally, injecting cold start information into metrics and logs is inherently tied to the Lambda execution model. For example, in Python, the only way to capture cold start is using a decorator on the Lambda handler. However, when running outside of Lambda, the concept of an event and context object may not exist or be applicable. Our tracer utility doesn't run outside of Lambda because the root segment thing.

leandrodamascena:
> Agree with the overall strategy of postponing the major refactor, I only have a question about the actual logic before we move forward.
> 
> After we've clarified that I can approve

Already replied that. Please let me know if you have any additional question.

dreamorosi (approved):

aws_lambda_powertools/logging/logger.py
dreamorosi: ok, makes sense.

Issue #3473, opened 2023-12-08
Feature request: Add support to mask/encrypt/decrypt Pydantic models, Dataclasses, and standard Python classes in the DataMasking utility


leandrodamascena:
### Use case

Currently, the DataMasking utility only supports operations with traversable types in Python, for example: Lists, Dict, str, and others. It is a limitation for customers who want to integrate DataMasking utility with their existing Pydantic models, data classes, or standard Python classes.

### Solution/User Experience

Add support for mask, encrypt and decrypt Pydantic models, Dataclasses, and standard Python classes.

### Acknowledgment

- [X] This feature request meets [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

rubenfonseca:
We've added this to our backlog, and we intend to work on this early next year.

dreamorosi:
Hey @leandrodamascena, whenever you have time this week, could you please leave a comment with some more details about what needs to be done / implemented as part of this issue?

This will help potential contributors to orient themselves.

leandrodamascena:
Hey everyone! Our current implementation supports dict and lists as they can be directly converted. However, we recognize that customers may want to use more complex data structures such as Pydantic Models, DataClasses, or custom Python classes.

To accommodate these cases, we need to implement a pre-processing step to prepare the data before submitting it for erase, encrypt, or decrypt. This solution addresses the input data challenge, but it's important to note that we cannot guarantee the data will remain a valid instance of the original Pydantic model after processing, for example.

The encryption/erase process may alter the data type in ways that break the model's validation rules. For example:

```python
from __future__ import annotations

from aws_lambda_powertools.utilities.data_masking import DataMasking

from pydantic import BaseModel

class MyModel(BaseModel):
    name: str
    age: int


data_masker = DataMasking()

data = MyModel(name="powertools", age=5)

erased = data_masker.erase(data, fields=["age"])  

print(erased)
# output: {'name': 'powertools', 'age': '*****'}
```

Not that `age` is now a string and no longer an int, breaking Pydantic's validation model.

[...]

PR #6412, opened 2025-04-06
Feat/masking input types support

VatsalGoel3:
**Issue number:** #3473

### Changes

This PR adds support to the DataMasking utility to handle complex Python input types such as:

- **Pydantic models**
- **Dataclasses**
- **Standard Python classes with `.dict()` method**

To support this, a new `prepare_data` function was introduced, which performs type introspection and converts the input data into a dictionary before processing.

This function is now invoked at the beginning of the `erase`, `encrypt`, and `decrypt` methods, allowing these methods to seamlessly accept structured objects in addition to primitive types like `dict`, `str`, `list`, etc.

### User experience

**Before:**
```python

class MyModel(BaseModel):
    name: str
    age: int

data = MyModel(name="powertools", age=5)
masker = DataMasking()
masked = masker.erase(data, fields=["age"])  # ❌ This raised errors or did not work
```

**After:**
```python
# ✅ Now works correctly and returns: {'name': 'powertools', 'age': '*****'}
masked = masker.erase(data, fields=["age"])
```

This allows customers to use the utility directly with modern application architectures that use type-safe data structures.

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**: N/A

Checklist:

* [x] Implement warnings (if it can live side by side)

</details>

## Acknowledgment


VatsalGoel3:
@leandrodamascena, for now I applied the prepare_data() function as you suggested in the Issue, but I have an idea for making the function more robust and covering more edge cases, it would be like

    """
    Recursively convert complex objects into dictionaries (or simple types) so that they can be
    processed by the data masking utility. This function handles:
    
    - Dataclasses (using dataclasses.asdict)
    - Pydantic models (using model_dump)
    - Custom classes with a dict() method
    - Fallback to using __dict__ if available
    - Recursively traverses dicts, lists, tuples, and sets
    - Guards against circular references

    Parameters
    ----------
    data : Any
        The input data which may be a complex type.
    _visited : set, optional
        Internal set of visited object IDs to prevent infinite recursion on cyclic references.

    Returns
    -------
    Any
        A primitive type, or a recursively converted structure (dict, list, etc.)
    """

If that is more relevant, will implement this one.

PR #6564, opened 2025-04-25
feat(bedrock_agent): add new Amazon Bedrock Agents Functions Resolver

### LINKED ISSUE #6300: Feature request: Amazon Bedrock Agents Functions
### Use case

RFC: #6081  
Issue in TS: [aws-powertools/powertools-lambda-typescript/issues/3710](https://github.com/aws-powertools/powertools-lambda-typescript/issues/3710)

Bedrock Agents allows you to define action groups for your agents in two ways: [OpenAPI schemas](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-api-schema.html), and [direct function](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-action-function.html) integration. This issue focuses on the latter.

As a customer I can create Bedrock Agents that have tools at their disposal. These tools, or functions, can be defined as AWS Lambda functions. One Lambda function can hold one or more tools and when looked at together, they are what makes an [action group](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-action-create.html).

When I build a Lambda function with multiple tools in it, I am responsible for parsing the payload sent by Bedrock and, based on certain fields, call the corresponding tool in my code (aka the tool use). The response of this tool use is then returned by my Lambda function handler according to a specific format that Bedrock expects.

This can result in some degree of boilerplate code that I have to repeat for each action group, specifically:

- parsing/validating the incoming Bedrock Agent request payload
- handling the event using the correct tool/function
- building the response according to the response Bedrock Agent payload schema

**Example**
```python
def get_current_time():
    ...

def greet_user(name):
    ...

def simple_calculator(a, b, operation):
    ...

    agent = event['agent']
    actionGroup = event['actionGroup']
    function = event['function']
    parameters = event.get('parameters', [])

    # Execute the appropriate function based on the 'function' parameter
    if function == 'get_current_time':
        result = get_current_time()
    elif function == 'greet_user':
        name = parameters[0]['value'] if parameters else "User"
        result = greet_user(name)
    elif function == 'simple_calculator':
        if len(parameters) >= 3:
            a = float(parameters[0]['value'])
            b = float(parameters[1]['value'])
            operation = parameters[2]['value']
            result = simple_calculator(a, b, operation)
        else:
            result = "Error: Insufficient parameters for simple_calculator"
    else:
        result = f"Unknown function: {function}"

    responseBody = {
        "TEXT": {
            "body": str(result)
        }
    }

    action_response = {
        'actionGroup': actionGroup,
        'function': function,
        'functionResponse': {
            'responseBody': responseBody
        }
    }

    function_response = {'response': action_response, 'messageVersion': event['messageVersion']}
    print("Response: {}".format(function_response))

    return function_response

```

As a customer I would like to abstract all of that and instead focus primarily on building the tools for my agents, which are the only part of this that is specific to my business.

### Solution/User Experience

When paired with a Lambda function via action group, Bedrock sends and expects payloads of known shapes.

**Payload example**
```json
{
    "messageVersion": "1.0",
    "agent": {
        "alias": "PROD",
        "name": "hr-assistant-function-def",
        "version": "1",
        "id": "1234abcd-56ef-78gh-90ij-klmn12345678"
    },
    "sessionId": "87654321-abcd-efgh-ijkl-mnop12345678",
    "sessionAttributes": {
        "employeeId": "EMP123",
        "department": "Engineering"
    },
    "promptSessionAttributes": {
        "lastInteraction": "2024-02-01T15:30:00Z",
        "requestType": "vacation"
    },
    "inputText": "I want to request vacation from March 15 to March 20",
    "actionGroup": "VacationsActionGroup",
    "function": "submitVacationRequest",
    "parameters": [{
        "employeeId": "EMP123",
        "startDate": "2024-03-15",
        "endDate": "2024-03-20",
        "vacationType": "annual"
    }]
}
```
[Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-lambda.html#agents-lambda-input)

**Response payload example**
```json
{
   "response":{
      "actionGroup":"SmarterAgentActionGroup",
      "function":"submitVacationRequest",
      "functionResponse":{
         "responseBody":{
            "TEXT":{
               "body":"Your vacation was scheduled!"
            }
         }
      }
   },
   "messageVersion":"1.0"
}
```
[Documentation](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-lambda.html#agents-lambda-response)

Since the input event includes both the function and parameters fields, we can abstract most/all the boilerplate and provide a more streamlined experience.

We could implement a `BedrockAgentFunctionResolver` resolver that provides a structured way to register functions, resolve function calls, and handle requests within Lambda.

```python
from aws_lambda_powertools.event_handler import BedrockAgentFunctionResolver

app = BedrockAgentFunctionResolver()

@app.tool(name="currentTime", description="Gets the current UTC time")
def get_current_time():
    ... 
@app.tool(description="Greets the user using name")
def greet_user(name):
...
@app.tool(description="Add two numbers")
def simple_calculator(a, b, operation):
...
def lambda_handler(event: dict, context: LambdaContext) → dict:
```

Similar to the experience for [Bedrock Agents with OpenAPI](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/bedrock_agents/), we need a decorator `tool` that have `name` parameter as `Optional`. If customers want to make this more resilient to refactoring or want to have special naming conventions they can optionally specify a `name` parameter and in that case that takes precedence over the actual function name. 

The implementation above allows customers to focus on defining and implementing the tools within the action group rather than the undifferentiated boilerplate required to parse an event, resolve which tool to use, call it, and build the response payload.

Finally, customers can also access the current request `event` in the tool definitions, as `app.current_event`, ang get information like `session_id` or `prompt_session_attributes ` similar to [Bedrock Agents OpenAPI - accessing custom request fields](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/bedrock_agents/#accessing-custom-request-fields)

### Alternative solutions

See this discussion for considerations on alternative solutions: [#6081](https://github.com/aws-powertools/powertools-lambda-python/discussions/6081)

### Acknowledgment

anafalcao:
**Issue number:** #6300

## Summary
This PR introduces support for Bedrock Agents Functions, providing the user a similar experience from the current [Bedrock Agents with OpenAPI Resolver](https://docs.powertools.aws.dev/lambda/python/latest/core/event_handler/bedrock_agents/#key-features). 

### Changes

- [x] Bedrock Agent Functions data class
- [x] BedrockFunctionsResponseBuilder
- [x] BedrockAgentFunctionResolver
- [x] Optional response fields - sessionAttributes, promptSessionAttributes, knowledgeBasesConfiguration
- [x] Parser
- [x] Additional validations
- [x] name on Tool decorator
- [x] Documentation with examples

### User experience

```python
from aws_lambda_powertools.event_handler import BedrockAgentFunctionResolver

app = BedrockAgentFunctionResolver()

@app.tool(name="currentTime", description="Gets the current UTC time")
def get_current_time():
    ... 
@app.tool(description="Greets the user using name")
def greet_user(name):
...
@app.tool(description="Add two numbers")
def simple_calculator(a, b, operation):
...
def lambda_handler(event: dict, context: LambdaContext) → dict:
```

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,204 @@
...
+                "function": event.function,
+                "functionResponse": {"responseBody": {"TEXT": {"body": str(body if body is not None else "")}}},
+            },
+        }
+
+        # Add responseState if it's an error
+        if self.status_code >= 400:
+            response["response"]["functionResponse"]["responseState"] = (
+                ResponseState.REPROMPT.value if self.status_code == 400 else ResponseState.FAILURE.value
+            )
```
leandrodamascena: Hi @anafalcao! I think we need to spend some time investigating the behavior of `REPROMPT` and `FAILURE` modes. I don't know if we can make some assumption based on the http_code response code or if we let the customer handle it. To me it makes more sense for clients to make their own choices.

anafalcao: Hi @leandrodamascena!
I did some tests with it, and as responseState is context-specific and not always required, I believe it’s best making responseState an optional field in BedrockFunctionResponse, rather than enforcing automatic behavior that may not align with their intent.
For example:
	•	User can "REPROMPT" when a parameter is present but invalid (e.g.  division by zero). 
	•	User can "FAILURE" when there’s an external error (e.g., failed call to a third-party API) 

Basic example without Powertools:
```python
    try:
        a = next((float(param['value']) for param in parameters if param['name'] == 'a'), None)
        b = next((float(param['value']) for param in parameters if param['name'] == 'b'), None)

        if a is None or b is None:
            return {
                'responseState': 'REPROMPT',
                'responseBody': {
                    'TEXT': {
                        'body': 'Both parameters "a" and "b" are required. Please provide them.'
                    }
                }
            }
            
    except ValueError:
        return {
            'responseState': 'FAILURE',
            'responseBody': {
                'TEXT': {
                    'body': 'Invalid input. Parameters "a" and "b" must be numbers.'
                }
            }
        }
```

leandrodamascena: Hi @anafalcao! Thanks a lot for looking into this. So you suggest we have a `responseState` field in the Response class and let the customer decide what it wants to do, right? I believe this is the right direction because the customer will have full control over the workflow and decide what reports to Bedrock.

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,188 @@
...
+            prompt_session_attributes = None
+            knowledge_bases = None
+            response_state = None
+
+        response: dict[str, Any] = {
+            "messageVersion": "1.0",
+            "response": {
+                "actionGroup": event.action_group,
+                "function": event.function,
+                "functionResponse": {"responseBody": {"TEXT": {"body": str(body if body is not None else "")}}},
```
leandrodamascena: It seems that `respondeBody` expects a dynamic MIME type instead of hard-coded "TEXT". Do you know what happens when you change this and return a json, for example?

image

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,188 @@
...
+    def __init__(
+        self,
+        body: Any = None,
+        session_attributes: dict[str, str] | None = None,
+        prompt_session_attributes: dict[str, str] | None = None,
+        knowledge_bases: list[dict[str, Any]] | None = None,
+        response_state: str | None = None,
+    ) -> None:
+        if response_state is not None and response_state not in ["FAILURE", "REPROMPT"]:
+            raise ValueError("responseState must be None, 'FAILURE' or 'REPROMPT'")
```
leandrodamascena:
```suggestion
            raise ValueError("responseState must be 'FAILURE' or 'REPROMPT'")
```

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,188 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        body: Any = None,
+        session_attributes: dict[str, str] | None = None,
+        prompt_session_attributes: dict[str, str] | None = None,
+        knowledge_bases: list[dict[str, Any]] | None = None,
+        response_state: str | None = None,
```
leandrodamascena: I think it makes more sense this field be a Literal of FAILURE AND REPROMPT and then give the customer a type hint in their IDE.

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,188 @@
...
+        ----------
+        description : str | None
+            Description of what the tool does
+        name : str | None
+            Custom name for the tool. If not provided, uses the function name
+        """
+
+        def decorator(func: Callable) -> Callable:
+            if not description:
+                raise ValueError("Tool description is required")
```
leandrodamascena: The `description` parameter can be "None", but is it mandatory here? Is the `description` mandatory to register a route? If not, I think we should not check this here.

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,188 @@
...
+
+            function_name = name or func.__name__
+            if function_name in self._tools:
+                raise ValueError(f"Tool '{function_name}' already registered")
```
leandrodamascena: Hmm I don't think we should raise here, but instead warning and register the most recent route.

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def _resolve(self) -> dict[str, Any]:
+        """Internal resolution logic"""
+        if self.current_event is None:
+            raise ValueError("No event to process")
+
+        function_name = self.current_event.function
+
+        try:
+            result = self._tools[function_name]["function"]()
```
leandrodamascena: A critical part of this agent's workflow is passing parameters to functions. These parameters contain values ​​that will be useful for the function in question. In this [discussion](https://github.com/aws-powertools/powertools-lambda-python/discussions/6081#discussion-7948767), we talk about this:

**Support for injecting parameters into the function signature.**

Can you pls add this support?

leandrodamascena (changes requested):
Hello @anafalcao! I finished the first review and we need to fix the parameters part before proceeding with another review. Can you please work on that?

aws_lambda_powertools/event_handler/bedrock_agent_function.py
anafalcao: Currently only "TEXT" is supported as the responseBody content type, added this as a comment so it's clear for the users.
https://docs.aws.amazon.com/bedrock/latest/userguide/agents-lambda.html

aws_lambda_powertools/event_handler/bedrock_agent_function.py
anafalcao: You're right, removed the requirement here! Thanks

aws_lambda_powertools/event_handler/bedrock_agent_function.py
anafalcao: Changed to a PowertoolsWarning

anafalcao:
> Hello @anafalcao! I finished the first review and we need to fix the parameters part before proceeding with another review. Can you please work on that?

Hey @leandrodamascena ! Thank you so much for all the feedbacks, I pushed a new commit addressing them.

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,216 @@
...
+            self.clear_context()
+
+            # Build and return the response
+            return BedrockFunctionsResponseBuilder(result).build(self.current_event)
+        except Exception as error:
+            # Return a formatted error response
+            error_response = BedrockFunctionResponse(body=f"Error: {str(error)}", response_state="FAILURE")
+            return BedrockFunctionsResponseBuilder(error_response).build(self.current_event)
+
+    def append_context(self, **additional_context):
```
dreamorosi, line 242: Are we ever calling this method to append the context?

aws_lambda_powertools/event_handler/bedrock_agent_function.py
```diff
@@ -0,0 +1,216 @@
...
+        function_name = self.current_event.function
+
+        try:
+            # Extract parameters from the event
+            parameters = {param.name: param.value for param in getattr(self.current_event, "parameters", [])}
+
+            func = self._tools[function_name]["function"]
+            # Filter parameters to only include those expected by the function
+            sig = inspect.signature(func)
+            valid_params = {name: value for name, value in parameters.items() if name in sig.parameters}
```
dreamorosi, line 227: We can't parse `array` types, but we can "trust" `number`, `integer`, `boolean` and `string` types and potentially cast them to the correct type before passing them to the handler.

Otherwise customers will always get strings.

docs/core/event_handler/bedrock_agents.md
```diff
@@ -26,72 +30,53 @@ Create [Agents for Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/us
 
 **[OpenAPI schema](https://www.openapis.org/){target="_blank"}** is an industry standard JSON-serialized string that represents the structure and parameters of your API.
 
+**Function details** consist of a list of parameters, defined by their name, [data type](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_agent_ParameterDetail.html), and whether they are required. The agent uses these configurations to determine what information it needs to elicit from the user.
+
 **Action group** is a collection of two resources where you define the actions that the agent should carry out: an OpenAPI schema to define the APIs that the agent can invoke to carry out its tasks, and a Lambda function to execute those actions.
```
dreamorosi: This could be confusing, with the new feature an action group is only a function and its tool definitions, without OpenAPI schema

docs/core/event_handler/bedrock_agents.md
```diff
@@ -26,72 +30,53 @@ Create [Agents for Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/us
...
-=== "Using AWS Serverless Application Model (SAM)"
-	Using [AWS SAM](https://aws.amazon.com/serverless/sam/){target="_blank"} you can create your Lambda function and the necessary permissions. However, you still have to create your Agent for Amazon Bedrock [using the AWS console](https://docs.aws.amazon.com/bedrock/latest/userguide/agents-create.html){target="_blank"}.
+An action group defines actions that the agent can help the user perform. You can define action groups as OpenAPI-based or Function-based.
 
-	```yaml hl_lines="18 26 34 61"
-	--8<-- "examples/event_handler_bedrock_agents/sam/template.yaml"
-	```
+| Aspect               | OpenAPI-based Actions                                           | Function-based Actions                                          |
+|----------------------|------------------------------------------------------------------|----------------------------------------------------------------|
+| Definition Style      | `@app.get("/path", description="")`<br>`@app.post("/path", description="")` | `@app.tool(name="")`                                            |
```
dreamorosi: Let's add `description` to the example of the tool as well, LLMs have higher chances to understand when to use a tool if we provide descriptions.

docs/core/event_handler/bedrock_agents.md
```diff
@@ -26,72 +30,53 @@ Create [Agents for Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/us
...
+| Parameter Handling    | Path, query, and body parameters                                 | Function parameters                                             |
+| Use Case              | REST-like APIs, complex request/response structures              | Direct function calls, simpler input/output                     |
```
dreamorosi: I think `simpler input` is probably the most correct.

Based on tests Bedrock is able to work with complex outputs just fine even with the function-based resolver, there's just not a schema but it does fine.

docs/core/event_handler/bedrock_agents.md
```diff
@@ -26,72 +30,53 @@ Create [Agents for Amazon Bedrock](https://docs.aws.amazon.com/bedrock/latest/us
...
+| Response object       | Via `BedrockResponse`                                            | Via `BedrockFunctionResponse`                                   |
+| Best For              | - Complex APIs with multiple endpoints<br>- When OpenAPI spec is required<br>- Integration with existing REST APIs | - Simple function-based actions<br>- Direct LLM-to-function mapping<br>- When function descriptions are sufficient |
```
dreamorosi: I'd remove this `When function descriptions are sufficient`.

In practice, you still need to configure Bedrock and specify the tools, parameters, and description - it just happens in the Agent config rather than in the function's code.

docs/core/event_handler/bedrock_agents.md
```diff
@@ -114,79 +99,74 @@ You are required to add a `description` parameter in each endpoint, doing so wil
...
 === "Output payload"
 
-	```json hl_lines="10"
-	--8<-- "examples/event_handler_bedrock_agents/src/getting_started_with_validation_output.json"
+	```json hl_lines="9"
+	--8<-- "examples/event_handler_bedrock_agents/src/getting_started_output_func.json"
 	```
 
-#### When validation fails
+??? note "What happens under the hood?"
```
dreamorosi: Based on the content of the mermaid diagram, this should probably be under the OpenAPI section since it mentions both the OpenAPI schema and validation, which don't happen in this new implementation.

docs/core/event_handler/bedrock_agents.md
```diff
@@ -114,79 +99,74 @@ You are required to add a `description` parameter in each endpoint, doing so wil
...
-
-=== "OpenAPI schema"
+    ```python hl_lines="4 9 12 21"
+    --8<-- "examples/event_handler_bedrock_agents/src/getting_started_functions.py"
+    ```
 
-	```json
-	--8<-- "examples/event_handler_bedrock_agents/src/getting_started_with_validation_schema.json"
-	```
+	1. `name` and `description` are optional here.
```
dreamorosi, line 109: I know in the RFC I was the one arguing for making these optional, but since then I learned a bit more about how LLMs and agents work and I think that `description` especially should be mandatory.

I mean technically it's not mandatory and the agent _might_ work, but best practices show defining tools with more exhaustive metadata brings better effectiveness, for example [this is an excerpt from Anthropic's guidance on agents](https://www.anthropic.com/engineering/building-effective-agents) (see Appendix 2):
> Tool definitions and specifications should be given just as much prompt engineering attention as your overall prompts.

and 

> Put yourself in the model's shoes. Is it obvious how to use this tool, based on the description and parameters, or would you need to think carefully about it? If so, then it’s probably also true for the model. A good tool definition often includes example usage, edge cases, input format requirements, and clear boundaries from other tools.
How can you change parameter names or descriptions to make things more obvious? Think of this as writing a great docstring for a junior developer on your team. This is especially important when using many similar tools.

With this in mind, it might better help our customers to make `description` mandatory, or at least strongly recommended.

dreamorosi (commented):
Left a few comments/questions

aws_lambda_powertools/event_handler/bedrock_agent_function.py
leandrodamascena: This is a function to allow customer to inject additional data - like dependency injection - to access this within the route (tool in this case).

aws_lambda_powertools/event_handler/bedrock_agent_function.py
leandrodamascena: Hmm make sense, I see your point. Ok, I'll cast them before passing to the function.

aws_lambda_powertools/event_handler/bedrock_agent_function.py
dreamorosi: Ok, but are we ever passing the `context` to the tool handler?

docs/core/event_handler/bedrock_agents.md
dreamorosi: I was wrong, it doesn't really matter if we pass it - what matters is how you declare it in your Bedrock Agent configuration - thank you @anafalcao.

docs/core/event_handler/bedrock_agents.md
dreamorosi: I was wrong, it doesn't really matter if we specify it here - what matters is how you declare it in your Bedrock Agent configuration - thank you @anafalcao.

dreamorosi (dismissed):

dreamorosi (approved):

leandrodamascena (approved):

PR #6794, opened 2025-06-11
fix(logger): fix exception on flush without buffer

### LINKED ISSUE #6793: Bug: Missing _buffer_config check before in flush_buffer
### Expected Behaviour

When `flush_buffer_on_uncaught_error` is `true` but no `buffer_config` is set, no `AttributeError` should be raised on an error.

### Current Behaviour

```
[ERROR] AttributeError: 'Logger' object has no attribute '_buffer_cache'
  File "/opt/python/aws_lambda_powertools/logging/logger.py", line 551, in decorate
    self.flush_buffer()
  File "/opt/python/aws_lambda_powertools/logging/logger.py", line 1205, in flush_buffer
    buffer = self._buffer_cache.get(tracer_id)
  File "/opt/python/aws_lambda_powertools/logging/logger.py", line 306, in __getattr__
    return getattr(self._logger, name)
```

### Code snippet

```python
from aws_lambda_powertools.logging import Logger

logger = Logger()

@logger.inject_lambda_context(flush_buffer_on_uncaught_error=True)
def handler(event, context):
    raise ValueError("foo")
```

### Possible Solution

like in other places, wrap self._buffer_cache.get like:

```python
if not self._buffer_config:
    return
```

### Steps to Reproduce

Deploy lambda and raise the exception.

latest

3.13

### Packaging format used

PyPi

### Debugging logs

```python

```

tonnico:
**Issue number:** https://github.com/aws-powertools/powertools-lambda-python/issues/6793

### Changes

earlier exit when no buffer config is set.

### User experience

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


leandrodamascena (changes requested):
Hey @tonnico thanks for sending this PR. Your fork doesn't allow modification, so, can you please add this test to make sure we are testing this situation?

https://github.com/tonnico/powertools-lambda-python/blob/c2dd26f297cfcb8834cc0ecd96f256259a5acd9d/tests/functional/logger/required_dependencies/test_powertools_logger_buffer.py

```python
def test_flush_buffer_without_buffer_config(stdout, service_name, lambda_context, monkeypatch):
    # Set initial trace ID for first Lambda invocation
    monkeypatch.setenv(constants.XRAY_TRACE_ID_ENV, "1-67c39786-5908a82a246fb67f3089263f")

    # GIVEN A logger without buffer configuration
    logger = Logger(level="DEBUG", service=service_name, stream=stdout)

    @logger.inject_lambda_context(flush_buffer_on_uncaught_error=True)
    def handler(event, context):
        # Log messages that should be flushed when an exception occurs
        logger.debug("this log line will be flushed after error - 1")
        logger.debug("this log line will be flushed after error - 2")
        raise ValueError("Test error")

    # WHEN Invoking the handler and expecting a ValueError
    # AND flush_buffer_on_uncaught_error is True but there is no logger buffer configuration
    with pytest.raises(ValueError):
        handler({}, lambda_context)

    # THEN Verify that buffered log messages are flushed without any exception
    log = capture_multiple_logging_statements_output(stdout)
    assert len(log) == 2, "Expected two log messages to be flushed"
```

tonnico:
> Hey @tonnico thanks for sending this PR. Your fork doesn't allow modification, so, can you please add this test to make sure we are testing this situation?

Hello @leandrodamascena. Thank you for the test. However, I have a slightly different view of the behaviour.

When no buffer configuration is set and the `flush_buffer_on_uncaught_error` parameter is set to True, nothing should be flushed because there is no buffer. Therefore, when the logger is set to `DEBUG` and a message is sent with `logger.debug`, I expect these messages to be sent immediately and not buffered.

I changed some comments and values and hope this is ok for you.

leandrodamascena:
> > Hey @tonnico thanks for sending this PR. Your fork doesn't allow modification, so, can you please add this test to make sure we are testing this situation?
> 
> Hello @leandrodamascena. Thank you for the test. However, I have a slightly different view of the behaviour.
> 
> When no buffer configuration is set and the `flush_buffer_on_uncaught_error` parameter is set to True, nothing should be flushed because there is no buffer. Therefore, when the logger is set to `DEBUG` and a message is sent with `logger.debug`, I expect these messages to be sent immediately and not buffered.
> 
> I changed some comments and values and hope this is ok for you.

Thanks for fixing that, I agree 100% with you! Just let the CI completes and I'll merge!!

leandrodamascena (approved):
APPROVED!

PR #6936, opened 2025-07-03
fix: enable response compression when using multi-value headers

### LINKED ISSUE #6912: Bug: REST response compression does not work when multiValueHeaders are enabled
### Expected Behaviour

Response compression is possible when "multiValueHeaders" are enabled.

### Current Behaviour

Compression does not work under any circumstances when "multiValueHeaders" are enabled.

### Code snippet

```python
import json
import unittest

app = ALBResolver()

@app.get("/poc", compress=True)
def test_poc():
    return {
        "statusCode": 200,
        "body": json.dumps({"foo": "bar"}),
    }

class TestLambdaHandler(unittest.TestCase):
    def setUp(self):
        self.event_base = {
            "httpMethod": "GET",
            "path": "/poc",
        }
        self.context = {}

    def test_lambda_handler_with_headers(self):
        event = self.event_base.copy()
        event["headers"] = {"accept-encoding": "gzip, deflate"}
        response = lambda_handler(event, self.context)
        assert "Content-Encoding" in response["headers"]

    def test_lambda_handler_with_multivalueheaders(self):
        event = self.event_base.copy()
        event["multiValueHeaders"] = {"accept-encoding": ["gzip", "deflate"]}
        response = lambda_handler(event, self.context)
        print(response)
        assert "multiValueHeaders" in response
        assert "headers" not in response
        assert "Content-Encoding" in response["multiValueHeaders"], response["multiValueHeaders"]

if __name__ == "__main__":
    unittest.main()

--
Result:

     assert "Content-Encoding" in response["multiValueHeaders"], response["multiValueHeaders"]
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: defaultdict(<class 'list'>, {'Content-Type': ['application/json']})
```

### Possible Solution

This seems to stem from the compressibility request header check only checking `event.headers`:
https://github.com/aws-powertools/powertools-lambda-python/blob/ed6d3d37c00e98775dff71e7b5b1e3387223a3ed/aws_lambda_powertools/event_handler/api_gateway.py#L919

`event.resolved_headers_field` is more generic, but requires some additional processing (e.g. "gzip, deflate" can be a single header value within the sequence of `event.multi_value_headers`).

### Steps to Reproduce

Code snippet above

latest

3.11

### Packaging format used

PyPi

### Debugging logs

```python

```

kimsappi:
**Issue number:** #6912

### Changes

Enabled REST response compression to work when multi-value headers are enabled for REST resolvers. Previously only the Lambda invocation event's `"headers"` field was considered, now `"multiValueHeaders"` is used when present, relying on generic Powertools functionality (`event.resolved_headers_field`).

### User experience

Previously response compression would not work if the Lambda function integration had multi-value headers enabled. Now it does.

## Checklist

* [ ] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -916,8 +916,14 @@ class ResponseBuilder(Generic[ResponseEventT]):
         bool
             True if compression is enabled and the "gzip" encoding is accepted, False otherwise.
         """
-        encoding = event.headers.get("accept-encoding", "")
-        if "gzip" in encoding:
+        encoding = event.resolved_headers_field.get("accept-encoding", "")
+        gzip_accepted = False
+        if isinstance(encoding, str):
+            gzip_accepted = "gzip" in encoding
+        elif isinstance(encoding, list):
+            gzip_accepted = "gzip" in ",".join(encoding)
+
+        if gzip_accepted:
             if response_compression is not None:
                 return response_compression  # e.g., Response(compress=False/True))
             if route_compression:
--- a/tests/functional/event_handler/_pydantic/test_openapi_swagger.py
+++ b/tests/functional/event_handler/_pydantic/test_openapi_swagger.py
@@ -26,7 +26,7 @@ def test_openapi_swagger():
 def test_openapi_swagger_compressed():
     app = APIGatewayRestResolver(enable_validation=True)
     app.enable_swagger(compress=True)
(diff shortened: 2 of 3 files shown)


leandrodamascena (approved):
Thanks a lot for this PR @kimsappi! I read your comment in the issue and I agree.

PR #7026, opened 2025-07-23
feat(metrics): add runtime validations for the metric name

### LINKED ISSUE #7010: Maintenance: Add runtime validation for the metric name
### Summary

While adding the runtime validations for the Typescript implementation of AWS Powertools Lambda Metrics, I noticed an issue when a metric name with either an empty string or with a string longer than 255 characters was used, the metrics were not being written. If this validation is not done in the runtime, the metrics will get dropped by Cloudwatch.

### Why is this needed?

As per the [MetricDatum API Reference](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html), the Metric Name has a length constraint of a minimum of length 1 and a maximum of length 255. Since, the runtime validation for the metric name is missing on the `add_metric` function in the `cloudwatch_emf` provider of the `Metric` utility. If we pass in an empty string or a string with more than 255 characters currently as the metric name, those metrics are being dropped by Cloudwatch silently.

Metrics

### Solution

1. Runtime Metric Name Type Validation

The Typescript implementation includes a runtime check to ensure that metric name is a string and the length constraint is satisfied. If an invalid value is passed, it raises a `RangeError`.

Typescript code:
```typescript
if (!isString(name)) throw new Error(`${name} is not a valid string`);
if (
  name.length < MIN_METRIC_NAME_LENGTH ||
  name.length > MAX_METRIC_NAME_LENGTH
)
  throw new RangeError(
    `The metric name should be between ${MIN_METRIC_NAME_LENGTH} and ${MAX_METRIC_NAME_LENGTH} characters`
  );
```

[Typescript Source Code](https://github.com/aws-powertools/powertools-lambda-typescript/blob/main/packages/metrics/src/Metrics.ts#L1081-L1088)

[Python Source - add_metric method](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L139-L151)

### Acknowledgment

sdangol:
**Issue number:** #7010 

## Summary

This PR adds runtime validation for the metric name. It ensures that the metric name meets the required constraints before they are added, providing better error handling and feedback to users. This was done to prevent potential issues that could have come up due to invalid metric name being passed to Cloudwatch.

### Changes

- Added runtime validation for the Metric Name: Validated according to the constraints in https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html
- Added constants for metric name length constraints (`MIN_METRIC_NAME_LENGTH` and `MAX_METRIC_NAME_LENGTH`)
- Added tests for the new validations

### User experience

Before this, for the users who were passing in an invalid metric name (empty string or a string of more than 255 characters), the metrics would get silently dropped. But, now the users will get an error.

## Checklist

* [X] Changes are documented

<details>

**RFC issue number**:

Checklist:

</details>

## Acknowledgment


tests/functional/metrics/required_dependencies/test_metrics_cloudwatch_emf.py
```diff
@@ -443,6 +446,38 @@ def test_schema_validation_no_namespace(metric, dimension):
             my_metric.add_dimension(**dimension)
 
 
+def test_schema_validation_empty_metric_name(metric, dimension, namespace):
```
leandrodamascena, line 449: Hey @sdangol all these 2 tests are working as expected, but codecov is complaining because we need to test the `EMFMetricProvider`, so, can you please make a small change and create a new instance of this provider?

Something like this:

`my_metrics = AmazonCloudWatchEMFProvider(namespace=namespace)`

Reference: https://github.com/aws-powertools/powertools-lambda-python/pull/7026/files#diff-53fb0e6eea8f9e23cdca06f0c22375183de0f1ddb845122e6fd73f25c9caa8f3R45

leandrodamascena (changes requested):
Hey @sdangol thanks a lot for submitting this PR!! I just left a small comment and we are good to merge.

tests/functional/metrics/required_dependencies/test_metrics_cloudwatch_emf.py
sdangol: Updated them to test the `AmazonCloudWatchEMFProvider`

leandrodamascena (approved):
Thanks a lot for working on this and addressing the feedback @sdangol!

PR #7058, opened 2025-07-29
feat(parameters): add support for retrieving batch of secrets

### LINKED ISSUE #4200: Feature request: Support for retrieving batch of secrets
### Use case

Have the same experience as SSM but Secrets Manager now that it's possible.

It was quietly launched recently: https://aws.amazon.com/about-aws/whats-new/2023/11/aws-secrets-manager-batch-retrieval-secrets/

### Solution/User Experience

Same experience as `get_parameters_by_name`, as each secret might have a different TTL associated with it: https://docs.powertools.aws.dev/lambda/python/latest/utilities/parameters/#getting_started_parameter_by_namepy

### Alternative solutions

_No response_

### Acknowledgment

leandrodamascena:
**Issue number:** #4200

### Changes

This pull request implements batch secrets retrieval functionality for AWS Secrets Manager using the `batch_get_secret_value` API. The implementation introduces a new `get_secrets_by_name()` function that provides a interface similar to the existing `get_parameters_by_name` function, allowing users to retrieve multiple secrets in a single API call for significant performance improvements and cost optimization.

### User experience

**Before this change:**

Users had to make individual API calls for each secret, resulting in poor performance and higher costs:

```python
# 3 separate API calls - slow and expensive
db_password = get_secret("database/password")
api_key = get_secret("api/key") 
jwt_secret = get_secret("jwt/secret")
```

**After this change:**
Users can retrieve multiple secrets in a single API call with the same familiar interface:

```python
# 1 API call for all secrets - fast and cost-effective
secrets = get_secrets_by_name([
    "database/password",
    "api/key", 
    "jwt/secret"
])
 
# With transforms and caching
secrets = get_secrets_by_name(
    names=["config-secret"],
    transform="json",
    max_age=600
)
```

## Checklist

* [x] Changes are documented

<details>

**RFC issue number**: N/A - This is a non-breaking additive feature

Checklist:
* [x] Migration process documented (No migration needed - existing code continues to work unchanged)
* [x] Implement warnings (No warnings needed - this is purely additive functionality)

</details>

## Acknowledgment


aws_lambda_powertools/utilities/parameters/secrets.py
```diff
@@ -126,11 +130,159 @@ def _get(self, name: str, **sdk_options) -> str | bytes:
...
+            When the parameter provider fails to retrieve secrets
+        TransformParameterError
+            When the parameter provider fails to transform a secret value
+        """
+        if not names:
+            raise GetSecretError("You must provide at least one secret name")
+
+        # Create a unique cache key for this batch of secrets
+        # Use sorted names to ensure consistent caching regardless of order
+        cache_key_name = "|".join(sorted(names))
```
dreamorosi, line 249: I'm wondering if it would make sense to cache the secrets individually rather than as a single group. Especially when working in more complex code bases where the secrets might be retrieved from different modules, having a more granular cache could be useful.

For example, assuming I'm retrieving `paramaters.get_secrets_by_name(names=["a", "b"])` instead of caching them together, each one would have its own cache key & entry. Thanks to this, if one is retrieved separately before/after it would benefit from caching instead of going to Secrets Manager multiple times.

With that said, I don't know if it's worth the effort at this stage - if it's too complex we can probably postpone it until there's demand.

leandrodamascena: I really would like to do it here, this is a great idea, but I see some considerations.

1/ I need to refactor this method to create a logic to remove from the array items that is already in cache and make `batch_get_secret_value` with the one that is not cached.

2/ I see some potential edge-cases with this change. Considering I'll use individual `SecretsName` as the cache key and the customer is running Lambda function in `us-east-1` and is using this to get secrets from sa-east-1:

```py
production_secrets = parameters.get_secrets_by_name(
        names=["app-secret"],
        Filters=[
            {"Key": "primary-region", "Values": ["sa-east-1"]},
            {"Key": "tag-value", "Values": ["production"]},
        ],
       max_age=60,
    )
```

And after that customer calls `parameters.get_secret("/app-secret", max_age=60,)`, the value will be returned from the cache, but the region will potentially be different, leading the customer to get wrong values. To solve this, I need to include filters in the cache key, but when using get_secret customers usually don't use filters.

Probably we can wait from customer demand to do this. What do you think?

dreamorosi: Maybe we should add the region as part of the cache key. This problem can happen even today if I call `get_secret` (or any of the utility functions really) in two different areas of the codebase.

I trust your judgment, if you think it's too complex to add now then we can do it in the future if there's demand.

leandrodamascena: I still think we can wait for customer demand. I can't assume the region will always be present in the `Filters` field, as the customer can use different keys such as tags, all, and others, and I always need to pass sdk_options/Filter as part of the key, and in this case the caching mechanism won't affect subsequent calls using `get_secret`.

Let's release it as is and see if clients actually need it.

leandrodamascena: Can you please review again?

dreamorosi: I think we could've done granular caching with the assumption that you only use one region - but ok. Will approve.

dreamorosi (approved):

PR #7199, opened 2025-08-18
docs(build_recipes): add cross build page

### LINKED ISSUE #7198: Docs: Create cross-platform page in the build recipes documentation

Customers are looking for comprehensive guidance on how to package Lambda functions with Powertools for AWS Lambda (Python) using different build tools and dependency managers. Currently, there's no dedicated documentation covering the various approaches and their trade-offs.

docs.powertools.aws.dev/lambda/python/latest

### How can we improve?

Add a new section to the new build recipes documentation with:

- Cross-platform build examples

### Got a suggestion in mind?

_No response_

### Acknowledgment

leandrodamascena:
**Issue number:** closes #7198 

### Changes

This PR creates the cross build page in the build recipes documentation.

### User experience

---

### DIFF
--- a/docs/build_recipes/cross-platform.md
+++ b/docs/build_recipes/cross-platform.md
@@ -1,6 +1,6 @@
 ---
 title: Cross-Platform Build Considerations
-description: Handle architecture differences when building Lambda packages
+description: Handle architecture differences when building AWS Lambda packages
 ---
 
 <!-- markdownlint-disable MD043 -->
@@ -8,7 +8,7 @@ description: Handle architecture differences when building Lambda packages
 Many modern Python packages include compiled extensions written in Rust or C/C++ for performance reasons. These compiled components are platform-specific and can cause deployment issues when building on different architectures.
 
 ???+ warning "Architecture Mismatch Issues"
-    Building Lambda packages on macOS (ARM64/Intel) for deployment on AWS Lambda (Linux x86_64 or ARM64) will result in incompatible binary dependencies that cause import errors at runtime.
+    Building AWS Lambda packages on macOS (ARM64/Intel) for deployment on AWS Lambda (Linux x86_64 or ARM64) will result in incompatible binary dependencies that cause import errors at runtime.
 
 ## Common compiled libraries
 
--- a/docs/build_recipes/cross-platform.md
+++ b/docs/build_recipes/cross-platform.md
@@ -116,17 +116,19 @@ pydantic-2.5.0-cp311-cp311-linux_x86_64.whl
 
 ## Lambda runtime environments
 
-Lambda runtimes use specific Amazon Linux versions with fixed GLIBC versions. Packages built on development machines with newer GLIBC versions will fail at runtime with import errors. Each Python runtime version corresponds to a specific Amazon Linux base system that determines compatible system library versions.
+<!-- markdownlint-disable MD013 -->
+Lambda managed runtimes use [specific Amazon Linux versions](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported) versions with fixed GLIBC versions. Packages built on development machines with newer GLIBC versions will fail at runtime with import errors. Each Python runtime version corresponds to a specific Amazon Linux base system that determines compatible system library versions.
+<!-- markdownlint-enable MD013 -->
 
 ### Amazon Linux versions and GLIBC compatibility
 
-| Python Runtime | Base System | GLIBC Version | Architecture Support | Status |
-|----------------|-------------|---------------|---------------------|---------|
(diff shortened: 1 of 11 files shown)


docs/build_recipes/cross-platform.md
```diff
@@ -0,0 +1,253 @@
...
+    ```
+
+=== "All extras (mixed dependencies)"
+
+    ```txt title="requirements.txt - Requires careful platform handling"
+    # These extras have minimal or no compiled dependencies
+    aws-lambda-powertools[tracer]==3.18.0      # aws-xray-sdk (mostly pure Python)
+    aws-lambda-powertools[aws-sdk]==3.18.0    # boto3 (pure Python)
+
+    ```
```
leandrodamascena, line 63: I added this example with code inline to avoid complaints in the scorecard that it's not using hashes. The reason it doesn't use hashes is because I want to make it clear to developers the difference between using extras when installing Powertools. Using hashes would completely negate the idea of making extras explicit to customers. The file could be something like this:

Image

leandrodamascena:
Hey @dreamorosi I need a review here. Please read this comment before the review: https://github.com/aws-powertools/powertools-lambda-python/pull/7199/files#r2282547105

docs/build_recipes/cross-platform.md
```diff
@@ -0,0 +1,253 @@
...
+title: Cross-Platform Build Considerations
+description: Handle architecture differences when building Lambda packages
+---
+
+<!-- markdownlint-disable MD043 -->
+
+Many modern Python packages include compiled extensions written in Rust or C/C++ for performance reasons. These compiled components are platform-specific and can cause deployment issues when building on different architectures.
+
+???+ warning "Architecture Mismatch Issues"
+    Building Lambda packages on macOS (ARM64/Intel) for deployment on AWS Lambda (Linux x86_64 or ARM64) will result in incompatible binary dependencies that cause import errors at runtime.
```
dreamorosi:
```suggestion
    Building AWS Lambda packages on macOS (ARM64/Intel) for deployment on Lambda (Linux x86_64 or ARM64) will result in incompatible binary dependencies that cause import errors at runtime.
```

docs/build_recipes/cross-platform.md
```diff
@@ -0,0 +1,253 @@
...
+**Why wheels matter for Lambda:**
+
+* **Consistent builds** - Same binary across environments
+* **Faster installs** - No compilation step required
+* **Predictable dependencies** - Known system library requirements
+* **Architecture safety** - Platform-specific binaries
+
+## Lambda runtime environments
+
+Lambda runtimes use specific Amazon Linux versions with fixed GLIBC versions. Packages built on development machines with newer GLIBC versions will fail at runtime with import errors. Each Python runtime version corresponds to a specific Amazon Linux base system that determines compatible system library versions.
```
dreamorosi:
```suggestion
Lambda managed runtimes use [specific Amazon Linux versions](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported) with fixed GLIBC versions. Packages built on development machines with newer GLIBC versions will fail at runtime with import errors. Each Python runtime version corresponds to a specific Amazon Linux base system that determines compatible system library versions.
```

docs/build_recipes/cross-platform.md
```diff
@@ -0,0 +1,253 @@
...
+
+### Amazon Linux versions and GLIBC compatibility
+
+| Python Runtime | Base System | GLIBC Version | Architecture Support | Status |
```
dreamorosi: I'm unsure about this table, I'd rather link to the [docs page](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported).

If we decide to keep it, I'd at least remove the `Status` column, since it doesn't add anything to the AL/GLIBC discussion and will save us some overhead to keep things up to date.

dreamorosi (changes requested):
Left a couple comments - but it's looking good.

docs/build_recipes/cross-platform.md
leandrodamascena: Good point about the status, I removed it. The reason I want to keep this table is because the Lambda documentation removed documentation about boto3 and other runtime libraries, for example, so I want to make this clear to developers. Even though I think it costs maintaining this page.

dreamorosi: I see your point, but you know why they removed it and the main reason is that we don't have an automated way of keeping it up to date. But I'm ok to leave it for now as long as it doesn't become a huge overhead.

dreamorosi (approved):

docs/build_recipes/cross-platform.md
leandrodamascena: This will be updated when a new runtime is added or when one is removed, so, we need to change it anyway.

PR #7246, opened 2025-08-26
fix(shared/functions): fix `slice_dictionary` implementation

### LINKED ISSUE #7252: Bug: slice_dictionary yields duplicate chunks instead of slicing dictionary correctly
### Expected Behaviour

slice_dictionary should yield consecutive chunks of dictionary items until the input dictionary is exhausted. Each chunk should contain up to chunk_size items without repetition.

### Current Behaviour

The slice_dictionary helper currently does not slice dictionaries into chunks properly.
Because itertools.islice(data, chunk_size) always starts from the beginning of the dictionary, each iteration yields the same first chunk_size keys, resulting in duplicate chunks.

### Code snippet

```python
import itertools

def slice_dictionary(data: dict, chunk_size: int):
    for _ in range(0, len(data), chunk_size):
        yield {dict_key: data[dict_key] for dict_key in itertools.islice(data, chunk_size)}

data = {"k0": 0, "k1": 1, "k2": 2}
# actual   [{'k0': 0, 'k1': 1}, {'k0': 0, 'k1': 1}]
# expected [{'k0': 0, 'k1': 1}, {'k2': 2}]
print(list(slice_dictionary(data, 2)))
```

### Possible Solution

_No response_

### Steps to Reproduce

Code snippet from above.

latest

3.9

### Packaging format used

PyPi

### Debugging logs

```python

```

Liam-DeVoe:
**Issue number:** #7252 

The implementation of `slice_dictionary` has a bug: it will keep repeating the first chunk, because the stop index `chunk_size` is not incremented as it iterates.

```python
import itertools

def slice_dictionary(data: dict, chunk_size: int):
    for _ in range(0, len(data), chunk_size):
        yield {dict_key: data[dict_key] for dict_key in itertools.islice(data, chunk_size)}

data = {"k0": 0, "k1": 1, "k2": 2}
# actual   [{'k0': 0, 'k1': 1}, {'k0': 0, 'k1': 1}]
# expected [{'k0': 0, 'k1': 1}, {'k2': 2}]
print(list(slice_dictionary(data, 2)))
```

fwiw: I found this bug during a research project I'm working on which uses LLMs to write property-based tests in [Hypothesis](https://github.com/HypothesisWorks/hypothesis). A property which found this bug was proposed and written "autonomously" by the agent. I wrote this PR myself and take full responsibility for it.

---

(*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*)


anafalcao:
Hey @tybug ! Thank you so much for opening this PR, we’ll review it and get back to you.

leandrodamascena:
> fwiw: I found this bug during a research project I'm working on which uses LLMs to write property-based tests in [Hypothesis](https://github.com/HypothesisWorks/hypothesis). A property which found this bug was proposed and written "autonomously" by the agent. I wrote this PR myself and take full responsibility for it.
> 
> (_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._)

Hey @tybug this code looks good to me. But I'm curious about the hypothesis you created. I have plans to introduce hypothesis framework in this project to exactly cover cases like this. Do you have a minimal repository where I can those tests?

leandrodamascena (approved):
Thanks a lot for working on this @Liam-DeVoe! APPROVED

Liam-DeVoe:
>  I have plans to introduce hypothesis framework in this project to exactly cover cases like this. Do you have a minimal repository where I can those tests?

As a Hypothesis maintainer, that makes me very happy to hear :)

Here's the full bug report written autonomously by the agent, which has the property-based test it wrote inside it: 
[bug_report_slice_dictionary_2025-08-18_23-20_k8f3.md](https://github.com/user-attachments/files/22031166/bug_report_slice_dictionary_2025-08-18_23-20_k8f3.md).

This test is...fine, but it's not particularly idiomatic. The way I might write it instead is:

```python
from hypothesis import given, strategies as st
from aws_lambda_powertools.shared.functions import slice_dictionary

@given(
    st.dictionaries(st.text(), st.integers()),
    st.integers(min_value=1)
)
def test(data, chunk_size):
    if not data:
        chunks = list(slice_dictionary(data, chunk_size))
        assert chunks == []
        return

    chunks = list(slice_dictionary(data, chunk_size))

    reconstructed = {}
    for chunk in chunks:
        reconstructed.update(chunk)

    assert reconstructed == data
```

and possibly the empty-list case could be removed/folded into the standard property as well (I would have to think about it a bit more).

PR #7270, opened 2025-08-28
docs: Fix copy to clipboard button

### LINKED ISSUE #7263: Docs: Copy to clipboard icon doesn't invoke copy

I was trying to use the "Copy to clipboard" button, but it does not work for me. I am using Firefox. Clicking the text works, but not the icon

Image

The gif kind of captures the issue. I'm clicking on both the text and icon. Neither works for me

_No response_

### How can we improve?

* Clicking the icon or the text should copy the text.

### Acknowledgment

ConnorKirk:
**Issue number:** closes #7263 

## Summary

The copy to clipboard functionality did not always work as expected. Clicking the clipboard icon did not work.
This seems to be because the event handler is added to the `copyMe` class. The clipboard icon is a sibling to this class, rather than a child of the class.

The typescript docs had a similar issue and we implemented the same fix
 Issue: https://github.com/aws-powertools/powertools-lambda-typescript/issues/3832
PR: https://github.com/aws-powertools/powertools-lambda-typescript/pull/3836

### Changes

* Remove clipboard icon from markdown.
* Define clipboard icon as css

---


ConnorKirk:
I need to confess I've hit a weird issue that stopped the markdown lint container image running in my pre-commit hook. I've skipped it, as I think the check is also run in the CI. Sorry for my bad behaviour 🙈

leandrodamascena (changes requested):
Hi @ConnorKirk, is this working in your environment? It doesn't seem to be working in Chrome. I don't think this is an issue with this change, but it's not working for me.

Can you please take a look?

ConnorKirk:
Weird 🤔 
It is working for me in both Chrome and Firefox locally.
I didn't include it in the gif, but I can see it appearing in my clipboard manager as well

Firefox:
2025-08-28_clipboard-firefox (1)


Chrome:
2025-08-28_clipboard-chrome (1)

leandrodamascena:
Actually you're right! There was something weird with Chrome. This is working now.

leandrodamascena (approved):
Thanks @ConnorKirk! APPROVED!

PR #7368, opened 2025-09-11
docs(event_handler): add info section about types

### LINKED ISSUE #7364: Bug: Use appropriate error when passing non string query params
### Expected Behaviour

A correctly raised exception with a descriptive message. For example:

```
TypeError: Query param value must be a string
```

### Current Behaviour

When using `enable_validation=True` and declaring integer query params in routes, passing string values results in the following error:

```
    return {k: v.split(",") for k, v in self.query_string_parameters.items()}
               ^^^^^^^
AttributeError: 'int' object has no attribute 'split'
```

This is really misleading and forces the developer to dig into Powertools source code to determine what is going on.

### Code snippet

```python
#### App Code

app = APIGatewayRestResolver(enabled_validation=True)


@app.get("/todos")
def get_todos(page_num: Annotated[int, Query()] = None):
    pass



#### Test Code

event = {
    # ...
    "queryStringParameters": {"page_num": 5}  # <-- Integer value
}
```

### Possible Solution

It looks like Powertools assumes values to be strings, based on the usage of `split()`. If so, then we should probably raise a `TypeError` when encountering a query param value that is not a string prior to automatically splitting.

### Steps to Reproduce

See Code Snippet

latest

3.9

### Packaging format used

PyPi

### Debugging logs

```python

```

leandrodamascena:
**Issue number:** closes #7364 

### Changes

This PR adds a new section to inform customers how to handle fields such as query strings and headers. 

### User experience

---

### DIFF
--- a/docs/build_recipes/performance-optimization.md
+++ b/docs/build_recipes/performance-optimization.md
@@ -7,7 +7,7 @@ description: Optimize Lambda functions for better performance and reduced costs
 
 Optimize your Lambda functions for better performance, reduced cold start times, and lower costs. These techniques help minimize package size, improve startup speed, and reduce memory usage.
 
-Always validate your function's behavior after applying optimizations to ensure an optimization hasn't introduced any issues with your packages. For example, removal of directories that appear to be unnecessary, such as `docs`, can break some libraries. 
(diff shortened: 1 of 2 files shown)


hjgraca (approved):
GTM

PR #7415, opened 2025-09-22
fix(event_handler): parse single list items in form data

### LINKED ISSUE #7342: Bug: Form() field does not handle list parameters
I'm trying to parse a POST based Form that has a multi-valued string property.
However this doesn't appear to work?. It does work for the string parameter, but not for the list.

Minimal example:

```python
from typing import Annotated
from aws_lambda_powertools.event_handler.openapi.params import Form

app = APIGatewayHttpResolver(
    enable_validation=True,
)

@app.post("/example")
def example(
    param: Annotated[
        str,
        Form(),
    ],
    param_list: Annotated[
        list[str],
        Form(),
    ] = []
):
     pass

```

Tested by generating a dummy event to the lambda with body content like this:
```python

import urlparse.parse

def test_example():
    # Also tested with tuple form [("param_list", ["value1", "value2"])]
    body  = urlparse.parse.urlencode({
        "param": "value",
        "param_list": ["value1", "value2"]
    }, doseq=True)
    event = {
        "headers": [
            "accept": "application/json",
            "content-type": "application/x-www-form-urlencoded"
        ],
        ...
        "body": body 
    }
   ... rest of test elided
```
But I always get response code of 422 like this:
```json
{
  "statusCode": 422,
  "detail": [
    {
      "loc": [
        "body","param_list"
      ],
      "type": "list_type"
    }
  ]
}
````
Thanks!

_Originally posted by @bdellegrazie in https://github.com/aws-powertools/powertools-lambda-python/discussions/7337_

### PR DESCRIPTION
**Pull request by tonnico**
**Issue number:** closes #7342

### Changes

- remove single value from `_parse_form_data`
- create single value from lists for non-sequence values

### User experience

---

### CODE REVIEW

**Comment by dreamorosi** [comment]
Hi @tonnico, thank you for the PR!

The static analysis shows a warning on the function touched by the changes. With the new additions, the cognitive complexity of the function went up because of the excessive branching logic.

Would it be possible to refactor it and reduce it?

Thank you.

**Review by dreamorosi — DISMISSED** [review]
See my [comment above](https://github.com/aws-powertools/powertools-lambda-python/pull/7415#issuecomment-3322713073) about reducing cognitive complexity and address the SonarCloud issue.

Let us know if you have any questions.

**Review by leandrodamascena — CHANGES_REQUESTED** [review]
Hey @tonnico thanks a lot for sending this PR. What do you think about adding a test with a field that is annotated as a list but contains only a single value when submitted?! Idk if we have this test and this was the cause of this issue.

Pls let me know what do you think.

File: tests/functional/event_handler/_pydantic/test_openapi_validation_middleware.py [hunk 1]
```diff
@@ -1481,20 +1481,33 @@ def handler_custom_route_response_validation_error() -> Model:
...
+    assert result["body"] == '{"name":"test","tags":["tag1","tag2"]}'
+
+    # WHEN sending a POST request with a single value for a list field
+    gw_event["body"] = "name=test&tags=tag1"
+
+    result = app(gw_event, {})
+
+    # THEN it should parse the form data correctly
     assert result["statusCode"] == 200
+    assert result["body"] == '{"name":"test","tags":["tag1"]}'
```
tonnico, line 1604: @leandrodamascena here is a test for a single list value.

tonnico:
> See my [comment above](https://github.com/aws-powertools/powertools-lambda-python/pull/7415#issuecomment-3322713073) about reducing cognitive complexity and address the SonarCloud issue.
> 
> Let us know if you have any questions.

I refactored the function.  
I also find some duplicated code in `_request_params_to_args`.  
Finally i sort some variables so that `_request_params_to_args` and `_request_body_to_args` looks like more the same.

dreamorosi:
Hi @tonnico, thank you for addressing the review feedback - @leandrodamascena will review this tomorrow or Friday at the latest.

Thank you for your patience and contribution!

leandrodamascena:
I'm reviewing it now! I see some errors in CI related to the glide library, probably because I merged a PR today morning. Let me see what's going on.

leandrodamascena (approved):
Hi @tonnico! Sorry for the slight delay in reviewing, but this PR is really awesome! Nothing to change, thank you very much for reducing the code complexity by moving some of the logic into separate functions.

APPROVED!

bdellegrazie:
@tonnico  - thank you for fixing this - much appreciated!

PR #7440, opened 2025-09-28
fix(batch): replace asyncio.get_event_loop() with future-compatible implementation for Python 3.12+

### LINKED ISSUE #7397: Tech debt: Note about `asyncio.get_event_loop()` failling in Python 3.12+
### Why is this needed?

While perusing the code to understand how `AsyncBatchProcessor` starts the event loop, I found the following note in the code for `BasePartialProcessor`:

https://github.com/aws-powertools/powertools-lambda-python/blob/4743b80c28dcc56df7af0a7dcc8f44bf97dde62f/aws_lambda_powertools/utilities/batch/base.py#L135

Now that Lambda has the Python 3.13 runtime, this should probably re-evaluated to determine if any work is needed here to solve the concern or if the note can be removed.

Batch processing

### Suggestion

Write a test to confirm whatever concern that prompted the note is or is not happening. If it happens, resolve the bug. If it doesn't happen, remove the note.

### Acknowledgment

kattakaha:
**Issue number:** closes #7397

## Summary

This PR addresses the tech debt noted in `AsyncBatchProcessor` regarding `asyncio.get_event_loop()` potentially failing in Python 3.12+. The fix replaces the deprecated `asyncio.get_event_loop()` call with a future-compatible implementation that works across Python versions.

## Changes made

- Replaced `asyncio.get_event_loop()` with try/except logic using `asyncio.get_running_loop()` and `asyncio.new_event_loop()`
- Updated code comments to explain the Python version compatibility approach
- Verified the fix works in Python 3.14.0rc2 where `asyncio.get_event_loop()` raises `RuntimeError`
- Maintained backward compatibility with existing Python versions

## User experience

- No breaking changes for existing users
- Prevents `RuntimeError` in Python 3.14+ Lambda environments
- Ensures AsyncBatchProcessor continues to work reliably across all supported Python versions
- Tests will be added in a follow-up commit to verify the fix

---


aws_lambda_powertools/utilities/batch/base.py
```diff
@@ -132,7 +132,16 @@ async def async_process_closure():
...
-            loop = asyncio.get_event_loop()  # NOTE: this might return an error starting in Python 3.12 in a few years
+            # In Python 3.12+, asyncio.get_event_loop() raises a warning when called without a running loop
+            # In Python 3.14+, it will raise an exception if no event loop is running
+            # Use try/except to handle both the current behavior and future Python versions
+            try:
+                loop = asyncio.get_running_loop()
+            except RuntimeError:
+                # No running loop, create a new one for Lambda environment
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
```
leandrodamascena, line 143: I think this code block is basically doing nothing with this try/except: trying to get a loop running or create a new one, but since Lambda doesn't have an active loop, it mostly falls into the "except" block, right? I haven't tested it yet, but I'm not sure if this is the correct implementation for it.

leandrodamascena (changes requested):
Hey @kattakaha! I left a small comment that probably we need to think a little bit more about this. I'm testing Python 3.14 to see if this is a problem.

PR #7468, opened 2025-10-02
fix(metrics): align flush_metrics behavior with raise_on_empty_metrics documentation

### LINKED ISSUE #7467: Inconsistent behavior of flush_metrics with raise_on_empty_metrics parameter
## Summary

The current implementation of [metrics.flush_metrics](https://github.com/aws-powertools/powertools-lambda-python/blob/adbd965671f9e2077eac36fd480b3814d7a9b978/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L383) does not fully align with its documentation regarding the raise_on_empty_metrics parameter.

## Details

According to the [documentation](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L388C8-L391C72):

```python
Parameters
----------
raise_on_empty_metrics : bool, optional
    raise exception if no metrics are emitted, by default False

``` 

This suggests that setting `raise_on_empty_metrics=False` should prevent exceptions from being raised, **regardless of whether metrics are present.**

However, in practice, when calling:
```python
metrics.flush_metrics(raise_on_empty_metrics=False)
``` 

The code [still evaluates the presence of metrics due to the current implementation](https://github.com/aws-powertools/powertools-lambda-python/blob/adbd965671f9e2077eac36fd480b3814d7a9b978/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L393), raising an error:
```python
if not raise_on_empty_metrics and not self.metric_set:
```

## Suggested Fix

To make the behavior consistent with the documented intent, the check could be simplified to:
```python
if not raise_on_empty_metrics:
```

This change ensures that no exception is raised if the user explicitly disables it, regardless of whether metrics are set.

### Rationale

The documentation indicates that the sole purpose of the raise_on_empty_metrics flag is to control whether an exception is thrown when no metrics are emitted. If users pass False, they are intentionally choosing to suppress exceptions, so the existence (or absence) of metrics should not alter that behavior.

### Environment
- Library: aws-lambda-powertools
- Version: [please fill in]
- Runtime: Python 3.12 on AWS Lambda

## Additional Context

This proposal would bring the implementation in line with the documented behavior, improving consistency and reducing potential confusion for developers.

jotafeldmann:
Addressing issue https://github.com/aws-powertools/powertools-lambda-python/issues/7467 

 
**Issue number:** closes #7467 

## Summary
Align `flush_metrics` behavior with the `raise_on_empty_metrics` [documentation](https://github.com/aws-powertools/powertools-lambda-python/blob/d34dead063f2396c1e0b73a8e84b5143d95876c6/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L391).

When `raise_on_empty_metrics=False`, no exception should be raised regardless of whether metrics exist.  

### Changes
- Simplified the conditional logic in `flush_metrics` from `if not raise_on_empty_metrics and not self.metric_set:` to `if not raise_on_empty_metrics:`  

#### Before:

Even when `raise_on_empty_metrics=False`, the code still considered the presence of metrics and could raise exceptions inconsistently.  

#### After:

When `raise_on_empty_metrics=False`, no exception is raised regardless of metric set state, which is consistent with the documented intent.

---


leandrodamascena:
This PR review is paused until we clarify this: https://github.com/aws-powertools/powertools-lambda-python/issues/7467#issuecomment-3370869391

jotafeldmann:
Closing due https://github.com/aws-powertools/powertools-lambda-python/issues/7467#issuecomment-3383946741

PR #7582, opened 2025-10-26
fix(typing): Fix type of Metrics.set_timestamp argument

### LINKED ISSUE #7581: Static typing: Metrics.set_timestamp does not allow datetime.datetime, although supported
### Static type checker used

mypy (project's standard)

3.13

latest

### Static type checker info

Docstring is correct, the underlying `provider.set_timestamp()` is correct as well, but `Metrics.set_timestamp()` only allows and `int` argument.

```sh
$ uvx --with aws_lambda_powertools mypy test.py
test.py:7:25: error: Argument 1 to "set_timestamp" of "Metrics" has incompatible type "datetime"; expected "int"  [arg-type]
Found 1 error in 1 file (checked 1 source file)
```

### Code snippet

```python
from datetime import UTC, datetime

now = datetime.now(UTC)
Metrics().set_timestamp(int(now.timestamp() * 1_000))  # ok
Metrics().set_timestamp(now)  # incompatible type

Metrics().provider.set_timestamp(int(now.timestamp() * 1_000))  # ok
Metrics().provider.set_timestamp(now)  # ok
```

### Possible Solution

Add `datetime.datetime` as a possible type to `Metrics.set_timestamp(timestamp: int)`.

czechnology:
**Issue number:** closes #7581

## Summary

The docstring is correct but the type hint is not.

The underlying `AmazonCloudWatchEMFProvider.set_timestamp` supports both `int` and `datetime.datetime`.

### Changes

Just a minor fix: extend type from `int` to `int | datetime.datetime`.

### User experience

No more type warnings if an instance of `datetime.datetime` is passed to `Metrics.set_timestamp()`.

---


leandrodamascena (approved):
Thanks for fixing that! APPROVED!

PR #7764, opened 2025-12-02
feat(idempotency): Allow durable functions to replay

### LINKED ISSUE #7762: Feature request: add support for durable function in Idempotency Utility
### Expected Behaviour

Idempotency utility detects it is a durable execution and invokes the handler

### Current Behaviour

Raise IdemptoencyAlreadyInProgressError. Function will never terminate successfully. IdempotencyRecord in the persistance provider will be left in `INPROGRESS` state

### Code snippet

```python
from aws_durable_execution_sdk_python import (
    DurableContext,
    StepContext,
    durable_execution,
    durable_step,
)
from aws_durable_execution_sdk_python.config import Duration
    DynamoDBPersistenceLayer,
    idempotent,
)

persistence_layer = DynamoDBPersistenceLayer(table_name="powertools-idempotency-table")


@durable_step
def greet_user(step_context: StepContext, name: str) -> str:
    """Generate a greeting."""
    return f"Hello {name}!"


@durable_execution
@idempotent(persistence_store=persistence_layer)
def lambda_handler(event: dict, context: DurableContext) -> str:
    """Simple durable function."""
    print(event, context)
    name = event.get("name", "World")
    greeting = context.step(greet_user(name))

    context.wait(duration=Duration.from_seconds(2))
    return greeting
```

### Possible Solution

_No response_

### Steps to Reproduce

1. Create lambda function with durable config specified
2. Invoke handler above
3. Observe error

latest

3.14

### Packaging format used

PyPi

### Debugging logs

```python

```

ConnorKirk:
**Issue number:** fixes #7762 

## Summary
This PR adds support for durable functions in the Idempotency utility. 

1. Introduce the `is_replay` variable. We use this to determine if a function is being replayed.
2. Correctly register the lambda context when a durable context is passed.

### Changes

* Add `is_replay` parameter to the `handle` method
* If there's an existing idempotency record, but the function is a replay, then allow the function to continue
* Add a stubbed `DurableContext` class
* Add tests

### User experience

Idempotency utility now allows durable functions to execute after suspension. 

## Notes

This leaves one issue remaining in the idempotency utility when used with durable functions.

We use `context.getTimeInMillis()` to calculate the expiry time of the `INPROGRESS` status. Duplicate payloads received after that expiry time will be allowed. In a durable execution, the length of the workflow can be up to 1 year, across multiple lambda invocations. This is longer than the maximum 15 minutes that an `INPROGRESS` status can currently have. Idempotency records for long running workflows will not be protected from duplicate invokes beyond the first 15 minutes of their life.

In future, we could
1. Push for the inclusion of the `DurableContext.getExecutionTimeoutInMillis()` method. Once this is added, we can calculate the workflow timeout in the same manner we do for a single invocation.
2. Allow the user to configure a timeout. We'd direct users to set this to match their durable execution timeout. I'm reluctant to add a new configuation option to the Handler/Config, especially if it might be redundant in the medium term (see #1)
3. Retrieve the durable execution configuration within the function and set the timeout. This appears simpler, but requires an additional IAM permission, and might be surprising for a user.

---


aws_lambda_powertools/utilities/typing/lambda_context.py
```diff
@@ -93,3 +93,16 @@ def tenant_id(self) -> str | None:
...
+    _lambda_context: LambdaContext
+    _state: object
+
+    @property
+    def lambda_context(self) -> LambdaContext:
+        return self._lambda_context
+
+    @property
+    def state(self) -> object:
+        return self._state
```
leandrodamascena: Changed from a concrete class to a Protocol with @runtime_checkable. This avoids type conflicts with the real `DurableContext` from the Durable SDK - users can use the SDK's class directly without import collisions and we avoid creating a type only for this.

```suggestion

# Protocol enables duck typing without competing with the real DurableContext from the SDK
@runtime_checkable
class DurableContextProtocol(Protocol):
    """Protocol for durable execution context compatibility."""

    @property
    def lambda_context(self) -> LambdaContext:
        """The underlying Lambda context."""
        ...

    @property
    def state(self) -> object:
        """The durable execution state."""
        ...
```

aws_lambda_powertools/utilities/idempotency/idempotency.py
```diff
@@ -91,7 +91,17 @@ def handler(event, context):
...
+    if hasattr(context, "state"):
+        # Extract lambda_context from DurableContext
+        durable_context = cast("DurableContext", context)
+        config.register_lambda_context(durable_context.lambda_context)
+        # Note: state.operations is accessed via duck typing at runtime
+        is_replay = len(durable_context.state.operations) > 1  # type: ignore[attr-defined]
+    else:
+        # Standard LambdaContext
+        config.register_lambda_context(context)
+        is_replay = False
```
leandrodamascena: I think if we change `hasattr(context, "state")` to `hasattr(context, "state") and hasattr(context, "lambda_context")`. We have more explicit check that ensures we only treat it as a durable context when both required attributes exist.

```suggestion
    # Check for durable context using duck typing
    if hasattr(context, "state") and hasattr(context, "lambda_context"):
        durable_context = cast("DurableContextProtocol", context)
        config.register_lambda_context(durable_context.lambda_context)
        # Note: state.operations is accessed via duck typing at runtime
        is_replay = len(durable_context.state.operations) > 1  # type: ignore[attr-defined]
    else:
        # Standard LambdaContext
        config.register_lambda_context(context)
        is_replay = False
```

leandrodamascena (commented):
Hey @ConnorKirk! Overall the PR looks good! I'd suggest considering these changes to use a Protocol instead of a concrete class - this avoids type conflicts with the real DurableContext from the Durable SDK and enables proper duck typing. I haven't fully tested with mypy, but we shouldn't have any issues.

If you think that make sense, you need to change the class name in all the other files.

Thanks a lot for working on this.

ConnorKirk:
Thank you for the review @leandrodamascena . I've incorporated your feedback about using protocols. I have also rebased the branch.
Finally - I slightly tightened the check when we determine whether a call is a replay.

leandrodamascena (approved):
thanks a lot for working on this @ConnorKirk! APPROVED

PR #7818, opened 2025-12-16
feat(parser): add model for the DynamoDB Stream Lambda invocation record

### LINKED ISSUE #7817: Feature request: data model for DynamoDB Stream Lambda invocation records
### Use case

When processing DynamoDB Stream events with a lambda, it may be necessary to enable an on-failure destination, e.g., an SQS DLQ, and re-process failed records with a different lambda. The [Lambda Developer Guide](https://docs.aws.amazon.com/lambda/latest/dg/services-dynamodb-errors.html) describes the format of the invocation records that a Lambda with a DynamoDB event source sends to a SNS, SQS or S3 destination for failed invocations. Having a data model for those records in Powertools would be useful for parsing and validating the events in the lambda that processes the failed records.

### Solution/User Experience

#7818 adds a data model for DynamoDB Stream Lambda invocation records.

### Alternative solutions

```markdown

```

### Acknowledgment

exg:
**Issue number:** closes #7817

## Summary

Data model for DynamoDB Stream Lambda Invocation records

### Changes

This commit adds a model for the invocation record that a Lambda with a DynamoDB event source sends to a SNS, SQS or S3 destination for failed invocations. The record format is described in

https://docs.aws.amazon.com/lambda/latest/dg/services-dynamodb-errors.html

The model can be used in the handler of a lambda that processes failed invocations from SNS or SQS by combining it with the source envelope model, e.g.,

```
from aws_lambda_powertools.utilities.parser import envelopes, event_parser
from aws_lambda_powertools.utilities.parser.models import DynamoDBStreamLambdaInvocationModel

@event_parser(model=DDBStreamInvocationModel, envelope=envelopes.SqsEnvelope)
def lambda_handler(event: list[DDBStreamInvocationModel], context: LambdaContext):
    ...
```

### User experience

---


svozza:
Thank you for creating this issue and the accompanying PR! We will review your PR shortly.

aws_lambda_powertools/utilities/parser/models/dynamodb.py
```diff
@@ -110,3 +110,83 @@ class DynamoDBStreamModel(BaseModel):
...
+        description="",
+        examples=["Unhandled"],
+    )
+    statusCode: int = Field(
+        description="The status code returned by the Lambda",
+    )
+
+
+# https://docs.aws.amazon.com/lambda/latest/dg/services-dynamodb-errors.html
+class DynamoDBStreamLambdaInvocationModel(BaseModel):
```
svozza: Is this the right name for this? There's no indication that this event is specifically for errors; as currently named this seems like it's a success record.

exg: Ah yes, good point. Maybe `DynamoDBStreamLambdaFailedInvocationModel`?

svozza: Yep, sounds reasonable to me!

leandrodamascena: I think for the first time we're creating new models for specific destination failure/success, and this doesn't just apply to DynamoDB. It is also valid for other streaming services, like Kafka, for example. So, I wonder if we should make this more explicit, like `DynamoDB...OnSuccessDestination` or `DynamoDB...OnFailureDestination`.

What do you both think?

exg: That sounds good to me. So it would be `DynamoDBStreamLambdaOnFailureDestinationModel` ?

leandrodamascena: Yes, I like the name. I only have two questions before we proceed with the final review.

1/ Could you add an alias for all fields? While it's good to keep the original name, we know that in Python we use snake_case and I'd like to offer both options to developers.

For example:

```py
requestId: str = Field(
description="The unique identifier of the request.",
alias="request_id"
)
```

2/ I don't remember offhand if the payload changes when we send it in case of success. If it does, it makes sense to keep the original name. Otherwise, we can use a more generic name, such as `DynamoDBStreamLambdaOnDestinationModel`. Can you test it pls? If you're not available, I can test this this morning and report back here.

Thank you very much for taking the time to this PR.

exg: > 1/ Could you add an alias for all fields? While it's good to keep the original name, we know that in Python we use snake_case and I'd like to offer both options to developers.

Sure, will do.

> 2/ I don't remember offhand if the payload changes when we send it in case of success. If it does, it makes sense to keep the original name. Otherwise, we can use a more generic name, such as DynamoDBStreamLambdaOnDestinationModel. Can you test it pls? If you're not available, I can test this this morning and report back here.

I have tested it; a destination for a DynamoDB event source mapping can be added only with the "On failure" condition.

leandrodamascena: Nice nice! So, pls add the alias and let me know to make a final review.

exg: Renamed the model and added the aliases (using a [Pydantic alias generator](https://docs.pydantic.dev/latest/concepts/alias/#using-alias-generators))

leandrodamascena: Thanks a lot @exg

leandrodamascena (approved):
APPROVED!

PR #7846, opened 2025-12-20
docs: clarify append_context_keys behavior with overlapping keys

### LINKED ISSUE #7690: Bug: `append_context_keys` drops previously appended keys when temporary keys overlap
### Expected Behaviour

After exiting `append_context_keys`, all keys that existed before the context (and their values) should be restored.

### Current Behaviour

`Logger.append_context_keys` [states](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/logging/logger.py#L816) to “temporarily add logging keys”, but when a temporary key shares the same name as an existing key, the original key/value is permanently removed after the context exits. This breaks typical request-level logging flows where immutable context (e.g., order_id, user_id) is set once and child functions use the context manager to override/add keys briefly. After the first overlap, all later log lines lose the original key.

### Code snippet

```python

logger = Logger(service="orders")

def ship_order(order_id: str):
    logger.append_keys(order_id=order_id)

    logger.info("Received order")

    with logger.append_context_keys(order_id="ORD-CHILD"):
        logger.info("Processing shipment")

    # Expected: original order_id still present.
    # Actual: order_id key vanished entirely.
    logger.info("Order completed")


if __name__ == "__main__":
    ship_order("ORD-123")
```

```
{"level":"INFO","location":"ship_order:9","message":"Received order","timestamp":"2025-11-15 13:18:57,825+0800","service":"orders","order_id":"ORD-123"}
{"level":"INFO","location":"ship_order:13","message":"Processing shipment","timestamp":"2025-11-15 13:18:57,825+0800","service":"orders","order_id":"ORD-CHILD"}
{"level":"INFO","location":"ship_order:17","message":"Order completed","timestamp":"2025-11-15 13:18:57,825+0800","service":"orders"}
```

### Possible Solution

Root cause appears in `LambdaPowertoolsFormatter.append_context_keys`: it adds the keys and unconditionally calls `remove_keys` on exit, rather than restoring previous values. A fix could snapshot overlapping keys before the context and restore them afterward, or only remove keys that didn’t exist beforehand.

### Steps to Reproduce

Run the repro scripts.

latest

3.12

### Packaging format used

PyPi

### Debugging logs

```python

```

> Note: This issue was identified by an automated testing tool for academic research and manually verified. If you have any concerns about this type of reporting, please let me know, and I will adjust my workflow accordingly.

oyiz-michael:
**Issue number:** closes #7690

## Summary

This PR clarifies the behavior of `append_context_keys()` when keys overlap with existing keys, addressing user confusion reported in #7690. This is a documentation-only fix that explains the current implementation without changing functionality.

### Changes

- **Updated docstrings** in `formatter.py` and `logger.py`:
  - Added Warning section explaining that all keys added within the context are removed on exit
  - Clarified that overlapping keys lose their original values after context exit
  - Recommended using `append_keys()` for persistent keys

- **Enhanced documentation** in `docs/core/logger.md`:
  - Added prominent "danger" callout box explaining key collision behavior
  - Included concrete code example showing what happens when keys overlap
  - Demonstrates the issue: persistent key `order_id="ORD-123"` is lost after using `append_context_keys(order_id="ORD-CHILD")`

### User experience

**Before:**
- Users expected Python context manager conventions (restore state on exit)
- Documentation said "temporarily add logging keys" without clarifying removal behavior
- When keys overlapped, original values were silently lost, breaking logging flows
- No guidance on avoiding key name collisions

**After:**
- Clear warnings in both code docstrings and user-facing documentation
- Concrete example showing the key collision problem
- Explicit recommendation: use `append_keys()` for persistent context, avoid overlapping key names with `append_context_keys()`
- Users understand this is current expected behavior, not a bug

---


leandrodamascena (approved):
Thanks fir the PR @oyiz-michael!

APPROVED!

PR #7910, opened 2026-01-06
ci: replace poetry with uv

### LINKED ISSUE #5624: Maintenance: Migrate our pipelines from poetry to uv
### Why is this needed?

The `uv` is an extremely fast Python package and project manager, written in Rust.

We need to migrate our pipelines from `poetry` to `uv` to make it fast and have some benefits that `uv` brings. I will be creating sub-issues to make this work progressively.

Other

### Solution

Migrate from poetry to uv.

### Acknowledgment

leandrodamascena:
**Issue number:** closes #5624 

## Summary

This PR migrates our pipelines from Poetry to uv.

Poetry examples in `examples/build_recipes/poetry/ `kept for users who still use it.

### Changes

### User experience

---


tests/e2e/utils/lambda_layer/powertools_layer.py
```diff
@@ -42,7 +42,23 @@ def build(self) -> str:
         self.before_build()
 
         if self._has_source_changed():
-            subprocess.run(self.build_command, shell=True, check=True)
+            # Build wheel first, then install with platform constraints
```
svozza: Just for my own curiosity, why does these new commands need to be added?

Makefile
```diff
@@ -5,25 +5,23 @@ target:
 	@$(MAKE) pr
 
 dev:
-	pip install --upgrade pip pre-commit poetry
-	@$(MAKE) dev-version-plugin
-	poetry install --extras "all redis datamasking valkey"
+	pip install --upgrade pip pre-commit uv
+	uv sync --all-extras
```
tonnico, line 9: I like the move to `uv`.

Maybe somthing like this?

```
pip install --upgrade uv
uv sync --all-extras
```

and putting `pre-commit` (maybe `pre-commit-uv`) to `extras`? I think, you do not need to upgrade `pip` anymore after installing `uv`. what do you think?

tests/e2e/utils/lambda_layer/powertools_layer.py
leandrodamascena: Honestly, I think this bunch of commands is overkill for managing dependencies. We're not concerned with cold start and things like that; we just want to ensure everything works as expected after any code change, regardless of whether we install libraries with debug symbols, binaries, or not. So, I think I'll refactor all these commands and use the `PythonFunction` constructor from the CDK, which takes care of all that.

Makefile
leandrodamascena: Yes, you're right! I'm focusing now on resolving the issue with our end-to-end tests that are overengenireeing dependencies install, and I'll take a look on this after that.
Thanks for the tip, @tonnico.

tests/e2e/utils/lambda_layer/powertools_layer.py
svozza: Makes sense!

Suresh-2017:
Issue: Incorrect setup-python Cache Configuration (CI Risk)
Files impacted:
.github/workflows/quality_check.yml
.github/workflows/run-e2e-tests.yml
.github/workflows/release-v3.yml
.github/workflows/quality_code_cdk_constructor.yml
Severity: Critical
The workflows still usecache: "poetry" inactions/setup-python, which appears to be a leftover from the previous Poetry-based setup. Since the project has migrated touv, this configuration is no longer valid and may lead to cache misses or inconsistent CI behavior.
Current configuration (outdated): yaml- name: Set up Python ${{ matrix.python-version }}
  uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548
  with: python-version: ${{ matrix.python-version }}
  cache: "poetry"
Recommended fix:
 Preferred option: Remove the cache setting entirely, asastral-sh/setup-uv withenable-cache: true already provides optimized caching.
yaml with: python-version: ${{ matrix.python-version }}
No cache key - handled by setup-uv
 Alternative (if retaining setup-python caching): yaml cache: "pip"
Note: Avoid duplicating caching layers. The setup-uv cache is sufficient and more appropriate for this setup.

PR #7913, opened 2026-01-06
feat: add HttpResolverAlpha resolver

### LINKED ISSUE #7912: Feature request: Add HttpResolverAlpha to run Event Handler APIs locally with ASGI servers
### Use case

Event Handler resolvers only work with Lambda event payloads. To test APIs locally, you need SAM CLI with Docker or deploy to AWS.

I want to run my API locally with uvicorn like I do with frameworks like FastAPI, Flask, and others.

### Solution/User Experience

New `HttpResolverAlpha` that implements the ASGI specification:

```python
from aws_lambda_powertools.event_handler import HttpResolverAlpha

app = HttpResolverAlpha(enable_validation=True)

@app.get("/hello/<name>")
def hello(name: str):
    return {"message": f"Hello, {name}!"}

# Run locally: uvicorn app:app --reload
# Deploy to Lambda: handler = app
```

ASGI is the industry standard for Python web apps (same spec FastAPI, Starlette, and Django use). Works with any ASGI server like uvicorn, hypercorn, daphne, or granian. No need to reinvent HTTP handling - we leverage existing battle-tested servers. Also async-ready for future improvements.

This enables faster local development without Docker or AWS deployment. 

This must be an Alpha version yet.

### Alternative solutions

```markdown
SAM CLI
LocalStack
```

### Acknowledgment

leandrodamascena:
**Issue number:** closes #7912 

## Summary

Adds `HttpResolverLocal` - a new resolver that implements ASGI for running Event Handler APIs locally with uvicorn or any ASGI server.

### Changes

- New `HttpResolverLocal` class in `aws_lambda_powertools/event_handler/http_resolver.py`
- Implements ASGI spec, no external dependencies
- All existing features work: routing, middleware, validation, OpenAPI, CORS, exception handling
- Marked as Alpha with runtime warning

### User experience

```python
from aws_lambda_powertools.event_handler import HttpResolverLocal

app = HttpResolverLocal(enable_validation=True)

@app.get("/hello/<name>")
def hello(name: str):
    return {"message": f"Hello, {name}!"}

# Run locally: uvicorn app:app --reload
# Deploy to Lambda: handler = app
```

---


dreamorosi:
I have some questions/concerns around whether this fits in the library.

Our first tenet says we focus on AWS Lambda runtime while this goes quite outside of that. I can see some value in having a regular resolver running locally for local development, but in the current shape the new resolver requires a code change which is creates a precedent in us adding a significant runtime implementation for a non-Lambda environment.

I am not opposing this feature, but I wonder if we can frame it differently or make it work differently.

As a quick win, I'd feel more comfortable if this was 1/ listed under the testing section of the docs, and 2/ named in a way that clearly conveys it's meant to be used for local testing only. In both cases my concern is that this gives the false impression that we want to support non-Lambda environments - which is not the case I hope.

Along the same lines, is there any way at all to get an existing resolver to work with the ASGI protocol when an env variable is set? I don't know the specifics but what I have in mind is basically customers using one of the already existing resolvers, and then when they set an env variable like `POWERTOOLS_EVENT_HANDLER_ASGI` the resolver works locally with this protocol.

Again, my goal here is mainly to make sure this is positioned in the correct way. Happy to discuss.

leandrodamascena:
Those are fair points. Thanks for raising them.

> I have some questions/concerns around whether this fits in the library.
> 
> Our first tenet says we focus on AWS Lambda runtime while this goes quite outside of that. I can see some value in having a regular resolver running locally for local development, but in the current shape the new resolver requires a code change which is creates a precedent in us adding a significant runtime implementation for a non-Lambda environment.
> 

Actually there is a minimal code change between local and Lambda

1/ Change the resolver
2/ Uncomment the Lambda Handler.

> 
> As a quick win, I'd feel more comfortable if this was 1/ listed under the testing section of the docs, and 2/ named in a way that clearly conveys it's meant to be used for local testing only. In both cases my concern is that this gives the false impression that we want to support non-Lambda environments - which is not the case I hope.

I am good to name this `HttpResolverLocal` and list it in the testing documentation, making it clear to use for test. This is a quick solution for clients who want to run it locally and create prototypes, especially those who are not onboarded in Lambda and are getting used to the Lambda programming model.

Tests, local run, CI and more.

> 
> Along the same lines, is there any way at all to get an existing resolver to work with the ASGI protocol when an env variable is set? I don't know the specifics but what I have in mind is basically customers using one of the already existing resolvers, and then when they set an env variable like `POWERTOOLS_EVENT_HANDLER_ASGI` the resolver works locally with this protocol.

Hmm no, there isn't because I'll change the code behavior and it's not ideal.

> 
> Again, my goal here is mainly to make sure this is positioned in the correct way. Happy to discuss.

Sure thing, I think your point about tests and naming make it clear.

WDYT?

dreamorosi (commented):
Sounds good, thanks for explaining.

I am ok with moving the section in the docs under testing plus renaming.

I'd suggest `HttpResolverLocalDev` to make it extra clear, but if it's too long `HttpResolverLocal` that you suggested + a call out in the docstring of the class is enough.

For future reference/readers, in similar cases where the boundary of what we build is not clear, we'll still want to evaluate whether something fits or not - so our tenet still stands.

leandrodamascena:
Hey @dreamorosi can I have another review here, pls?

aws_lambda_powertools/event_handler/http_resolver.py
```diff
@@ -0,0 +1,485 @@
...
+
+    # Run locally with uvicorn:
+    # uvicorn app:app --reload
+
+    # Deploy to Lambda (sync only):
+    # handler = app
+    ```
+    """
+
+    def __init__(
```
svozza, line 209: How does the sync handler part of this work? It looks like all the methods in this class are async. I'm just curious really.

leandrodamascena: haha trick question @svozza! But this is the Python being Python..

The `HttpResolverLocal` has two entry points depending on where it runs:

1/ When deployed to Lambda, it uses the sync `resolve(event, context)` method inherited from `ApiGatewayResolver` - same as any other resolver, nothing async involved.

1/ When running locally with `uvicorn`, it uses the async `__call__(scope, receive, send)` method which is the ASGI interface. This triggers the async flow internally.

The route handlers can be sync or async and both work. Surprised? PYTHON! hahahah. The decision happens in `final_handler` (https://github.com/aws-powertools/powertools-lambda-python/pull/7913/changes/BASE..290024a454f9228afd2d717a110d3a0aec55bf0c#diff-58f8308da12bb101ce3a2e8fd284fe46165eaea84e1e413dd05d5e1e224ce32dR302) inside `_run_middleware_chain_async`:

```python
result = route.func(**route_args)
if inspect.iscoroutine(result):
    result = await result
```

I call the handler and check if the result is a coroutine. If it is, I await it. If not, I just use the value directly because it's a sync invocation

An async function can call sync functions without any issues - it only needs `await` for coroutines. The reverse is what doesn't work (sync can't await async). So a sync handler like `def hello()` works fine when called from the async ASGI flow.

I took some time to make it work with both, because this way we can prepare the Event Handler to be async if Lambda decides to move to this.

Let me know if I was clear or not.

svozza: Ah yes, that makes sense. I was wondering what the `iscoroutine` check later on was doing and now it's clear!

svozza (approved):

dreamorosi (approved):

PR #7920, opened 2026-01-07
feat(openapi): add support for micro Lambda pattern

### LINKED ISSUE #7919: RFC: Support OpenAPI generation for micro functions via CLI/script
### Discussed in https://github.com/aws-powertools/powertools-lambda-python/discussions/5914

<sup>Originally posted by **ran-isenberg** February  3, 2024</sup>

_No response_

Event Handler - REST API

### Summary

Hi,
I'd like to use event handler with event validation turned on for my micro functions service. At the moment, the /swagger generation is done via one endpoint and unless that handler defines all routes in the app, there's no way to generate all the routes for the OpenAPI.

### Use case

I have a service with multiple micro function, I'm not a fan of mono lambda for large service.
In addition, I'm more interested in generating the OpenAPI doc offline and saving it to my service repository where it can be later published to various location, and not necessarily served via the rest endpoint.

### Proposal

Create a CLI that I can run locally on my service in the shell that generates the OpenAPI doc. It will need to init the event handler and it's all uses in the different lambda handlers and then generate the document and save it to a destination folder in a JSON format.
I have created a CLI that works with the current implementation - it accesses the  deployed API GW at the /swagger?format=json and saves the output. Now, this is obviously not the solution, but it's an example of a possible interface (replace stack output input with the lambda handlers folder patch and it might be enough.

Here's the current CLI:
https://github.com/ran-isenberg/aws-lambda-handler-cookbook/blob/main/generate_openapi.py

### Out of scope

I have added this script as a makefile command of my AWS Lambda handler cookbook and now it's part of my CI/CD pipeline where I can fail PRs if developers added/changed the REST API but forgot to update the documentation. Expect a blog post about it very soon :)

### Potential challenges

Thanks @leandrodamascena  for this section: 
1. - OS compatibility: It's a CLI, we'll make sure it works on at least Windows, Linux and Mac.
2. - Security: what are the possible problems when we provide a CLI (even if it is pure python code) to install on someone' computer?
3 - Performance: do we need to worry about any performance issues? - i dont think so, it runs locally,. can be several seconds at max.

Please note that it can even be just a script that i can run from powertools library.

_No response_

### Alternative solutions

_No response_

### Acknowledgment

- [X] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)</div>

leandrodamascena:
**Issue number:** closes #7919 

**Original Disucssion:** https://github.com/aws-powertools/powertools-lambda-python/discussions/5914

**Documentation will come in a separate PR. I'm refactoring the OpenAPI docs.**

## Summary

This PR adds support for generating unified OpenAPI schemas from multiple Lambda handlers.

This is an old desire from customers that don't use Fat Lambda architecture. When you have micro-functions (one Lambda per route), each Lambda has its own resolver and generates its own OpenAPI spec. This feature allows merging all specs into a single unified schema.

### Implementation and security concerns

Some customers suggested a CLI approach, but that could introduce security concerns with arbitrary code execution.

This is pure Python with the customer in control of paths, patterns, and exclusions. Discovery uses AST analysis to detect resolver instances - no code is executed during discovery, just static analysis. The actual schema generation only happens when the customer explicitly calls `get_openapi_schema()`.

## Changes

New `OpenAPIMerge` class and `configure_openapi_merge()` method to discover and merge OpenAPI schemas from multiple Lambda function handlers into a single specification.

**1. `OpenAPIMerge` class** - Standalone class to generate the schema file via Python code. Write a simple script, run it, get your `openapi.json`.

**2. `configure_openapi_merge()` method** - For customers who want a dedicated Lambda to serve the OpenAPI spec and Swagger UI. This Lambda only serves the merged schema - it won't execute the actual route handlers since each route lives in its own Lambda function. The routes in the schema point to API Gateway, which routes to the correct Lambda.

### Conflict handling

When the same path+method exists in multiple handlers, use `on_conflict`:

- `warn` (default): logs warning, keeps first
- `error`: raises OpenAPIMergeError  
- `first`: silently keeps first
- `last`: uses last (override)

Files are sorted alphabetically, so precedence is deterministic.

### Supported resolvers

All of them: APIGatewayRestResolver, APIGatewayHttpResolver, ALBResolver, LambdaFunctionUrlResolver, VPCLatticeResolver, VPCLatticeV2Resolver, BedrockAgentResolver.

### Standalone

```py
from aws_lambda_powertools.event_handler.openapi import OpenAPIMerge

merge = OpenAPIMerge(title="My API", version="1.0.0")
merge.discover(path="./functions", pattern="handler.py", recursive=True)

with open("openapi.json", "w") as f:
    f.write(merge.get_openapi_json_schema())
```

### Via resolver (runtime, Swagger UI)

```py

app.configure_openapi_merge(
    path="./functions",
    pattern="handler.py",
    recursive=True,
    title="My API",
    version="1.0.0",
)
app.enable_swagger(path="/docs")
```

---


leandrodamascena:
Hey @dreamorosi and @svozza I'd like a review here. Pls read the PR body before the review.

I'm still fixing some coverage issues.

ran-isenberg:
yes, this is so overdue!!!! Thank you!

dreamorosi (commented):
Agree with the approach suggested of having a runtime implementation over a CLI or similar.

I'll leave the technical review of the PR to Stefano - if you want a second pair of eyes you can also involve @sdangol since he's been working on Event Handler as well.

tests/functional/event_handler/_pydantic/test_openapi_merge.py
```diff
@@ -0,0 +1,340 @@
...
+        "queryStringParameters": {"format": "json"},
+        "headers": {},
+        "requestContext": {"stage": "prod", "path": "/prod/swagger"},
+    }
+    response = app.resolve(event, {})
+
+    # THEN it should return merged schema
+    body = json.loads(response["body"])
+    assert body["info"]["title"] == "Swagger Merge API"
+    assert "/users" in body["paths"]
```
svozza, line 220: Is this validation sufficient? Don't these paths have other properties we might want to validate?

leandrodamascena: Good point, but this is already tested in a bunch of other places, like [here](https://github.com/aws-powertools/powertools-lambda-python/pull/7920/changes/BASE..3d2cb7fa9038c38a5e5a11aeddffef34bf231009#diff-51779f6e92994c021d6a5872ff91d533fc5829798e2095981aabe0d0ecf2d58eR281) and in other methods that uses the swagger.
I think we are safe testing only these props.

svozza: Ah, sorry I missed that bit below!

hjgraca (dismissed):

dreamorosi:
Let's wait for @svozza final review before merging pls

hjgraca (approved):

svozza (approved):

leandrodamascena:
This PR is ready to be merged. @ran-isenberg offered to test it today with his Cookbook project - https://github.com/ran-isenberg/aws-lambda-handler-cookbook - and I would take the opportunity to find bugs and everything else before the official release.

That said, @svozza, could you prioritize this review a bit? I would greatly appreciate the help.

PR #7940, opened 2026-01-11
feat: Add a flag to ALBResolver to URL-decode query parameters

### LINKED ISSUE #7621: Feature request: Add a flag to ALBResolver to URL-decode query parameters
### Use case

(follow up from Discord thread: https://discord.com/channels/1006478942305263677/1432874977454985288)

Background: According to https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html, for ALB requests
> If the query parameters are URL-encoded, the load balancer does not decode them. You must decode them in your Lambda function

This decoding should happen before input validation; I think `datetime` types are especially problematic for this if you're trying to pass them as ISO-format strings. For example: `.../?start=2025-12-20T16%3A56%3A02.032000%2000%3A00&end=2025-12-30T16%3A56%3A02.032000%2000%3A00`

### Solution/User Experience

It would be nice for ALBResolver to optionally do this decoding for you (I understand it could be a behavior change, so you probably wouldn't want it to be enabled by default).

An example of this behavior in some other frameworks that process ALB events:
https://github.com/brefphp/bref/issues/456
https://github.com/fastify/aws-lambda-fastify/pull/91

### Alternative solutions

```markdown
I was able to work around this by adding a middleware for my ALB handler:


@lambda_handler_decorator
def alb_query_parameter_decoder(
    handler: Callable[[dict, LambdaContext], dict],
    event: dict,
    context: LambdaContext,
) -> dict:
    query_params = event.get("queryStringParameters")
    if query_params:
        for k, v in query_params.items():
            try:
                query_params[k] = unquote(v)
            except Exception:
                pass

```

### Acknowledgment

chriselion:
**Issue number:** closes #7621

## Summary
From https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html, for ALB requests

> If the query parameters are URL-encoded, the load balancer does not decode them. You must decode them in your Lambda function

This adds a flag to ALBResolver to decode the query parameters before they are validated.

### Changes

* Add a `decode_query_parameters` flag to `ALBResolver` initializer.
* Add a ` decode_query_parameters` to `ALBEvent`. This defaults to false.
  * I thought about adding a `**kwargs` argument to the `BaseProxyEvent` initializers, but didn't love that.
* Add `ALBResolver._to_proxy_event` (overrides the parent method). This calls the parent method, and sets a decode_query_parameters flag on the resulting ALBEvent.
* Added functional test to ensure that an encoded datetime string could be passed to a route that accepts a datetime parameter, and a unit test for the ALBEvent decoding.

### User experience

No change to the user experience - default behavior is the same, and users have to opt-in to get the new behavior.

---


aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -3316,8 +3154,34 @@ def __init__(
         enable_validation: bool = False,
         response_validation_error_http_code: HTTPStatus | int | None = None,
         json_body_deserializer: Callable[[str], dict] | None = None,
+        decode_query_parameters: bool = False,
     ):
-        """Amazon Application Load Balancer (ALB) resolver"""
+        """Amazon Application Load Balancer (ALB) resolver
+
+
+        Parameters
```
chriselion, line 3324: These are copied from `ApiGatewayResolver` (except for `decode_query_parameters`, obviously)

aws_lambda_powertools/event_handler/api_gateway.py
```diff
@@ -3354,3 +3219,10 @@ def _to_response(self, result: dict | tuple | Response | BedrockResponse) -> Res
             result.body = ""
 
         return super()._to_response(result)
+
+    @override
+    def _to_proxy_event(self, event: dict) -> BaseProxyEvent:
+        proxy_event = super()._to_proxy_event(event)
+        if isinstance(proxy_event, ALBEvent):
```
chriselion, line 3388: This should always be the case, but I think mypy would complain if I didn't.

Alternatively, I could just create the ALBEvent here, but if the base method ever changes, that could break.

aws_lambda_powertools/utilities/data_classes/alb_event.py
```diff
@@ -30,13 +33,27 @@ class ALBEvent(BaseProxyEvent):
     - https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html
     """
 
+    @override
+    def __init__(self, data: dict[str, Any], json_deserializer: Callable | None = None):
+        super().__init__(data, json_deserializer)
+        self.decode_query_parameters = False
```
chriselion, line 39: This gets overridden in ALBResolver._to_proxy_event.

aws_lambda_powertools/utilities/data_classes/alb_event.py
```diff
@@ -30,13 +33,27 @@ class ALBEvent(BaseProxyEvent):
...
     def resolved_query_string_parameters(self) -> dict[str, list[str]]:
-        return self.multi_value_query_string_parameters or super().resolved_query_string_parameters
+        params = self.multi_value_query_string_parameters or super().resolved_query_string_parameters
+        if not self.decode_query_parameters:
+            return params
+
+        # Decode the parameter keys and values
+        decoded_params = {}
+        for k, vals in params.items():
+            decoded_params[unquote(k)] = [unquote(v) for v in vals]
```
chriselion, line 54: Decoded the keys for consistency. That's what it looks like other frameworks do, although I'm not sure it would be useful here.

tests/functional/event_handler/_pydantic/test_openapi_validation_middleware.py
```diff
@@ -1,4 +1,5 @@
 import base64
+import datetime
```
chriselion, line 2: Is this the best file for this test? I originally started here, but not sure if something like `tests/functional/event_handler/_pydantic/test_api_gateway.py` is better.

aws_lambda_powertools/event_handler/api_gateway.py
chriselion: Coverage is complaining about this, since I don't test the negative branch. Is a `# pragma: no cover` OK here?

leandrodamascena:
Hey @chriselion I'm attending some internal training and will review this PR later this week. Thanks a lot for working on this.

aws_lambda_powertools/event_handler/api_gateway.py
leandrodamascena: Hey @chriselion this is an indirect change, so thats ok we keep this.

leandrodamascena (approved):
Hey @chriselion thanks a lot for working on this! APPROVED!!!

hjgraca (approved):

PR #7952, opened 2026-01-19
fix(event-handler): prevent OpenAPI schema bleed when reusing response dictionaries

### LINKED ISSUE #7711: Bug: OpenAPI schema return types bleed across routes when reusing response dictionaries
### Expected Behaviour

When building an API using the `APIGatewayRestResolver`, I should be able to reuse common response dictionary objects across multiple routes. The generated OpenAPI schema should accurately reflect the unique return type annotation for each function, even when they share the same response configuration dictionary object.

For example, given these two routes:

| Route | Return Type | Expected OpenAPI Schema Response Model |
| :--- | :--- | :--- |
| `/exams` | `Response[list[ExamSummary]]` | A list of `ExamSummary` objects |
| `/exams/<exam_id>` | `Response[ExamConfig]` | A single `ExamConfig` object |

Each route should display its own correct return type in the OpenAPI schema.

### Current Behaviour

The framework leverages response dictionaries by reference (since multiple routes share the same dictionary object) and mutates them when injecting the return type schema.

As a result, the OpenAPI schema generated is incorrect due to a schema bleed: the return type from one of the routes persists in the shared dictionary and is then erroneously used for other routes.

In my specific local tests with the code snippet below, it appears the schema for `/exams/<exam_id>` is being injected first (or whichever route is processed first by the OpenAPI generator):

1. The `/exams` endpoint **incorrectly** shows a response model for **`ExamConfig`** instead of the expected `list[ExamSummary]` type.
2. The `/exams/<exam_id>` endpoint **correctly** shows a response model for **`ExamConfig`**.

### Code snippet

```python
class Responses:
    """Pre-configured OpenAPI response schemas."""

    # Base responses
    OK = {200: OpenAPIResponse(description="Successful operation")}

    #...

    # Common combinations
    STANDARD_ERRORS = {**NOT_FOUND, **CONFLICT, **VALIDATION_ERROR, **SERVER_ERROR}

    @classmethod
    def combine(cls, *response_dicts: dict[int, OpenAPIResponse]) -> dict[int, OpenAPIResponse]:
        """Combine multiple response dictionaries."""
        result = {}
        for response_dict in response_dicts:
            result.update(response_dict)
        return result


@app.get(
    "/exams",
    tags=["Exams"],
    responses=Responses.combine(Responses.OK, Responses.STANDARD_ERRORS),
)
def list_exams() -> Response[list[ExamSummary]]:
    """Lists all available exams."""
    # ... returns list[ExamSummary] but the OpenAPI says it returns ExamConfig


@app.get(
    "/exams/<exam_id>",
    tags=["Exams"],
    responses=Responses.combine(Responses.OK, Responses.STANDARD_ERRORS), # Reusing the shared Responses.OK dictionary
)
def get_exam_config(exam_id: str) -> Response[ExamConfig]:
    """Get the configuration for a specific exam"""
    # ... returns ExamConfig
```

### Possible Solution

The issue appears to be in the `_get_openapi_path` method in `api_gateway.py`, where the response dictionary is mutated directly without creating a copy first.

Suggested fix - perform a deep copy before mutation:

```python
# Add the response to the OpenAPI operation
if self.responses:
    for status_code in list(self.responses):
        # Current implementation (mutates the shared dictionary):
        # response = self.responses[status_code] 
        
        # Proposed fix - deep copy to prevent mutation of the source object:
        response = copy.deepcopy(self.responses[status_code])

        # Case 1: there is not 'content' key
        if "content" not in response:
            # This mutation would now only affect the local copy
            response["content"] = { 
                _DEFAULT_CONTENT_TYPE: self._openapi_operation_return(...)
            }
```

### Steps to Reproduce

1. Create a shared response dictionary in a `Responses` class (e.g., `Responses.OK`)
2. Define multiple routes that reuse the same response dictionary via `Responses.combine()`
3. Use different return type annotations for each route (e.g., `Response[list[ExamSummary]]` vs `Response[ExamConfig]`)
4. Generate the OpenAPI schema
5. Inspect the response models in the OpenAPI schema - observe that one route's return type incorrectly appears in another route's schema

latest

3.13

### Packaging format used

PyPi

### Debugging logs

```python

```

oyiz-michael:
**Issue number:** closes #7711

### Changes

This PR fixes a bug where OpenAPI schema return types were bleeding across routes when reusing response dictionaries. The issue occurred when multiple routes shared the same response dictionary object (e.g., via `Responses.combine()`), causing the schema generator to mutate the shared dictionary and incorrectly apply one route's return type schema to all routes.

**Root cause:** The `_add_route_to_openapi_routes` method was directly referencing the shared response dictionary, so modifications made during schema generation affected all routes using that dictionary.

**Solution:** Changed line 670 in `api_gateway.py` to use `copy.deepcopy()` to create an independent copy of the response dictionary before any modifications, ensuring each route's schema remains isolated.

**Files changed:**
- `aws_lambda_powertools/event_handler/api_gateway.py`: Added `import copy` and modified response dictionary handling to use `deepcopy()`
- `tests/functional/event_handler/_pydantic/test_openapi_shared_response_bleed.py`: Added comprehensive regression tests

### User experience

**Before:**
```python

responses = Responses.combine(
    Responses.success,
    Responses.not_found
)

@app.get("/exams", responses=responses)
def list_exams() -> list[ExamSummary]:
    ...

@app.get("/exam/<exam_id>/config", responses=responses)
def get_exam_config(exam_id: str) -> ExamConfig:
    ...
```

The generated OpenAPI schema would incorrectly show both routes returning the same type (whichever was processed last), instead of list[ExamSummary] and ExamConfig respectively.

After:
Each route correctly maintains its own return type schema in the OpenAPI specification, regardless of shared response dictionaries. The /exams route shows list[ExamSummary] and the /exam/<exam_id>/config route shows ExamConfig as expected.

Testing:
Added 2 new regression tests that verify no schema bleed occurs
Code quality checks pass: ruff format, ruff check, mypy
All 323 existing event handler tests pass

Disclaimer: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

### DIFF
--- a/aws_lambda_powertools/event_handler/api_gateway.py
+++ b/aws_lambda_powertools/event_handler/api_gateway.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import base64
+import copy
 import json
 import logging
 import re
@@ -666,7 +667,8 @@ class Route:
         # Add the response to the OpenAPI operation
         if self.responses:
             for status_code in list(self.responses):
-                response = self.responses[status_code]
+                # Create a deep copy to prevent mutation of the shared dictionary
+                response = copy.deepcopy(self.responses[status_code])
 
                 # Case 1: there is not 'content' key
                 if "content" not in response:
--- a/tests/functional/event_handler/_pydantic/test_openapi_response_combined.py
+++ b/tests/functional/event_handler/_pydantic/test_openapi_response_combined.py
@@ -1,6 +1,4 @@
-"""Test for bug #7711: OpenAPI schema return types bleed across routes when reusing response dictionaries"""
-
-from __future__ import annotations
+from typing import Dict, List
 
 from pydantic import BaseModel
 
@@ -39,7 +37,7 @@ class Responses:
     STANDARD_ERRORS = {**NOT_FOUND, **VALIDATION_ERROR, **SERVER_ERROR}
 
     @classmethod
-    def combine(cls, *response_dicts: dict[int, OpenAPIResponse]) -> dict[int, OpenAPIResponse]:
+    def combine(cls, *response_dicts: Dict[int, OpenAPIResponse]) -> Dict[int, OpenAPIResponse]:
         """Combine multiple response dictionaries."""
         result = {}
         for response_dict in response_dicts:
@@ -62,7 +60,7 @@ def test_openapi_shared_response_no_bleed():
         tags=["Exams"],
         responses=Responses.combine(Responses.OK, Responses.STANDARD_ERRORS),
     )
-    def list_exams() -> Response[list[ExamSummary]]:
+    def list_exams() -> Response[List[ExamSummary]]:
         """Lists all available exams."""
         return Response(
             status_code=200,
--- a/tests/functional/event_handler/_pydantic/test_openapi_shared_response_bleed.py
+++ b/tests/functional/event_handler/_pydantic/test_openapi_shared_response_bleed.py
@@ -1,0 +1,173 @@
+"""Test for bug #7711: OpenAPI schema return types bleed across routes when reusing response dictionaries"""
+
+from __future__ import annotations
+
+from pydantic import BaseModel
+
+from aws_lambda_powertools.event_handler import APIGatewayRestResolver, Response
+from aws_lambda_powertools.event_handler.openapi.types import OpenAPIResponse
+
+
+class ExamSummary(BaseModel):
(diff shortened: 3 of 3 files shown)


leandrodamascena (approved):
Hey @oyiz-michael thanks a lot for this PR! Approved.

hjgraca (approved):

PR #7965, opened 2026-01-23
feat(event-handler): add per-route validation support

### LINKED ISSUE #6983: Feature request: Add enable_validation per route
### Discussed in https://github.com/aws-powertools/powertools-lambda-python/discussions/4771

<sup>Originally posted by **Sheemap** July 16, 2024</sup>
Hello! We have a fairly large monolithic lambda that does not have enable_validation set.

We tried it out on a different lambda of ours, validating the schema with an OpenAPI spec, and it was a great experience!

We want to get the monolith using it as well, but Im not finding  any way to do it incrementally. Aside from making a new lambda, and its hard to find the time to migrate the whole thing at once.

Would it be possible to allow validation to be enabled on a split Router or so?</div>

oyiz-michael:
Issue number: closes #6983

Summary

Changes
This PR adds per-route  enable_validation control to Event Handler, allowing routes to override the resolver-level validation setting. This enables incremental adoption of OpenAPI validation in large monolithic Lambda functions.

Implementation:

Added enable_validation parameter (bool | None) to the Route class
Updated all route decorators to accept enable_validation
Modified middleware stack builder to check route-level setting before resolver-level setting
Extended support to BedrockAgentResolver and async middleware chains
Route-level setting overrides resolver-level when explicitly set; None inherits from resolver
Fully backwards compatible - existing code works without changes

User experience
Before:
```
# Validation was all-or-nothing at the resolver level

# ALL routes must have valid OpenAPI schemas
# No way to incrementally adopt validation in a large monolithic Lambda
@app.get("/new-endpoint")  # Must be validated
def new_handler(): ...

@app.get("/legacy-endpoint")  # Also must be validated - breaks if not ready!
def legacy_handler(): ...
```

Users with large monolithic Lambdas couldn't adopt validation incrementally without migrating all endpoints at once, which is time-consuming and risky.

After:
```
# Enable validation globally

# New/critical endpoints use validation (inherit from resolver)
@app.get("/new-endpoint")
def new_handler() -> NewResponse:
    return NewResponse(...)

# Legacy endpoints can opt-out during migration
@app.get("/legacy-endpoint", enable_validation=False)
def legacy_handler():
    return {"status": "ok"}  # No validation required

# Gradually migrate routes over time without breaking existing functionality

```

This enables teams to:
Enable validation globally on the resolver
Selectively disable it for legacy routes not yet ready
Migrate routes incrementally at their own pace
Avoid breaking changes during gradual adoption

Disclaimer: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.


dreamorosi (approved):
Hi @oyiz-michael, thank you for the PR and for the patience in the review.

I've looked at the implementation and as far as I can see the changes proposed make sense and address the issue.

Thank you!

PR #7980, opened 2026-02-07
fix(event_handler): return 415 status_code for unsupported content-type headers

### LINKED ISSUE #7978: Bug: Event handler with unsupported content-type headers raises an exception instead of returning a 4xx error
### Expected Behaviour

Passing unsupported content-type headers (for example, `"headers": {"content-type": "text/xml"}`) to an event handler (tested with `ALBResolver`, but probably affects others) should result in some sort of client error, for example 422.

### Current Behaviour

Currently, an uncaught exception is raised, resulting in a 500 status code:
```
request returned status 500
  error: Traceback (most recent call last):
  File "/Users/celion/.cache/uv/environments-v2/validation-repro-e88f9191e633f920/lib/python3.12/site-packages/aws_lambda_powertools/event_handler/api_gateway.py", line 2734, in _call_route
    route(router_middlewares=self._router_middlewares, app=self, route_arguments=route_arguments),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/celion/.cache/uv/environments-v2/validation-repro-e88f9191e633f920/lib/python3.12/site-packages/aws_lambda_powertools/event_handler/api_gateway.py", line 509, in __call__
    return self._middleware_stack(app)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/celion/.cache/uv/environments-v2/validation-repro-e88f9191e633f920/lib/python3.12/site-packages/aws_lambda_powertools/event_handler/api_gateway.py", line 1639, in __call__
    return self.current_middleware(app, self.next_middleware)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/celion/.cache/uv/environments-v2/validation-repro-e88f9191e633f920/lib/python3.12/site-packages/aws_lambda_powertools/event_handler/middlewares/base.py", line 121, in __call__
    return self.handler(app, next_middleware)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/celion/.cache/uv/environments-v2/validation-repro-e88f9191e633f920/lib/python3.12/site-packages/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py", line 102, in handler
    received_body=self._get_body(app),
                  ^^^^^^^^^^^^^^^^^^^
  File "/Users/celion/.cache/uv/environments-v2/validation-repro-e88f9191e633f920/lib/python3.12/site-packages/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py", line 132, in _get_body
    raise NotImplementedError("Only JSON body or Form() are supported")
NotImplementedError: Only JSON body or Form() are supported
```

### Code snippet

```python
# /// script
# dependencies = [
#   "aws-lambda-powertools==3.24.0",
#   "pydantic"
# ]
# ///

import pydantic
import aws_lambda_powertools

class MyRequest(pydantic.BaseModel):
    foo: str = ""

class MyResponse(pydantic.BaseModel):
    bar: str

app = ALBResolver(enable_validation=True)

@app.put("/test/")
def test_endpoint(
    body: MyRequest,
) -> MyResponse:
    return MyResponse(bar=body.foo)


def main():
    print(f"Running on aws-lambda-powertools=={aws_lambda_powertools.__version__} pydantic=={pydantic.__version__}")
    event = {
        "httpMethod": "PUT",
        "path": "/test/",
        "requestId": "some_request_id",
        "requestContext": {"elb": {"targetGroupArn": ":target:"}},
        # Set non-JSON non-form headers
        "headers": {"content-type": "text/xml"}
    }

    try:
        response = app.resolve(event, context=None)
        print(f"request returned status {response['statusCode']}")
        if response["statusCode"] != 200:
            print(f"  error: {response['body']}")
    except Exception as e:
        print(f"Raise an exception resolving event: {e}")

if __name__ == "__main__":
    main()
```

### Possible Solution
The relevant code is here:
https://github.com/aws-powertools/powertools-lambda-python/blob/705e41600d2b74cf2fc319005dd9ec71b2dab1fc/aws_lambda_powertools/event_handler/middlewares/openapi_validation.py#L117-L132
An easy fix would be to raise a different exception type - maybe `RequestValidationError`?

### Steps to Reproduce

No special installation steps are required. I can reproduce the error with the latest release (3.24.0) 

3.24.0 (current latest)

3.12

### Packaging format used

PyPi

### Debugging logs

```python
n/a, can reproduce locally
```

chriselion:
**Issue number:** closes #7978

## Summary
For handlers with a body parameter, previously we would raise an uncaught NotImplementedError, which results in a 500 response.

Now, raise a new subclass of `NotImplementedError` (to maintain existing behavior), and handle the new exception and return a 415 (Unsupported Media Type) response.

### Changes

* Added a new Exception, `RequestUnsupportedContentType`
* Raise `RequestUnsupportedContentType` when trying to parse the request body but the content-type header isn't one of the known ones.
* Update the default exception handler to check for RequestUnsupportedContentType, and return a response with 415 status code.

### User experience

Previously, requests for some endpoints would receive a 500 response if they passed unexpected headers.

Now, they will receive a 415 response, which is more indicative of a client error.

---


aws_lambda_powertools/event_handler/openapi/exceptions.py
```diff
@@ -49,3 +49,14 @@ class SchemaValidationError(ValidationException):
 
 class OpenAPIMergeError(Exception):
     """Exception raised when there's a conflict during OpenAPI merge."""
+
+
+class RequestUnsupportedContentType(NotImplementedError):
```
leandrodamascena:
```suggestion
class RequestUnsupportedContentType(NotImplementedError, ValidationException):
    def __init__(self, msg: str, errors: Sequence[Any]) -> None:
        NotImplementedError.__init__(self, msg)
        ValidationException.__init__(self, errors)

```

Minor suggestion: could we make `RequestUnsupportedContentType` inherit from both `NotImplementedError` and `ValidationException`?

This keeps backward compatibility (except `NotImplementedError` still catches it) while also following the same hierarchy as `RequestValidationError` and `ResponseValidationError`. That way except `ValidationException` catches all validation-related errors consistently.

Let me know what do you think.

leandrodamascena (changes requested):
Thanks for the PR! I think returning 415 is the correct behavior for unsupported content types - the previous 500 was an intentional design decision at the time, but this is a better approach. For most customers this won't break anything. However, anyone asserting `status_code == 500` in their tests for this specific scenario, or monitoring/alerting on 500s from unsupported content types, will see a change in behavior. We'll keep an eye on customer feedback and revert if needed.

I left a minor suggestion on the exception hierarchy - I'll merge once you've had a chance to review it.

leandrodamascena (changes requested):
Thanks for the PR! I think returning 415 is the correct behavior for unsupported content types - the previous 500 was an intentional design decision at the time, but this is a better approach. For most customers this won't break anything. However, anyone asserting `status_code == 500` in their tests for this specific scenario, or monitoring/alerting on 500s from unsupported content types, will see a change in behavior. We'll keep an eye on customer feedback and revert if needed.

I left a minor suggestion on the exception hierarchy - I'll merge once you've had a chance to review it.

aws_lambda_powertools/event_handler/openapi/exceptions.py
chriselion: Makes sense. I try to avoid multiple inheritance, so let me just make sure this works.

leandrodamascena: Yeah, I don't like either, especially in this case. Pls just add a comment to remove this in Powerrtools v4 and just inherit from `ValidationException`

chriselion: Updated - I'm not sure if there's a specific structure you want for the comments, so I just tried to copy from one of the ones I found.

leandrodamascena:
Mypy is reporting an error that is unrelated to this PR. Let me check it.

leandrodamascena (approved):
Thanks for addressing the feedback @chriselion! APPROVED!

PR #6119, opened 2025-02-18
fix(openapi): validate response serialization when falsy

anafalcao:
**Issue number:** #5887

### Changes

This PR fixes a bug where OpenAPI validation was not validating response serialization when the response body was a falsy value (e.g., False, 0, empty string, etc.), making the the response body validation is more strict.
It also removes Pydantic V1 compatibility. 

### User experience
Before:
```python
class Todo(BaseModel):
    title: str
    status: Literal["Pending", "Done"]

@app.get("/none")
def return_none() -> Todo:
    return None # Returns 200

@app.get("/empty_list")
def return_empty_list() -> Todo:
    return list() # Returns 200
```
After:
```python
class Todo(BaseModel):
    title: str
    status: Literal["Pending", "Done"]

@app.get("/none")
def return_none() -> Optional[Todo]:
    return None # Returns 200 -> Optional 

@app.get("/empty_list")
def return_empty_list() -> Todo:
    return list() # Returns 422 - validation error
```

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented
* [x] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #7990, opened 2026-02-10
fix(openapi): correct response validation for falsy objects

### LINKED ISSUE #5887: Bug: Open API Validation not validating response serialization when body is Falsy.
### Expected Behaviour

When:
* resolver (of type APIGatewayResolver) has `enable_validation=True` and,
* route's function is annotated with a return type `X` (e.g. a Pydantic model) and,
* route's function return value is not of type X (and cannot be casted)

Expected:
raise a 422 (`HTTPStatus.UNPROCESSABLE_ENTITY`) as the result of error in response serialization

### Current Behaviour

If function's return value is Falsy, the route is resolved to a 200 OK.

### Code snippet

```python
from typing import Literal

from aws_lambda_powertools.event_handler.api_gateway import APIGatewayRestResolver
import pytest

class Todo(BaseModel):
    title: str
    status: Literal["Pending", "Done"]


@app.get("/none")
def return_none() -> Todo:
    return None


@app.get("/empty_list")
def return_empty_list() -> Todo:
    return list()


@app.get("/empty_dict")
def return_empty_dict() -> Todo:
    return dict()


@app.get("/empty_string")
def return_empty_string() -> Todo:
    return ""


# --- Tests below ---


@pytest.fixture()
def event_factory():
    def _factory(path: str):
        return {
            "httpMethod": "GET",
            "path": path,
        }

    yield _factory


@pytest.mark.parametrize(
    "path, body",
    [
        ("/empty_dict", "{}"),
        ("/empty_list", "[]"),
        ("/none", "null"),
        ("/empty_string", ""),
    ],
    ids=["empty_dict", "empty_list", "none", "empty_string"],
)
def test_unexpected_response_validation(path, body, event_factory):
    """Test to demonstrate cases where a non-OK response is expected from APIGatewayResolver."""
    event = event_factory(path)
    response = lambda_handler(event, None)
    assert response["body"] == body


if __name__ == "__main__":
    pytest.main(["-v", __file__])
```

### Possible Solution

Replace the falsy valuation of the response body in the `_handle_response` method of the `OpenAPIValidationMiddleware` class with a more robust class/type comparison.

### Steps to Reproduce

* save snippet above to a file
* install the following dependencies
    - "aws-lambda-powertools>=3.4.1"
    - "pydantic>=2.10.5"
    - "pytest>=8.3.4"
* run the file

latest, 3.4.1

3.12

### Packaging format used

PyPi

### Debugging logs

```python

```

amin-farjadi:
**Issue number:** closes #5887

## Summary

Fix openapi response validation behaviour on Falsy return objects.

### Changes

in `openapi_validation.py`, `OpenAPIResponseValidationMiddleware` class:
- behaviour of `_handle_response` method modified
- behaviour of `_serialize_response` method modified
- `_serialize_response` method renamed to `_serialize_response_with_validation`

in `test_openapi_validation_middleware.py`:
- remove the skip from `test_validation_error_none_returned_non_optional_type` to cover issue #5887
- added `test_validation_error_different_list_returned_non_optional_type` and `test_validate_list_response` to cover additional scenarios

in `test_http_resolver_pydantic.py`:
- `test_async_handler_with_validation` marked to skip due to a separate issue

### User experience

Now Falsy response will be validated againt the OpenAPI return field (is there is any).

---


amin-farjadi:
Hi @leandrodamascena is there anything I can do for this?

dreamorosi (dismissed):
Thanks for the PR and for proposing a fix.

If I understand the issue correctly, this fix might be more complex than necessary. 

The root cause is that falsy values are being skipped due to `if response.body and response.is_json()`. 

A minimal fix would be:

```diff
- if response.body and response.is_json():
+ if response.is_json():
```

This single line change should fix both issues without more significant changes.

Do you think we could simplify the approach?

amin-farjadi:
Hi @dreamorosi,

Your suggestion will not work for `test_validation_error_none_returned_non_optional_type`

dreamorosi:
I've tested the one-liner change I suggested above and it works correctly for the `test_validation_error_none_returned_non_optional_type` test.

image

`None` is JSON-serializable and becomes `"null"`, `response.is_json()` returns `True` for `None` and the validation middleware runs and correctly fails with `422`.

leandrodamascena (changes requested):
Hey @amin-farjadi and @dreamorosi thanks a lot for working on this! I looked into both approaches here.

Andrea's suggestion works well and passes all the tests, but we had a similar solution for the PR #6119 and it ended up causing a breaking change for some customers.

I prefer your approach here @amin-farjadi. Instead of relying on the response content-type to decide if we should validate, you check if there's a return type annotation - that feels like the right thing to do and avoids the kind of edge cases where it validate the response and not check the type annotation.

@amin-farjadi before merging, please re-enable the two skipped test blocks in `test_openapi_validation_middleware.py` - they should work now:

- `test_none_returned_for_falsy_return`
- `test_custom_response_validation_error_http_code_invalid_response_none` 

Just remove the `@pytest.mark.skipif(reason="Test temporarily disabled until falsy return is fixed")` decorators from both.

amin-farjadi:
thank you for your replies @dreamorosi @leandrodamascena 

Removed the skips, Leonardo 🚀

leandrodamascena (approved):
Hey @amin-farjadi thanks a lot for fixing this! I'll approve but won't merge now because I'll release a new version next week with this fix #7901 and I don't want to change a lot of stuff in the same release.

I'll merge this PR next week and release in 2, 3 weeks max.

PR #7994, opened 2026-02-14
feat(buffer-handler): add buffering support for external loggers

### LINKED ISSUE #7918: Feature request: introduce handler-based buffering for external python loggers in powertools Logger
### Use case

I am using copy_config_to_registered_loggers from the utils module in aws_lambda_powertools.logging.

My application bundles external libraries (for example, botocore) that emit logs via Python’s standard logging APIs (stdout/stderr). I was able to successfully register the botocore loggers using the following call:

```python
from aws_lambda_powertools.logging import utils

utils.copy_config_to_registered_loggers(
    source_logger=LOGGER,
)
```

After doing this, the external loggers correctly inherit the formatter and log level from the source logger, as expected.
However, what appears to be missing is log buffering support for external libraries when buffering is enabled on the lamda.

Based on my tests, `copy_config_to_registered_loggers` does not appear to attach the buffering mechanism to external loggers such as `botocore`. As a result, their logs are not stored in memory and flushed only on error, which is the behavior I would expect given the configuration.

To be explicit:

- Buffering and flush-on-error work correctly for logs emitted through the Powertools Logger
- The same behavior does not apply to logs emitted by external libraries, even after registering them via copy_config_to_registered_loggers

### Solution/User Experience

@leandrodamascena suggested a possible implementation defining a `BufferingHandler` which would intercept logs at the `emit()` level 

```python
class BufferingHandler(logging.Handler):
    def __init__(self, buffer_cache: LoggerBufferCache, buffer_config: LoggerBufferConfig):
        super().__init__()
        self.buffer_cache = buffer_cache
        self.buffer_config = buffer_config

    def emit(self, record: logging.LogRecord) -> None:
        do_logic
		...
```
In this way as far as i understood when external logger perform a call eg. `logger.info()` it'd be possible to catch the event and buffer.
There would be then a new flag available for `copy_config_to_registered_loggers` which would accept a new parameter `include_buffering` by default set to `False` to keep backward-compatibility 

### Alternative solutions

```markdown

```

### Acknowledgment

acascell:
**Issue number:** closes #7918

## Summary
This PR implements log buffering support for external libraries when using AWS Lambda Powertools Logger with buffering enabled.

### Changes

1. **New `BufferingHandler` class** (`aws_lambda_powertools/logging/buffer/handler.py`):
   - Custom logging handler that intercepts logs from external libraries extending `logging.Handler`
   - Override and extends emit() method delegating to the source Logger's `_add_log_record_to_buffer()` function
   - Shares the same buffer cache and tracer_id as the Powertools Logger
   - Ensures external library logs are buffered and flushed together with application logs

2. **Enhanced `copy_config_to_registered_loggers()` function** (`aws_lambda_powertools/logging/utils.py`):
   - Added new `include_buffering` parameter (default: `False`)
   - Maintains backward compatibility - existing behavior unchanged when parameter is `False`

3. **Updated configuration logic** (`` - ``):
   - adds `BufferingHandler` on `include_buffering` flag

4. **Included unit and functional tests** 
- `tests/functional/logger/required_dependencies/test_logger_utils.py` 
-  `tests/unit/logger/required_dependencies/test_logger_buffer_handler.py`

### User Experience

**Before this change:**
When using Logger with buffering enabled, external library logs (botocore, mangum, etc.) were NOT buffered - they appeared immediately in CloudWatch, defeating the purpose of buffering:
```python
from aws_lambda_powertools.logging import utils
from aws_lambda_powertools.logging.config import LoggerBufferConfig

# Enable buffering on Powertools Logger
buffer_config = LoggerBufferConfig(
    max_bytes=20480,
    buffer_at_verbosity="DEBUG",
    flush_on_error_log=True
)
LOGGER = Logger(service="payment", buffer_config=buffer_config)

# Configure external loggers
utils.copy_config_to_registered_loggers(source_logger=LOGGER)

# Get external logger
botocore_logger = logging.getLogger("botocore")

def handler(event, context):
    LOGGER.debug("Processing payment")              # Buffered
    LOGGER.info("Validating request")               # Buffered
    botocore_logger.info("Making API call")         # NOT buffered (appears immediately)
    botocore_logger.debug("Request headers set")    # NOT buffered (appears immediately)
    LOGGER.error("Payment failed")                  # Flushes only LOGGER's logs
```

**Result:** External library logs appeared immediately in CloudWatch, while application logs were buffered.

**After this change:**
External library logs are now buffered alongside application logs when `include_buffering=True`:
```python
from aws_lambda_powertools.logging import utils
from aws_lambda_powertools.logging.config import LoggerBufferConfig

# Enable buffering on Powertools Logger
buffer_config = LoggerBufferConfig(
    max_bytes=20480,
    buffer_at_verbosity="DEBUG",
    flush_on_error_log=True
)
LOGGER = Logger(service="payment", buffer_config=buffer_config)

# Configure external loggers WITH buffering
utils.copy_config_to_registered_loggers(
    source_logger=LOGGER,
    include_buffering=True  # Enable buffering for external loggers
)

# Get external logger
botocore_logger = logging.getLogger("botocore")

def handler(event, context):
    LOGGER.debug("Processing payment")              # Buffered
    LOGGER.info("Validating request")               # Buffered
    botocore_logger.info("Making API call")         # Buffered
    botocore_logger.debug("Request headers set")    # Buffered
    LOGGER.error("Payment failed")                  # Flushes ALL logs (app + external)
```

**Result:** All logs (application + external libraries) are buffered together and flushed as a unit

---


dreamorosi:
Hi @acascell - thanks for the PR and sorry for the delayed review.

@leandrodamascena will most likely review this next week once he's back.

aws_lambda_powertools/logging/buffer/handler.py
```diff
@@ -0,0 +1,60 @@
...
+            Buffer configuration from the source logger
+        source_logger : Logger
+            The Powertools Logger instance to delegate buffering logic to
+        """
+        super().__init__()
+        self.buffer_cache = buffer_cache
+        self.buffer_config = buffer_config
+        self.source_logger = source_logger
+
+    def emit(self, record: logging.LogRecord) -> None:
```
leandrodamascena, line 47: I noticed the handler buffers everything regardless of level, but our Logger checks `_check_minimum_buffer_log_level` before deciding to buffer or emit. So if someone sets `buffer_at_verbosity="DEBUG"`, the Logger emits INFO logs right away - but external loggers through this handler would buffer them instead. That inconsistency could cause logs to be silently lost if the buffer is never flushed.

The fix is to add the same level check:

```python
from aws_lambda_powertools.logging.buffer.functions import _check_minimum_buffer_log_level

def emit(self, record: logging.LogRecord) -> None:
    level_name = logging.getLevelName(record.levelno)

    # If log level exceeds buffer threshold, emit directly through source logger
    if _check_minimum_buffer_log_level(self.buffer_config.buffer_at_verbosity, level_name):
        self.source_logger._logger.handle(record)
        return

    self.source_logger._add_log_record_to_buffer(
        level=record.levelno,
        msg=record.getMessage(),
        args=(),
        exc_info=record.exc_info,
        stack_info=False,
    )
```

aws_lambda_powertools/logging/utils.py
```diff
@@ -102,6 +114,17 @@ def _configure_logger(
...
+    buffer_config = getattr(source_logger, "_buffer_config", None)
+    if include_buffering and buffer_config is not None:
+        buffer_handler = BufferingHandler(
+            buffer_cache=source_logger._buffer_cache,
+            buffer_config=buffer_config,
+            source_logger=source_logger,
+        )
+        logger.addHandler(buffer_handler)
+        LOGGER.debug(f"Logger {logger} configured with BufferingHandler")
+        return  # exit earlier and don't add source handlers, would cause double logging
```
leandrodamascena, line 126: When `include_buffering=True`, the external logger gets no source handlers (the `return` skips the handler copy loop). This is fine for buffered logs since they flush through the source logger's handler. But if you add the level check from comment 1, logs above the buffer threshold need a handler to emit through - otherwise they go nowhere.

You'd need to also attach the source handlers:

```python
if include_buffering and buffer_config is not None:
    buffer_handler = BufferingHandler(
        buffer_cache=source_logger._buffer_cache,
        buffer_config=buffer_config,
        source_logger=source_logger,
    )
    logger.addHandler(buffer_handler)

    for source_handler in source_logger.handlers:
        logger.addHandler(source_handler)
    return
```

But then you'd need the `BufferingHandler` to suppress the record from propagating to the other handlers when it buffers - otherwise you get double logging. A cleaner approach might be to handle both paths inside `emit()` itself and keep the early return as-is.

leandrodamascena (changes requested):
Hey @acascell thanks a lot for working on this! I ran some tests and I think we need to fix small changes before we merge this. I just left some comments, can you please read them and see if this make sense?

Thanks a lot.

leandrodamascena (approved):
Hey @acascell thanks a lot for working on this. I just fixed small bugs and it's ready to merge.

APPROVED

hjgraca (approved):

PR #8001, opened 2026-02-20
fix(data_classes): support path parameters in APIGatewayAuthorizerResponse

### LINKED ISSUE #7979: Bug: APIGatewayAuthorizerResponse allow_route do not support /{proxy+} resource
### Expected Behaviour

Allowing route with a resource `/{proxy+}` should not produce this error:
`Invalid resource path: /{proxy+}. Path should match ^[/.a-zA-Z0-9-_\\*]+$`

```python
policy.allow_route(http_method="GET", resource="/{proxy+}")
# No error raised
```

### Current Behaviour

Currently a `ValueError` is raised when passing `"/{proxy+}"` at 

### Code snippet

```python
def lambda_handler(event: APIGatewayAuthorizerRequestEvent, _context: LambdaContext) -> dict[str, Any]:
        arn = event.method_arn
        policy = APIGatewayAuthorizerResponse.from_route_arn(
            arn=arn,
            principal_id="user",
        )
        # with event.resource = "/{proxy+}"
        policy.allow_route(http_method=event.http_method, resource=event.resource)
        return policy.asdict()
```

### Possible Solution

Modify the regex pattern affected to `path_regex` inside `APIGatewayAuthorizerResponse` class definition to accept the value `/{proxy+}`.

It can be passed to `r"^([/.a-zA-Z0-9-_\*]+)|(\{proxy\+\})$"` as example.

### Steps to Reproduce

No special installation steps are required. I can reproduce the error with the latest release (3.24.0)

latest

3.13

### Packaging format used

PyPi

### Debugging logs

```python

```

abhu85:
## Summary

Update the `path_regex` pattern in `APIGatewayAuthorizerResponse` to support API Gateway path parameters including:
- Standard path parameters: `{param}`
- Greedy path parameters: `{proxy+}`

Previously, the regex `r"^[/.a-zA-Z0-9-_\*]+$"` did not allow `{`, `}`, or `+` characters, causing `allow_route` and `deny_route` to reject valid API Gateway resource paths like `/{proxy+}`.

## Root Cause

The regex pattern at line 440 in `api_gateway_authorizer_event.py` was too restrictive and didn't account for path parameter syntax that API Gateway supports.

## Changes

- Updated `path_regex` to `r"^[/.a-zA-Z0-9-_\*{}+]+$"` to allow path parameters
- Added docstring explaining supported path formats
- Added 3 test cases for path parameter scenarios

## Test Plan

- [x] All 16 tests pass (`pytest tests/unit/data_classes/required_dependencies/test_api_gateway_authorizer.py`)
- [x] Verified exact issue case works: `builder.allow_route(http_method='GET', resource='/{proxy+}')`
- [x] Linting passes (`ruff check`)
- [x] Formatting passes (`ruff format --check`)

Fixes #7979

---

*By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.*

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/api_gateway_authorizer_event.py
+++ b/aws_lambda_powertools/utilities/data_classes/api_gateway_authorizer_event.py
@@ -437,8 +437,15 @@ class APIGatewayAuthorizerResponse:
     - https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html
     """
 
-    path_regex = r"^[/.a-zA-Z0-9-_\*]+$"
-    """The regular expression used to validate resource paths for the policy"""
+    path_regex = r"^[/.a-zA-Z0-9-_\*{}+]+$"
+    """The regular expression used to validate resource paths for the policy.
+
+    Supports standard path characters and API Gateway path parameters:
+    - Standard: `/path/to/resource`
+    - Wildcard: `/path/*` or `/path/*/resource`
+    - Path parameter: `/path/{param}`
+    - Greedy path parameter: `/{proxy+}` or `/path/{proxy+}`
+    """
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


dreamorosi:
Hi there, the issue linked to this PR has not been triaged, so proposing a fix is premature.

The contributing guidelines explain that this is a requirement, thanks for the PR but we'll be closing it for now.

PR #8036, opened 2026-03-16
feat(event_handler): add Request object for middleware access to resolved route and args

### LINKED ISSUE #4609: Feature request: HTTP resolver middleware should have access to path parameters
### Use case

HTTP resolver middlewares should have access to path parameters to help with use cases like resource-specific authnz, logging enrichment, etc. Access is already available in an undocumented manner that may be considered "private" according to PEP 8 (see "User Experience" below).

### Solution/User Experience

Current:
```python
def middleware(app: APIGatewayRestResolver, next_middleware: NextMiddleware) -> Response:
    path_params = app.context.get("_route_args")
    logger.append_keys(path_params=path_params)
    return next_middleware(app)
```
New:
```python
def middleware(app: APIGatewayRestResolver, next_middleware: NextMiddleware) -> Response:
    path_params = app.context.path_parameters
    logger.append_keys(path_params=path_params)
    return next_middleware(app)
```

### Alternative solutions

Perhaps it would be ideal to have the parsed path parameters available via `app.current_event` instead (as with API Gateway resolvers), but using `app.context` is certainly easier and more in line with the current state. Providing data in the powertools `app.current_event` that is not present in a typical Lambda handler's [`event`](https://docs.aws.amazon.com/lambda/latest/dg/services-alb.html) could also lead to confusion.

### Acknowledgment

### LINKED ISSUE #7992: Feature request: give ApiGatewayRestResolver middleware access to resolved route and route args
### Use case

We are using a few "lambdaliths" behind API Gateway (up to 10s of routes each, behind seperate base resources). We came from single lambdas for every api gateway endpoint and then moved to this approach. However we had not gotten rid of all the individual api gateway methods and integrations. We are looking to replace all of them with /<base_resource>/{proxy+} with ANY method. This all seemed to work fine, untill we ran into an issue with the authorizer middleware we created.

We created some ACLs on a resource/endpoint level, for which the simplest example looks like:
```yaml
/applications/{application_id}:
  PUT: [admin]
```
Where `/applications` is the base resource and one lambda handles all the routes behind this. To check which route is being called, our middleware accessed:
```python
app.current_event.resource
```
For which the value would be `/applications/{application_id}` if that is the route being called.
However when moving to `/applications/{proxy+}`, we no longer have access to this information. It looks like there is no simple/standard way to get the information we'd need for our authorizaion from the `app.current_event` . Besides the actual route, we also need the path parameters, since in this case we use the `application_id` for authorization. In the `app.current_event` these are now:
```python
{'proxy': '4da715ee-79d4-4e52-81cb-1ecc464708fb'}
```
Instead of what we used to have:
```python
{'application_id': '4da715ee-79d4-4e52-81cb-1ecc464708fb'}
```

However we know the information is available (powertools needs to route the request to the correct handler after the middleware after all), and can see this from the `app.context`:
```python
{'_route': <aws_lambda_powertools.event_handler.api_gateway.Route object at 0xffffb3cef410>, '_path': '/applications/4da715ee-79d4-4e52-81cb-1ecc464708fb', '_route_args': {'application_id': '4da715ee-79d4-4e52-81cb-1ecc464708fb'}}
```
And the `Route` object in the `_route` key also contains exactly the endpoint we'd need:
```python
{'method': 'GET', 'path': '/applications/<application_id>', 'openapi_path': '/applications/{application_id}', ... }
```

### Solution/User Experience

We are unsure how stable the app.context usage would be, since they are "private". So I think ideally we would just like those to be public. Either stable in the `app.context` or in the `app.current_event`.

### Alternative solutions

Another alternative would be use (parts of) the ApiGatewayResolver._resolve method, such that we can match the app.current_event.path to a route ourselves, but this seems much more involved and roundabout.

Related issue I found: https://github.com/aws-powertools/powertools-lambda-python/issues/4609

### Acknowledgment

oyiz-michael:
**Issue number:** closes #7992, closes #4609

### Changes

- Added a new `Request` class (`aws_lambda_powertools/event_handler/request.py`) that provides structured access to the resolved route pattern, path parameters, HTTP method, headers, query parameters, and body
- Added `app.request` property on `BaseRouter` — available inside middleware after route resolution
- Added automatic `Request` injection into route handlers via type annotation (declare a parameter typed as `Request` and it gets injected automatically)
- Updated `_registered_api_adapter` to detect `Request`-annotated handler parameters and inject the object (with per-route caching to avoid repeated signature inspection)
- Updated OpenAPI `get_dependant()` to skip `Request`-typed parameters so they don't appear in generated schemas
- Exported `Request` from `aws_lambda_powertools.event_handler`

### User experience

**Before:** Middleware had no access to the resolved route pattern or extracted path parameters. `app.current_event.path_parameters` only contained the raw API Gateway parameters (e.g. `{"proxy": "..."}` for `{proxy+}` routes), not the Powertools-resolved values.

**After:** Middleware and route handlers can access a `Request` object with the resolved route and parameters:

```python
from aws_lambda_powertools.event_handler import APIGatewayRestResolver, Request, Response

# Middleware usage
def auth_middleware(app: APIGatewayRestResolver, next_middleware):
    req = app.request
    print(req.route)            # "/applications/{application_id}"
    print(req.path_parameters)  # {"application_id": "4da715ee-..."}
    print(req.method)           # "PUT"
    return next_middleware(app)

app.use(middlewares=[auth_middleware])

# Handler injection via type annotation
@app.get("/applications/<application_id>")
def get_application(application_id: str, request: Request):
    user_agent = request.headers.get("user-agent")
    return {"id": application_id, "user_agent": user_agent}
```

---


leandrodamascena (approved):
Hey @oyiz-michael, thanks for this PR! This is a great addition. The `Request` object is much cleaner than having middleware dig into `app.context` or deal with API Gateway's raw `{proxy+}` path parameters. Really nice work on the tests too, you covered a lot of ground.

I went through the code and made a few improvements on top of your work.

I'll open a follow-up issue for documentation updates after we merge this. The middleware docs should show `app.request` as the recommended way to access resolved route info.

Thanks again for the contribution!

APPROVED!

PR #8126, opened 2026-04-07
refactor(event_handler): refactoring encoder file

### LINKED ISSUE #8121: Tech debt: Simplify jsonable_encoder
### Why is this needed?

`jsonable_encoder` in `openapi/encoders.py` has 11 return paths. A type-based dispatch pattern (dict of type to encoder function) would make it more readable and extensible.

Steps:
1. Create dispatch dicts for exact type matches and isinstance checks
2. Refactor the function to use dispatch instead of cascading isinstance
3. Remove the noqa

Event Handler - REST API

### Suggestion

_No response_

### Acknowledgment

leandrodamascena:
**Issue number:** closes #8121 

## Summary

Refactor json encoder function.

### Changes

### User experience

---


aws_lambda_powertools/event_handler/openapi/encoders.py
```diff
@@ -346,21 +335,22 @@ def decimal_encoder(dec_value: Decimal) -> int | float:
...
     Enum: lambda o: o.value,
-    frozenset: list,
-    deque: list,
-    GeneratorType: list,
-    Path: str,
+    PurePath: str,
     Pattern: lambda o: o.pattern,
     SecretBytes: str,
     SecretStr: str,
-    set: list,
```
svozza, line 363: How come we're removing all the list types?

leandrodamascena: Good catch as always 😄 ! The previous `list()` mapping didn't recursively encode elements, so something like `deque([Decimal("1.5")])` would keep the `Decimal` objects as-is instead of converting them.

Without them in `ENCODERS_BY_TYPE`, they fall through to `_dump_sequence` (line 132) which encodes each element properly.

svozza: Ah nice!

svozza (approved):

Issue #7895, opened 2026-01-05
Maintenance: Address SonarCloud code quality issues


leandrodamascena:
### Why is this needed?

SonarCloud is reporting code quality issues that should be addressed to maintain code health and reduce technical debt. Fixing these issues will improve code maintainability, readability, and potentially catch bugs early.

### Which area does this relate to?

Automation

### Solution

Review and fix SonarCloud reported issues. This is an umbrella issue and sub-issues will be created for each specific problem to track progress individually.

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

PR #8128, opened 2026-04-07
feat(event_handler): add Dependency injection with Depends()

### LINKED ISSUE #8099: Feature: Dependency Injection with `Depends()` for Event Handler
### Use case

I've been thinking about how to improve the Event Handler developer experience and make it more Pythonic. Today the only way to inject shared logic or resources into route handlers is `append_context()`, which is essentially an untyped `dict[str, Any]`. It works, but it has real problems:

- No type safety or IDE autocomplete -`app.context["tenant_id"]` can `KeyError` at runtime
- Auth checks, tenant extraction, and resource setup end up duplicated across handlers or pushed into `lambda_handler` as manual setup
- Testing requires building the context dict by hand or monkeypatching globals
- No composition - you can't express that "orders_table depends on dynamodb_client"

[FastAPI solved this years ago with `Depends()`](https://fastapi.tiangolo.com/tutorial/dependencies/). The pattern is well understood, widely adopted, and maps naturally to Lambda's execution model. This would be the single most impactful improvement to the Event Handler developer experience.

`append_context()` would remain for backward compatibility, but `Depends()` would make it unnecessary for most use cases.

### Solution/User Experience

```python
from aws_lambda_powertools.event_handler.openapi.params import Depends

# Dependency: extract tenant from authorizer
def get_tenant(request: Request) -> str:
    ctx = request.current_event.request_context.authorizer.get_context()
    tenant_id = ctx.get("tenant_id")
    if not tenant_id:
        raise UnauthorizedError("Missing tenant")
    return tenant_id

# Dependency: DynamoDB table (cacheable across invocations)
def get_orders_table() -> Table:
    return boto3.resource("dynamodb").Table(os.environ["ORDERS_TABLE"])

# Handler: clean, typed, testable
@app.get("/orders")
def list_orders(
    tenant_id: str = Depends(get_tenant),
    table: Table = Depends(get_orders_table),
):
    return table.query(KeyConditionExpression=Key("pk").eq(tenant_id))["Items"]
```

Dependencies can depend on other dependencies, forming a composable tree:

```python
def get_dynamodb() -> DynamoDBClient:
    return boto3.resource("dynamodb")

def get_users_table(db: DynamoDBClient = Depends(get_dynamodb)) -> Table:
    return db.Table(os.environ["USERS_TABLE"])

@app.get("/users/me")
def get_profile(
    tenant_id: str = Depends(get_tenant),
    table: Table = Depends(get_users_table),
):
    return table.get_item(Key={"pk": tenant_id})["Item"]
```

Testing becomes trivial with dependency overrides - no mocks, no monkeypatching:

```python
def test_list_orders():
    app.dependency_overrides[get_tenant] = lambda: "test-tenant"
    app.dependency_overrides[get_orders_table] = lambda: mock_table
    result = app.resolve(apigw_event, context)
    app.dependency_overrides.clear()
```

### How it compares to `append_context`

| | `append_context` | `Depends()` |
|---|---|---|
| Type safety | `dict[str, Any]` | Return type of the function |
| IDE autocomplete | No | Yes |
| Missing dependency | `KeyError` at runtime | Clear resolution error |
| Where logic lives | Outside, in `lambda_handler` | In the dependency function |
| Testability | Build dict manually | `dependency_overrides` |
| Composition | Flat | Recursive tree |
| Reusability | Copy-paste `append_context` calls | Import the function |

### Implementation plan

Split into 3 PRs, each independently mergeable.

**1. Core `Depends` class and dependency resolution**

1. Add `Depends` class to `openapi/params.py` with `dependency: Callable` and `use_cache: bool`
2. Extend `get_dependant()` in `openapi/dependant.py` to detect `Depends` in function signatures and recursively build sub-dependant trees
3. Add `_solve_dependencies()` to resolve the tree bottom-up with per-invocation cache
4. Wire into the route resolution flow in `api_gateway.py` - after parameter extraction, before handler call
5. Dependencies that type-hint `Request` receive the current request object automatically
6. No changes to existing parameter handling - `Depends` is purely additive

**2. Dependency overrides for testing**

1. Add `dependency_overrides: dict[Callable, Callable]` to `ApiGatewayResolver`
2. During resolution, check overrides before calling the original dependency
3. Router composition: child routers inherit parent's overrides

**3. OpenAPI integration**

1. When a dependency's parameter has a security type (OAuth2, APIKey, etc.), propagate to the route's OpenAPI security requirements
2. Security scheme dependencies auto-register in the OpenAPI components
3. This makes security enforcement and documentation a single declaration

### What we would NOT do (vs FastAPI)

- **Generator/yield dependencies** - In FastAPI these handle setup/teardown (e.g. DB sessions). Lambda invocations are short-lived, so this adds complexity with little value
- **Async dependencies** - Powertools is sync. No need

### Caching strategy

Dependencies are cached per invocation (`resolve()` call) by default. If the same dependency is used by multiple handlers or sub-dependencies in the same invocation, it gets resolved once and the result is reused. The cache is cleared at the end of `resolve()`.

- `use_cache=True` (default): cache within the same invocation
- `use_cache=False`: resolve every time it's called

For expensive resources that should persist across invocations (boto3 clients, table references), customers can continue using module-level variables as they do today - that pattern works well and doesn't need to be replaced.

### Alternative solutions

```markdown
The current `append_context()` approach works but is untyped and requires manual setup in `lambda_handler`. The alternative would be to improve `append_context` with type hints (e.g. a TypedDict), but that still wouldn't give us composition, automatic resolution, or dependency overrides for testing, etc.
```

### Acknowledgment

leandrodamascena:
**Issue number:** closes #8099 

## Summary

This PR add `Depends()` for type-safe dependency injection in route handlers using the `Annotated` pattern. Dependencies are declared inline in the function signature - no decorators, no global state. This also supports nested dependencies, per-invocation caching, `Request` injection, and `dependency_overrides` for easy testing.

### User experience

Before this, sharing services like a DynamoDB table across handlers required `append_context` - untyped, no IDE support, manual setup in every handler.

Now you declare what a handler needs directly in its signature:

```python
import os
from typing import Any

import boto3
from typing_extensions import Annotated

from aws_lambda_powertools.event_handler.openapi.params import Depends

def get_dynamodb_resource():
    return boto3.resource("dynamodb")


def get_orders_table(dynamodb: Annotated[Any, Depends(get_dynamodb_resource)]):
    return dynamodb.Table(os.environ["ORDERS_TABLE"])


def get_users_table(dynamodb: Annotated[Any, Depends(get_dynamodb_resource)]):
    return dynamodb.Table(os.environ["USERS_TABLE"])


@app.get("/orders/<user_id>")
def get_user_orders(
    user_id: str,
    orders_table: Annotated[Any, Depends(get_orders_table)],
    users_table: Annotated[Any, Depends(get_users_table)],
):
    user = users_table.get_item(Key={"pk": user_id})["Item"]
    orders = orders_table.query(KeyConditionExpression="pk = :uid", ExpressionAttributeValues={":uid": user_id})
    return {"user": user["name"], "orders": orders["Items"]}


```

The resolver handles the rest: resolving, caching, and injecting. Dependencies can depend on other dependencies, and the whole tree is resolved automatically.

For testing, swap any dependency without monkeypatching:

```python
app.dependency_overrides[get_dynamodb_table] = lambda: mock_table
```

---

### DIFF
--- a/aws_lambda_powertools/event_handler/depends.py
+++ b/aws_lambda_powertools/event_handler/depends.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import inspect
 from typing import TYPE_CHECKING, Annotated, Any, get_args, get_origin, get_type_hints
 
 if TYPE_CHECKING:
@@ -10,6 +9,10 @@ if TYPE_CHECKING:
 
     from aws_lambda_powertools.event_handler.openapi.params import Dependant
     from aws_lambda_powertools.event_handler.request import Request
+
+
+class DependencyResolutionError(Exception):
+    """Raised when a dependency cannot be resolved."""
 
 
 class Depends:
@@ -48,26 +51,39 @@ class Depends:
     """
 
     def __init__(self, dependency: Callable[..., Any], *, use_cache: bool = True) -> None:
+        if not callable(dependency):
+            raise DependencyResolutionError(
+                f"Depends() requires a callable, got {type(dependency).__name__}: {dependency!r}",
+            )
         self.dependency = dependency
         self.use_cache = use_cache
+
+
+class _DependencyNode:
+    """Lightweight node in a dependency tree — used by ``build_dependency_tree``."""
+
+    def __init__(self, *, param_name: str, depends: Depends, sub_tree: DependencyTree) -> None:
+        self.param_name = param_name
+        self.depends = depends
+        self.dependant = sub_tree
 
 
 class DependencyTree:
     """Lightweight dependency tree — no pydantic required.
 
     This mirrors the shape that ``solve_dependencies`` expects (a ``.dependencies``
-    attribute containing ``DependencyParam`` objects), but can be built without
-    importing pydantic.
+    attribute containing nodes with ``.param_name``, ``.depends``, and ``.dependant``),
+    but can be built without importing pydantic.
     """
 
-    def __init__(self, *, dependencies: list[DependencyParam] | None = None) -> None:
-        self.dependencies: list[DependencyParam] = dependencies or []
+    def __init__(self, *, dependencies: list[_DependencyNode] | None = None) -> None:
+        self.dependencies: list[_DependencyNode] = dependencies or []
 
 
 class DependencyParam:
-    """Holds a dependency's parameter name and its resolved dependency sub-tree."""
+    """Holds a dependency's parameter name and its resolved Dependant sub-tree (OpenAPI path)."""
 
-    def __init__(self, *, param_name: str, depends: Depends, dependant: Dependant | DependencyTree) -> None:
+    def __init__(self, *, param_name: str, depends: Depends, dependant: Dependant) -> None:
         self.param_name = param_name
         self.depends = depends
         self.dependant = dependant
@@ -84,16 +100,12 @@ def _get_depends_from_annotation(annotation: Any) -> Depends | None:
 
 def _has_depends(func: Callable[..., Any]) -> bool:
     """Check if a callable has any Depends() parameters, without importing pydantic."""
-    signature = inspect.signature(func)
-    globalns = getattr(func, "__globals__", {})
+    try:
+        hints = get_type_hints(func, include_extras=True)
+    except Exception:
+        return False
 
-    for param in signature.parameters.values():
-        annotation = param.annotation
-        if isinstance(annotation, str):  # pragma: no cover - from __future__ annotations
-            try:
-                annotation = eval(annotation, globalns)  # noqa: S307
-            except Exception:
-                continue
+    for annotation in hints.values():
         if _get_depends_from_annotation(annotation) is not None:
             return True
     return False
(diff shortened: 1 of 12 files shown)


aws_lambda_powertools/event_handler/depends.py
```diff
@@ -0,0 +1,222 @@
...
+    """Lightweight node in a dependency tree — used by ``build_dependency_tree``."""
+
+    def __init__(self, *, param_name: str, depends: Depends, sub_tree: DependencyTree) -> None:
+        self.param_name = param_name
+        self.depends = depends
+        self.dependant = sub_tree
+
+
+class DependencyTree:
+    """Lightweight dependency tree — no pydantic required.
```
svozza, line 72: Just for my own curioisty, are mimicking the Pydantic interface here so that it stays compatible without us having to actually subclass any Pydantic classes?

leandrodamascena: Yep! I'm mimicking the Pydantic Dependant class, so `solve_dependencies` can works seamlessly with both:

1/ `DependencyTree` (no pydantic needed) - Customers can use Depends without Pydantic.
2/ `Dependant` (when pydantic is installed for OpenAPI). 

This way this feature works regardless of whether customers have pydantic installed or not.

leandrodamascena:
Hey @svozza can you pls review again?

svozza (approved):

svozza:
Not sure if you want to handle the Sonarqube waring about complexity but other than that looks good to me

leandrodamascena:
> Not sure if you want to handle the Sonarqube waring about complexity but other than that looks good to me

Thanks for approving this! I have this issue https://github.com/aws-powertools/powertools-lambda-python/issues/7895. I'll start working on this in the next week. There are a bunch of them that I want to fix.

PR #4502, opened 2024-06-13
feat(parser): Allow primitive data types to be parsed using TypeAdapter

sthulb:
**Issue number:** #2734

## Summary

This PR allows customers to use union types as per #2734. 

### Changes

> Please provide a summary of what's being changed

This change adopts the `TypeAdapter` from Pydantic v2 to allow the parser and base envelope to validate between the potentially different types.

### User experience

> Please share what the user experience looks like before and after this change

```python
from aws_lambda_powertools.utilities.parser import event_parser
from pydantic import BaseModel, Field
from typing import Annotated, Any, Literal, Union


class SuccessfulCallback(BaseModel):
    status: Literal["succeeded"]
    name: str
    breed: Literal["Husky", "Labrador"]


class FailedCallback(BaseModel):
    status: Literal["failed"]
    error: str


DogCallback = Annotated[
    Union[SuccessfulCallback, FailedCallback], Field(discriminator="status")
]

@event_parser(model=DogCallback)
def lambda_handler(event: DogCallback, _: Any) -> str:
    if isinstance(event, FailedCallback): # alternatively `if event.status == "failed"` (if your type checker is on top of things)
        return f"Uh oh. Had a problem: {event.error}"

```

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [X] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [X] I have performed a self-review of this change
* [X] Changes have been tested
* [x] Changes are documented
* [X] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #8175, opened 2026-04-26
fix(parser): type hints should reflect primitive types support

### LINKED ISSUE #8174: Static typing: `parser` utility type hints should reflect primitive types support
### Static type checker used

pyright/pylance

3.13

latest

### Static type checker info

```
Method "parse" overrides class "BaseEnvelope" in an incompatible manner
  Parameter 3 type mismatch: base parameter is type "type[T@parse]", override parameter is type "type[Model@parse]"
  Return type mismatch: base method returns type "NotImplementedType", override returns type "list[Model@parse | None]"
    Type "T@parse" is not assignable to type "BaseModel"
      "object*" is not assignable to "BaseModel"
    "list[Model@parse | None]" is not assignable to "NotImplementedType"
```

### Code snippet

This is identifiable by analyzing the library's source code itself.

### Possible Solution

I've created a PR that tackles this: #8175.

catarinacps:
**Issue number:** closes #8174

## Summary

As per #4502, primitive types are supported by the `parser` utility. However, the type hints utilized in the functions and envelopes there do not reflect that at the moment, making type checkers lack when inferring the return types, for example. This aims to improve this situation by adjusting the type hints there.

### Changes

This PR thus changes the `TypeVar` used in the envelope types' functions to `T`, introduced in #4502. It also sets the `BaseEnvelope` `parser(...)` method with envelope overload return type hint to `T | list[T | None] | None` to allow those subclasses to overload the method correctly.

### User experience

Before:

image
<br>

After the change the type checker will be able to narrow the type to `list[EUMMessage]`, as that is the actual return type that should be shown.

---


leandrodamascena (dismissed):
Oi @catarinacps, tudo bem? :D 

This is a great contribution, clean and consistent. Swapping `Model` (bound to `BaseModel`) to unbound `T` across all envelopes finally makes type checkers aware of the primitive types support we added in #4502. Love it!

APROVADO!

leandrodamascena (approved):

PR #8194, opened 2026-05-02
fix(event_handler): fix ALB resolver returns when response body is None

### LINKED ISSUE #8186: Bug: Inconsistent validation of empty responses between ALBResolver and APIGatewayHttpResolver
### Expected Behaviour

I'm trying to update to a newer version of `aws-lambda-powertools`, and I'm hitting problems for endpoints that return "empty" 204 responses (e.g. for a DELETE endpoint). For example:

```python
def delete_endpoint() -> Response[None]:
    return Response(
        status_code=204,
        body=None,
    )
```
seems to follow the recommendations for the "Using Response with data validation?" section in the docs [here](https://docs.aws.amazon.com/powertools/python/latest/core/event_handler/api_gateway/#fine-grained-responses).

I would expect both ALBResolver and APIGatewayHttpResolver to treat this the same, and they agree in `aws-lambda-powertools==3.25.0` but not in `aws-lambda-powertools==3.26.0` and higher.

My suspicion is that this modification of the response body is the cause:
https://github.com/aws-powertools/powertools-lambda-python/blob/08c99219e000738c7be3e5b22338566dcb76b987/aws_lambda_powertools/event_handler/api_gateway.py#L3361-L3364

### Current Behaviour

The ALBResolver returns a response like
```python
{
  "statusCode": 422,
  "body": "{\"statusCode\":422,\"detail\":[{\"loc\":[\"response\"],\"type\":\"none_required\"}]}",
  "isBase64Encoded": false,
  "headers": {
    "Content-Type": "application/json"
  }
}
```

### Code snippet

```python
# /// script
# dependencies = [
#   "pydantic==2.12.0",
#   "aws-lambda-powertools==3.28.0",
# ]
# ///
import json
from typing import Any
from aws_lambda_powertools.event_handler import ALBResolver, APIGatewayHttpResolver, Response


def build_alb_event(*, path: str, method: str) -> dict[str, Any]:
    event: dict[str, Any] = {
        "httpMethod": method,
        "path": path,
        "requestId": "some_request_id",
        "requestContext": {"elb": {"targetGroupArn": ":target:"}},
    }
    return event


def build_api_gateway_event(*, path: str, method: str) -> dict[str, Any]:
    event: dict[str, Any] = {
        "rawPath": path,
        "requestContext": {
            "requestContext": {"requestId": "some_request_id"},
            "http": {"method": method},
            "stage": "$default",
        },
    }

    return event

alb_app = ALBResolver(enable_validation=True)
api_gateway_app = APIGatewayHttpResolver(enable_validation=True)

# I tried multilple return types for the endpoint
# - No return type: OK
# - typing.Any: OK
# - None: 422 from ALB
# - Response[None]: 422 from ALB
# - Response[str]: 422 from API Gateway
# - Response[str] with body="": OK

def delete_endpoint() -> Response[None]:
    return Response(
        status_code=204,
        body=None,
    )

alb_app.route("test", "DELETE")(delete_endpoint)
api_gateway_app.route("test", "DELETE")(delete_endpoint)

alb_resp = alb_app.resolve( build_alb_event(path="test", method="DELETE"), context={} )
api_resp = api_gateway_app.resolve( build_api_gateway_event(path="test", method="DELETE"), context={})

print(f"alb: {json.dumps(alb_resp, indent=2)}")
print(f"api: {json.dumps(api_resp, indent=2)}")
```

### Possible Solution

As noted in the snippet, I tried a few different return type annotations on the function. Omitting the return type, or using `typing.Any` seems to fix (or at least hide) the problem. The only fix that maintains typing while fixing both ALB and APIGateway behavior seems to returning an empty string instead of `None` for the `Response` body, and using `Response[str]` as the type annotation. 

### Steps to Reproduce

Run the attached script. I used `uv run repro_204.py`, but installing with `pip` should also work.

3.26.0 and higher

3.12

### Packaging format used

Lambda Layers

### Debugging logs

```python

```

leandrodamascena:
**Issue number:** closes #8186

## Summary

ALB resolver converts `body=None` to `body=""` to prevent ALB 5xx errors, but this conversion happens before the response validation middleware runs. Validation then sees `""` instead of `None` and fails with `none_required` for handlers returning `-> None`.

Fix: in the validation middleware, treat empty string body as `None` when the return type is `None`.

### Changes

### User experience

---


tests/functional/event_handler/_pydantic/test_openapi_validation_middleware.py
```diff
@@ -4227,3 +4227,21 @@ def handler(session_id: Annotated[str, Cookie()]):
...
+def test_alb_response_none_body_with_validation(gw_event_alb):
+    # GIVEN an ALBResolver with validation enabled
+    app = ALBResolver(enable_validation=True)
+
+    gw_event_alb["path"] = "/no-content"
+    gw_event_alb["httpMethod"] = "DELETE"
+
+    # WHEN a handler returns Response with body=None and return type is None
+    @app.delete("/no-content")
+    def handler() -> None:
```
chriselion, line 4241: Should this be `-> Response[None]`?

leandrodamascena: Hi @chriselion, thank you so much for that! You're right, I hadn't noticed that test. I added the test and made some changes to the code to make the response type more robust.

dreamorosi (approved):

PR #8243, opened 2026-05-22
feat(batch): support FIFO entire batch failure override

### LINKED ISSUE #7596: Feature request: Support raise_on_entire_batch_failure in SqsFifoPartialProcessor
### Use case

`BatchProcessor` supports a `raise_on_entire_batch_failure` option (through its parent class, `BasePartialBatchProcessor`). `SqsFifoPartialProcessor` extends from `BatchProcessor`, but doesn't allow the `raise_on_entire_batch_failure` option to be set, so the default value of `True` is always used.

For the same reason that `raise_on_entire_batch_failure` is a useful option on `BatchProcessor` (as [explained in the docs](https://docs.aws.amazon.com/powertools/python/latest/utilities/batch/#working-with-full-batch-failures)), that option would be equally useful on `SqsFifoPartialProcessor`. We want to configure both Standard and FIFO queues to behave the same (not raise `BatchProcessingError` when all records in the batch fail to process), but right now `aws-lambda-powertools` only supports that option for Standard queues.

### Solution/User Experience

A simple solution would be to add `raise_on_entire_batch_failure` to the constructor of `SqsFifoPartialProcessor` and pass it when the superclass constructor is invoked:

```python
class SqsFifoPartialProcessor(BatchProcessor):

    def __init__(self, model: BatchSqsTypeModel | None = None, skip_group_on_error: bool = False, raise_on_entire_batch_failure: bool = True):
        """
        Initialize the SqsFifoProcessor.

        Parameters
        ----------
        model: BatchSqsTypeModel | None
            An optional model for batch processing.
        skip_group_on_error: bool
            Determines whether to exclusively skip messages from the MessageGroupID that encountered processing failures
            Default is False.
        raise_on_entire_batch_failure: bool
            Raise an exception when the entire batch has failed processing.
            When set to False, partial failures are reported in the response

        """
        self._skip_group_on_error: bool = skip_group_on_error
        self._current_group_id = None
        self._failed_group_ids: set[str] = set()
        super().__init__(EventType.SQS, model, raise_on_entire_batch_failure)
```

### Alternative solutions

```markdown

```

### Acknowledgment

rohan-patnaik:
Fixes #7596

## Summary
This lets `SqsFifoPartialProcessor` use the existing `raise_on_entire_batch_failure` option that `BatchProcessor` already supports.

## Why
Before this change, FIFO batch processing always raised `BatchProcessingError` when every record ended up failed, even when callers wanted the Lambda partial failure response instead. That made the FIFO processor behave differently from the standard batch processor for the same option.

## Test plan
- `.venv/bin/python -m pytest tests/functional/batch/required_dependencies/test_utilities_batch.py -k 'sqs_fifo_batch_processor'`
- `.venv/bin/python -m ruff format --check aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py tests/functional/batch/required_dependencies/test_utilities_batch.py`
- `.venv/bin/python -m ruff check aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py tests/functional/batch/required_dependencies/test_utilities_batch.py`

---

### DIFF
--- a/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
+++ b/aws_lambda_powertools/utilities/batch/sqs_fifo_partial_processor.py
@@ -66,7 +66,12 @@ class SqsFifoPartialProcessor(BatchProcessor):
         None,
     )
 
-    def __init__(self, model: BatchSqsTypeModel | None = None, skip_group_on_error: bool = False):
+    def __init__(
+        self,
+        model: BatchSqsTypeModel | None = None,
+        skip_group_on_error: bool = False,
+        raise_on_entire_batch_failure: bool = True,
+    ):
         """
         Initialize the SqsFifoProcessor.
 
@@ -77,12 +82,15 @@ class SqsFifoPartialProcessor(BatchProcessor):
         skip_group_on_error: bool
             Determines whether to exclusively skip messages from the MessageGroupID that encountered processing failures
             Default is False.
+        raise_on_entire_batch_failure: bool
+            Raise an exception when the entire batch has failed processing.
+            When set to False, partial failures are reported in the response.
 
         """
         self._skip_group_on_error: bool = skip_group_on_error
         self._current_group_id = None
         self._failed_group_ids: set[str] = set()
-        super().__init__(EventType.SQS, model)
+        super().__init__(EventType.SQS, model, raise_on_entire_batch_failure)
 
     def _process_record(self, record):
         self._current_group_id = record.get("attributes", {}).get("MessageGroupId")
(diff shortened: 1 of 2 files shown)


rohan-patnaik:
@hjgraca, is this still something you’d like to move forward with? I can refresh the branch and resolve the current conflict. Thanks!

rohan-patnaik:
@hjgraca, refreshed against develop and resolved the constructor conflict while preserving positional logger compatibility; all 47 batch tests pass.
Could you take another look when you have a chance? Happy to adjust anything else.Thanks!

PR #8253, opened 2026-05-30
ci: add ActionScope GitHub Actions security exposure scan

r12habh:
## What

Adds a lightweight ActionScope workflow to scan GitHub Actions, Terraform, and IAM/policy JSON changes for CI/CD security exposure.

The workflow is intentionally conservative:

- runs only when workflow/action/IaC/policy files change, plus manual dispatch
- uses only `contents: read`
- does not call AWS APIs or require cloud credentials
- pins `actions/checkout` to a full commit SHA
- installs `actionscope>=0.3.5,<1.0` from PyPI
- fails only on `critical` findings, so the current non-critical findings do not block CI

## Why

I ran ActionScope locally against this repository and it found workflow-level CI/CD security signal around AWS credentials, token permissions, environments, or action supply-chain posture.

```text
Overall Risk: 🟠 HIGH
Workflows scanned: 29
AWS credential sources: 31
Critical: 0 | High: 26 | Medium: 22 | Low: 106 | Info: 20
```

Because this uses `--fail-on critical`, this PR should add visibility without changing the current pass/fail posture.

### DIFF
--- a/.github/workflows/actionscope.yml
+++ b/.github/workflows/actionscope.yml
@@ -1,0 +1,39 @@
+name: ActionScope
+on:
+  pull_request:
+    paths:
+      - ".github/actions/**"
+      - ".github/workflows/**"
+      - "**/*.tf"
+      - "**/*.tf.json"
+      - "**/*iam*.json"
+      - "**/*policy*.json"
+  push:
+    branches:
+      - develop
+    paths:
+      - ".github/actions/**"
+      - ".github/workflows/**"
+      - "**/*.tf"
+      - "**/*.tf.json"
+      - "**/*iam*.json"
+      - "**/*policy*.json"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  scan:
+    name: GitHub Actions security exposure
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Install ActionScope
+        run: python3 -m pip install "actionscope>=0.3.5,<1.0"
+
+      - name: Scan workflow exposure
+        run: actionscope scan . --fail-on critical --no-color


leandrodamascena:
Hey @r12habh, thanks for opening this.

I looked into this and I have to be very transparante with you: the package you're adding (`actionscope`) is authored by you, was first published two weeks ago, and had multiple releases just today. This PR looks like self-promotion rather than a genuine security contribution.

A few things that stand out:

- There's no issue opened first to discuss whether we need this
- The PR doesn't show any actual findings, it just show a generic summary
- The tool has no adoption or independent validation yet
- We already have security scanning in place for our CI/CD pipelines

If you genuinely found security concerns in our workflows, the right approach would be to open an issue (or use private disclosure for sensitive findings) describing what you found, not to add your own tool to our CI.

We're going to close this PR. No hard feelings, but we can't add third-party tools to our pipeline without a clear need and proper vetting, especially from packages with no track record.

Closing this.

r12habh:
Hey @leandrodamascena, thank you for the direct response. Every one of your points is fair and I owe you a real answer rather than excuses.

**The disclosure I should have led with:** I'm the author of ActionScope. The PR didn't make that clear, and that's the single biggest miss here. Conflict of interest belongs on line one of any tooling-promotion PR, not implied or omitted.

**On the broader pattern you noticed:** Yes, this was part of a batch. I scanned a list of AWS-heavy public repos and opened similar PRs across several at once. From your side that reads as a coordinated promotion campaign, not a security contribution, and I should have predicted that. The bot warning on this PR about "no related issue" was exactly the signal to stop and I didn't.

**On the project being 2 weeks old with several releases today:** Accurate. The releases today came from each batch of real-repo live scans, including yours, surfacing actual bugs in the tool that I fixed and shipped: silent JSON file-cap truncation, false-positive SHA-pin findings, etc. In isolation that's just shipping discipline. From the outside it reads exactly like "unstable project from an unknown author asking me to introduce a dependency." That's the right read.

**Briefly, why I built it:** I kept seeing threads on Hacker News and Reddit where developers were asking the same question but getting no good answer: if my GitHub Actions workflow is compromised, what can an attacker actually do in my AWS account? People would post their CI setup and nobody could tell them. `actionlint` validates YAML syntax, `zizmor` and Scorecard catch workflow patterns, Checkov audits IAM policies in isolation, but none of them cross the workflow to IAM boundary. I built ActionScope to try to answer that specific question. I also ran a public measurement study across 493 repos and 3,981 workflows before building, partly to verify the problem was real and partly because I wanted the design grounded in actual data. Source and study: https://github.com/r12habh/ActionScope/tree/main/research

**The actual findings on this repo** (not a templated summary):

```
Workflows: 28 | Credential Sources: 31 | Overall Risk: HIGH

  All 31 AWS credential sources use OIDC (zero static keys, excellent)
  0 known-compromised actions
  0 script-injection patterns
  0 artifact-poisoning patterns

  9 GitHub Environment hardening findings (MEDIUM):
    on_push_docs.yml → release-docs (deploy_without_environment)
    publish_v3_layer.yml → beta, prod, sar-beta, sar-prod (4 jobs)
    layer_govcloud*.yml → download, copy_east, copy_west (4 jobs)

  25 HIGH-risk GITHUB_TOKEN permissions (mostly contents:write and
  packages:write where you would expect them, not regression material)
```

The honest takeaway is that your CI is already in great shape. The only meaningful surface is the 9 environment-protection findings. For the layer-publish workflows specifically, declaring `environment: production` or equivalent would let you gate prod and SAR pushes behind Required Reviewers in Settings → Environments. Real value but small.

**I'm closing this PR.** If ActionScope ever earns the kind of track record that would justify proposing it as a CI dependency to a project like this one, the path is opening an issue first to discuss whether you want it, not a templated PR. The other open PRs from the same batch are getting the same treatment.

Thanks again for taking the time. The course correction is on me.

PR #8268, opened 2026-06-10
fix(event-handler): handle CORS preflight OPTIONS in HttpResolverLocal

### LINKED ISSUE #8267: Bug: OPTIONS call returning 500 in HttpResolverLocal
### Expected Behaviour

When cors=CORSConfig(...) is passed to HttpResolverLocal, browser preflight (OPTIONS) requests should return 204 with the appropriate Access-Control-* headers, just like APIGatewayRestResolver does.

### Current Behaviour

OPTIONS requests fall through to _handle_not_found_async, which has no CORS preflight branch. They get treated as 404s, and because the not-found path uses lookup_exception_handler(NotFoundError) which walks the MRO, any generic @app.exception_handler(Exception) handler picks them up and returns 500 with no CORS headers. The browser then blocks the actual request.

### Code snippet

```python
import json
from aws_lambda_powertools.event_handler import HttpResolverLocal, Response
from aws_lambda_powertools.event_handler.api_gateway import CORSConfig

app = HttpResolverLocal(cors=CORSConfig(allow_origin="*"))

@app.post("/items")
def create_item():
    return {"ok": True}

@app.exception_handler(Exception)
def handle_server_error(ex: Exception):
    return Response(
        status_code=500,
        body=json.dumps({"error": "internal"}),
    )
```

### Possible Solution

Mirror the CORS preflight branch from ApiGatewayResolver._handle_not_found into _handle_not_found_async:
```
if self._cors and self.current_event.http_method.upper() == "OPTIONS":
    headers = {
        "Access-Control-Allow-Methods": CORSConfig.build_allow_methods(self._cors_methods),
    }
    response = Response(status_code=204, content_type=None, headers=headers, body="")
else:
    # existing not-found logic
```

### Steps to Reproduce

```
uvicorn app:app --port 3001
curl -i -X OPTIONS http://127.0.0.1:3001/items \
  -H 'Origin: http://localhost:3000' \
  -H 'Access-Control-Request-Method: POST'
# → 500, no Access-Control-* headers
```

latest

3.10

### Packaging format used

PyPi

### Debugging logs

```python

```

Avinm:
**Issue number:** closes #8267

### Changes

`HttpResolverLocal` was overriding `_resolve_async`, `_call_route_async`, `_run_middleware_chain_async`, and `_handle_not_found_async`, duplicating the parent's logic but without the CORS preflight branch present in `ApiGatewayResolver._handle_not_found_async`. `OPTIONS` requests fell through to the not-found handler and, if a generic `@app.exception_handler(Exception)` was registered, returned 500 with no CORS headers — blocking browser preflight checks.

Replaces all four overrides with a single thin `_resolve_async` that calls `super()._resolve_async()` and serializes the returned `ResponseBuilder` to `dict` for the ASGI layer. The parent already handles route matching, CORS preflight (`OPTIONS` → 204), not-found, and exception handling correctly.

### User experience

**Before:** `OPTIONS` preflight to `HttpResolverLocal` with `cors=CORSConfig(...)` returns 500 with no CORS headers. Browser blocks the actual request.

**After:** `OPTIONS` preflight returns 204 with correct `Access-Control-Allow-Origin` and `Access-Control-Allow-Methods` headers, matching `APIGatewayRestResolver` behavior.

---


leandrodamascena (approved):
Amazing! APPROVED

PR #8269, opened 2026-06-12
docs(event-handler): add import path gotcha for dependency_overrides testing

### LINKED ISSUE #8215: Docs: Depends import gotcha

Absolutely loving that new Depends in the REST API handler, massive improvement to how we've done things.

Something we've noticed is when testing with dependency overrides, the imports for the dependency function have to match exactly between the test file and the handler file.

For example doing dependency overrides with get_config in a test with imports like this (despite it being the same actual function) means the override won't be applied and will instead resolve the actual get_config.

```python
# in handler.py
from depends import get_config

# in test_handler.py
from my_app.api_handler.depends import get_config
```

what you need to do would be

```python
# in handler.py
from depends import get_config

# in test_handler.py
from depends import get_config
```

This is a funky gotcha that's not immediately obvious or debuggable, so a note in the docs could be helpful.

https://docs.aws.amazon.com/powertools/python/latest/core/event_handler/api_gateway/#testing-with-dependency-overrides

### How can we improve?

I think having a call-out about this gotcha could be helpful in the testing section.

### Got a suggestion in mind?

When using `dependency_overrides`, the imported dependency reference in your test must exactly match the one used by the handler. (With sensible example, like what I've provided)

### Acknowledgment

hirenkumar-n-dholariya:
**Issue number:** fixes #8215

## Summary

Add a warning callout in the "Testing with dependency overrides" section
explaining that the import path for `dependency_overrides` must exactly
match between the handler and test file.

## Changes

Added `???+ warning` admonition with ✅/❌ tabbed examples showing:
- Correct usage: import path matches exactly between handler and test
- Wrong usage: different import path causes override to be silently ignored

## User experience

**Before:** Users encounter a non-obvious bug where `dependency_overrides`
silently has no effect when import paths don't match

**After:** Clear warning with examples explaining why import paths must
match exactly, helping users avoid this gotcha

## Acknowledgment

- [x] This PR meets the [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [x] I have read the [CONTRIBUTING](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/CONTRIBUTING.md) guide

---


leandrodamascena (approved):
Thanks @hirenkumar-n-dholariya! Approved.

hirenkumar-dholariya:
Thank you @leandrodamascena 
Your response is really prompt and motivating. Sincerely appreciate it!

Issue #7776, opened 2025-12-05
Documentation: add `serializer` to `BedrockAgentFunctionResolver` docstring


xming-cn:
### Summary

BedrockAgentFunctionResolver currently uses json.dumps as its default serializer without setting ensure_ascii=False.
As a result, all non-ASCII characters (Chinese, Japanese, emoji, etc.) are escaped into \uXXXX sequences.

This significantly reduces the quality and readability of tool outputs in Bedrock Agent workflows, and can negatively impact LLM reasoning, especially for multilingual scenarios.

Additionally, the current documentation does not mention that the serializer is configurable, making this behavior surprising and difficult for users to diagnose.

Why this is a problem
- LLM agents typically expect natural UTF-8 output, not escaped Unicode.
- Bedrock models (Claude, Llama, Titan) all fully support Unicode — no need for escaping.
- Escaped text makes responses harder to read and lowers the effectiveness of agent tools that return text.
- Non-ASCII languages become difficult to use with the default resolver.
- The serializer parameter is undocumented, so users have no indication that this behavior can be changed.

This behavior appears to be inherited from standard Python JSON defaults rather than intentional design for LLM agent workflows.

### Expected Behaviour

The BedrockAgentFunctionResolver default output serializer should preserve Unicode characters, e.g.: 
```python
lambda body: json.dumps(body, ensure_ascii=False)
```

This aligns with typical expectations for modern LLM/agent systems where multilingual natural text is common.

### Current Behaviour

Example:
```python
return {"msg": "你好"}
```

becomes:
```json
{"msg": "\u4f60\u597d"}
```

### Code snippet

```python
from aws_lambda_powertools.event_handler import BedrockAgentFunctionResolver

app = BedrockAgentFunctionResolver()

@app.tool(name="hello", description="Returns a greeting")
def hello():
    return {"msg": "你好"}

def lambda_handler(event, context):
    return app.resolve(event, context)
```

Output: `{"msg": "\u4f60\u597d"}`
Expected: `{"msg": "你好"}`

### Current Workaround

It is possible to override the serializer manually:

```python
import json
from aws_lambda_powertools.event_handler import BedrockAgentFunctionResolver

app = BedrockAgentFunctionResolver(
    serializer=lambda x: json.dumps(x, ensure_ascii=False)
)
```

This restores correct Unicode output.

However:
1. This capability is undocumented.
2. Users have no way to know the default serializer is the cause of the issue.
[...]

Issue #6394, opened 2025-04-03
Feature request: Allow customizing ApiGatewayResolver serializer instantiation parameters


alexminza:
### Use case

By default `json.dumps` is initialized with `ensure_ascii: bool = True` which escapes all non-ASCII output.
To overcome this, one has to recreate the internal serializer from scratch and pass it as a custom serializer:

```python
serializer = functools.partial(json.dumps, separators=(",", ":"), cls=json_encoder.Encoder, ensure_ascii=False)
app = APIGatewayHttpResolver(serializer=serializer)
```

This hardcodes many of the library design choices on the app side.

### Solution/User Experience

Allow passing additional parameters to the serializer initialization:

```python
app = APIGatewayHttpResolver(serializer_kwargs={'ensure_ascii': False})
```

https://github.com/aws-powertools/powertools-lambda-python/blob/develop/aws_lambda_powertools/event_handler/api_gateway.py#L1553

```python
def __init__(self, serializer_kwargs: dict):
    # Allow for a custom serializer or a concise json serialization
    self._serializer = serializer or partial(json.dumps, separators=(",", ":"), cls=Encoder, **serializer_kwargs)
```

Proposed solution is similar to the LangChain library design:
https://api.python.langchain.com/en/latest/chains/langchain.chains.llm.LLMChain.html#langchain.chains.llm.LLMChain.llm_kwargs

### Alternative solutions

Hardcode serializer args:

```python
self._serializer = serializer or partial(json.dumps, separators=(",", ":"), cls=Encoder, ensure_ascii=False)
```

### Acknowledgment

- [x] This feature request meets [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

leandrodamascena:
Hey @alexminza! Thanks for opening this! I see your point here, but since we already provide a way for customers to bring their own serializer, don't you think allowing `kwargs` in the default serializer might complicate the overall experience? Additionally, we may lose control of the default serializer and customers may make mistakes unintentionally.

> To overcome this, one has to recreate the internal serializer from scratch and pass it as a custom serializer:

[...]

PR #8271, opened 2026-06-13
feat(event_handler): allow custom serializer on BedrockAgentResolver

### LINKED ISSUE #8270: Feature request: allow customizing serializer on BedrockAgentResolver (parity with BedrockAgentFunctionResolver)
Thanks for Powertools — it's our default for Lambda.

## Use case

A Bedrock Agent action group that returns non-ASCII text (e.g. Korean `잔액은 1,000원입니다` or emoji) is `\uXXXX`-escaped in `responseBody.body`, because `BedrockResponseBuilder.build()` (`event_handler/bedrock_agent.py:42`) serializes via the default `json.dumps(..., ensure_ascii=True)` serializer (`api_gateway.py:1625`).

Unlike `ApiGatewayResolver` and the sibling `BedrockAgentFunctionResolver` — both of which expose a `serializer=` argument (as noted by the maintainer in #7776) — `BedrockAgentResolver.__init__` hardcodes `serializer=None` (`bedrock_agent.py:102-111`) with no way to override it. So there's no supported way to emit readable non-ASCII from a Bedrock Agent resolver.

## Solution/User Experience

Expose the parent's existing `serializer` parameter on `BedrockAgentResolver` (no new abstraction — the same `serializer` param #6394 endorsed and the sibling function resolver already accepts). Users could then pass `partial(json.dumps, ensure_ascii=False)`.

## Alternative solutions

Defaulting the builder to `ensure_ascii=False` would be a behavior change; exposing the existing param is additive and lower-risk.

I'd be happy to open a small PR with a regression test.

---

_This issue was prepared with the help of an AI agent (Claude Code) and reviewed by a human._

kimnamu:
### Changes

`BedrockAgentResolver` is the only OpenAPI resolver that doesn't expose `serializer`, so non-ASCII Bedrock Agent responses can't avoid `\uXXXX` escaping. This exposes the parent `ApiGatewayResolver`'s existing `serializer` argument on `BedrockAgentResolver` and passes it through — matching the sibling `BedrockAgentFunctionResolver`, which already accepts `serializer` (and the approach the maintainer endorsed in #7776 / #6394).

### User experience

| Item | Before | After |
|---|---|---|
| `BedrockAgentResolver(serializer=...)` | ❌ `TypeError: unexpected keyword 'serializer'` | ✅ accepted |
| Korean/emoji response body | ❌ `잔액...` escaped | ✅ `잔액은 1,000원입니다 💰` (with `ensure_ascii=False`) |
| Default behavior (no serializer) | escapes non-ASCII | ✅ unchanged |
| Public API / other args / response shape | — | ✅ unchanged (additive optional kwarg) |

Out of scope (follow-up): defaulting the builder to `ensure_ascii=False` (a behavior change) — intentionally not done here.

## Checklist

- Tests: 2 new GIVEN/WHEN/THEN tests in `_pydantic/test_bedrock_agent.py` (default escapes; custom `partial(json.dumps, ensure_ascii=False)` preserves non-ASCII). Reverting the source change (keeping tests) → `TypeError: ... unexpected keyword argument 'serializer'`; restored → 17 passed. `ruff`/`mypy` clean.

---

_This contribution was prepared with the help of an AI agent (Claude Code); a human reviewed the change, rationale, and test results before submission._

---


leandrodamascena (approved):
Thanks @kimnamu! Approved!

PR #8316, opened 2026-07-01
feat(circuit_breaker): support composite primary key in DynamoDB persistence

### LINKED ISSUE #8315: Feature request: Circuit breaker DynamoDB persistence, support single-table (composite key) reuse
### Use case

`CircuitBreakerDynamoDBPersistence` lets us rename attributes but assumes a partition-key-only table. Our single-table design uses a **composite key**, and DynamoDB requires the sort key on every call so the persistence layer can't operate against it. It would be inert, failing open on every call (logged at WARNING) and never trips.

### Solution/User Experience

Match the Idempotency utility's `sort_key_attr` + `static_pk_value`:

```python
CircuitBreakerDynamoDBPersistence(
    table_name="my-single-table",
    key_attr="PK", sort_key_attr="SK", static_pk_value="CIRCUIT_BREAKER",
)
```

`sort_key_attr` stays **optional**, omit it and behavior is unchanged (partition-key-only, as today). Only when you supply it does the layer write the composite key: `static_pk_value` into the partition, circuit name into the sort key. No separate table, consistent with the rest of Powertools.

Happy to do a PR for this.

### Alternative solution

A dedicated single-attribute table works but adds infra/IAM and breaks the single-table model, the same reason Idempotency added composite-key support.

### Acknowledgment

Iamrodos:
**Issue number:** closes #8315

### Changes

`CircuitBreakerDynamoDBPersistence` lets you rename attributes but assumes a partition-key-only table. Against a single-table design that uses a composite (partition + sort) key, DynamoDB requires the sort key on every call, so the layer can't operate: it fails open on every invocation (logged at WARNING) and the breaker never trips.

This adds optional `sort_key_attr` and `static_pk_value` parameters, matching the composite-key support the Idempotency utility already offers, so there's nothing new to learn for anyone using both. Set `sort_key_attr` and the circuit name is written to the sort key, with `static_pk_value` in the partition key (default `circuit_breaker#{LAMBDA_FUNCTION_NAME}`, which namespaces circuits per function; set it to a fixed value to share one circuit across functions). Omit `sort_key_attr` and behavior is unchanged — the circuit name stays the partition key, exactly as today.

The composite key is threaded through every DynamoDB path the utility uses, not just the read: the `get_item` read, the conditional `put_item` that elects the single half-open probe, and the `update_item` on state transitions. So the breaker works end to end on a composite table.

Also included: docs (`docs/utilities/circuit_breaker.md`) with a "Using a composite primary key" section, a sample-item table, and share-vs-namespace guidance; and a new example, `examples/circuit_breaker_alpha/src/working_with_composite_key.py`.

### User experience

Before, a composite-key table couldn't be reused — every call failed and the circuit never tripped:

```python
# Fails open on every invocation against a PK + SK table
persistence = CircuitBreakerDynamoDBPersistence(table_name="my-single-table")
```

Now you opt into the existing single-table layout:

```python
persistence = CircuitBreakerDynamoDBPersistence(
    table_name="my-single-table",
    key_attr="PK",
    sort_key_attr="SK",
    static_pk_value="CIRCUIT_BREAKER",
)
```

Omitting `sort_key_attr` keeps today's partition-key-only behavior, so this is fully backward compatible.

### Validation

New Stubber-based functional tests cover `get_item` / `put_item` / conditional `put_item` / `update_item` in both single-key and composite modes, and assert the read-back name comes from the sort key in composite mode. They also cover the `sort_key_attr == key_attr` guard and the namespaced default value.

```bash
poetry run pytest tests/functional/circuit_breaker_alpha tests/unit/circuit_breaker_alpha
```

---


leandrodamascena (changes requested):
Really nice work. One small thing: `static_pk_value` only does anything when `sort_key_attr` is set. If someone passes it alone it's silently ignored, which is annoying to debug. Could we warn?

```python
if static_pk_value is not None and sort_key_attr is None:
    warnings.warn(
        "static_pk_value is ignored unless sort_key_attr is also set.",
        category=PowertoolsUserWarning,
        stacklevel=2,
    )
```

Iamrodos:
Checking errors. Duplicated file name for the example. Fixing.

Iamrodos:
> Really nice work. One small thing: `static_pk_value` only does anything when `sort_key_attr` is set. If someone passes it alone it's silently ignored, which is annoying to debug.

Yes, it came up in my own reviews and I skipped doing in. Serves me right. Will also add a test like there already is for the other guards. Doing now.

leandrodamascena (approved):
Thanks a lot for addressing it @Iamrodos! APPROVED!

PR #8319, opened 2026-07-02
fix(circuit_breaker): normalize handled/ignored exceptions to a validated tuple at construction

### LINKED ISSUE #8317: Bug (circuit_breaker): handled_exceptions silently accepts a non-tuple, then raises a cryptic TypeError only when a circuit would trip
### Expected Behaviour

`CircuitBreakerConfig(handled_exceptions=...)` should either accept the common convenience forms (a single exception type, or any iterable of types) and normalize them, or reject a bad value **at construction time** with a clear error. Whatever it accepts should work for the whole lifecycle of the circuit — including when a protected call actually fails.

### Current Behaviour

If `handled_exceptions` is passed as a **`list`** (instead of a `tuple`), construction succeeds silently. The problem only surfaces later, the first time a protected call raises, from inside `CircuitBreakerConfig.counts_as_failure`:

```
TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

Because `counts_as_failure` raises *before* the failure is counted, two things happen at the worst possible moment:

1. The real downstream exception is masked by a `TypeError`.
2. The failure is never counted, so **the circuit never opens** — the breaker silently stops protecting you exactly when the dependency is failing.

**Root cause.** `counts_as_failure` uses the value directly in `isinstance`, and `__init__` stores the argument verbatim (`self.handled_exceptions = handled_exceptions`) with no normalization or validation, even though the type hint says `tuple[type[Exception], ...] | None`. A `list` passes construction and only fails at the `isinstance` call — on the failure path.

**Why this matters (and how we found it).** `counts_as_failure` only runs **when a protected call raises** — i.e. when the downstream is already unhealthy. A wrong-but-accepted value is dormant through every happy-path test and every healthy day in production, and only detonates during an outage, masking the real error and preventing the circuit from opening. We hit this while integration-testing our own adoption: we wrapped an SMTP send and drove the full `CLOSED → OPEN → HALF_OPEN → CLOSED` cycle by injecting downstream failures. A happy-path test never executes `counts_as_failure` and would never surface this — it took deliberate failure-injection. That's why the library should catch it at construction rather than leave it to each adopter's depth of testing.

### Code snippet

```python
from aws_lambda_powertools.utilities.circuit_breaker_alpha import (
    circuit_breaker,
    CircuitBreakerConfig,
)
from aws_lambda_powertools.utilities.circuit_breaker_alpha.persistence import (
    CircuitBreakerDynamoDBPersistence,
)

persistence = CircuitBreakerDynamoDBPersistence(table_name="cb")

# handled_exceptions given as a LIST, not a tuple:
config = CircuitBreakerConfig(handled_exceptions=[ConnectionError, TimeoutError])

@circuit_breaker(name="demo", persistence_store=persistence, config=config)
def call_downstream():
    raise ConnectionError("downstream is down")

# Import + build + config construction all succeed — nothing complains here.
# The bug appears only when the protected call raises:
call_downstream()
# TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
#   -> CircuitBreakerConfig.counts_as_failure -> isinstance(exception, self.handled_exceptions)
```

The relevant library code:

```python
# aws_lambda_powertools/utilities/circuit_breaker_alpha/config.py
def counts_as_failure(self, exception: Exception) -> bool:
    ...
    if self.handled_exceptions is not None:
        return isinstance(exception, self.handled_exceptions)   # a list is invalid here
```

### Possible Solution

Normalize **and** validate at construction (in `CircuitBreakerConfig.__init__` / `_validate`):

```python
def _normalize_exceptions(value, name):
    if value is None:
        return None
    # accept a single exception type OR any iterable of them
    types = (value,) if isinstance(value, type) else tuple(value)
    for exc in types:
        if not (isinstance(exc, type) and issubclass(exc, BaseException)):
            raise CircuitBreakerConfigError(f"{name} must be exception types, got {exc!r}")
    return types

# in __init__:
self.handled_exceptions = _normalize_exceptions(handled_exceptions, "handled_exceptions")
self.ignored_exceptions = _normalize_exceptions(ignored_exceptions, "ignored_exceptions")
```

This **coerces** a `list`/`set`/single type → `tuple`, **validates** each element is an exception type (clear `CircuitBreakerConfigError` instead of a cryptic `TypeError`), and runs **at build time** — moving detection off the failure path. Cost is a one-time O(n) at construction; `counts_as_failure` is unaffected.

**Consistent with powertools' own conventions** — the event handler already accepts exception classes as a single class *or* a list and normalizes them:

```python
# aws_lambda_powertools/event_handler/exception_handling.py
def register_exception_handler(func: Callable):
    if isinstance(exc_class, list):
        for exp in exc_class:
            self._exception_handlers[exp] = func
    else:
        self._exception_handlers[exc_class] = func
    return func
```

(same scalar-or-list normalization appears for `content_type`, `encoding`, `_strip_prefixes`, cookie headers). So the strict-tuple-or-cryptic-failure behaviour is the outlier. Happy to open a PR with this change plus tests.

### Steps to Reproduce

1. Build a `CircuitBreakerConfig` with `handled_exceptions` as a **list** and decorate a function with `@circuit_breaker(...)` — everything imports and constructs fine.
2. Call the protected function so it raises one of those exceptions.
3. Observe `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union` from `counts_as_failure` instead of the failure being counted; the circuit never opens.

`3.31.x` — utility is `circuit_breaker_alpha` (alpha namespace).

Python 3.14

### Packaging format used

PyPi

Iamrodos:
**Issue number:** closes #8317

### Changes

`CircuitBreakerConfig` stored `handled_exceptions`/`ignored_exceptions` verbatim and used them directly in `isinstance(exception, self.handled_exceptions)`. Passing a **list** constructed fine but raised `TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union` later — only when a circuit tripped, i.e. only under a real downstream failure. Because `counts_as_failure` raises before the failure is counted, it masks the original error and the circuit never opens.

Both params now go through a `_normalize_exceptions` step in `__init__` that accepts a single exception type or any iterable of them (normalizing to a tuple), and validates the result is a non-empty tuple of exception types — raising `CircuitBreakerConfigError` at construction instead of a cryptic `TypeError` later. Mirrors how `event_handler`'s exception handler already accepts a single class or a list. A bare `tuple(...)` (as in #8318) isn't enough: it regresses the single-type case (`tuple(SomeException)` raises) and skips validation.

### User experience

Before:

```python
CircuitBreakerConfig(handled_exceptions=[TimeoutError])   # ok at construction
# ...first time the downstream fails:
# TypeError: isinstance() arg 2 must be a type, a tuple of types, or a union
```

After: the list just works; genuinely bad input (a non-exception element, a non-iterable, an empty list, a str) fails immediately at construction with a clear `CircuitBreakerConfigError`.

### Tests

New unit tests in `test_config_and_states.py`: list/generator → tuple, single type → tuple (the #8318 regression), and rejection of an empty iterable, a non-exception element, a non-iterable, and a str. `pytest` (unit + functional), `make mypy`, ruff, and the security/complexity baselines all pass.

---

One thing I noticed while in here — totally non-blocking, just curious: the exceptions have a slightly different feel across the utility. Config validation raises `CircuitBreakerConfigError`, but `CircuitBreakerDynamoDBPersistence`'s `sort_key_attr == key_attr` check raises a bare `ValueError`, so `except CircuitBreakerError` catches every construction/config problem *except* that one. It's tiny at runtime (fails fast at setup), but since it's alpha it felt like the cheap moment to raise it, before changing it would be a breaking change. Would you want construction-time validation unified under `CircuitBreakerError`, or is the `ValueError` deliberate? Happy to fold it into this PR or do a quick follow-up — whatever suits.

---

### DIFF
--- a/aws_lambda_powertools/utilities/circuit_breaker_alpha/config.py
+++ b/aws_lambda_powertools/utilities/circuit_breaker_alpha/config.py
@@ -4,7 +4,12 @@ Configuration for the Circuit Breaker utility.
 
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 from aws_lambda_powertools.utilities.circuit_breaker_alpha.exceptions import CircuitBreakerConfigError
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
 
 
 class CircuitBreakerConfig:
@@ -24,12 +29,14 @@ class CircuitBreakerConfig:
     success_threshold : int
         Number of *consecutive* probe successes required to close a half-open circuit.
         Defaults to 3.
-    handled_exceptions : tuple[type[Exception], ...] | None
+    handled_exceptions : type[Exception] | Iterable[type[Exception]] | None
         Allowlist: only these exception types count as failures; anything else
-        propagates without affecting the circuit. Mutually exclusive with
+        propagates without affecting the circuit. Accepts a single exception type or
+        an iterable of them (normalized to a tuple). Mutually exclusive with
         ``ignored_exceptions``. Defaults to ``None`` (treated as ``(Exception,)``).
-    ignored_exceptions : tuple[type[Exception], ...] | None
-        Denylist: every exception counts as a failure *except* these. Mutually
+    ignored_exceptions : type[Exception] | Iterable[type[Exception]] | None
+        Denylist: every exception counts as a failure *except* these. Accepts a single
+        exception type or an iterable of them (normalized to a tuple). Mutually
         exclusive with ``handled_exceptions``. Defaults to ``None``.
     local_cache_max_age : int
         Seconds a circuit's state is cached in the execution environment before a
@@ -38,8 +45,9 @@ class CircuitBreakerConfig:
     Raises
     ------
     CircuitBreakerConfigError
-        If both ``handled_exceptions`` and ``ignored_exceptions`` are provided, or a
-        numeric tunable is not a positive integer.
+        If both ``handled_exceptions`` and ``ignored_exceptions`` are provided, a
+        numeric tunable is not a positive integer, or an exception allowlist/denylist
+        is empty or contains a value that is not an exception type.
 
     Example
     -------
@@ -57,10 +65,16 @@ class CircuitBreakerConfig:
         failure_threshold: int = 5,
         recovery_timeout: int = 30,
         success_threshold: int = 3,
-        handled_exceptions: tuple[type[Exception], ...] | None = None,
-        ignored_exceptions: tuple[type[Exception], ...] | None = None,
+        handled_exceptions: type[Exception] | Iterable[type[Exception]] | None = None,
+        ignored_exceptions: type[Exception] | Iterable[type[Exception]] | None = None,
         local_cache_max_age: int = 5,
     ):
+        # Normalize first: a single exception type or any iterable becomes a tuple, and a
+        # bad value fails here (at construction) rather than as a cryptic isinstance
+        # TypeError later, the first time the circuit evaluates a failure.
+        handled_exceptions = self._normalize_exceptions(handled_exceptions, "handled_exceptions")
+        ignored_exceptions = self._normalize_exceptions(ignored_exceptions, "ignored_exceptions")
+
         self._validate(
             failure_threshold=failure_threshold,
             recovery_timeout=recovery_timeout,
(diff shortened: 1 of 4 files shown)


leandrodamascena (changes requested):
Hey @Iamrodos thanks a look for sending this PR.

Since #8316 already merged, that stray `ValueError` you flagged is on `develop` now, so let's just fix it here rather than open a follow-up. I checked and it's the only raw `ValueError` left in the utility, so this one change wraps it up.

Only open question is which exception to use. `CircuitBreakerConfigError` is really about `CircuitBreakerConfig`, and `CircuitBreakerPersistenceError` is for write-path store errors, so neither is a perfect fit for a construction check. I'd go with `CircuitBreakerConfigError` or the base `CircuitBreakerError`, but you decide, I'm happy with both. Add a small test and I'll approve.

Iamrodos:
Done. Selected `CircuitBreakerConfigError` and updated existing test.

leandrodamascena:
> Done. Selected `CircuitBreakerConfigError` and updated existing test.

Thanks! Pls fix the small error with ty and then we can merge. Let me know if you need any help.

leandrodamascena (approved):
Thanks a lot @Iamrodos! APPROVED!

PR #8321, opened 2026-07-03
fix(parser): fix DynamoDB Stream Lambda invocation record model

### LINKED ISSUE #8322: Bug: The DynamoDB Stream Lambda invocation record model rejects valid records
### Expected Behaviour

The `responseContext.functionError` field of a DynamoDB Stream Lambda Invocation record is null if the error did not occur during function execution. The model should accept records with a null function error.

### Current Behaviour

The model rejects records with a null function error.

### Code snippet

```python

```

### Possible Solution

_No response_

### Steps to Reproduce

Validate a record with a null function error.

latest

3.14

### Packaging format used

PyPi

### Debugging logs

```python

```

exg:
**Issue number:** closes #8322

## Summary

The PR fixes an issue in the DynamoDB Stream Lambda Invocation record data model introduced in #7818.

### Changes

The responseContext.functionError field of a DynamoDB Stream Lambda Invocation record is null if the error did not occur during function execution.

### User experience

---

### DIFF
--- a/aws_lambda_powertools/event_handler/events_appsync/appsync_events.py
+++ b/aws_lambda_powertools/event_handler/events_appsync/appsync_events.py
@@ -321,7 +321,7 @@ class AppSyncEventsResolver(Router):
                     if isinstance(ret, Exception)
                     else {"id": e.get("id"), "payload": ret}
                 )
-                for e, ret in zip(self.current_event.events, results)
+                for e, ret in zip(self.current_event.events, results, strict=True)
             ],
         )
 
--- a/aws_lambda_powertools/utilities/parser/models/dynamodb.py
+++ b/aws_lambda_powertools/utilities/parser/models/dynamodb.py
@@ -175,7 +175,8 @@ class ResponseContext(BaseModel):
         description="The version of the Lambda executed",
         examples=["$LATEST"],
     )
-    function_error: str = Field(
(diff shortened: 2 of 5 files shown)


leandrodamascena (changes requested):
Thanks, and you're right the field is optional.

One thing: I'd make it `Optional[str] = Field(default=None, ...)`. On its own `Optional[str]` accepts an explicit `null` but still rejects the field being absent, and AWS tends to just drop these fields rather than send `null` in some cases. The default covers both. A test for the missing field would round it out, then happy to approve.

exg:
> Thanks, and you're right the field is optional.
> 
> One thing: I'd make it `Optional[str] = Field(default=None, ...)`. On its own `Optional[str]` accepts an explicit `null` but still rejects the field being absent, and AWS tends to just drop these fields rather than send `null` in some cases. The default covers both. A test for the missing field would round it out, then happy to approve.

Thanks, updated 👍 

I noticed that the ruff target version is behind so I bumped it. After the bump ruff reports 2 B905 errors and 1017 errors from ruff upgrade rules. I fixed the former and added temporary ignores for the latter, to avoid adding noise to the PR. The upgrade errors are automatically fixable, so I can easily fix them in a separate commit if you prefer.

leandrodamascena (approved):
Thanks for working on this @exg! Approved!

PR #4165, opened 2024-04-19
refactor(event_source): fix type hint

robmoss2k:
**Issue number:** #3568

### Changes

Add the `Any` return type to the event_source function definition. The `handler` argument already specifies that it will return `Any` and `event_source` returns whatever `handler` returns.

### User experience

`mypy --strict` won't throw `error: Untyped decorator makes function "handler" untyped  [misc]` any longer.

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [X] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [X] I have performed a self-review of this change
* [X] Changes have been tested
* [X] Changes are documented
* [X] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #8332, opened 2026-07-11
fix(event_source): mypy strict mode compliance

### LINKED ISSUE #8331: Static typing: event_source makes handler untyped under mypy --strict
### Static type checker used

mypy (project's standard)

3.14

3.31.0

### Static type checker info

`mypy 2.1.0` & `mypy_extensions==1.1.0`

### Code snippet

```python
"""Example handler."""

from aws_lambda_powertools.utilities.data_classes import SQSEvent, event_source


@event_source(data_class=SQSEvent)
def lambda_handler(event: SQSEvent, _context: LambdaContext) -> None:
    """..."""
```

### Possible Solution

Re-opening of #3568 with new approach for solving this.

#### Diagnosis

`event_source` defines return types as `Any`, rather than using a type variable, plus `lambda_handler_decorator`'s return type is simply `Callable` (equivalent to `Callable[..., Any]`) that mypy gets for the usage of the wrapper.

It's quite hard annotating `lambda_handler_decorator` accurately for all uses cases of it, but the specific one in `event_source.py` is simpler.

#### Proposed solution

Define the resultant type of `lambda_handler_decorator(_event_source)` as decorated, since this has `data_class` as keyword only I think this is only possible with a `Protocol`. We then cannot annotate the wrapped function directly so just `cast` afterwards.

stenczelt:
**Issue number:** closes #8331

### Changes

`aws_lambda_powertools.utilities.data_classes.event_source` complies with `mypy` in `--strict` mode as well.

Removed a redundant casting to the dataclass in `examples/event_sources/src/kinesis_firehose_response_exception.py`.

Changes are localised to `event_source.py`, rather than trying to annotate `lambda_handler_decorator` for all use cases, unlike earlier attempts, e.g. #4165. Since the decorator only takes the dataclass as a keyword argument, I've defined a protocol and cast the result to this.

### User experience

Having a handler using the `@event_source` decorator, in `main.py`

```python
from aws_lambda_powertools.utilities.data_classes import SQSEvent, event_source


@event_source(data_class=SQSEvent)
def lambda_handler(event: SQSEvent, _context: LambdaContext) -> None:
    return None

```

Before:
```
$ mypy --strict main.py

main.py:5:2: error: Untyped decorator makes function "lambda_handler" untyped  [untyped-decorator]
```

After:
```
$ mypy --strict main.py 

Success: no issues found in 1 source file
```

---

### DIFF
--- a/aws_lambda_powertools/utilities/data_classes/event_source.py
+++ b/aws_lambda_powertools/utilities/data_classes/event_source.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Protocol, TypeVar, cast
 
 from aws_lambda_powertools.middleware_factory import lambda_handler_decorator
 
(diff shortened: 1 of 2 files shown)


leandrodamascena (approved):
Amazing work @stenczelt! Thanks a lot, APPROVED!

PR #8366, opened 2026-08-05
feat(tracer): add OpenTelemetryProvider implementation (#7003)

### LINKED ISSUE #7003: Feature request: OpenTelemetry Support
### Use case

We're actively considering bringing OpenTelemetry support to Powertools for AWS Lambda 🎉 

If this is something that you'd like to see us working on, please take some time to read this RFC aws-powertools/powertools-lambda#90 and share your feedback. If you prefer, you can also email us at [aws-powertools-maintainers@amazon.com](mailto:aws-powertools-maintainers@amazon.com) using your work email ;)

This RFC focuses on Tracer and all examples are in Python, but we really want to hear your opinion and make this the best possible OTEL experience on AWS Lambda!

### Solution/User Experience

See https://github.com/aws-powertools/powertools-lambda/discussions/90

### Alternative solutions

```markdown

```

### Acknowledgment

Alwaysgaurav1:
**Issue number:** closes #7003

### Changes
- Implemented `OpenTelemetrySegment` and `OpenTelemetryProvider` in `aws_lambda_powertools.tracing.opentelemetry`.
- Exported `OpenTelemetryProvider` and `OpenTelemetrySegment` from `aws_lambda_powertools.tracing`.
- Added unit tests covering segment attribute set, exception recording, subsegment context managers, and `Tracer` decorator integration in `tests/unit/test_opentelemetry.py`.

### User experience
Users can now pass an `OpenTelemetryProvider` to `Tracer` to send trace data to OpenTelemetry backends:

```python
from aws_lambda_powertools import Tracer
from aws_lambda_powertools.tracing import OpenTelemetryProvider

provider = OpenTelemetryProvider()
tracer = Tracer(service="my-service", provider=provider)

@tracer.capture_method
def handler(event, context):
    return "ok"

### CODE REVIEW

File: aws_lambda_powertools/tracing/opentelemetry.py [hunk 1]
```diff
@@ -0,0 +1,107 @@
...
+            with self._tracer.start_as_current_span(name) as span:
+                yield OpenTelemetrySegment(span)
+        else:
+            yield OpenTelemetrySegment(None)
+
+    @contextmanager
+    def in_subsegment_async(self, name: str | None = None, **kwargs) -> Generator[BaseSegment, None, None]:
+        name = name or "subsegment"
+        if self._tracer is not None:
+            with self._tracer.start_as_current_span(name) as span:
```
iblancasa, line 77: `in_subsegment_async` is decorated with `@contextmanager`, which creates a synchronous context manager. Powertools uses this method with async with, so every async `@capture_method` call will fail at runtime

aws_lambda_powertools/tracing/opentelemetry.py
```diff
@@ -0,0 +1,107 @@
...
+        if self.span and hasattr(self.span, "record_exception"):
+            self.span.record_exception(exception)
+
+
+class OpenTelemetryProvider(BaseProvider):
+    """Tracing provider utilizing OpenTelemetry for Powertools Tracer."""
+
+    def __init__(self, tracer: Any | None = None):
+        if tracer is None:
+            try:
```
iblancasa, line 56: When OpenTelemetry is not installed, this silently creates a no-op provider. Users can deploy successfully but receive no traces.
I guess something should be, at least, printed

aws_lambda_powertools/tracing/opentelemetry.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def put_metadata(self, key: str, value: Any, namespace: str = "default") -> None:
+        if self.span and hasattr(self.span, "set_attribute"):
+            attr_key = f"{namespace}.{key}" if namespace else key
+            self.span.set_attribute(attr_key, str(value))
+
+    def add_exception(
+        self,
+        exception: BaseException,
+        stack: list[traceback.StackSummary] | None = None,
```
iblancasa, line 44: Converting metadata with str(value) loses structured data and produces Python representations rather than stable telemetry values.
It can also create very large span attributes. Please use bounded JSON serialization or record metadata as a bounded span event. The same issue exists in `put_metadata`

aws_lambda_powertools/tracing/opentelemetry.py
```diff
@@ -0,0 +1,107 @@
...
+    def put_metadata(self, key: str, value: Any, namespace: str = "default") -> None:
+        try:
+            from opentelemetry import trace
+
+            span = trace.get_current_span()
+            if span and hasattr(span, "set_attribute"):
+                attr_key = f"{namespace}.{key}" if namespace else key
+                span.set_attribute(attr_key, str(value))
+        except ImportError:
+            pass
```
iblancasa, line 101: Tracer enables automatic patching by default, but patch and patch_all silently do nothing for this provider. This means customers may assume HTTP, AWS SDK, and database calls are instrumented when they are not

aws_lambda_powertools/tracing/opentelemetry.py
```diff
@@ -0,0 +1,107 @@
...
+            attr_key = f"{namespace}.{key}" if namespace else key
+            self.span.set_attribute(attr_key, str(value))
+
+    def add_exception(
+        self,
+        exception: BaseException,
+        stack: list[traceback.StackSummary] | None = None,
+        remote: bool = False,
+    ):
+        if self.span and hasattr(self.span, "record_exception"):
```
iblancasa, line 47: Recording an exception does not set the OpenTelemetry span status to `ERROR`. Failed handlers and methods may therefore appear successful in tracing backends

iblancasa (commented):
The provider abstraction is a useful direction, but I think this needs more work before it is ready to merge.

Issue #7003 and the linked RFC describe a first-class OpenTelemetry experience that can work alongside Lambda autoinstrumentation. This PR currently provides basic child-span creation, but it does not yet define the expected behavior for dependencies, asynchronous functions, error status, metadata handling, or automatic instrumentation.

Could you please address the inline runtime issues and add tests using a real OpenTelemetry SDK/in-memory exporter rather than only mocks? In particular, we should verify async decorators, parent-child context, error status, metadata serialization and size limits and the behavior of `patch`/`patch_all`.

It would also help to document how customers select this provider and whether this is intended as a complete OpenTelemetry tracing solution or only a lower-level provider building block.

I know it has been a while since you created the PR and maybe you are not interested anymore in the feature, your priorities changed or something else. If you don't have bandwidth for working on this PR, let me know and I can take it.

PR #8367, opened 2026-08-08
fix(tracer): declare in_subsegment_async as an async context manager

### LINKED ISSUE #8365: Bug: Tracer's BaseProvider declares in_subsegment_async as a synchronous context manager
### Expected Behaviour

The type checker accepts `async with tracer.provider.in_subsegment_async(...)`.

The abstract method `in_subsegment_async` in
`aws_lambda_powertools/tracing/base.py` returns an asynchronous context
manager. Its annotation shows this:
`-> AbstractAsyncContextManager[BaseSegment]`. The declaration does not
have the `@contextmanager` decorator.

### Current Behaviour

The abstract method has the `@contextmanager` decorator. Its return
annotation is `Generator[BaseSegment, None, None]`. These declarations
tell the type checker that the method returns a synchronous context
manager.

This declaration does not agree with three parts of Powertools:

1. `Tracer._decorate_async_function` enters the result with `async with`.
2. The `capture_method` documentation shows `in_subsegment_async` with
   `async with` as the escape hatch for concurrent asynchronous functions.
3. The default provider is the recorder from `aws-xray-sdk`. Its
   `in_subsegment_async` returns an asynchronous context manager.

Thus correct code causes a type error. The user must add a `cast` to
remove the error.

### Code snippet

```python
from aws_lambda_powertools import Tracer

tracer = Tracer()


async def handler_body() -> None:
    async with tracer.provider.in_subsegment_async(name="## example"):
        ...

mypy reports:

error: "_GeneratorContextManager[BaseSegment, None, None]" has no attribute "__aenter__"  [attr-defined]
error: "_GeneratorContextManager[BaseSegment, None, None]" has no attribute "__aexit__"; maybe "__exit__"?  [attr-defined]
```

### Possible Solution

Change the declaration in BaseProvider:

1. Remove the @contextmanager decorator from the abstract method.
2. Change the return annotation to
AbstractAsyncContextManager[BaseSegment].

This change makes the interface agree with _decorate_async_function,
with the documentation, and with the shipped X-Ray provider.

### Steps to Reproduce

1. Save the code snippet above to a file.
2. Run mypy on the file.
3. See the two errors above.

3.22.0

3.14

### Packaging format used

PyPi

### Debugging logs

```python

```

vishwakt:
**Issue number:** closes #8365

### Changes

`BaseProvider.in_subsegment_async` was decorated with `@contextmanager` and annotated to return `Generator[BaseSegment, None, None]`, which declares a synchronous context manager. The method is entered with `async with` in three places: `Tracer._decorate_async_function`, the documented escape hatch for concurrent async functions in `capture_method`, and user code targeting the default X-Ray provider, whose `in_subsegment_async` returns an asynchronous context manager.

This PR removes the `@contextmanager` decorator from the abstract method and changes the return annotation to `AbstractAsyncContextManager[BaseSegment]`. No runtime change, the abstract method body is only a docstring.

### User experience

Before, mypy rejects correct code:

```python
async with tracer.provider.in_subsegment_async(name="## example"):
    ...
```

```
error: "_GeneratorContextManager[BaseSegment, None, None]" has no attribute "__aenter__"  [attr-defined]
error: "_GeneratorContextManager[BaseSegment, None, None]" has no attribute "__aexit__"; maybe "__exit__"?  [attr-defined]
```

After, the same code type checks without a cast. Verified with mypy on the snippet from the issue, on `aws_lambda_powertools/tracing`, and on `examples/tracer/src/capture_method_async_concurrency.py`. All unit tests in `tests/unit/test_tracing.py` pass.

---


leandrodamascena (approved):
Thanks for working on this @vishwakt !

PR #8389, opened 2026-08-15
fix(streaming): _S3SeekableIO.__next__/__iter__ don't advance self._position, corrupting subsequent seek()/read()

### LINKED ISSUE #8388: S3Object streaming: iterating a stream corrupts subsequent seek()/read() by silently returning the wrong bytes
### Expected Behaviour

Consuming an `S3Object`/`_S3SeekableIO` stream via the iterator protocol (`for line in s3_object: ...` or `next(s3_object)`) should keep the stream's position tracking accurate, so any subsequent `seek()` or `read()` operates from the correct byte offset.

### Current Behaviour

`_S3SeekableIO.__next__` and `__iter__` bypass position tracking entirely, so any `seek()` after iterating the stream computes the wrong S3 byte range, and the following `read()` silently returns the wrong slice of the object — no exception is raised, the data is just wrong (skipped or duplicated bytes).

### Root cause

`aws_lambda_powertools/utilities/streaming/_s3_seekable_io.py`:

Every other read path in this class carefully advances `self._position` by the number of bytes actually consumed, specifically so `seek()` can compute the correct absolute S3 byte offset:

```python
def read(self, size: int | None = -1) -> bytes:
    size = None if size == -1 else size
    data = self.raw_stream.read(size)
    if data is not None:
        self._position += len(data)
    return data

def readline(self, size: int | None = None) -> bytes:
    data = self.raw_stream.readline(size)
    self._position += len(data)
    return data
```

But `__next__`/`__iter__` don't do this bookkeeping at all:

```python
def __next__(self):
    return self.raw_stream.__next__()

def __iter__(self):
    return self.raw_stream.__iter__()
```

They hand iteration control directly to botocore's `StreamingBody`, whose own iterator reads chunks straight off the underlying stream and tracks its own internal counter — one `_S3SeekableIO` never looks at. `self._position` stays at whatever it was before iteration began.

`seek()` uses `self._position` as ground truth to build the next `Range` header on `raw_stream` access:

```python
@property
def raw_stream(self) -> PowertoolsStreamingBody:
    if self._raw_stream is None:
        range_header = f"bytes={self._position}-"
        ...
```

So after iterating N bytes and then seeking, the resulting Range header is computed from the stale pre-iteration position instead of the true current position — the reopened stream starts at the wrong byte offset.

### Concrete reproduction

```python
s3obj = S3Object(bucket=bucket, key=key)

for line in s3obj:          # consumes bytes via __iter__/__next__; self._position never advances
    ...

s3obj.seek(5, io.SEEK_CUR)  # computed as 0 + 5, not (actual bytes consumed) + 5
s3obj.read(...)             # silently reads the wrong slice of the object
```

Note: this is masked when a `GzipTransform`/`CsvTransform` is applied, because `GzipFile`/`TextIOWrapper` internally call `.read()` (which is position-tracked correctly) rather than `.__iter__()`. The bug is exposed specifically on the raw/untransformed stream's iteration protocol — a fully supported, documented usage (`S3Object`'s class docstring advertises being both seekable and streamable/iterable).

### Possible Solution

Route `__next__` through the already position-tracked `readline()`, and have `__iter__` return `self` — matching the standard Python file-iterator protocol. I have a fix + regression tests ready and will open a PR referencing this issue.

### Steps to Reproduce

Reproducible via unit test against `_S3SeekableIO`: iterate a multi-line payload fully, then check `tell()` — it stays at 0 instead of reflecting bytes consumed. A subsequent `seek()` + `raw_stream` access then requests the wrong `Range` header.

latest (`develop`, v3.34.0)

3.12

### Debugging logs

N/A — logic bug reproducible via unit test, not runtime-log-dependent.

Adityaj0:
### Issue number

Fixes #8388

### Summary

Fixes a bug where iterating an `_S3SeekableIO`/`S3Object` stream (`for line in s3_object: ...` or `next(s3_object)`) leaves `self._position` stuck at its pre-iteration value, so a subsequent `seek()` computes the wrong S3 byte range and the following `read()` silently returns the wrong slice of the object.

### Changes

`__next__` and `__iter__` previously delegated straight to `raw_stream.__next__()`/`raw_stream.__iter__()`:

```python
def __next__(self):
    return self.raw_stream.__next__()

def __iter__(self):
    return self.raw_stream.__iter__()
```

Unlike `read()`/`readline()`/`readlines()` (which all advance `self._position` by the number of bytes actually consumed), this bypassed position bookkeeping entirely. `seek()` uses `self._position` as ground truth to build the `Range` header on the next `raw_stream` access, so any `seek()` after iterating would compute an offset relative to a stale position.

Fix: route `__next__` through the already position-tracked `readline()`, and have `__iter__` return `self`, matching the standard Python file-iterator protocol:

```python
def __next__(self):
    line = self.readline()
    if not line:
        raise StopIteration
    return line

def __iter__(self):
    return self
```

This is functionally equivalent for consumers that only care about line-by-line iteration (the common case), while keeping position tracking correct for anyone who seeks after iterating.

### User experience

Before this fix, code like:
```python
for line in s3_object:
    ...
s3_object.seek(5, io.SEEK_CUR)
s3_object.read(...)
```
would silently read the wrong bytes after the seek, with no exception raised. After this fix, position tracking stays correct across iteration, so `seek()`/`read()` behave correctly regardless of whether the stream was previously consumed via `.read()` or via iteration.

### Checklist

- [x] My changes are covered by tests
- [x] I have added/updated the `develop` branch docs, if applicable — N/A, internal bug fix, no public API/behavior change for correctly-functioning code paths

**NO**. `__iter__` returning `self` instead of `raw_stream`'s iterator, and `__next__` routing through `readline()`, are both drop-in-compatible with the existing documented iteration usage (`for line in s3_object`, `next(s3_object)`) — only the previously-broken position tracking changes.

### Tests

Added to `tests/functional/streaming/_boto3/test_s3_seekable_io.py`:
- `test_next_advances_position` — verifies `tell()` correctly reflects bytes consumed via `next()`
- `test_iter_returns_self` — standard iterator protocol check
- `test_seek_after_iteration_uses_correct_range` — reproduces the real failure mode: iterate the stream, then seek, and verify the resulting `Range` header reflects true consumed position rather than the stale pre-iteration value

All three fail against the pre-fix code (verified by reverting the fix locally and re-running; the seek test shows the exact bug live: expected `bytes=28-`, received `bytes=5-`) and pass after the fix.

Full suite `tests/functional/streaming/_boto3/test_s3_seekable_io.py`: 24 passed (independently re-verified).

---

### DIFF
--- a/aws_lambda_powertools/utilities/streaming/_s3_seekable_io.py
+++ b/aws_lambda_powertools/utilities/streaming/_s3_seekable_io.py
@@ -169,10 +169,16 @@ class _S3SeekableIO(IO[bytes]):
         return self._closed
 
     def __next__(self):
-        return self.raw_stream.__next__()
+        # Route through readline() (not raw_stream.__next__()) so self._position is kept in sync with
+        # bytes actually consumed. Without this, a seek() following iteration computes offsets relative
+        # to a stale self._position, causing subsequent reads to silently return the wrong byte range.
+        line = self.readline()
+        if not line:
+            raise StopIteration
+        return line
 
     def __iter__(self):
-        return self.raw_stream.__iter__()
+        return self
 
     def __enter__(self):
         return self
--- a/tests/functional/streaming/_boto3/test_s3_seekable_io.py
+++ b/tests/functional/streaming/_boto3/test_s3_seekable_io.py
@@ -163,6 +163,65 @@ def test_next(s3_seekable_obj, s3_client_stub):
         next(s3_seekable_obj)
 
 
+def test_next_advances_position(s3_seekable_obj, s3_client_stub):
+    payload = b"hello world\nworld hello"
+    streaming_body = PowertoolsStreamingBody(raw_stream=io.BytesIO(payload), content_length=len(payload))
+
+    s3_client_stub.add_response(
+        "get_object",
+        {"Body": streaming_body},
+        {"Bucket": s3_seekable_obj.bucket, "Key": s3_seekable_obj.key, "Range": "bytes=0-"},
+    )
+
+    assert next(s3_seekable_obj) == b"hello world\n"
+    # Regression: __next__ must advance self._position the same way read()/readline() do.
+    # Before the fix, __next__ delegated straight to raw_stream.__next__(), leaving
+    # self._position stuck at 0 -- so tell() (and any seek() relative to it) would be wrong.
+    assert s3_seekable_obj.tell() == len(b"hello world\n")
+
+    assert next(s3_seekable_obj) == b"world hello"
+    assert s3_seekable_obj.tell() == len(payload)
+
+    with pytest.raises(StopIteration):
+        next(s3_seekable_obj)
+
+
+def test_iter_returns_self(s3_seekable_obj):
+    # Standard Python iterator protocol: __iter__ must return the object itself, not a
+    # separate iterator over the underlying raw stream (which would bypass position tracking).
+    assert iter(s3_seekable_obj) is s3_seekable_obj
+
+
+def test_seek_after_iteration_uses_correct_range(s3_seekable_obj, s3_client_stub):
+    payload = b"hello world\nworld hello"
+    streaming_body = PowertoolsStreamingBody(raw_stream=io.BytesIO(payload), content_length=len(payload))
+
+    s3_client_stub.add_response(
+        "get_object",
+        {"Body": streaming_body},
+        {"Bucket": s3_seekable_obj.bucket, "Key": s3_seekable_obj.key, "Range": "bytes=0-"},
+    )
+
+    # Consume the stream via the iterator protocol (e.g. "for line in s3_object: ...")
+    consumed = list(s3_seekable_obj)
+    total_consumed = sum(len(line) for line in consumed)
+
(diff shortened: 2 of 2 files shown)


leandrodamascena (changes requested):
Thanks for catching the position tracking issue. The bug is real, but I think the fix needs one adjustment before we merge it.

`StreamingBody` iteration returns 1 KiB chunks, not lines. Routing `__next__()` through `readline()` changes the existing behavior. I reproduced this with the same payload:

- current behavior: `[1024, 481]`
- this PR: `[2, 1501, 2]`

For an object without newlines, `next()` can also read the entire remaining object instead of a 1 KiB chunk.

Could we preserve the current iteration contract and only add the missing position tracking?

```python
def __next__(self):
    chunk = next(self.raw_stream)
    self._position += len(chunk)
    return chunk

def __iter__(self):
    return self
```

The tests should also use a payload larger than 1 KiB, confirm the chunk size remains unchanged, and test a partial iteration followed by `SEEK_CUR` and an actual `read()`. The current seek test consumes the whole object and then seeks beyond EOF, so it verifies the generated range but not the returned data.

leandrodamascena (approved):

PR #8391, opened 2026-08-15
fix(idempotency): is_missing_idempotency_key iterates dict keys instead of values

### LINKED ISSUE #8390: Bug: is_missing_idempotency_key iterates dict keys instead of values, missing all-None-value payloads
### Expected Behaviour

`BasePersistenceLayer.is_missing_idempotency_key` should correctly detect when the JMESPath-extracted idempotency payload is effectively empty (all values absent), regardless of whether `event_key_jmespath` resolves to a scalar, list/tuple, or dict.

### Current Behaviour

For `dict` input, `is_missing_idempotency_key` iterates `data` directly:

```python
@staticmethod
def is_missing_idempotency_key(data) -> bool:
    if isinstance(data, (tuple, list, dict)):
        return all(x is None for x in data)
    elif isinstance(data, (int, float, bool)):
        return False
    return not data
```

`for x in data` over a `dict` walks its **keys**, not its **values**. It works correctly for `list`/`tuple`, but silently checks the wrong thing for `dict`.

This matters because `event_key_jmespath` can be a multi-select expression that projects several event fields into a dict, e.g. `'{user: headers.user_id, order: body.order_id}'`. If the event lacks both fields, JMESPath still returns `{"user": None, "order": None}` — the keys are ordinary non-`None` strings, so the (wrong) key-based check evaluates `False`, i.e. "not missing," even though every value is `None`.

With `raise_on_no_idempotency_key=True`, the safety check meant to raise `IdempotencyKeyError` in this situation silently doesn't fire. With the default `False`, no warning is emitted either, and the persistence layer hashes `{"user": None, "order": None}` into a real idempotency key — so unrelated invocations that both fail to populate those fields collapse onto the same idempotency key and get incorrectly deduplicated against each other.

The existing unit test for this function only covers `{None: None}` (`None` as the *key*), which happens to still pass under the buggy key-iterating behavior (since it iterates over the single key, which is `None`) — that's why this went unnoticed.

### Code snippet

```python
from aws_lambda_powertools.utilities.idempotency.persistence.base import BasePersistenceLayer

BasePersistenceLayer.is_missing_idempotency_key({"user": None, "order": None})
# False  (expected: True)
```

End-to-end, with the safety check that should reject this:

```python
from aws_lambda_powertools.utilities.idempotency.persistence.base import BasePersistenceLayer
from aws_lambda_powertools.utilities.idempotency.config import IdempotencyConfig

class DummyLayer(BasePersistenceLayer):
    def _put_record(self, *a, **kw): pass
    def _update_record(self, *a, **kw): pass
    def _delete_record(self, *a, **kw): pass
    def _get_record(self, *a, **kw): pass

layer = DummyLayer()
layer.configure(
    IdempotencyConfig(
        event_key_jmespath="{user: headers.user_id, order: body.order_id}",
        raise_on_no_idempotency_key=True,
    ),
    function_name="test",
)

layer._get_hashed_idempotency_key(data={"headers": {}, "body": {}})
# Returns 'test-func.test#b7d94249a37b613fc73b975e0d56d7c4' instead of raising IdempotencyKeyError
```

### Possible Solution

Iterate `data.values()` for `dict` input instead of `data` itself:

```python
if isinstance(data, dict):
    return all(x is None for x in data.values())
elif isinstance(data, (tuple, list)):
    return all(x is None for x in data)
```

I have a fix and a regression test ready and will open a PR shortly.

### Steps to Reproduce

See code snippet above.

latest (reproduced on current `develop`)

python3.12

### Packaging format used

PyPi

### Debugging logs

_No response_

Adityaj0:
**Issue number:** closes #8390

### Changes

`is_missing_idempotency_key` iterated `data` directly for `dict` input:

```python
if isinstance(data, (tuple, list, dict)):
    return all(x is None for x in data)
```

`for x in data` over a `dict` walks its keys, not its values. This is correct for `list`/`tuple`, but wrong for `dict` — a dict produced by a JMESPath multi-select expression (e.g. `event_key_jmespath="{user: headers.user_id, order: body.order_id}"`) over fields that are all absent from the event resolves to `{"user": None, "order": None}`: ordinary non-`None` keys, all-`None` values. The key-based check incorrectly reports this as "not missing."

Consequence: with `raise_on_no_idempotency_key=True`, the safety check silently doesn't fire. With the default `False`, no warning is emitted and the all-None dict gets hashed into a real idempotency key, causing unrelated invocations that both fail to populate those fields to collapse onto the same key and get incorrectly deduplicated.

Fix: iterate `data.values()` for `dict` input.

```python
if isinstance(data, dict):
    return all(x is None for x in data.values())
elif isinstance(data, (tuple, list)):
    return all(x is None for x in data)
```

Added `is_missing_idempotency_key({"user": None, "order": None})` and `is_missing_idempotency_key({"user": "abc"})` cases to the existing `test_is_missing_idempotency_key` in `tests/functional/idempotency/_boto3/test_idempotency.py`. The existing test only covered `{None: None}` (`None` as the key), which happens to still pass under the buggy key-iterating code — that's why it never caught this. Confirmed the new assertion fails against the pre-fix code and passes with the fix:

```
$ python -m pytest tests/functional/idempotency/_boto3/test_idempotency.py -q
108 passed

$ python -m pytest tests/functional/idempotency --ignore=tests/functional/idempotency/_redis -q
128 passed
```

(`_redis` excluded from my local run only because of an unrelated missing `multiprocess` dependency in my environment, not related to this change.)

### User experience

Before: a dict-shaped idempotency key extraction (multi-select JMESPath) whose values are all missing is silently treated as present, either bypassing `raise_on_no_idempotency_key` entirely or generating a real (and collision-prone, since it's a constant hash) idempotency key from empty data.

After: correctly detected as missing, matching the existing behavior for list/tuple/scalar inputs.

---


leandrodamascena (approved):
Hey @Adityaj0 thank you for the contribution! I rebased the PR onto the latest `develop`, simplified the comments, and added regression coverage for missing dictionary keys.

Issue #8047, opened 2026-03-18
Maintenance: Update ruff target version to 3.10


maxrabin:
### Why is this needed?

The project currently supports all supported Python versions which, as of now, is 3.10-3.14. Despite this, the `ruff` config targets 3.8 which is not supported.

### Which area does this relate to?

Other

### Solution

Ruff target version should be bumped up to 3.10

### Acknowledgment

- [x] This request meets [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [ ] Should this be considered in other Powertools for AWS Lambda languages? i.e. [Java](https://github.com/aws-powertools/powertools-lambda-java/), [TypeScript](https://github.com/aws-powertools/powertools-lambda-typescript/), and [.NET](https://github.com/aws-powertools/powertools-lambda-dotnet/)

leandrodamascena:
Hey @maxrabin I created the first two issues and lets see how we can make progress with this!

VatsalGoel3:
Hi, I’d like to work on this. @leandrodamascena

leandrodamascena:
I think this was fixed, but pls take a look.

VatsalGoel3:
@leandrodamascena Thanks, I took a look.

The main Ruff target-version change does appear to be fixed already in the current branch: ruff.toml now has target-version = "py310", and make lint / make format-check both pass locally.

I did notice that the merged change keeps temporary Ruff ignores for UP006, UP007, UP035, and UP045, which currently hide a larger set of pyupgrade findings. I also found a few stale Python 3.8 references in tox.ini and docs.

So I think #8047 itself may be resolved if the scope was only bumping the Ruff target version. If you’d like, I can either:

open a small cleanup PR for the stale tox.ini / docs Python 3.8 references, or
work on a scoped plan for removing the temporary pyupgrade ignores safely.

Please let me know which direction would be most useful.

leandrodamascena:
Hey @VatsalGoel3, thanks a lot for digging into this and for the clear write-up.

Yeah, #8047 itself is basically done, the target bump is already in. The interesting part is the leftover you spotted: bumping to 3.10 turned on a bunch of pyupgrade rules and we silenced them (`UP006`, `UP007`, `UP035`, `UP045`) instead of actually fixing them. I'd love to clean those up now.

[...]

PR #8399, opened 2026-08-24
refactor(typing): use PEP 604 syntax for Optional annotations (UP045)

### LINKED ISSUE #8398: Maintenance: Fix UP045 ruff lint violations (Optional[X] to X | None)
### Why is this needed?

Bumping the ruff target to `py310` turned on a set of pyupgrade rules that were silenced rather than fixed. `ruff.toml` still carries temporary ignores for `UP006`, `UP007`, `UP035` and `UP045`.

In #8047 @leandrodamascena proposed clearing these one rule per PR, safest first, with each PR removing its own ignore so the fix and the removal land together. This issue covers `UP045` (`Optional[X]` to `X | None`), which was listed as one of the two safest to start with.

Current numbers on `develop`:

- 285 violations across 65 files (24 in `aws_lambda_powertools`, 25 in `examples`, 15 in `tests`)
- 283 fixed automatically by the ruff autofix
- 2 need manual work, both in `aws_lambda_powertools/utilities/batch/types.py`

The two manual cases are runtime assignments rather than annotations:

```python
BatchTypeModels = Optional[Union[...]]
BatchSqsTypeModel = Optional[Type[SqsRecordModel]]
```

Ruff correctly refuses to autofix these, since rewriting them changes the object that is actually built at import time.

Opening this as a scoped issue so a PR has something to close. Happy to fold it into the umbrella issue instead if you would rather track all four rules in one place.

Static typing

### Solution

1. Apply the `UP045` autofix across the repository.
2. Convert the two runtime assignments in `batch/types.py` by hand, verifying the resulting object compares equal to the previous definition.
3. Remove the `Optional` imports left unused by the conversion, which `F401` flags.
4. Remove `"UP045"` from `lint.ignore` in `ruff.toml`.

`Union` is deliberately left in place, since that belongs to `UP007` and keeping it out preserves the one rule per PR split.

### Acknowledgment

manshahH:
**Issue number:** closes #8398

## Summary

Clears the first of the four temporary pyupgrade ignores in `ruff.toml`, following the one rule per PR plan @leandrodamascena outlined in #8047.

`Optional[X]` becomes `X | None` across the repository, and `"UP045"` is removed from `lint.ignore` so the rule is enforced from here on.

### Changes

- 285 violations across 65 files: 24 in `aws_lambda_powertools`, 25 in `examples`, 15 in `tests`, plus `ruff.toml`
- 283 fixed by the ruff autofix
- 2 converted by hand
- 63 `Optional` imports left unused by the conversion were removed, since `F401` flags them

**The two manual cases.** Both are in `aws_lambda_powertools/utilities/batch/types.py`, where `Optional[...]` sits in a runtime assignment rather than an annotation:

```python
BatchTypeModels = Optional[Union[...]]
BatchSqsTypeModel = Optional[Type[SqsRecordModel]]
```

Ruff correctly refuses to autofix these, because the expression is evaluated at import time and a rewrite changes the object that actually gets built. I converted them by hand and checked the result against the previous definitions:

```text
Optional[Union[...]] == BatchTypeModels             -> True
Optional[Type[SqsRecordModel]] == BatchSqsTypeModel -> True
```

`Union` is deliberately left in place there, since it belongs to `UP007`. Removing it here would blur the one rule per PR split.

**Docstrings.** Twenty parameter descriptions and section comments inside files this PR already touches still referred to `Optional[...]` while the code below them had changed, so they were brought in line. Files not otherwise modified by this PR were left alone.

### User experience

No user facing change. This is internal typing syntax only, and the runtime behaviour of every converted annotation and alias is unchanged.

For contributors, `UP045` is now enforced, so the older `Optional[X]` spelling gets caught by `make lint` instead of being silently accepted.

### Verification

- `ruff check` passes with `UP045` enforced
- `ruff format --check`: 1310 files already formatted
- Full suite excluding e2e: 2497 passed. Six failures are pre existing Windows only path and formatting issues that reproduce identically on a clean `develop`
- `mypy aws_lambda_powertools examples`: same error count before and after, no new errors introduced
- bandit security baseline and xenon complexity baseline both unaffected

Happy to carry on with `UP007` and then `UP035` in the same shape once this one looks right.

## Acknowledgment

- [x] This PR meets the [Powertools for AWS Lambda (Python) Tenets](https://docs.powertools.aws.dev/lambda/python/latest/#tenets)
- [x] I have read the [CONTRIBUTING](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/CONTRIBUTING.md) guide
- [x] I have followed the [Code of Conduct](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/CODE_OF_CONDUCT.md)

---


leandrodamascena (changes requested):
Thanks for tackling this. I like this direction because it resolves the Ruff issue instead of adding another exception we would need to maintain as technical debt.

I found two small typos that currently fail at import time:

- `tests/e2e/utils/data_fetcher/logs.py:57`
- `tests/e2e/utils/data_fetcher/traces.py:109`

Could you please change `any | None` to:

```python
Any | None
```

Lowercase `any` is Python's built-in function, so the current annotation raises a `TypeError`.

manshahH:
Good catch, thank you. Fixed in 06b30b6.

Your diagnosis is right. `Optional[any]` was tolerated because `typing` never validates its argument, so the lowercase `any` sat there unnoticed. Once it became `any | None` the expression is evaluated at import time, and neither module has `from __future__ import annotations`, so importing them raised:

```text
TypeError: unsupported operand type(s) for |: 'builtin_function_or_method' and 'NoneType'
```

Both signatures are now `Any | None`, with `Any` added to the typing import in `logs.py`. I also updated the matching docstring line just below the first one, at `logs.py:64`, which carried the same lowercase `any`.

This slipped past me because `make test` runs with `--ignore tests/e2e`, so these two modules are never imported by the suite. To be sure nothing else of the same shape was hiding in the diff, I walked the AST of all 65 changed files and checked every converted annotation for a left operand that is a builtin function rather than a type. 275 annotations checked, and those were the only two.

All four `data_fetcher` modules now import cleanly, and `ruff check`, `ruff format --check` and the test suite are unchanged.

leandrodamascena (approved):
Thanks for fixing this and for checking the remaining converted annotations as well. I pulled the latest commit and confirmed that both modules now evaluate the annotations correctly, with Ruff checks also passing.

This fully addresses my review and resolves the Ruff rule without leaving another exception for us to maintain.

Thank you very much for the work. APPROVED!

PR #5653, opened 2024-11-26
feat(metrics): warn when overwriting dimension

anafalcao:
**Issue number:**
#5638 

### Changes

> Please provide a summary of what's being changed

When using dimensions with the Metrics class, the default behavior is to overwrite existing dimensions if a new dimension with the same name/key is added. To improve troubleshooting and prevent silent data loss, we've implemented a warning system that alerts users when dimension overwriting occurs.

### User experience

> Please share what the user experience looks like before and after this change

```python
from aws_lambda_powertools.metrics import Metrics

metrics = Metrics(namespace="MyApp")

metrics.add_dimension(name="test", value="foo")
metrics.add_dimension(name="test", value="bar")  # "bar" replaces "foo"

# A warning is now issued:
# UserWarning: Dimension 'test' has already been added. The previous value will be overwritten.
```

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [x] I have performed a self-review of this change
* [x] Changes have been tested
* [x] Changes are documented
* [x] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #8403, opened 2026-08-26
fix(metrics): stop spurious overwrite warnings from set_default_dimensions

### LINKED ISSUE #8402: Bug: metrics.set_default_dimensions emits PowertoolsUserWarning: Dimension 'Key' has already been added. The previous value will be overwritten.
### Expected Behaviour

When adding new default dimensions to a `Metrics` instance using `metrics.set_default_dimensions(Key=value)` there should be no WARNING logs if they were not set before.

### Current Behaviour

When adding new default dimensions to a `Metrics` instance using `metrics.set_default_dimensions(Key=value)` there are two WARNING logs being emitted:
```
[WARNING] aws_lambda_powertools/metrics/metrics.py:124: PowertoolsUserWarning: Dimension 'Key' has already been added. The previous value will be overwritten.
  self.provider.add_dimension(name=name, value=value)

[WARNING] aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py:571: PowertoolsUserWarning: Dimension 'Key' has already been added. The previous value will be overwritten.
  self.add_dimension(name, value)
```

### Code snippet

```python
metrics = Metrics()
    metrics.set_default_dimensions(Environment=os.environ["STAGE"], Resource="MyJob")
    metrics.add_metric(name="MyJobDuration", unit=MetricUnit.Seconds, value=42)
    metrics.flush_metrics()
```

### Possible Solution

https://github.com/aws-powertools/powertools-lambda-python/blob/8db13c761d6f592a4e26dab74035af29412d6634/aws_lambda_powertools/metrics/metrics.py#L176-L201

In the Metrics.set_default_dimensions method, these two code snippets seem to be doing duplicated work:
```
        self.provider.set_default_dimensions(**dimensions)
```
and
```
        for name, value in dimensions.items():
            self.add_dimension(name, value)
```
as both end up calling add_dimension in the provider, which will happen twice. The solution is not totally clear to me as I didn't go deeper into understanding the self.dimension_set, self.default_dimensions, self.provider.dimension_set and self.provided.default_dimensions relationship. Does it make sense to have these structures duplicated in self and self.provider?

### Steps to Reproduce

1. Set the following environment variables in the Lambda function:
   ```
   POWERTOOLS_METRICS_NAMESPACE: MyNamespace
   POWERTOOLS_PARAMETERS_MAX_AGE: 900
   POWERTOOLS_PARAMETERS_SSM_DECRYPT: true
   POWERTOOLS_SERVICE_NAME: MyService
   STAGE: dev
2. Add the code snipped to the Lambda function code.
3. Run the Lambda function.
4. Observe the logs containing the WARNING entries for the Environment and Resource keys.

3.34.0

3.14

### Packaging format used

Lambda Layers

### Debugging logs

```python

```

vishwakt:
**Issue number:** closes #8402

### Changes

There are two causes, and the second makes this worse than reported: even with default dimensions set once at module level, the recommended pattern, the warning fires for every default dimension on every flush.

1. `Metrics.set_default_dimensions` called `self.provider.set_default_dimensions(**dimensions)` and then re-added every dimension through `add_dimension`. The second pass always found the keys already registered, so the warning fired even on the very first call. The redundant loop is removed and the method now delegates to the provider. The docstring, which sat after the first statement and was not treated as a docstring by Python, moved to the top of the method.

2. The provider re-registers default dimensions internally: `clear_metrics` re-adds them after every flush, and warm invocations repeat `set_default_dimensions`. The previous condition in `AmazonCloudWatchEMFProvider.add_dimension` warned for any name already present in `dimension_set` or `default_dimensions`, so these internal paths warned too. It now warns only when a dimension is overwritten with a different value, which matches the warning message ("The previous value will be overwritten") and the intent of #5653, whose test covers overwriting with a different value.

3. The provider constructor replaced a falsy `default_dimensions` argument with a new dict, so the initially empty dict `Metrics` passes in was silently swapped out and updates made through the provider never reached the dict `Metrics` owns. It now keeps the given dict unless `None` is passed. Credit to @ericbn, who found this in #8404; his fix is applied here verbatim with commit co-authorship, as requested in review.

### User experience

Before: one warning per dimension on the first `set_default_dimensions` call, plus one per default dimension on every flush and on every warm re-registration.

After: no warnings when nothing is overwritten. Overwriting a dimension with a different value still warns once, including when done via `set_default_dimensions`.

Seven tests added: no warning on first call, no warning on unchanged re-registration, exactly one warning when a default value changes, no warnings across `log_metrics` invocations, no warning when re-adding a dimension with the same value, and two regression tests confirming a provided empty `default_dimensions` dict remains shared between `Metrics` and the provider. The existing `test_add_dimension_overwrite_warning` passes unchanged.

---

### DIFF
--- a/aws_lambda_powertools/metrics/metrics.py
+++ b/aws_lambda_powertools/metrics/metrics.py
@@ -174,7 +174,6 @@ class Metrics:
         )
 
     def set_default_dimensions(self, **dimensions) -> None:
-        self.provider.set_default_dimensions(**dimensions)
         """Persist dimensions across Lambda invocations
 
         Parameters
@@ -195,9 +194,7 @@ class Metrics:
             def lambda_handler():
                 return True
         """
-        for name, value in dimensions.items():
-            self.add_dimension(name, value)
-
+        self.provider.set_default_dimensions(**dimensions)
         self.default_dimensions.update(**dimensions)
 
     def clear_default_dimensions(self) -> None:
--- a/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py
+++ b/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py
@@ -87,7 +87,7 @@ class AmazonCloudWatchEMFProvider(BaseProvider):
     ):
         self.metric_set = metric_set if metric_set is not None else {}
         self.dimension_set = dimension_set if dimension_set is not None else {}
-        self.default_dimensions = default_dimensions or {}
+        self.default_dimensions = default_dimensions if default_dimensions is not None else {}
         self.namespace = resolve_env_var_choice(choice=namespace, env=os.getenv(constants.METRICS_NAMESPACE_ENV))
         self.service = resolve_env_var_choice(choice=service, env=os.getenv(constants.SERVICE_NAME_ENV))
         self.function_name = function_name
--- a/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py
+++ b/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py
@@ -317,7 +317,7 @@ class AmazonCloudWatchEMFProvider(BaseProvider):
             )
             return
 
-        if name in self.dimension_set or name in self.default_dimensions:
+        if name in self.dimension_set and self.dimension_set[name] != value:
             warnings.warn(
                 f"Dimension '{name}' has already been added. The previous value will be overwritten.",
                 category=PowertoolsUserWarning,
--- a/tests/functional/metrics/required_dependencies/test_metrics_cloudwatch_emf.py
+++ b/tests/functional/metrics/required_dependencies/test_metrics_cloudwatch_emf.py
@@ -1193,6 +1193,31 @@ def test_log_metrics_with_default_dimensions_no_warning_across_invocations(names
     assert not [warning for warning in w if "has already been added" in str(warning.message)]
 
 
+def test_provider_keeps_provided_default_dimensions_dict(namespace):
+    # GIVEN a provider constructed with an empty default dimensions dict e.g., the one Metrics shares
+    shared_default_dimensions: dict = {}
+    my_provider = AmazonCloudWatchEMFProvider(namespace=namespace, default_dimensions=shared_default_dimensions)
+
+    # WHEN default dimensions are set through the provider
+    my_provider.set_default_dimensions(environment="test")
+
+    # THEN the provided dict remains in use and receives the update
+    assert my_provider.default_dimensions is shared_default_dimensions
+    assert shared_default_dimensions == {"environment": "test"}
+
+
+def test_metrics_shares_default_dimensions_with_provider(namespace):
+    # GIVEN a Metrics instance with the default provider
+    my_metrics = Metrics(namespace=namespace)
+
+    # WHEN default dimensions are set
+    my_metrics.set_default_dimensions(environment="test")
+
+    # THEN Metrics and the provider hold the same dict, both with the update
+    assert my_metrics.default_dimensions is my_metrics.provider.default_dimensions
+    assert my_metrics.default_dimensions == {"environment": "test"}
+
+
 def test_add_dimension_no_warning_when_value_unchanged(namespace):
     # GIVEN a Metrics instance with a dimension added
     my_metrics = Metrics(namespace=namespace)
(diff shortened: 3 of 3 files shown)


ericbn:
This change is still allowing the add_dimension method in the provider to be called twice. I see there's another redundant call: `self.set_default_dimensions(**self.default_dimensions)` in the clear_metrics method in the provider.

The change to hide the warning when the new value is the same as the old one is nice, but it's hiding the redundant call that's still happening. Not sure if I'd change the condition when the warning is emitted there.

Also, I think it might be worth investigating this whole code better. I don't get why Metrics keeps its own data and I don't understand the explanation in the comment in https://github.com/aws-powertools/powertools-lambda-python/blob/8db13c761d6f592a4e26dab74035af29412d6634/aws_lambda_powertools/metrics/metrics.py#L77-L81

There's even a side effect here: since Metrics passes references of all its data structures to the provider, they're both (the Metrics and the provider) sharing the same instances of data. In this case mutating them from Metrics and from the provider is also redundant. And this does not happen if a Metrics is initialized with a provider instance given in the constructor by the user. So there's a lot of confusing complexity here, I'd say most of it possibly unintentional.

(Disclaimer: this was all reviewed and written without the use of AI)

vishwakt:
Good catch on clear_metrics, but that call isn't actually redundant. serialize_metric_set only reads dimension_set, never default_dimensions, so the re-registration in clear_metrics is what brings the defaults back after every flush. Drop it and default dimensions would only show up in the first invocation's output (test_log_persist_default_dimensions covers this).

That's also why I changed the warning condition instead of removing the call: the re-registration is doing real work, it just shouldn't be reported to the user as an overwrite. And the old condition was inaccurate on its own anyway, since it warned "the previous value will be overwritten" even when the value hadn't changed. Real overwrites still warn.

The cleaner design would be to merge default_dimensions at serialize time and never store them in dimension_set, but that touches MAX_DIMENSIONS accounting and add_dimension's override semantics, so I kept it out of this PR.

Agreed the Metrics vs provider state sharing is confusing. The class attributes are intentional (shared state across Metrics() instances, see the NOTE at metrics.py:77), but the default provider aliasing those dicts while a custom provider diverges does feel accidental. Probably worth its own issue rather than growing this PR.

ericbn:
Here's my take on this: #8404

leandrodamascena (changes requested):
Thanks for putting this together and for keeping the fix focused. Also, thank you @ericbn for spotting an important detail in #8404.

I think we should bring [this `default_dimensions` initialization change](https://github.com/aws-powertools/powertools-lambda-python/blob/11ca80ae39f9000ee1a950396d0348254c64c307/aws_lambda_powertools/metrics/provider/cloudwatch_emf/cloudwatch.py#L90) into this PR before merging it.

Once `Metrics.set_default_dimensions` delegates only to the provider, the provider must preserve the dictionary passed by `Metrics`. Today we have:

```python
self.default_dimensions = default_dimensions or {}
```

When `Metrics` passes its initially empty shared dictionary, it is falsy, so the provider replaces it with a new dictionary. Calling `provider.set_default_dimensions(...)` then updates only that new dictionary, while the dictionary owned by `Metrics` remains empty.

Eric fixed it in #8404 with:

```python
self.default_dimensions = default_dimensions if default_dimensions is not None else {}
```

Could you please bring this small fix into the PR and add a regression test confirming that a provided empty dictionary remains shared after setting default dimensions?

vishwakt:
Done @leandrodamascena. 

I brought in Eric's fix as-is, with commit co-authorship so the credit is on the record. 

The provider now keeps the passed dict unless `None` is given. Added two regression tests, one at the provider level confirming a provided empty dict stays shared and receives updates through `set_default_dimensions`, and one confirming `Metrics` and the default provider hold the same dict afterwards.

@ericbn thanks for catching that one. The falsy empty dict swap explains the order-dependent sharing I stumbled over earlier in this thread, and you're right that it needed to land together with this change now that `Metrics` delegates to the provider.

leandrodamascena:
Thanks @vishwakt for the work and patience here. I was unsure at first because @ericbn raised some good points, so I spent more time comparing both approaches.

This PR fixes the warning without changing how `Metrics` exposes and shares its state. Setting the same value again stays silent, while changing the value still warns. This is the behavior I expect.

@ericbn, thanks for the effort and for catching the empty dictionary problem. It was important to include that fix here and keep the co-authorship.

There is another valid problem with non-string default dimensions, but I’ll handle that separately.

I’ll approve and merge this one.

leandrodamascena (approved):

PR #8405, opened 2026-08-27
perf(parser): lazy-load envelopes to reduce Lambda cold start latency by ~900ms

### LINKED ISSUE #8406: perf(parser): eagerly loading all 16 envelopes on import adds ~900ms Lambda cold start latency
## Expected Behavior

Importing `parse` or `event_parser` from `aws_lambda_powertools.utilities.parser` should not load envelope modules unless an envelope is actually used.

## Current Behavior

`parser/__init__.py` eagerly imports `envelopes` and `BaseEnvelope` at module level:

```python
from aws_lambda_powertools.utilities.parser import envelopes
from aws_lambda_powertools.utilities.parser.envelopes import BaseEnvelope
```

This causes all 16 envelope modules (apigw, sqs, sns, kinesis, dynamodb, cloudwatch, ...) to load on every cold start, even when no envelope is ever used.

## Steps to Reproduce

Any Lambda function with:

```python
from aws_lambda_powertools.utilities.parser import parse
```

Profile with `PYTHONPROFILEIMPORTTIME=1` on Lambda and observe:

```
import time:  30627 |   1413849 | aws_lambda_powertools.utilities.parser
import time:   6263 |    295176 |   aws_lambda_powertools.utilities.parser.envelopes
import time:    989 |    267728 |     aws_lambda_powertools.utilities.parser.envelopes.apigw
import time:    855 |    285029 |     aws_lambda_powertools.utilities.parser.envelopes.apigw  (apigwv2)
# ... 14 more envelope modules
```

## Observed Impact

~900ms added to Lambda INIT_DURATION for every cold start on functions that use `parse()` without an envelope. Measured on AWS Lambda arm64, Python 3.13, 1024MB.

## Possible Solution

Use Python's module-level `__getattr__` to defer loading `envelopes` and `BaseEnvelope` until they are actually accessed. Public API is fully preserved.

I have submitted PR #8405 with the fix and a regression test.

## Environment

- **Powertools version:** 3.x
- **Python version:** 3.13
- **Lambda runtime:** arm64

ErezMizrahi:
**Issue number:** closes #8406

### Changes

- Replace eager imports of `envelopes` and `BaseEnvelope` in `parser/__init__.py` with a module-level `__getattr__` that defers loading until those names are actually accessed
- Add a functional test asserting that importing `parse` does not load any envelope modules

### User experience

**Before:** Any Lambda function importing `parse` or `event_parser` from `aws_lambda_powertools.utilities.parser` caused all 16 envelope modules to load eagerly at cold start — even when no envelope was ever used.

**After:** Envelope modules are only imported on first access (e.g. when passing `envelope=SqsEnvelope` to `parse()`). Functions that use `parse()` without an envelope pay zero envelope-loading cost.

**Measured impact:** ~900ms reduction in Lambda INIT_DURATION for API Gateway handlers using `parse()` without an envelope. Profiled using `PYTHONPROFILEIMPORTTIME=1` on AWS Lambda (arm64, Python 3.13, 1024MB):

```
# Before
import time:  30627 |   1413849 | aws_lambda_powertools.utilities.parser
import time:   6263 |    295176 |   aws_lambda_powertools.utilities.parser.envelopes
# ... 14 more envelope modules loaded unnecessarily

# After
import time:    312 |     41203 | aws_lambda_powertools.utilities.parser
# No envelope modules loaded
```

**Public API is fully preserved** — `from aws_lambda_powertools.utilities.parser import envelopes` and `from aws_lambda_powertools.utilities.parser import BaseEnvelope` continue to work exactly as before.

---


aws_lambda_powertools/utilities/parser/__init__.py
```diff
@@ -1,11 +1,30 @@
...
 from aws_lambda_powertools.utilities.parser.parser import event_parser, parse
 
+if TYPE_CHECKING:
+    from aws_lambda_powertools.utilities.parser import envelopes as envelopes
+    from aws_lambda_powertools.utilities.parser.envelopes import BaseEnvelope
+
+
+def __getattr__(name: str) -> object:
+    if name == "envelopes":
+        from aws_lambda_powertools.utilities.parser import envelopes as _envelopes  # noqa: PLC0415
```
leandrodamascena: Hi Erez, thank you for profiling this and documenting the cold-start impact. I really like the direction of this change.

I pulled the branch locally and tried the existing public import in a fresh Python process:

```python
from aws_lambda_powertools.utilities.parser import envelopes
```

It currently ends in a `RecursionError`. `__getattr__("envelopes")` imports the same attribute from the same package, which invokes `__getattr__` again. `from ...parser import *` fails for the same reason.

Could we load the submodule with `importlib.import_module(f"{__name__}.envelopes")` and cache the resolved value in `globals()`? I tested that approach locally and it preserves both the lazy import and the existing public API.

tests/functional/parser/test_parser.py
```diff
@@ -308,3 +308,30 @@ def handler(event: SqsModel, _: LambdaContext):
...
+    """Importing parse from parser __init__ must not eagerly load all envelope modules.
+
+    Envelopes are only needed when envelope= is passed to parse()/event_parser().
+    Eager loading all 16 envelopes adds ~900ms to Lambda cold start for functions
+    that only use parse() without an envelope.
+    """
+    import sys
+
+    # Remove any previously cached parser modules to simulate a fresh import
+    parser_modules = [key for key in sys.modules if "aws_lambda_powertools.utilities.parser.envelopes" in key]
```
leandrodamascena: One concern with this test: by the time it runs, the conftest and this module have already imported envelope classes. Removing entries from `sys.modules` can leave references behind, so the result may depend on test order.

Could we run this check in a fresh subprocess instead? I would also cover the existing public imports for `envelopes` and `BaseEnvelope`, since those are the compatibility paths affected by this change.

leandrodamascena (changes requested):
Thanks for working on this. The lazy-loading direction is valuable, but the current implementation breaks an existing public import path. I am requesting changes so we preserve compatibility and make the import test reliably isolated.

ErezMizrahi:
Hi @leandrodamascena,

Thank you for the thorough review and valuable feedback! I've addressed all your comments:

✅ **Fixed RecursionError**: Now using `importlib.import_module()` with `globals()` caching as you suggested
✅ **Improved test isolation**: All tests now run in fresh subprocesses to ensure reliable isolation
✅ **Comprehensive test coverage**: Added tests for all public import paths:
   - `from parser import parse` (should NOT load envelopes)
   - `from parser import envelopes` (should load envelopes)
   - `from parser import BaseEnvelope` (should load envelopes)
   - `from parser import *` (should work correctly)

All tests pass and the lazy-loading behavior is fully preserved while maintaining backward compatibility.

Thanks again for taking the time to review this! 🙏

aws_lambda_powertools/utilities/parser/__init__.py
ErezMizrahi: You’re absolutely right I changed it to work with importlib and globals

tests/functional/parser/test_parser.py
ErezMizrahi: I changed it and now each case in the test runs on a subprocess

leandrodamascena (dismissed):
Thank you, Erez. This was really good work.

I ran the benchmark myself using real Lambda functions and confirmed the improvement. On Python 3.13 ARM64, the median Init Duration dropped from 786.65 ms to 553.04 ms across 20 cold starts per variant. The exact number depends on the environment, but the improvement is clear.

I noticed one small compatibility gap: the lazy exports were missing from `dir(parser)` and `inspect.getmembers(parser)`. I added `__dir__` and a regression test directly to your branch. This keeps the envelopes lazy while preserving normal Python introspection.

I am approving this and will merge it once CI is green. Really nice contribution.

leandrodamascena (approved):

PR #8422, opened 2026-09-03
chore(ci): mirror commercial layer visibility

### LINKED ISSUE #8421: Maintenance: mirror source layer permissions across AWS partitions
### Why is this needed?

Partition layer workflows currently add a public `lambda:GetLayerVersion` permission to every copied layer version. That is correct for released commercial layers, but not for private alignment placeholders used to advance a new runtime to the shared layer version number.

Python 3.14 commercial versions 1 through 26 are private placeholders. Publishing those versions publicly in AWS China or GovCloud exposes artifacts that are not usable Powertools releases and makes partition behavior differ from the commercial source.

The new partition layer balancer inherited this unconditional permission behavior from the regular partition deployment workflow and the existing commercial layer-balancer. China Gamma testing exposed the mismatch before any Python 3.14 placeholders were published to China production.

Automation

### Solution

Capture the commercial source version policy during the read-only download stage and carry it with the layer artifact. Partition publishing should add a public permission only when the commercial source version has an unconditional public `lambda:GetLayerVersion` statement.

Apply the same behavior to both `Layer Balancer (Partitions)` and `Layer Deployment (Partitions)` so placeholders remain private while released versions remain public. Verify the resulting target policy after publication.

### Acknowledgment

- [ ] This should be considered in other Powertools for AWS Lambda languages

---

dreamorosi:
**Issue number:** closes #8421

### Changes

Carry each commercial layer version policy with its partition artifact and classify whether it has an unconditional public `lambda:GetLayerVersion` permission. `Layer Balancer (Partitions)` and `Layer Deployment (Partitions)` now grant public access only when the commercial source is public, then verify that the target visibility matches.

Fail on malformed policies, unexpected public target placeholders, target account mismatches, or a published version number that differs from the requested commercial version.

### User experience

Private alignment placeholders remain private when copied to AWS partitions, while released commercial layers retain public access. Maintainers can review the source visibility in workflow logs before a partition layer is published.

---

### DIFF
--- a/.github/workflows/layers_partitions.yml
+++ b/.github/workflows/layers_partitions.yml
@@ -189,6 +189,7 @@ jobs:
               )
           ' '${{ matrix.layer }}-${{ matrix.arch }}.policy.json' > /dev/null; then
             echo 'public=true' >> "$GITHUB_OUTPUT"
+            echo 'Commercial source layer is public'
           else
             RESULT=$?
             if (( RESULT > 1 )); then
@@ -196,6 +197,7 @@ jobs:
               exit 1
             fi
             echo 'public=false' >> "$GITHUB_OUTPUT"
+            echo 'Commercial source layer is private'
           fi
       - id: transform
         run: |
@@ -229,6 +231,7 @@ jobs:
         id: create-layer
         env:
           SOURCE_IS_PUBLIC: ${{ steps.source_policy.outputs.public }}
+          VERSION: ${{ inputs.version }}
         run: |
           jq '{"LayerName": "${{ matrix.layer }}-${{ matrix.arch }}", "Description": .Description, "CompatibleRuntimes": .CompatibleRuntimes, "CompatibleArchitectures": .CompatibleArchitectures, "LicenseInfo": .LicenseInfo} | with_entries(select(.value != null))' '${{ matrix.layer }}-${{ matrix.arch }}.json' > input.json
 
@@ -237,6 +240,11 @@ jobs:
             --cli-input-json file://./input.json \
             --query 'Version' \
             --output text)
+
+          if (( LAYER_VERSION != VERSION )); then
+            echo "Expected to publish as version ${VERSION}, received ${LAYER_VERSION}"
+            exit 1
+          fi
 
           echo "LAYER_VERSION=$LAYER_VERSION" >> "$GITHUB_OUTPUT"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/layers_partitions.yml
```diff
@@ -146,10 +168,37 @@ jobs:
...
+        with:
+          name: ${{ matrix.layer }}-${{ matrix.arch }}.policy.json
       - name: Verify Layer Signature
         run: |
           SHA=$(jq -r '.Content.CodeSha256' '${{ matrix.layer }}-${{ matrix.arch }}.json')
           test "$(openssl dgst -sha256 -binary ${{ matrix.layer }}-${{ matrix.arch }}.zip | openssl enc -base64)" == "$SHA" && echo "SHA OK: ${SHA}" || exit 1
+      - id: source_policy
+        name: Source Layer Visibility
+        run: |
+          if jq -e '
```
svozza: Seems like this function is repeated several times in the PR, is there any way we can centralised?

.github/workflows/layers_partitions.yml
```diff
@@ -160,8 +209,29 @@ jobs:
...
+          AWS_ACCOUNT: ${{ secrets[format('AWS_ACCOUNT_{0}', steps.transform.outputs.CONVERTED_REGION)] }}
+          PARTITION: ${{ needs.setup.outputs.partition }}
+          REGION: ${{ matrix.region }}
+        run: |
+          if [[ ! "$AWS_ACCOUNT" =~ ^[0-9]{12}$ ]]; then
+            echo "AWS account secret for ${REGION} is missing or invalid"
+            exit 1
+          fi
+
+          CALLER_ACCOUNT=$(aws --region "$REGION" sts get-caller-identity --query Account --output text)
```
svozza: Nit: this could be one call: `aws --region "$REGION" sts get-caller-identity --query '[Account, Arn]' --output text`

.github/workflows/layers_partitions.yml
dreamorosi: Centralized this in `.github/scripts/layer_policy_visibility.sh`. Both workflows now call the same fail-safe classifier, including malformed-policy handling.

.github/workflows/layers_partitions.yml
dreamorosi: Updated both partition workflows to query `[Account, Arn]` in one `get-caller-identity` call.

dreamorosi:
@svozza please, could you review again?

svozza (approved):

PR #8428, opened 2026-09-03
fix(feature_flags): handle same-hour midnight rollover and reject malformed HH:MM

### LINKED ISSUE #8423: Bug: SCHEDULE_BETWEEN_TIME_RANGE mishandles same-hour midnight rollover and accepts malformed HH:MM
### Expected Behaviour

A time range whose end is earlier than its start on the clock should be treated as crossing midnight regardless of whether the hour digits differ, and only strings that are exactly `HH:MM` should pass schema validation.

### Current Behaviour

Two related defects in the same code path.

1. Rollover is detected by comparing hours only (`if int(end_hour) < int(start_hour)` in `comparators.py`). A range like `START: "23:30", END: "23:00"` has equal hours, so it takes the same-day branch `start <= now <= end`, which can never be true. The rule silently never matches.
2. `TIME_RANGE_PATTERN` is `2[0-3]:[0-5]\d|[0-1]\d:[0-5]\d` with no anchors and is used with `re.match`. `"10:00abc"` passes validation. At evaluation time `int("00abc")` raises, `_match_by_action` swallows the exception and returns `False`, so the misconfiguration is invisible.

### Code snippet

```python
from aws_lambda_powertools.utilities.feature_flags import FeatureFlags, RuleAction
from aws_lambda_powertools.utilities.feature_flags.schema import SchemaValidator

# 1. same-hour rollover: never matches, at any time of day
features = {
    "late_night": {
        "default": False,
        "rules": {
            "almost all day": {
                "when_match": True,
                "conditions": [{
                    "action": RuleAction.SCHEDULE_BETWEEN_TIME_RANGE.value,
                    "key": "CURRENT_TIME",
                    "value": {"START": "23:30", "END": "23:00"},
                }],
            }
        },
    }
}

# 2. malformed time passes validation
SchemaValidator({
    "f": {"default": False, "rules": {"r": {"when_match": True, "conditions": [{
        "action": RuleAction.SCHEDULE_BETWEEN_TIME_RANGE.value,
        "key": "CURRENT_TIME",
        "value": {"START": "10:00abc", "END": "12:00"},
    }]}}}
}).validate()  # no error raised
```

### Possible Solution

Compare full minutes-since-midnight (`start_h*60+start_m` vs `end_h*60+end_m`) to decide rollover, and anchor the pattern (`^(?:2[0-3]|[01]\d):[0-5]\d$`) or use `fullmatch`.

Context: we're porting this utility to Powertools for TypeScript (aws-powertools/powertools-lambda-typescript#5614) and want both runtimes to evaluate the same document identically.

### Steps to Reproduce

Run the snippet. For (1), evaluate `late_night` at any time and observe `False`. For (2), observe `validate()` returns without error.

latest

3.13

### Packaging format used

PyPi

### Debugging logs

```python

```

dreamorosi:
**Issue number:** closes #8423

### Changes

`SCHEDULE_BETWEEN_TIME_RANGE` now treats any range whose end is earlier than its start on the clock as crossing midnight, and schema validation only accepts values that are exactly `HH:MM`.

Rollover was decided by comparing hours alone, so a range like `START: "23:30", END: "23:00"` (equal hours) fell into the same-day branch `start <= now <= end`, which can never hold. The comparator now compares minutes since midnight.

Separately, `TIME_RANGE_PATTERN` was unanchored and used with `re.match`, so `"10:00abc"` validated and then raised inside `int()` at evaluation time, where `_match_by_action` swallows the error and returns `False`. The pattern is now anchored (`^(?:2[0-3]|[01]\d):[0-5]\d$`) so the misconfiguration surfaces as a `SchemaValidationError`.

Regression tests cover a same-hour overnight range at a time inside it (10:00) and at the only uncovered gap (23:15), plus malformed `START`/`END` strings with trailing or leading garbage and extra digits. All four new cases fail against the previous code.

This also aligns evaluation with the TypeScript port in aws-powertools/powertools-lambda-typescript#5614.

### User experience

**Before**

- A rule with `{"START": "23:30", "END": "23:00"}` silently never matched, at any time of day.
- A rule with `{"START": "10:00abc", "END": "12:00"}` passed `SchemaValidator().validate()` and then silently evaluated to `False`.

**After**

- `{"START": "23:30", "END": "23:00"}` matches from 23:30 through 23:00 the next day, i.e. everything except 23:00-23:30.
- `{"START": "10:00abc", "END": "12:00"}` raises `SchemaValidationError: 'START' and 'END' must be a valid time format, time_format=%H:%M, rule=...` at validation time.

Existing valid `HH:MM` configurations are unaffected.

---

### DIFF
--- a/aws_lambda_powertools/utilities/feature_flags/comparators.py
+++ b/aws_lambda_powertools/utilities/feature_flags/comparators.py
@@ -55,10 +55,15 @@ def compare_time_range(context_value: Any, condition_value: dict) -> bool:
     start_time = current_time.replace(hour=int(start_hour), minute=int(start_min))
     end_time = current_time.replace(hour=int(end_hour), minute=int(end_min))
 
-    if int(end_hour) < int(start_hour):
+    # Compare full minutes-since-midnight so that ranges within the same hour (e.g. 23:30 -> 23:00)
+    # are correctly detected as crossing midnight.
+    start_minutes = int(start_hour) * 60 + int(start_min)
+    end_minutes = int(end_hour) * 60 + int(end_min)
+
+    if end_minutes < start_minutes:
         # In normal circumstances, we need to assert **both** conditions
         """
-        # When the end hour is smaller than start hour, it means we are crossing a day's boundary.
+        # When the end time is earlier than start time, it means we are crossing a day's boundary.
         # In this case we need to assert that current_time is **either** on one side or the other side of the boundary
         #
         # ┌─────┐                                    ┌─────┐                                  ┌─────┐
--- a/aws_lambda_powertools/utilities/feature_flags/constants.py
+++ b/aws_lambda_powertools/utilities/feature_flags/constants.py
@@ -9,5 +9,5 @@ CONDITION_VALUE = "value"
 CONDITION_ACTION = "action"
 FEATURE_DEFAULT_VAL_TYPE_KEY = "boolean_type"
 TIME_RANGE_FORMAT = "%H:%M"  # hour:min 24 hours clock
-TIME_RANGE_PATTERN = re.compile(r"2[0-3]:[0-5]\d|[0-1]\d:[0-5]\d")  # 24 hour clock
+TIME_RANGE_PATTERN = re.compile(r"^(?:2[0-3]|[01]\d):[0-5]\d$")  # 24 hour clock, exactly HH:MM
 HOUR_MIN_SEPARATOR = ":"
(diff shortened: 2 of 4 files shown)


leandrodamascena:
I'll review this by tomorrow.

leandrodamascena (approved):
Thanks Andrea, this looks good. I pushed a small follow-up to use `fullmatch()`, since Python’s `$` still accepts a final newline.

Valid `HH:MM` configurations are not affected. Invalid values are now rejected as expected, and the rollover fix behaves correctly. I’ll merge this.

PR #2052, opened 2023-03-26
fix(feature_flags): handle expected falsy values in conditions

ran-isenberg:
**Issue number:** #2051 

## Summary
Rule Conditions allow users to evaluate context attributes against values and actions defined in Rule Conditions.
The condition can easily be pivoted on a certain value in the context evaluating to falsy values
The values attribute should be able to accept any value except null
The condition should also honor any value for context key variable.

### Changes

- Added tests
- changed the validation from if true to if none
- changed error message to none instead of empty

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [x] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [ ] I have performed a self-review of this change
* [ ] Changes have been tested
* [ ] Changes are documented
* [x] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

Issue #2051, opened 2023-03-25
Bug: Feature Flag Rule Conditions Values and Context Key Variables cannot have Falsy values


ajwad-shaikh:
### Expected Behaviour

Rule Conditions allow users to evaluate context attributes against values and actions defined in Rule Conditions. 
The condition can easily be pivoted on a certain value in the context evaluating to falsy values. Consider the following example -

Context-
```json
{
   "is_paid": 0,
}
```

Feature Flag - 
```json
{
  "sales_offer": {
    "default": false,
    "rules": {
      "unpaid_users": {
        "when_match": true,
        "conditions": [ {"action": "EQUALS", "key": "is_paid", "value": 0} ]
      }
   }
}
```

The `values` attribute should be able to accept any value except `null`
The condition should also honor any value for context key variable.

#### Definition - Falsy Values

In python, the following values evaluate to False in boolean context - these are known as falsy values.
```py
empty_list = []
empty_string = ""
zero = 0
false_boolean = False
```

#### Falsy Values in Condition `value`

The `values` attribute in conditions cannot accept any of the falsy values, some of which are essential to framing conditions in practical use-cases.

If any of the falsy values are present in the condition - the utility throws the error below 
```py
File ~/venv/lib/python3.8/site-packages/aws_lambda_powertools/utilities/feature_flags/schema.py:330, in ConditionsValidator.validate_condition_value(condition, rule_name)
    328 value = condition.get(CONDITION_VALUE, "")
    329 if not value:
--> 330     raise SchemaValidationError(f"'value' key must not be empty, rule={rule_name}")
    331 action = condition.get(CONDITION_ACTION, "")
    333 # time actions need to be parsed to make sure date and time format is valid and timezone is recognized

SchemaValidationError: 'value' key must not be empty, rule=unpaid_users
```

#### Falsy Values in context `value`

The conditions simply return evaluate to **false** if the value for the key passed in context is falsy. These are also extremely essential to framing conditions in practical use-cases.

Example - 

> 🚨 The condition below will evaluate to False even if the _context value (0) is **NOT EQUAL** to condition value (1)_ 🚨 

```json
{
  "sales_offer": {
    "default": false,
    "rules": {
      "unpaid_users": {
        "when_match": true,
        "conditions": [ {"action": "NOT_EQUALS", "key": "is_paid", "value": 1} ]
      }
   }
}
```

#### Falsy Values in Condition `value`

```js
// context.json

{
[...]

PR #563, opened 2021-07-26
refactor(feature_flags): optimize UX and maintenance

heitorlessa:
**Issue #, if available:** #494

## Description of changes:

**Checklist**

 

* [x] [Meet tenets criteria](https://awslabs.github.io/aws-lambda-powertools-python/#tenets)
* [x] Update tests
* [x] Update docs
* [x] PR title follows [conventional commit semantics](https://github.com/awslabs/aws-lambda-powertools-python/blob/376ec0a2ac0d2a40e0af5717bef42ff84ca0d1b9/.github/semantic.yml#L2)

## New UX

```python
from typing import List

from aws_lambda_powertools.utilities.feature_flags import FeatureFlags, AppConfigStore

app_config = AppConfigStore(
    environment="test",
    application="powertools",
    name="test_conf_name",
    cache_seconds=600,
)

feature_flags: FeatureFlags = FeatureFlags(store=app_config)

is_my_feature_active: bool = feature_flags.evaluate(name="my_feature", context={}, default=False)

# TODO: This needs reviewing, possibly meant as List[Dict[str, bool]]???
all_features: List[str] = feature_flags.get_enabled_features(context={})
```

**Using envelope argument to pluck feature flags in an arbitrary key from fetched config**

```python
app_config = AppConfigStore(
    environment="test",
    application="powertools",
    name="test_conf_name",
    cache_seconds=600,
    envelope="features"    # feature toggles are under `features` key in fetched config
)

feature_flags: FeatureFlags = FeatureFlags(store=app_config)
```

## Changes

Major changes made during initial doc writing - Due to the size, we're splitting this PR to separate actual docs.

- [x] General
  - [x] ~~Rename `feature_toggles` to `feature_flags`~~
  - [x] ~~Refactor error loggers to contextual exceptions~~
- [x] Improve exceptions
  - [x] ~~Consider an additional fields to improve exception handling such as error type, feature, etc~~
    - [x] ~~String concatenation ended up slowing things down, dropping it~~ 
  - [x] ~~Consider fine grained ConfigurationError instead of a single ConfigurationError~~
  - [x] ~~Rename it to `InvalidSchemaError`~~
- [x] Refactor Schema to ease maintenance
  - [x] ~~Simplify schema by removing `features` key~~
  - [x] ~~Make each rule in `rules` unique by using Dict over List, and remove `rule_name`~~
  - [x] ~~Review long and redundant names such as `"feature": { "feature_default_value"...}`, `"rules": [{"rule_name"...}]`~~
[...]

PR #8429, opened 2026-09-03
fix(feature_flags): missing context key never satisfies a condition

### LINKED ISSUE #8424: Bug: a missing context key satisfies negative conditions (NOT_EQUALS, NOT_IN, KEY_NOT_IN_VALUE, VALUE_NOT_IN_KEY)
### Expected Behaviour

A condition whose `key` is absent from the evaluation context should not match, regardless of action. A rule written as "tier is not premium" should not fire for a request that carries no `tier` at all.

### Current Behaviour

`context.get(key)` returns `None` for an absent key, and `None` is then compared as an ordinary value. `None != "premium"` is `True`, `None not in [...]` is `True`, so negative actions (`NOT_EQUALS`, `NOT_IN`, `KEY_NOT_IN_VALUE`, `VALUE_NOT_IN_KEY`) match missing keys. Rules intended to target one segment fire for anonymous or partially-populated traffic. The behaviour is not documented and no test asserts it.

### Code snippet

```python
from aws_lambda_powertools.utilities.feature_flags import FeatureFlags, RuleAction

class InMemoryStore:
    def __init__(self, d): self.d = d
    def get_configuration(self): return self.d
    @property
    def get_raw_configuration(self): return self.d

features = {
    "discount_banner": {
        "default": False,
        "rules": {
            "non premium users": {
                "when_match": True,
                "conditions": [{
                    "action": RuleAction.NOT_EQUALS.value,
                    "key": "tier",
                    "value": "premium",
                }],
            }
        },
    }
}

flags = FeatureFlags(store=InMemoryStore(features))
print(flags.evaluate(name="discount_banner", context={}, default=False))               # True (unexpected)
print(flags.evaluate(name="discount_banner", context={"tier": "free"}, default=False))  # True
```

### Possible Solution

In `_evaluate_conditions`, if `condition["key"]` is not in `context`, treat the condition as not matched before calling the comparator. Time-based keys are already special-cased and would be unaffected.

This changes results for documents that, knowingly or not, rely on the current behaviour. We think it qualifies as a defect fix in the same spirit as #2051 (falsy values), where the old result was wrong rather than a designed behaviour, but we'd understand if maintainers prefer to hold it for a major. Context: the TypeScript port (aws-powertools/powertools-lambda-typescript#5614) is adopting "missing key never matches" and we'd like the two runtimes to agree.

### Steps to Reproduce

Run the snippet; first print is `True`.

latest

3.13

### Packaging format used

PyPi

### Debugging logs

```python

```

dreamorosi:
**Issue number:** closes #8424

### Changes

A condition whose `key` is absent from the evaluation context no longer matches, regardless of action.

`_evaluate_conditions` read the context with `context.get(key)`, so an absent key was compared as `None`. Negative actions then matched trivially: `None != "premium"` is `True`, `None not in [...]` is `True`. Rules such as "tier is not premium" fired for anonymous or partially populated requests that carried no `tier` at all.

`_match_by_action` now takes a `context_key_present` flag. The comparator runs exactly as before; if the key was absent and the comparator returned normally, the result is forced to `False`. Running the comparator first matters: actions such as `ANY_IN_VALUE` raise `ValueError` for a `None` context value, and that exception must keep reaching any `validation_exception_handler` the user registered, with the handler's return value honoured. Only the "comparator quietly returned `True` for an absent key" path changes.

A key that is present with an explicit `None` value is still compared as before, so this is about absence rather than falsiness. Time-based actions are unaffected since they never read the user context.

#### Why this is a regression fix, not a behaviour change

From the utility's introduction (#563, 2021) until v2.11.0, `_match_by_action` opened with:

```python
if not context_value:
    return False
```

so a missing key (`None`) could never satisfy a condition, for any action. #2052 (fixing #2051) removed that guard so that falsy values such as `0`, `""`, and `False` could be compared. That was the intended scope of the change; allowing an absent key to satisfy `NOT_EQUALS` and friends was a side effect, never documented, and no test asserted it. This PR restores the original "missing key never matches" semantics while keeping the #2051 behaviour for keys that are present with a falsy value.

Tests cover: the four negative actions with a missing key; an explicit `None` value still being compared; a missing key with `ANY_IN_VALUE` and a registered `ValueError` handler, asserting the handler is called and its result used; and the same without a handler, asserting no match. The behaviour is also now documented in the rule actions section of the feature flags docs.

### User experience

Given:

```json
{
  "discount_banner": {
    "default": false,
    "rules": {
      "non premium users": {
        "when_match": true,
        "conditions": [{ "action": "NOT_EQUALS", "key": "tier", "value": "premium" }]
      }
    }
  }
}
```

| `context`             | Before v2.11.0 | v2.11.0 to now | After   |
| --------------------- | -------------- | -------------- | ------- |
| `{}`                  | `False`        | `True`         | `False` |
| `{"tier": None}`      | `False`        | `True`         | `True`  |
| `{"tier": ""}`        | `False`        | `True`         | `True`  |
| `{"tier": "free"}`    | `True`         | `True`         | `True`  |
| `{"tier": "premium"}` | `False`        | `False`        | `False` |

Only the first row changes. `validation_exception_handler` behaviour is unchanged: comparators raise for the same inputs as before and handlers see the same exceptions.

---

### DIFF
--- a/aws_lambda_powertools/utilities/feature_flags/feature_flags.py
+++ b/aws_lambda_powertools/utilities/feature_flags/feature_flags.py
@@ -82,10 +82,26 @@ class FeatureFlags:
         self.logger = logger or logging.getLogger(__name__)
         self._exception_handlers: dict[Exception, Callable] = {}
 
-    def _match_by_action(self, action: str, condition_value: Any, context_value: Any) -> bool:
+    def _match_by_action(
+        self,
+        action: str,
+        condition_value: Any,
+        context_value: Any,
+        context_key_present: bool = True,
+    ) -> bool:
         try:
             func = RULE_ACTION_MAPPING.get(action, lambda a, b: False)
-            return func(context_value, condition_value)
+            matched = func(context_value, condition_value)
+
+            if not context_key_present:
+                # A key absent from the context never satisfies a condition. Without this, `None` would be
+                # compared as a regular value and negative actions (NOT_EQUALS, NOT_IN, ...) would match.
+                # We still run the comparator first so that any exception it raises (e.g. ANY_IN_VALUE on a
+                # non-list) reaches registered validation exception handlers exactly as it did before.
+                self.logger.debug(f"context key not present, condition does not match: action={action}")
+                return False
+
+            return matched
         except Exception as exc:
             self.logger.debug(f"caught exception while matching action: action={action}, exception={str(exc)}")
 
@@ -118,6 +134,7 @@ class FeatureFlags:
             cond_key = condition.get(schema.CONDITION_KEY, "")
             cond_action = condition.get(schema.CONDITION_ACTION, "")
             cond_value = condition.get(schema.CONDITION_VALUE)
+            context_key_present = True
 
             # time based rule actions have no user context. the context is the condition key
             if cond_action in (
@@ -126,18 +143,16 @@ class FeatureFlags:
                 schema.RuleAction.SCHEDULE_BETWEEN_DAYS_OF_WEEK.value,
             ):
                 context_value = cond_key  # e.g., CURRENT_TIME
-            elif cond_key not in context:
-                # A key absent from the context never satisfies a condition. Without this guard, `None`
-                # would be compared as a regular value and negative actions (NOT_EQUALS, NOT_IN, ...) would match.
-                self.logger.debug(
-                    f"rule did not match, context key not found, rule_name={rule_name}, "
-                    f"rule_value={rule_match_value}, name={feature_name}, key={cond_key}",
-                )
-                return False
             else:
-                context_value = context[cond_key]
+                context_key_present = cond_key in context
+                context_value = context.get(cond_key)
 
-            if not self._match_by_action(action=cond_action, condition_value=cond_value, context_value=context_value):
+            if not self._match_by_action(
+                action=cond_action,
+                condition_value=cond_value,
+                context_value=context_value,
+                context_key_present=context_key_present,
+            ):
                 self.logger.debug(
                     f"rule did not match action, rule_name={rule_name}, rule_value={rule_match_value}, "
                     f"name={feature_name}, context_value={str(context_value)} ",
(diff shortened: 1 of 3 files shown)


leandrodamascena (changes requested):
Thanks for working on this. There is an existing Feature Flags behavior that this change does not take into account, and the result is a regression.

Users can register a `validation_exception_handler` to catch errors raised while evaluating conditions and decide how to handle them:

```python
@feature_flags.validation_exception_handler(ValueError)
def handle_invalid_context(exc):
    return True
```

For example, evaluating `ANY_IN_VALUE` without the expected context key currently reaches the comparator, raises `ValueError`, and calls the registered handler.

With this PR, the missing-key check returns `False` before `_match_by_action` runs. The registered handler is never called. I reproduced the difference:

- `develop`: the handler is called and its result is used.
- This branch: the handler is skipped and the condition returns `False`.

Could you fix this and add a test covering a missing key with a registered handler? Please hold off merging for now. Ping me when it is ready and I will review it again.

dreamorosi:
@leandrodamascena thanks, good catch. Fixed in 95094f96.

The guard now lives inside `_match_by_action` and runs *after* the comparator. The comparator is invoked exactly as on `develop`, so `ANY_IN_VALUE` (and `ALL_IN_VALUE`, `NONE_IN_VALUE`) still raise `ValueError` for a `None` context value and the registered handler is called with its return value honoured. Only the case where the comparator returns normally with `True` for an absent key is forced to `False`.

Added two tests at the end of `test_feature_flags.py`:
- `test_flags_missing_context_key_still_invokes_validation_exception_handler`: your scenario, asserts the handler is called once and `evaluate` returns the handler's `True`. This fails on the previous revision of the branch.
- `test_flags_missing_context_key_no_match_without_handler_for_raising_action`: same rule, no handler, asserts `False`.

Ready for another look whenever you have time.

leandrodamascena (approved):
Thanks Andrea, this is fixed now. Missing keys no longer match negative actions, while an explicit `None` keeps the previous behavior. The `validation_exception_handler` is also called and its return value is respected.

I updated the branch with `develop` and fixed the test conflict. Good work on this one. I’ll merge it.

PR #8430, opened 2026-09-03
chore(feature_flags): warn on empty schema and empty rules

### LINKED ISSUE #8427: Tech debt: Feature Flags schema accepts odd-but-harmless documents silently
### Why is this needed?

`RulesValidator.validate()` exits early on `if not self.rules`, so `rules: []` (a list, where the docs specify a dict) passes validation; a non-empty list fails. An empty top-level document `{}` is also accepted with no output. These aren't bugs in evaluation, but they hide authoring mistakes (a typo'd feature name yields `{}` after an envelope query, and nothing says so).

Feature Flags

### Suggestion

Emit a `warning` log for: empty top-level document, `rules` present but falsy and not a dict, and any feature with `rules` present but empty. Do not raise, to stay backward compatible. Add tests that pin the warnings.

Context: the TypeScript port (aws-powertools/powertools-lambda-typescript#5614) warns on the same cases and throws only where Python throws, so documents stay portable.

### Acknowledgment

dreamorosi:
**Issue number:** closes #8427

### Changes

`SchemaValidator` now emits a `warning` log for documents that are valid but almost certainly not what the author meant. Nothing is raised, so every document that validated before still validates.

Three cases are covered:

- Empty top-level document (`{}`). Typically the result of an envelope query that matched nothing, e.g. a typo'd feature group name.
- A feature whose `rules` key is present but empty (`{}` or `None`).
- A feature whose `rules` key is present, empty, and not a dictionary (`[]`, `""`). This gets a more specific message because the same value, once populated, would be rejected with `SchemaValidationError`; the empty form only slipped through because `if not self.rules` short-circuits first.

Omitting `rules` entirely stays silent, since that is the documented way to declare a static flag.

While wiring the warning I found that `RulesValidator` derived `feature_name` from `next(iter(self.feature))`, i.e. the feature's first key (`"default"` in practice), not its name. `FeaturesValidator` now passes the real name in through a new optional `feature_name` argument, so the new warnings and the existing `Feature rules must be a dictionary, feature=...` error name the right feature. The old fallback is kept so direct construction without the argument keeps working.

Tests pin each warning's level and message, and pin the two silent paths (well-formed document, `rules` omitted) so the warnings do not become noisy.

The TypeScript port (aws-powertools/powertools-lambda-typescript#5614) warns on the same cases and throws only where Python throws, so documents stay portable between runtimes.

### User experience

No behaviour change for `evaluate` or `get_enabled_features`. The only difference is in logs:

| Document                                        | Before | After                                                                                |
| ----------------------------------------------- | ------ | ------------------------------------------------------------------------------------ |
| `{}`                                            | silent | `WARNING Feature flags schema is empty, no features to validate`                     |
| `{"f": {"default": false, "rules": {}}}`        | silent | `WARNING Feature has 'rules' but it is empty, feature=f`                             |
| `{"f": {"default": false, "rules": []}}`        | silent | `WARNING Feature 'rules' should be a dictionary but is an empty list, feature=f`     |
| `{"f": {"default": false}}`                     | silent | silent                                                                               |
| `{"f": {"default": false, "rules": "4"}}`       | raises | raises, now with `feature=f` instead of `feature=default`                            |

---

### DIFF
--- a/aws_lambda_powertools/utilities/feature_flags/schema.py
+++ b/aws_lambda_powertools/utilities/feature_flags/schema.py
@@ -212,6 +212,12 @@ class SchemaValidator(BaseValidator):
         if not isinstance(self.schema, dict):
             raise SchemaValidationError(f"Features must be a dictionary, schema={str(self.schema)}")
 
+        if not self.schema:
+            # Often the result of an envelope query that matched nothing (e.g. a typo'd feature name).
+            # Harmless for evaluation, so warn rather than raise.
+            self.logger.warning("Feature flags schema is empty, no features to validate")
+            return
+
         features = FeaturesValidator(schema=self.schema, logger=self.logger)
         features.validate()
 
@@ -232,7 +238,12 @@ class FeaturesValidator(BaseValidator):
         for name, feature in self.schema.items():
             self.logger.debug(f"Attempting to validate feature '{name}'")
             boolean_feature: bool = self.validate_feature(name, feature)
-            rules = RulesValidator(feature=feature, boolean_feature=boolean_feature, logger=self.logger)
+            rules = RulesValidator(
+                feature=feature,
+                boolean_feature=boolean_feature,
+                logger=self.logger,
+                feature_name=name,
+            )
             rules.validate()
 
     # returns True in case the feature is a regular feature flag with a  boolean default value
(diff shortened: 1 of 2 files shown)


leandrodamascena:
Thanks Andrea. The validation changes and the feature name fix make sense. There is one difference between Python and TypeScript here.

In Python, the standard way to emit a user-facing warning is `warnings.warn()`. Internal logs under `aws_lambda_powertools.*` are intentionally suppressed by default with `NullHandler`, so `logger.warning()` is not visible unless users configure the package logger.

I think this follows the TypeScript implementation, where using the logger makes more sense. I’m sending a small commit using `PowertoolsUserWarning` and updating the tests to use `pytest.warns`. The feature name fix will stay as it is.

leandrodamascena (approved):

PR #8431, opened 2026-09-03
perf(feature_flags): validate schema once per fetched document

### LINKED ISSUE #8426: Tech debt: Feature Flags validates the full schema on every evaluate(), including cache hits
### Why is this needed?

`FeatureFlags.get_configuration()` constructs a new `SchemaValidator` and walks the whole document on every call. `evaluate()` and `get_enabled_features()` each call it, so a handler that evaluates five flags validates the entire document five times per invocation, even when `AppConfigStore` served the document from the Parameters cache without a network call. For documents with many flags and rules this is wasted CPU on every request.

Feature Flags

### Suggestion

Validate once per fetched document. Options: memoise on `id()` of the returned dict (the Parameters cache returns the same object until expiry), or have `StoreProvider` expose a version/etag and cache the validation result against it. Either keeps the "always validate fresh configuration" guarantee while skipping repeat work on cache hits.

### Acknowledgment

dreamorosi:
**Issue number:** closes #8426

### Changes

`FeatureFlags.get_configuration()` now skips schema validation when the store hands back the same document object it validated last time. A new document is always validated, so the "always validate fresh configuration" guarantee is unchanged; the only work removed is repeat validation of an unchanged, cached document.

The check is object identity (`config is self._last_validated_config`). The Parameters cache stores the transformed dict and returns that same object on every hit until `max_age` expires (`parameters/base.py`, `fetch_from_cache`), so identity is a faithful signal for "same fetch". `FeatureFlags` keeps a strong reference to the last validated document, so its `id()` cannot be recycled by an unrelated object. Third-party `StoreProvider` implementations that return a fresh dict per call get today's behaviour: validation on every call.

`AppConfigStore` with an `envelope` would have defeated the check on its own, since the JMESPath query produces a new object each call. The store now memoises the extracted result against the raw document's identity, so envelope users get the same benefit.

Failed validation is not cached: `_last_validated_config` is only assigned after `validate()` returns, so an invalid document raises on every call as before.

Tests cover: one validation across five `evaluate` calls plus `get_enabled_features`; re-validation when the store returns a new document and no re-validation when it returns the same one again; invalid documents raising on every call and not poisoning the cache; and the envelope path returning the same extracted object and validating once.

### User experience

No API or behavioural change for `evaluate`, `get_enabled_features`, or `get_configuration`. Return values are identical.

For a handler evaluating N flags per invocation against a cached document, schema validation runs once per cache refresh instead of N times per invocation. The saving scales with document size (features x rules x conditions).

---

### DIFF
--- a/aws_lambda_powertools/utilities/feature_flags/appconfig.py
+++ b/aws_lambda_powertools/utilities/feature_flags/appconfig.py
@@ -86,6 +86,10 @@ class AppConfigStore(StoreProvider):
             boto3_client=boto3_client,
             boto3_session=boto3_session,
         )
+        # Memoised envelope extraction: (raw document, extracted config). The Parameters cache hands back the
+        # same raw dict until expiry, so we can reuse the extracted result rather than re-running the JMESPath
+        # query and producing a new object on every call.
+        self._last_extracted: tuple[dict[str, Any], dict[str, Any]] | None = None
 
         # Override the user agent to use "feature_flags" instead of "parameters"
         self._register_feature_flags_user_agent()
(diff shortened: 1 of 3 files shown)


leandrodamascena:
I had the feeling this assumes the returned object is immutable, so I checked that case.

The standard `AppConfigStore` creates a new dictionary after a refresh, so this works fine. It only becomes a problem if someone changes the returned dictionary manually, or if a custom store updates the same dictionary in place.

I am fine considering this incorrect usage. The returned configuration must be treated as read-only, and custom stores must return a new dictionary when the configuration changes. I will add a commit to make this clear in the documentation.

leandrodamascena (approved):
Thanks for working on this @dreamorosi! APPROVED

PR #7083, opened 2025-07-31
fix(parameters): fix _transform_and_cache_get_parameters_response

ericbn:
**Issue number:** #7084 

### Changes

`_transform_and_cache_get_parameters_response` was passing positional arguments to `transform_value` in the wrong order, making the latter fail miserably with

    TransformParameterError: Unable to transform value using "the value"

because `value` was being passed as the expected `transform` argument. Use keyword arguments instead and cover this with a test.

### User experience

`get_parameters_by_name` fails when used with `transform`. It's surprising that I might the first one trying to use this. 😅 

Example:

```python
    parameters = get_parameters_by_name(
        {"/dev/param1": {}, "/dev/param2": {}}, transform="json",
    )
```

Fails with:
`aws_lambda_powertools.utilities.parameters.exceptions.TransformParameterError: Unable to transform value using '"value of /dev/param2"' transform: 'NoneType' object is not callable`

## Checklist

If your change doesn't seem to apply, please leave them unchecked.

* [X] [Meet tenets criteria](https://docs.powertools.aws.dev/lambda/python/#tenets)
* [X] I have performed a self-review of this change
* [X] Changes have been tested
* [ ] Changes are documented
* [X] PR title follows [conventional commit semantics](https://github.com/aws-powertools/powertools-lambda-python/blob/develop/.github/semantic.yml)

<details>
<summary>Is this a breaking change?</summary>

**RFC issue number**:

Checklist:

* [ ] Migration process documented
* [ ] Implement warnings (if it can live side by side)

</details>

## Acknowledgment

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

**Disclaimer**: We value your time and bandwidth. As such, any pull requests created on non-triaged issues might not be successful.

PR #8436, opened 2026-09-04
fix(parameters): honor auto transform in batched get_parameters_by_name

### LINKED ISSUE #8435: Bug: get_parameters_by_name ignores transform="auto" for batched parameters
`get_parameters_by_name(..., transform="auto")` returns raw strings for batched parameters. Nothing raises, so the unparsed value reaches caller code silently.

The split shows inside a single call. A `decrypt=True` parameter routes through `BaseProvider.get()`, which passes `key=name`, so `auto` works. The rest route through `SSMProvider._transform_and_cache_get_parameters_response`, which calls `transform_value()` with no `key`, so the suffix is read from an empty string and `auto` degrades to the identity transform.

On `develop`, stubbed client, one call with `transform="auto"`:

```
decrypt -> BaseProvider.get()   parsed=True   dict  {'value': 'hello'}
batch   -> GetParameters        parsed=False  str   '{"value": "hello"}'
```

```python
get_parameters_by_name(
    parameters={"/dev/a.json": {}, "/dev/b.json": {"decrypt": True}},
    transform="auto",
)
```

Fix is `key=name` in that `transform_value()` call. #7083 fixed the argument-order bug in the same function and covered `transform="json"`, so `auto` survived it.

Version: latest. Runtime: Python 3.13. Packaging: PyPi.

MohammedAlkindi:
**Issue number:** closes #8435

`transform="auto"` infers the transform from the parameter name suffix, but `_transform_and_cache_get_parameters_response` called `transform_value()` without `key=name`. The batched `GetParameters` path therefore inferred from an empty string and fell back to the identity transform, returning raw strings with nothing raised. Parameters routed through `BaseProvider.get()` (`decrypt=True`) already pass the key, so a single `get_parameters_by_name` call could return a parsed dict for some names and a raw string for others. #7083 fixed the argument-order bug in this function and covered `transform="json"`; `auto` was not covered by that test. This was the only `transform_value()` call site omitting the key: `BaseProvider.get()` and `secrets.py` pass it, and `get_multiple()` transforms a dict keyed per item.

Fail-before, source reverted and the new test kept: 1 failed, the assertion showing the raw JSON string against the expected dict. Pass-after: 1 passed. `tests/functional/parameters/`: 125 passed. `ruff format --check` and `ruff check`: clean.

Not run here: `make mypy`, `make ty`, and the e2e suite.

---


leandrodamascena (approved):
Thanks for working on this @MohammedAlkindi! Approved.

PR #8440, opened 2026-09-04
fix(ci): make partition layer deployment resumable

### LINKED ISSUE #8433: Tech debt: make partition layer deployment safe to resume
### Why is this needed?

`.github/workflows/layers_partitions.yml` performs twelve concurrent, irreversible Lambda layer publications without verifying each target layer is at the expected preceding version. It checks the returned version only after publishing, when an incorrect version number can no longer be reused.

The copy matrix also uses the default fail-fast behavior. A failure can cancel sibling jobs after publication but before permission or verification, leaving runtimes and architectures at different versions. The workflow is not idempotent: rerunning a partially successful deployment republishes completed layers at the next version, while rerunning all jobs can conflict with immutable workflow artifacts.

Additional gaps make recovery and operation harder:

- `version` is not validated and is interpolated directly into shell commands;
- AWS retry behavior is not configured for long cross-partition uploads;
- publication and public-permission creation are coupled in one non-atomic step;
- the reusable trigger omits the required `partition` input;
- source downloads do not fail or retry cleanly on HTTP errors;
- failed verification does not preserve target metadata artifacts.

These risks can permanently break layer version parity in China or GovCloud. Until addressed, maintainers must use `Layer Balancer (Partitions)` separately for each runtime because it provides target-position preflight, resumability, SHA checks, visibility mirroring, retries, and non-fail-fast execution.

Automation

### Suggestion

Bring the regular partition deployment workflow up to the balancer safety model:

- validate and normalize all dispatch/call inputs;
- require the target latest version to equal `version - 1` before publishing;
- disable matrix fail-fast and consider serial target publication;
- skip existing versions only when their SHA and visibility match the commercial source;
- make artifact uploads and permission creation idempotent;
- configure standard AWS retries;
- preserve partial-run metadata for recovery;
- complete or remove the broken `workflow_call` interface.

### Acknowledgment

- [ ] This should be considered in other Powertools for AWS Lambda languages

---

dreamorosi:
**Issue number:** closes #8433

### Changes

- Validate partition deployment inputs before any AWS operation.
- Preflight immutable target versions and require the expected preceding version before publication.
- Treat existing matching versions as resumable no-ops and repair missing public permissions.
- Separate publication, permission creation, and final SHA and visibility verification.
- Disable matrix fail-fast, add safe download retries, and make artifacts overwrite-safe.
- Preserve target metadata after partial failures for recovery.
- Complete the reusable workflow partition input.

### User experience

Maintainers can safely rerun a partially completed China or GovCloud deployment without republishing successful layers at the next immutable version. Existing versions must match the commercial SHA and visibility, missing public permissions can be repaired independently, and unsafe target positions fail before publication.

This PR is stacked on #8439 so deployment matrices retain only supported Python runtimes.

---

### DIFF
--- a/.github/workflows/layers_partitions.yml
+++ b/.github/workflows/layers_partitions.yml
@@ -4,7 +4,8 @@
 #
 # We pull each the version of the layer and store them as artifacts, the we upload them to each of the Partitioned AWS accounts.
 #
-# A number of safety checks are performed to ensure safety.
+# Every target publication is preflighted against the latest version already present in the partition, so a partially
+# failed run can be rerun safely: existing versions are verified instead of republished.
 
 on:
   workflow_dispatch:
@@ -26,6 +27,7 @@ on:
         options:
           - China
           - GovCloud
+        required: true
   workflow_call:
     inputs:
       environment:
@@ -34,6 +36,10 @@ on:
         required: true
       version:
         description: Layer version to duplicate
+        type: string
+        required: true
+      partition:
+        description: Partition to deploy to
         type: string
         required: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/layers_partitions.yml
```diff
@@ -215,73 +281,192 @@ jobs:
...
           fi
 
           if [[ "$TARGET_IS_PUBLIC" != "$SOURCE_IS_PUBLIC" ]]; then
-            echo "Published layer visibility does not match its commercial source"
+            echo "${LAYER_NAME}:${VERSION} in ${REGION} does not match its commercial source visibility"
             exit 1
           fi
-
       - name: Store Metadata - ${{ matrix.region }}
+        if: ${{ always() && !cancelled() }}
```
svozza, line 465: I think this can be simplfied to just `!cancelled`

PR #8442, opened 2026-09-05
feat[kafka]: avro offline schema handling

### LINKED ISSUE #8385: Feature request: [Kafka] Skip N leading bytes of a Kafka Avro payload before deserialization
#### Current Sate

When Kafka messages are produced with a schema-registry-aware serializer (for example
Confluent's `KafkaAvroSerializer` or AWS Glue's Avro serializer), the producer prepends a
short wire-format wrapper to every payload — typically 5 bytes for Confluent
(1-byte magic byte `0x00` + 4-byte big-endian schema ID) or 18 bytes for Glue
(1-byte header version + 1-byte compression + 16-byte UUID schema-version ID).

When my consumer uses the ESM Schema Registry integration, ESM strips those bytes upstream
and populates `valueSchemaMetadata.schemaId` — the existing `AvroDeserializer` behaves
correctly.

#### Problem 

I have my **Avro schema offline** (checked into my
Lambda), and I do not use the ESM Schema Registry integration. As a result:

- The raw payload — including the 5-byte wire-format prefix — reaches my Lambda function.
- `valueSchemaMetadata.schemaId` is not populated on the event, so the protobuf-style
  auto-detection in `deserializer/protobuf.py` does not translate.
- `AvroDeserializer.deserialize` hands the prefixed bytes straight to
  `avro.io.BinaryDecoder`. Because the prefix is legal Avro varint noise, the decoder
  either silently returns corrupted values (empty strings, zero ints) or raises
  `KafkaConsumerDeserializationError`, depending on the specific prefix bytes.

I want to keep using Powertools' `kafka_consumer` decorator with my offline Avro schema
without having to hand-roll a base64-decode + slice + re-encode dance in every handler.

### Solution/User Experience

Add two new optional parameters on `SchemaConfig`:

- `value_schema_id_prefix_length: int = 0`
- `key_schema_id_prefix_length: int = 0`

They tell the Avro deserializer how many leading bytes to skip after base64-decoding the
payload and before handing it to the Avro `BinaryDecoder`. Default is `0`, preserving
today's behaviour for anyone not using a schema-registry wire format.

```python
from aws_lambda_powertools.utilities.kafka import SchemaConfig, kafka_consumer

schema_config = SchemaConfig(
    value_schema_type="AVRO",
    value_schema=AVRO_SCHEMA,           # offline, checked into the Lambda
    value_schema_id_prefix_length=5,    # skip Confluent's magic byte + schema ID
)

@kafka_consumer(schema_config=schema_config)
def handler(event, context):
    for record in event.records:
        # record.value is the fully-deserialized Avro payload,
        # with the 5-byte prefix stripped transparently.
        ...
```

Scope of the change:

- Applies to **both** value and key sides, symmetrical with every other `SchemaConfig`
  option.
- Applies to **Avro only** for now. Protobuf already has schema-registry framing logic
  keyed on `schemaId` length; JSON-with-registry is uncommon and can follow later if there
  is demand.
- Non-negative-integer validation on the parameters at `SchemaConfig` construction time.
- The offset is included in the deserializer cache key so that concurrent handlers using
  the same Avro schema with different prefix lengths do not collide.

### Alternative solutions

* Require users to pre-process the payload themselves (base64-decode, slice, re-encode before the decorator runs). Works today but is boilerplate every user of an offline schema has to repeat, and defeats the ergonomic value of `kafka_consumer`.
* Add ESM Schema Registery integration. For a lamdba that's integrating with 1 Kafka stream, it is not worth adding more infrastructure.

### Acknowledgment

amin-farjadi:
**Issue number:** closes #8385

### Changes

Added a `value_schema_wire_format` arguement to the `SchemaConfig` of Kafka utility, enabling the use of local (packaged with lambda) AVRO value schema for payloads generated by schama-registery aware Confluent producers (i.e. payloads serialised by `KafkaAvroSerializer`).

### User experience

An optional argument has been added to `SchemaConfig`. Documentation has been updated, showing how and when this argument can be used.

---


leandrodamascena:
I'm reviewing this.

leandrodamascena:
Hi Amin, I reviewed the implementation and tested the cases we discussed in the issue.

The overall work is very good. The use case is clear, the API direction makes sense, and the implementation was close to complete. Instead of adding several review comments, I made the final changes directly in your branch with commit `e10e4701`.

I preserved compatibility for positional `SchemaConfig` calls, added Confluent header validation, added the same option for keys, corrected the documentation, and expanded the regression tests.

All 44 Kafka functional tests pass, together with Ruff, mypy, and the documentation build. The new parameters remain optional, so this does not break the current API.

I am going to merge the PR.

leandrodamascena (approved):

PR #8452, opened 2026-09-12
fix(event_handler): isolate local ASGI request state

### LINKED ISSUE #8451: Bug: HttpResolverLocal shares request state across concurrent ASGI requests
### Expected behaviour

Concurrent requests to the same `HttpResolver` / `HttpResolverLocal` ASGI application should retain their own bodies, headers, and routing context. Invalid or cancelled requests should not clear another in-flight request's context.

### Current behaviour

With native Pydantic validation enabled, six concurrent requests with distinct bodies all return the last request's body. This reproduces with 3.34.0 and current `develop`. An overlapping validation failure or cancellation can also clear a pending handler's context.

### Reproduction

Install `aws-lambda-powertools==3.34.0`, `pydantic`, and `httpx`, then run:

```python
import asyncio
import httpx
from aws_lambda_powertools.event_handler import HttpResolver

class Body(BaseModel):
    name: str

app = HttpResolver(enable_validation=True)

@app.post('/echo')
async def echo(body: Body) -> dict:
    await asyncio.sleep(0)
    return {'name': body.name}

async def main():
    async with httpx.AsyncClient(transport=httpx.ASGITransport(app=app), base_url='http://test') as client:
        responses = await asyncio.gather(*(
            client.post('/echo', json={'name': str(i)}) for i in range(6)
        ))
    names = [response.json()['name'] for response in responses]
    print(names)
    assert names == [str(i) for i in range(6)]

asyncio.run(main())

```

Expected: `['0', '1', '2', '3', '4', '5']`

Observed: `['5', '5', '5', '5', '5', '5']`, followed by the assertion failure.

### Cause and proposed fix

`HttpResolverLocal.asgi_handler()` assigns `BaseRouter.current_event` and `BaseRouter.lambda_context`, then awaits resolution. Routing and validation also mutate a shared `self.context` dictionary. Validation middleware runs through a thread bridge, allowing another ASGI request to overwrite this state before the first consumes it. Request cleanup can clear another request's context.

A request-scoped state object carried by `ContextVar` can isolate local ASGI resolution, including middleware threads, without serializing all local requests. I am preparing a focused PR with regressions for distinct bodies, overlapping validation failures, and cancellation.

### Scope

This concerns overlapping requests to the local ASGI server, not concurrent invocations within a Lambda execution environment. I understand the Lambda execution model; local ASGI servers nevertheless accept concurrent HTTP requests. This is separate from #8187 / #8196 (the validation-exception deadlock fixed in 3.29.0).

### Environment

Python 3.13; Powertools 3.34.0 and current develop; local ASGI, no AWS resources required.

wuodar:
**Issue number:** Fixes #8451

## Summary

Concurrent HTTP requests to one `HttpResolverLocal` ASGI app can consume another request's body or lose routing context when another request completes. With Pydantic validation enabled, six distinct requests in the reproduction all return the last body.

This draft isolates state on the local ASGI path. It does not introduce concurrent Lambda invocations or change `APIGatewayHttpResolver`.

### Changes

- Give each ASGI request a state object containing its event, Lambda-context stand-in, routing context, and processed middleware frames.
- Carry that object through a `ContextVar`, including `asyncio.to_thread` middleware and callbacks scheduled back onto the event loop.
- Reset the request binding on completion/cancellation instead of clearing a shared dictionary; preserve the synchronous `resolve()` path outside ASGI requests.
- Add functional regressions for concurrent distinct validated bodies, a validation failure overlapping a pending request, and cancellation overlapping another/subsequent request.
- Widen ASGI scope annotations to the mapping interfaces accepted by HTTP clients.

### User experience

A local development server can handle overlapping requests without those requests overwriting the resolver's event/context. Resolution stays concurrent rather than putting a lock around every HTTP request.

### Validation

- All three new regressions fail against unchanged `develop` and pass with this patch.
- `python -m pytest tests/functional/event_handler -o addopts='' -q`: **863 passed, 1 skipped**.
- Ruff check and format check pass for both changed files.
- `mypy aws_lambda_powertools/event_handler/http_resolver.py --follow-imports=silent`: passes.
- Full `make pr` was attempted but could not start because Poetry is not installed in the contribution environment. The checks above used a separate virtual environment; full repository CI has not been claimed as passing.

### Maintainer review / remaining scope

This is a draft pending issue triage and agreement on the state-isolation boundary. The patch covers state accessed through the resolver and resolver-bound dependency providers. Direct `Router.current_event` / `Router.context` access on an included upstream Router still uses legacy shared state and needs a design decision before treating this as a complete upstream solution. Existing synchronous resolver and event-handler tests remain green.

Cancellation isolation does not imply that this patch cancels handler tasks already scheduled by the existing synchronous middleware bridge; it ensures their state cannot be cleared by another request.

---


aws_lambda_powertools/event_handler/http_resolver.py
```diff
@@ -212,7 +226,46 @@ def __init__(
...
-        self._is_async_mode = False
+
+    @property
+    def _state(self) -> _RequestState:
+        return self._request_state.get() or self._startup_state
+
+    # Powertools declares these as mutable attributes. Properties preserve that
+    # interface while directing each task to its own state. asyncio.to_thread
+    # propagates the ContextVar, so middleware sees the same request dictionary.
+    @property
```
leandrodamascena, line 237: Thanks for working on this, Kacper. I reproduced the concurrency issue and I agree that `ContextVar` is the right direction.

I found one problem with included routers. `router.current_event.path` works on `develop`, but raises `AttributeError` with this change. The request state also needs to be available through the included `Router`.

aws_lambda_powertools/event_handler/http_resolver.py
```diff
@@ -274,25 +327,27 @@ async def asgi_handler(self, scope: dict, receive: Callable, send: Callable) ->
...
-        BaseRouter.lambda_context = context
-
-        self._is_async_mode = True
-
+        # Never write BaseRouter's class attributes: another ASGI request may
+        # enter while validation or the handler is awaiting I/O.
+        state = _RequestState(
+            event=self._to_proxy_event(event._data),
+            lambda_context=context,
+            context=self._startup_state.context.copy(),
```
leandrodamascena, line 335: One more thing I found while testing: this copy disconnects `app.context` from `router.context`.

After `app.append_context(value="request")`, the router can read the value on `develop`, but gets `None` with this change. We should keep both pointing to the same request context.

tests/functional/event_handler/_pydantic/test_http_resolver_pydantic.py
```diff
@@ -365,3 +365,85 @@ def create_user(user: UserModel) -> UserResponse:
...
+        entered = {name: asyncio.Event() for name in ("first", "second", "later")}
+        release = {name: asyncio.Event() for name in entered}
+        pending = []
+
+        @app.post("/concurrent")
+        async def echo(user: UserModel) -> dict:
+            app.append_context(name=user.name)
+            entered[user.name].set()
+            await release[user.name].wait()
+            return {"name": app.context["name"], "header": app.current_event.headers["x-request-name"]}
```
leandrodamascena, line 508: Could you please add a test using an included `Router`? That should cover both `router.current_event` and the context shared between the app and router. Thanks again for taking care of this concurrency issue.

leandrodamascena (changes requested):
I reproduced the concurrency issue and agree with the direction. I found two compatibility regressions with included Router instances; details inline.

leandrodamascena (commented):
Thanks for working on this @wuodar! Amazing work, approved!

leandrodamascena (approved):

PR #8453, opened 2026-09-13
fix(event_handler): match generic alias response models in OpenAPI schema

### LINKED ISSUE #8450: Bug: OpenAPI schema generation raises `StopIteration` when a custom response `model` is a generic alias such as `list[Model]`
### Expected Behaviour

Declaring a list body in a custom response, e.g.

```python
responses={200: {"content": {"application/json": {"model": list[Item]}}}}
```

should produce an array schema for that response, as it did through 3.24.0.

### Current Behaviour

Since 3.25.0, `get_openapi_schema()` (and therefore `get_openapi_json_schema()`, `enable_swagger()`, and OpenAPI validation setup) raises an unhandled `StopIteration` for any route whose custom response `model` is a generic alias. Routes whose `model` is a plain Pydantic class are unaffected.

```
  File "aws_lambda_powertools/event_handler/openapi/schema_generator.py", line 251, in _build_responses
    operation_responses[resp_code] = _build_custom_response(
        response=copy.deepcopy(responses[resp_code]),
        ...
  File "aws_lambda_powertools/event_handler/openapi/schema_generator.py", line 517, in _build_custom_response
    response["content"][content_type] = _resolve_response_payload(
        ...
  File "aws_lambda_powertools/event_handler/openapi/schema_generator.py", line 539, in _resolve_response_payload
    return_field = next(
        filter(
            lambda model: model.type_ is model_payload_typed["model"],
            dependant.response_extra_models,
        ),
    )
StopIteration
```

## Root cause

#7952 (shipped in 3.25.0, fixing #7711) started deep-copying each custom response dict before resolving its `model`:

```python
response=copy.deepcopy(responses[resp_code]),
```

`_resolve_response_payload` then locates the matching `ModelField` with an identity comparison:

```python
lambda model: model.type_ is model_payload_typed["model"]
```

`copy.deepcopy` returns the same object for a class, so `model: Item` still matches. It returns a **new** `types.GenericAlias` for `list[Item]`, so the identity check never matches, `next()` exhausts the filter, and `StopIteration` escapes. The `AssertionError("Model declared in custom responses was not found")` guard on the next line is never reached.

```python
>>> import copy
>>> copy.deepcopy(list[int]) is list[int]
False
>>> copy.deepcopy(list[int]) == list[int]
True
```

The regression tests added in #7952 use list return annotations on the handler but a plain class as the custom response `model`, so this path was not covered.

### Code snippet

```python

class Item(BaseModel):
    id: int


@app.get(
    "/items",
    responses={200: {"content": {"application/json": {"model": list[Item]}}}},
)
def list_items() -> list[Item]:
    return []


app.get_openapi_schema()  # StopIteration on >= 3.25.0, works on <= 3.24.0
```

### Possible Solution

Compare by equality instead of identity in `_resolve_response_payload`; `GenericAlias` implements `__eq__` structurally and classes compare by identity anyway:

```python
lambda model: model.type_ == model_payload_typed["model"]
```

With that one-line change the snippet above and a 100+ route application that hit this both generate correctly.

Alternatives: resolve the `model` against `response_extra_models` *before* deep-copying, or exclude the `model` value from the copy (it is only read, never mutated).

### Steps to Reproduce

1. `pip install "aws-lambda-powertools==3.34.0" pydantic`
2. Run the snippet above.
3. Observe `StopIteration`.
4. `pip install "aws-lambda-powertools==3.24.0"` and run it again: the schema is generated.

Verified on 3.24.0 (works), 3.25.0 (fails), 3.34.0 (fails) with Python 3.13.7 and Pydantic 2.12.5. The identity check is still present on `develop` at the time of writing.

3.34.0

3.13

### Packaging format used

PyPi

### Debugging logs

```python

```

TanbirRamim:
**Issue number:** closes #8450

### Changes

Since custom responses started being deep-copied before their `model` is resolved, a generic alias such as `list[Model]` comes back as a new `GenericAlias` object, so the identity check in `_resolve_response_payload` never found the matching field and `get_openapi_schema()` raised `StopIteration`. This compares the model with `==` instead of `is`; plain classes still compare by identity, and generic aliases compare structurally. I added a regression test in `test_openapi_responses.py` covering `list[Model]` and `dict[str, Model]`, which fails with `StopIteration` before the change and passes after; the event_handler functional tests, ruff and mypy all pass locally.

### User experience

Before, any route with a generic alias as a custom response model broke OpenAPI schema generation, Swagger and validation setup with `StopIteration`. After, those routes produce the expected array or object schema again, as they did up to 3.24.0.

---


leandrodamascena (approved):
Thanks Tanbir, this is a very good fix. I reproduced the `StopIteration` on `develop` and confirmed that both `list[Model]` and `dict[str, Model]` generate the expected schemas with this change.

Using equality here makes sense because `deepcopy` creates a new generic alias, while regular model classes keep the same behavior. I’ll approve and merge this.

PR #8457, opened 2026-09-14
test(shared): compare abs_lambda_path against a normalised path

### LINKED ISSUE #8449: abs_lambda_path returns native separators on Windows, failing its own unit test
On Windows, `tests/unit/test_shared_functions.py::test_abs_lambda_path_empty_envvar` fails on an unmodified `develop` checkout (3bd47e2):

```
assert abs_lambda_path() == default_lambda_path
AssertionError: assert '\var\task' == '/var/task'
```

`abs_lambda_path()` builds its result with `pathlib.Path`, which is `WindowsPath` off-Linux and rewrites the separators. `LAMBDA_TASK_ROOT` is always POSIX-shaped because the runtime is Linux, so the value the docstring says is returned unchanged comes back as `\var\task`.

Deployed functions are unaffected. What it does affect is contributors running the suite on Windows, and the Redis-certs example in `docs/utilities/idempotency.md`, which interpolates `f"{abs_lambda_path()}/certs/..."` and so yields mixed separators. Your workflows are `ubuntu-latest` only, so CI cannot see either.

Possible solution: use `PurePosixPath` when `LAMBDA_TASK_ROOT` is set, keeping `Path.cwd()` for the unset case. `test_abs_lambda_path_w_filename_envvar` also needs its oracle corrected, since it currently builds the expected value with the same `Path` call. I have both working locally and can open a PR if useful.

To reproduce: check out `develop` on Windows and run `pytest tests/unit/test_shared_functions.py -k abs_lambda_path`.

AI-assisted; I reproduced this and ran the suite myself.

MohammedAlkindi:
**Issue number:** closes #8449

`test_abs_lambda_path_empty_envvar` compared `abs_lambda_path()` against a hardcoded `/var/task`, which only holds on Linux, so the test failed on a Windows checkout.

**Changes** The assertion now compares against `str(Path(default_lambda_path))`, so it normalises to whatever separators the function returns on the platform running the suite. Runtime code is unchanged: as @leandrodamascena notes, Lambda runs on Linux and `abs_lambda_path` should keep returning native paths.

**User experience** No change. The fix is test-only.

---


leandrodamascena (changes requested):
Thanks @MohammedAlkindi . Lambda runs on Linux, so forcing POSIX paths in the runtime code does not make sense. The problem was only the test hardcoding a Linux path.

I pushed a change to keep the runtime code as it was and fix only the test.

Please add the acknowledgment section back. The check is failing, and I cannot merge the PR without it.

leandrodamascena (approved):

PR #8470, opened 2026-09-16
fix(feature_flags): warn on comparator evaluation errors

### LINKED ISSUE #8425: Feature request: surface comparator exceptions as misconfiguration instead of silent non-match
### Use case

`_match_by_action` catches every `Exception` from a comparator, logs it at `debug`, and returns `False`. A type mismatch between a context value and a condition value (`STARTSWITH` on an `int`, `KEY_GREATER_THAN_VALUE` between `str` and `int`, a non-list context for `ANY_IN_VALUE`) therefore looks identical to "condition evaluated and did not match". In production, with debug logging off, a misconfigured rule is indistinguishable from a rule that correctly excludes the user, and it can stay that way for months.

### Solution/User Experience

Two steps:

1. **Now (minor, no behaviour change):** log the swallowed exception at `warning` rather than `debug`, including feature name, rule name, key, and action, so the problem is visible in default log output.
2. **Next major:** raise a dedicated error (e.g. `ConditionEvaluationError`, in the `SchemaValidationError` family) naming feature, rule, key and action. Keep `validation_exception_handler` as the opt-in to restore "return a boolean instead of raising" for anyone who wants the old behaviour.

Context: the TypeScript port (aws-powertools/powertools-lambda-typescript#5614) throws on incompatible operands from day one; this proposal brings Python to the same place across a major.

### Alternative solutions

```markdown
Keep swallowing but add an `on_comparator_error` callback. Functionally the same as the existing handler; doesn't address the default being silent.
```

### Acknowledgment

Shubham-Padkonde:
**Issue number:** closes #8425

This covers step 1 of #8425 (warning-only). Step 2, a dedicated exception in the next major version, is still open, so please reopen the issue or split it out if you want to track that separately.

### Changes

Emit a warning when a feature-flag comparator raises while evaluating a present context key. Include the feature, rule, key, action, and exception type, without exposing operand values or exception messages. Pass that diagnostic context through the private comparator helper and document the warning.

### User experience

Incompatible operands, such as `STARTSWITH` on an integer, currently look like ordinary non-matches unless debug logging is enabled. They now produce an actionable warning. Existing non-match results and registered exception handlers are preserved; missing keys and valid non-matches stay quiet.

This implements the first step proposed in the issue. Introducing a new exception remains a separate major-version decision. The tested approach was described in the issue before submission.

Validation:
- Initial functional regression: 7 failed, 2 passed before the change; 9 passed after. An additional exception-message privacy regression brings the final focused suite to 10 passing cases.
- Full non-performance suite on Python 3.12/Linux: 2,579 passed, 4 skipped, 10 deselected; coverage 96.69%.
- All 10 performance checks passed.
- Repository Ruff checks, mypy (897 source files), security and complexity baselines, and modified-Python pre-commit checks passed.
- `make pr` stops at its Docker-based documentation command because Docker/WSL integration is unavailable here. Running markdownlint directly passes the edited page; the full docs check reports the pre-existing `MD053` error in unchanged `docs/changelog.md:1`. The remaining check categories were run separately as listed above. AWS end-to-end and Redis integration tests were not run.

Prepared with OpenAI Codex assistance.

---


Shubham-Padkonde:
This implements only the warning-only first step of #8425; the proposed major-version exception behavior remains open. The related-issue check currently requires a development link or a closing keyword. Could a maintainer link #8425 manually (or advise the preferred tracking issue for this first step)? I have avoided a closing keyword so merging this PR does not inadvertently close the remaining major-version work.

